PR #17, opened 2021-06-02
Add restructured code

pz-max:
The following adjustments were made:
- osm extraction script now in WP5 
- extraction script is optimised
- major cleaning process of extraction script is put into a separate jupyter script which can be converted into a py script after being more stable.
- all the code base is made linter conform


mnm-matin:
Please only change directory of files sparingly as the africa files have to be downloaded and processed again which can take about two hours.

pz-max:
Hey Matin, uploaded the updates & checked that everything works.
If not obvious, we discuss the reasons for the changes at the Code Meeting later.

Issue #28, opened 2021-07-05
Data Cleaning need to be added to new script


mnm-matin:
See TODO comments in cleaning scripts. Would have to refine and add rules for cleaning inclusive of comments.

pz-max:
Are you talking about the osm_data_cleaning.py script? If so, you have to push the comments. 
Should we maybe make first the jupyter run perfectly before making the cleaning script?

mnm-matin:
Just confirmed the TODO: comments are present as a placeholder in the script. Absolutely, let's craft the rules in the jupyter notebook first. Best!

mnm-matin:
@pz-max linting can really only be enforced once we have decided on a formatting standard for the code base. Best!

pz-max:
Guess sticking to the super-linter suggestions should be fine for now. 
Its a learning experience and leads to more homogenous code

PR #36, opened 2021-07-19
Update cleaning file

davide-f:
I propose a cleaned version of the cleaning script, as discussed in #28 

Please comment, modify and adjust as you please.

Davide


davide-f:
Note: I debugged the file and crosscheched the outputs of _substations and _lines cable that are identical (except lines for the filtering criterion voltage > or >= threshold) [I use the latter, while the original file was with the former; yet it is fast to change]

pz-max (approved):
Great job Davide. The code looks clean and useful.

pz-max:
Hi @davide-f, we only need to correct the linting errors. Linting is there to keep our code uniform and easy to read.
If you click on the "X" marked Lint Code Base details, then you will see where you have linting issues.
Could you please correct them? (let me know if you have any question)

davide-f:
I made some changes based on the lint log, but there must be something left for sure; please, could you run the flows so that I can proceed?

PR #41, opened 2021-07-28
Update linter.yml

pz-max:
Removes snakemake linting


pz-max:
Removes snakemake linting

Issue #34, opened 2021-07-17
Create rule build_shapes - necessary for base_network


pz-max:
See [PyPSA-Eur](https://github.com/PyPSA/pypsa-eur/blob/e2c3118f08bccddbd2eb47002b9d89d1f9eb3087/Snakefile)

PR #54, opened 2021-07-31
First runnable build_shapes.py linked and tested with snakemake

davide-f:
First prototype of build_shapes.py to answer #34 

Currently, the website to download gadm files are down. To prevent further problems, probably, it could be a good idea to keep some raw dataset just to make tests and avoid such problems in the future.

Problems with the current version:
- the simil-NUTS3 regions shall be further debugged to uniformize the index
- as described in #46 the current implmentation of the "NUTS3" regions require the integration of gpd and population data
- currently, the script downloads the geopackages of data internally. Probably this should be moved to retrive_data in the future
- the file of the eez zones may be downloaded by the user inlcuding data from the website ( https://www.marineregions.org/download_file.php?name=World_EEZ_v11_20191118_gpkg.zip ), We may include this file in the package to avoid the user to manually download it


scripts/build_shapes.py
```diff
@@ -0,0 +1,312 @@
...
+    ret_df = gpd.GeoSeries({
+        k: v
+        for k, v in ret_df.iteritems() if v.distance(country_shapes[k]) < tol
+    })
+    ret_df.index.name = "name"
+
+    return ret_df
+
+
+def nuts3(update=False):
```
pz-max, line 220: Would remove nuts3 and any EEZ (European Economic Zone) naming, wherever it makes sense at the moment.
These are EU namings. I would name it "gadm" and "aez". 

As you said, we need to work on the GADM level description. Guess that can be a function input somewhere. But that's work ongoing. Otherwise, it looks good for now :+1: Well done!

pz-max (commented):
Well done. Made a note somewhere in the build_shapes file. Hope you can see it.

scripts/build_shapes.py
davide-f: Ok! I agree, I kept the original nomenclature on the similarities with PyPSA-Eur.
Regarding gadm, also countries are derived from GADM, but maybe we can name them gadm to refer to the small regions.
Regarding EEZ, the first E stands for Exclusive, thus I would keep the "EEZ" nomenclature

PR #91, opened 2021-09-15
Fixing absence of Namibia in the build_bus_regions output and plotting tools

davide-f:
This pull requests solves the issue on the missing Namibia data and with a Notebook provides useful features for debugging the plotting issues we are experiencing.
Still there is work to do, but this is a first step


davide-f:
Merged because it extends the original main with no major difference with new-main

Issue #99, opened 2021-09-17
On missing country areas and off_shore shapes in build_bus_regions


davide-f:
As discussed with @pz-max , it is well known that the build_bus_regions appear to have wide blank ares, mainly related to Somalina, Ciad, Central African Republic, South Sudan and Somalia.
For Somalia, there are really no data, but for the other countries there are only few data (1-5 lines/buses), so that when the voronoi cells are calculated, vast areas of the countries become blank.

My feeling is that this issue depends on the implementation of voronoi_partition_pts which limits the size of each voronoi cell. When the number of points is large and widespread all over the region this is ok, but in our case where in some countries we have 1-3 points very close each other, the above does not work.

I see two options:
- we temporary develop our own voronoi_partition_pts and we notify the issue in github
- we post the issue in github and we wait for the modification

The version (a) may be faster

Regarding the off_shore areas, the same problem may apply also there, however, some countries may not have off shore buses.
I am not sure whether we should introduce fake nodes and lines to do so at this stage. Maybe we should tackle that issue when we will address the "under_construction" lines or similar.

davide-f:
As a clarification for the above, see the image below (reproduceable by the build_bus_regions_verifybuseslines notebook)
![image](https://user-images.githubusercontent.com/67809479/133790447-53aa680d-8c44-4352-87c1-4cd19e1a60b2.png)

davide-f:
@pz-max do you like it? XD

The only thing left is the off_shore regions of countries where no offshore bus is in place. For the moment, I think we could move on with this implementation.
They will become important when we consider optimal planning purposes including the development of new bus.

Probably, we have now other priorities, so maybe it could be a good idea to close this issue and leave the issue on the off_shore areas just as a reminder.
Anyway the issue about installing new buses could be relevant also for onshore areas of countries with very limited number of buses.

![image](https://user-images.githubusercontent.com/67809479/133794232-4b56e45c-7df9-4b3e-b8b7-916f79054229.png)

pz-max:
### missing countries
Looks good! FYI, for the other readers: Senegal, Gambia and West-Sahara are excluded for now. See issue #88 

### voronoi_partition_pts
I think it's fine to implement a temporary function for the Voronoi partition and wait until the new code is pushed into FRENSA code.

### missing offshore shapes
We have to review how the offshore buses are defined and fix that

davide-f:
[...]

PR #100, opened 2021-09-17
Final fix build_bus_regions to avoid holes in countries with data

davide-f:
This pull request solves the major issues related to #99 .

The output of build_bus_shapes is the following:
image

The missing elements left are the empty off-shore areas for countries without off shore buses as discussed in the issue.


pz-max (commented):
Changes look fine to me. Some questions:
- The new Voronoi partition function comes from the original FRESNA code?
- Do you think we can stop updating the geojson files? Since we implemented the snakemake integration these datafiles should be easy to reproduce.

davide-f:
Below the answers:
1) yeah, the copied code is a customization or the code from FRESNA. I also
submitted an issue there with a similar code to ours. Depending on the
output, we can easily remove the custom function and use the fresna one.
2) to fix the issue, I touched many files along the chain and the geojsons
already pushed needed to be updated. The point is: should we keep the
geojsons or not? The build shape code takes some time to process gdp and
pop. I thought it was better for me to process everything and give that
clear for all of us to use.

pz-max:
1) Perfect
2) I think, we can keep the current files. Yep, makes sense to update the files

Issue #53, opened 2021-07-31
Need to split Senegal-Gambia data in separate shapefiles. Related to issue #40 


pz-max:
We download generators, towers, lines, etc. from Geofabrik. 
BY default geofabrik assigns one country shapes for "Senegal-Gambia".
According to the ISO these should be represented be splitted.

We need to do the following:
- Take the country shapefile from @davide-f where SN and GM are separate shapes
- Split SN-GM data into SN and GM respectively. This could work with shapely intersect. Though must be carefully performed. Some edge-cases might be interconnector lines

mnm-matin:
Lines for SN-GM
![image](https://user-images.githubusercontent.com/45293386/128031082-3d91f539-1758-46b6-9a95-383e1efc69fa.png)

Substations for SN-GM
![image](https://user-images.githubusercontent.com/45293386/128031322-1914ba39-ad88-4cf1-a19f-988deca49e19.png)

Due to the sparsity of data, I would highly recommend keeping SN-GM as merged area. Best!

pz-max:
From a long-term perspective, if data becomes more available, splitting SN-GM seems more sustainable.
It would also avoid that we have to merge GDP rasters, GADM shapefiles for SN-GM.
I think we exclude SN-GM for now - until someone wants to split them

mnm-matin:
What do you think about the line circle in red above. How do you propose splitting it? I'm asking because even if more data becomes available it is more likely (in my [opinion](https://www.esi-africa.com/industry-sectors/generation/omvg-interconnection-on-course-to-connect-the-gambia-by-2019/)) that this will be the case for the newly acquired data. 

Also do you think it might be necessary to merge other countries in africa? (for example lesotho and south-africa)

pz-max:
Yes, indeed a challenge with interconnectors. I saw Nigeria also had OSM lines being tagged as "Nigeria" but crossed other countries. So it's a problem we generally face. Maybe PyPSA-Eur handled this special case already.

Regarding its impact on the energy system model, I guess if a line is tagged "SN" or "GM" might have no influence on the optimisation. It might only matter if someone wants to quantify the current value £$€ of assets in a country (there I am sure we can find solutions).

Would keep countries split as far as possible for now. Maybe good to know Lesotho (33km2) is a bit larger than Beligum (30km2), and roughly 7 times smaller than UK (210km2). Belgium is a country which we model in PyPSA-Eur.

PR #104, opened 2021-09-20
Methodology to specify the naming of the assets by the country_shapes and using the ISO coding

davide-f:
This pull request attempts to solve issues #88 and #53 in a general way that can be applied to different regions.

The assumption of this code are:
- countries in the config file are in ISO name for the whole code to work. Subsets of functions can be easily made work also with geofabrik standard (especially the osm_extractor and filtering)
- geofabrik-unique names are assumed to have a >2 coding
- the referencing between iso- and the corresponding geofabrik-codes of the country data not available with the standard iso coding in geofabrik must be described with a special dictionary implemented in osm_data_cleaning. Currently, the senegal-gambia and western-sahara regions are implemented.

Code tested and working for the entire africa with this output:
image
Empty areas are related to Gambia, Guinea-Bissau, Liberia and Somalia:
- Liberia has no voltage data for lines
- Guinea-Bissau and Somalia no data
- Gambia no detected data

Not yet updated the geojsons. I can add that commit if we agree on that.

Update: As signaled in #113 we clarify that with the current implementation, we currently chose to keep crossborder lines, as shared between countries inner and outer the selected areas.


pz-max (commented):
cool stuff

PR #110, opened 2021-10-05
Network simplifiction

hazemakhalek:


davide-f:
I saw the good work! Awesome! sorry for the partial absense, deadline last sunday...
I will crosscheck the changes for Africa.

I have some comments for the future commits/pull requests:
- please, especially when working in debug, do not update the geojsons. I am trying to build the geojsons for entire africa, so that any other user could calibrate on their specific regions. Clearly you are focusing on a region and that's ok, but please do not update the geojsons
- the os at the beginning of the files shall be kept commented in the pushed version, so that it can work with snakemake. It can be decommented in the debug phase.
- before merging pull requests, check for the style checks to be completed

As a general recommendation, when changes are relatively heavy, pull requests are useful for letting also us review the changes before merging. For minor changes, commits are fine.
In the next days, I'll verify the above on the africa.
Good job!

hazemakhalek:
- Updating the geojsons was not intended, sorry about that. I had a talk with Max about it earlier today and I'm wondering why actively update the geojsons? why shouldn't we add them to the gitignore file and one will generate them once they run the right script.
- True
- Okay



In general yes, I agree with the pull requests approach for big changes. It's just this time we, Max and I, discussed the code briefly and ran it before merging. Next time for sure!

I'll test it out again after the merge in the next couple of hours and open a couple of issues to explain what is still needed.
We'll talk about it soon ;)

davide-f:
Regarding the first point, the idea is exactly avoid to replace them and place them related to whole africa once for all the users.
The reason about why storing data in github is to avoid the user and all of you to dowload all the data needed to run the code, in the first case.
Running the entire code from scratch for the entire africa involves downloading >10-30Gb data. Instead, if we provide them with these data, there is no need to do so.
In the future, we have to wonder about intermediate approaches. We could discuss about that as well.

hazemakhalek:
One thing I can suggest is that we rename these big all-africa files to a different name, add a '-allafrica' or '-default' substring to the file names. This way they will only have to rename the big files and we don't face this issue, maybe then gitignore the files we generate. What do you think?

PR #116, opened 2021-10-12
Integrate powerplantmatching

koen-vg:
Add a snakemake rule to generate "resources/powerplants.csv", which uses a fork of powerplantmatching. The resulting powerplants are added to the network in the "add_electricity" rule.

I will get in touch with upstream powerplantmatching to see if we can address our needs upstream, so we don't have to maintain our own fork.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -185,12 +185,24 @@ rule build_renewable_profiles:
     script: "scripts/build_renewable_profiles.py"
 
 
+rule build_powerplants:
+    input:
+        base_network="networks/base.nc",
+        pm_config="powerplantmatching_config.yaml",
+        custom_powerplants="data/custom_powerplants.csv"
+    output: "resources/powerplants.csv"
+    log: "logs/build_powerplants.log"
+    threads: 1
+    resources: mem=500
+    script: "scripts/build_powerplants.py"
+
+
 rule add_electricity:
     input:
         base_network='networks/base.nc',
         tech_costs=COSTS,
         regions="resources/regions_onshore.geojson",
-        # powerplants='resources/powerplants.csv',
+        powerplants='resources/powerplants.csv',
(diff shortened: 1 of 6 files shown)


pz-max:
Hey @koen-vg, I am not sure but I think powerplantmatching had a pull request recently that fixed the initial issue we had (probably avoids our fork)?

koen-vg:
> Hey @koen-vg, I am not sure but I think powerplantmatching had a pull request recently that fixed the initial issue we had (probably avoids our fork)?

They fixed one issues, but other issues still remain, so right at this point we still need the fork. There are two issues (addressed in the two most recent commits to the fork, see https://github.com/koen-vg/powerplantmatching/commits/pypsa-africa):
- powerplantmatching still had a bug related to custom config, but that should be easy to address upstream.
- More seriously, I modified powerplantmatching to use exclusively alpha2 country codes, due to country name inconsistencies between different databases. This breaks compatibility, so we could suggest to upstream to include this as an options (which would require a little bit of extra work).

pz-max:
> > Hey @koen-vg, I am not sure but I think powerplantmatching had a pull request recently that fixed the initial issue we had (probably avoids our fork)?
> 
> They fixed one issues, but other issues still remain, so right at this point we still need the fork. There are two issues (addressed in the two most recent commits to the fork, see https://github.com/koen-vg/powerplantmatching/commits/pypsa-africa):
> 
>     * powerplantmatching still had a bug related to custom config, but that should be easy to address upstream.
> 
>     * More seriously, I modified powerplantmatching to use exclusively alpha2 country codes, due to country name inconsistencies between different databases. This breaks compatibility, so we could suggest to upstream to include this as an options (which would require a little bit of extra work).

Great detail Koen, let's try to get the issue fixed upstream. Happy to discuss that more on Friday :)

Issue #115, opened 2021-10-12
Open issues Data WorkFlow


davide-f:
This issue is meant to represent the major list of open issues across the dataworkflow, to present a summary and provide guidance.
This thread is meant to discuss data workflow structures and architecture, not solving the specific issues, whose discussion will be tackled by different independent issues.

# **Core-related issues by script**
## *retrieve_databundle_light*
Script to download data from google drive
- [x] (optional) Rather than download all files in folders, a selective download may be created, to distinguish test cases, all africa case, etc.
- [x] (optional) Add progress bar in the downloading process

## *build_shapes*
Script to create country shapes and populate them with GDP and Population data

- [x] Improve the modelling of GDP for the shapes, by improving the resolution of tiff images #56 
- [x] (optional) Add progress bar in the shape processing #56 
- [x] (optional) Parallel processing could be add to speed up the calculations #56 

## *osm_pbf_power_data_extractor*
Script to download OpenStreetMap data from openfabrik

- [x] #132

## *osm_data_cleaning*
Script to filter the raw data downloaded from openfabrik

- [x] Output filenames to be picked from snakemake
- [x] Verify how to tackle lines without endings (or accept the current formulation)
- [x] Add line endings to substations list (activity currently done in base_network) #113 
- [ ] Verify definition of underground/under_construction/tag_frequency, which are currently initialized only
- [x] Verify how to manage the substation datasets, in agreement to the future implementation of build_powerplants: #121 build_powerplants loads the raw file downloaded from OSM 
- [x] Fix None geometry shapes #131 

## *osm_built_network*
Script to create the network description

- [x] Filter out generator lines: sometimes lines linking the generator to the closest substation are set. It may makes sense to remove that. However, this shall be of second priority: we have to see how simplify network will act on such lines and that may automatically takes the issue into consideration
- [ ] Check how to properly tackle buses with/without low voltage
- [ ] Implement option parameters in the config and the corresponding methodologies about how to create the network, based on the data
- [x] Move the part that adds substations based on the line endings to osm_data_cleaning #121 
- [x] Need for merging buses corresponding to the same substation_id by voltage. The lines and buses datasets may lead to create buses whose latitude/longitude coordinates are very close but not identical. That creates problems that are now fixed in the base_network and shall be moved to the cleaning phase #121 
- [x] Improve the merging process of the buses taking into consideration the type of the node AC/DC
- [x] Fix None geometry shapes #131

## *base_network*
[...]

Issue #106, opened 2021-09-28
remove snakemake dependecies from functions


pz-max:
Functions should also work also outside the snakemake workflow. (See PyPSA-Eur PR here https://github.com/PyPSA/pypsa-eur/pull/275). Instead of using snakemake in a function, we should simply assign it to a variable.

An example of do's and don'ts
**Don't:**
`def load(n, cost=snakemake.input["cost"]):`

**Don't:**
>`def load(n):`
>>cost=snakemake.input["cost"]

**Do:**
>`def load(n, costs):`
>> cost=costs

and later at the execution important the i.e. costs from snakemake:
costs = snakemake.input["cost"]

pz-max:
@Hazem-IEG, @davide-f  We already partially applied the do's. At some point we should clean up function by function.

davide-f:
The only major function where it is heavily used is base_network; I've cleaned in other functions in a PR #580

davide-f:
Revision: they are all revised; I think we can close this issue once the PR is merged

PR #121, opened 2021-10-27
Various fixing, in particular buses and lines filtering

### LINKED ISSUE #113: Buses mislabeled/mislocated to different countries
While building the network for only 2 countries: Morocco and Algeria, I noticed that some of the buses on the base_network plot are located outside the 2 countries, specifically in Tunisia and Libya.
 
NA_base_network

There is at least 4 mislocated/mislabeled buses in each of Tunisia and Libya. The coordinates of these buses in the base network actually lie outside Algeria, so there's no significant deviation with the plot.

Inspecting the raw generators, lines and substations, it seems that this might be, at least for most of the cases, a result of the cross border lines. The algorithm might be adding the bus to the end of these lines?
image

This issue still persists after clustering, below is a yet-to-be-perfect 100-node clustered network of Morocco and Algeria:
image

In the next few days I will tryout the network with Tunisia and Lybia as well and see to which countries these buses are labeled. 
I believe this is probably just an issue that occurs at the borders of the network when the model is built for only a part of Africa.

davide-f:
This issue solves these issues:
- automatic data workflow for build_powerplants: the csv input is the output of the data processing
- simplify non-snakemake run of selected scripts: make the scripts runnable easily by just clicking run. To run the function mock_snakemake, the current working directory of python should be the root folder pypsa-africa; when running the scripts in the scripts folder, it is needed to change the current folder to its parent one; that has been better automated
- Fix the procedure used to identify buses in off-shore areas, before the procedure was bugged and the output was improperly evaluated: it run but the output wasn't the expected one. Now the check has been finalized
- the input file "eez_v11.gpkg" has been added as an input for the build_shapes function so that retrieve_databundle is run before build_shapes
- the option "add_line_endings" has been added to the osm_cleaning_data so that when true, the endings of each line are added to the substation list, before the country labeling. This solves the problem shown in #113 . To close #113 it is only left to understand how to manage the interconnection lines between the selected countries and the others; my personal view is to remove them. Next fix to do
-  in osm_build_network the methodology to assign station_ids has been improved to group together substations within a desired tolerance tol and by voltage; in the future we shall also include by type AC/DC. Nodes at different voltages are currently connected by fake lines that basically represent transformers. A development will be to add transformers as PyPSA Europe does
- in osm_build_network  the function set_line_ids has been created to properly allocate and define the bus corresponding to each line ending. The procedure links each line ending to the closest node having the same voltage as the line and modifies the linestring so that the path is complete and the ending of each line terminates to a bus
- osm_build_network outputs and inputs are now taken from snakemake, instead of being hard coded; in alignment to #106 
- Other fixing has been performed to remove problems, e.g. number of circuits = 0, under_construction=True improperly, etc.
- In build_natura_raster, it has been added a featyre so that the user can set as inputs in the snakemake a folder and the code automatically loads any shape file contained in that folder and their nested folders

The notebook "simplify_and_cluster_network" has been improved to derive the images noted in this pull request.

These changes are aligned to #115

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -37,6 +37,7 @@ datafiles = [
         "data/raw/copernicus/PROBAV_LC100_global_v3.0.1_2019-nrt_Discrete-Classification-map_EPSG-4326.tif",
         "data/raw/gebco/GEBCO_2021_TID.nc",
         "data/raw/eez/eez_v11.gpkg",
+        "data/costs.csv",
         "data/raw/landcover/WDPA_WDOECM_Oct2021_Public_AF.zip",
         "data/raw/hydrobasins/hybas_lake_af_lev04_v1c.shp",
         ]
@@ -94,6 +95,7 @@ rule build_shapes:
         # nuts3='data/bundle/NUTS_2013_60M_SH/data/NUTS_RG_60M_2013.shp',
         # nuts3pop='data/bundle/nama_10r_3popgdp.tsv.gz',
         # nuts3gdp='data/bundle/nama_10r_3gdp.tsv.gz',
+        eez='data/raw/eez/eez_v11.gpkg'
     output:
         country_shapes='resources/country_shapes.geojson',
         offshore_shapes='resources/offshore_shapes.geojson',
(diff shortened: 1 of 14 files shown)


pz-max (commented):
Looks good Davide Great job! :)
Only need to solve the merge conflicts and that's it

PR #134, opened 2021-11-06
Alternative clustering

hazemakhalek:
Two bugs still remain here:

First, within my example countries (MA and DZ) there is one bus falsely located to the wrong country, this is btw much better than before, it used to be around 12 buses. Hardcoded for now (Line 483 in `cluster_network`), proper fix still needed

Second in the substations, I had to overwrite all the substation values to 0, Line 299 in `cluster_network.py`


doc/api_reference.rst
```diff
@@ -7,83 +7,85 @@ API Reference
 _helpers
```
davide-f: Hazem, good work, however, maybe you should revert the documentation files that have been modified while you were editing your branch.
You may just copy and paste the files in the doc folder from the new-main into your branch and that should fix that issue.

scripts/base_network.py
```diff
@@ -38,14 +38,14 @@
...
 - ``data/parameter_corrections.yaml``: Corrections for ``data/entsoegridkit``
 - ``data/links_p_nom.csv``: confer :ref:`links`
 - ``data/links_tyndp.csv``: List of projects in the
-  `TYNDP 2018 <https://tyndp.entsoe.eu/tyndp2018/>`_ that are at least *in permitting*
-  with fields for start- and endpoint (names and coordinates), length, capacity,
-  construction status, and project reference ID.
+`TYNDP 2018 <https://tyndp.entsoe.eu/tyndp2018/>`_ that are at least *in permitting*
+with fields for start- and endpoint (names and coordinates), length, capacity,
+construction status, and project reference ID.
 - ``resources/country_shapes.geojson``: confer :ref:`shapes`
```
davide-f, line 49: Similarly to above, please, set the docstring here as in new-main

scripts/build_bus_regions.py
```diff
@@ -145,7 +147,35 @@ def custom_voronoi_partition_pts(points,
     polygons_arr[:] = polygons
     return polygons_arr
 
-
+def get_gadm_shape(onshore_locs):
+    def locate_bus(coords):
+        try:
+            return gadm_shapes[gadm_shapes.contains(Point(coords['x'],
```
davide-f: gadm_shapes may be given as an argument in the function, and when the function is called, we can give them the proper argument
This may be more readable and manutentable in my opinion; do you agree?

scripts/build_natura_raster.py
```diff
@@ -64,28 +64,6 @@
 # Requirement to set path to filepath for execution
 # os.chdir(os.path.dirname(os.path.abspath(__file__)))
```
davide-f: This file has been modified while you were working on your branch, please, revert it to the one in new main.

scripts/build_bus_regions.py
```diff
@@ -189,6 +220,13 @@ def custom_voronoi_partition_pts(points,
         # print(shapely.validation.explain_validity(onshore_shape), onshore_shape.area)
         onshore_locs = n.buses.loc[c_b & n.buses.substation_lv, ["x", "y"]]
         # print(onshore_locs.values)
+        if snakemake.config['alternative_clustering']:
```
davide-f: The approach seems to be:
- if alternative_clustering is in place: then (a) set bus_regions shapes using gadm, instead of voronoi, (b) re-process the results in build_renewable_profiles and (c) perform the clustering with the shapes.
- otherwise, do the typical workflow

I was wondering that maybe the (a-c) may not need to be cross-linked, they may be two separate options:
- in bus_regions: use voronoi or gadm shapes; when gadm shapes are processed, the inputs may be prepared to avoid the special option in build_renewable profiles maybe
- in cluster_network we provide an optional shape that is used to clusterize. When gadm option is used in bus_regions and the same shape is used in cluster_network we shall obtain the same result of the current process, but the user is not constrained to use both at the same time only. What do you think?
To me the above may be simple to implement; yet, we could decide whether to do it in this pull request or later.

scripts/build_renewable_profiles.py
```diff
@@ -238,8 +238,15 @@
 
     cutout = atlite.Cutout(paths["cutout"])
     regions = gpd.read_file(paths.regions).set_index("name").rename_axis("bus")
+    regions_crs=regions.crs
     # TODO: Drop NAN maybe somewhere else? NaN lead to error otherwise
     regions = regions.dropna(axis="index", subset=["geometry"])
+    if snakemake.config['alternative_clustering']:
```
davide-f: In https://github.com/pypsa-meets-africa/pypsa-africa/pull/134/files#r744323030 I was referring to this if

scripts/cluster_network.py
```diff
@@ -342,14 +381,23 @@ def clustering_for_n_clusters(
...
-                                       focus_weights, algorithm)
+        
+        if snakemake.config['alternative_clustering']:
+            n, busmap = busmap_for_gadm_clusters(n, 
+                    snakemake.config['build_shape_options']['gadm_layer_id'])                              #TODO make func only return busmap, and get level from config
+        else:
+            busmap = busmap_for_n_clusters(n, n_clusters, solver_name,
+                                               focus_weights, algorithm)
 
+        if snakemake.config['alternative_clustering']:                          #TODO conv. generators must be aggregated manually
```
davide-f: True/True, is it related to the TODO? maybe a better explanation may be needed

davide-f:
Good job @Hazem-IEG ! I am reviewing the request, yet they are easy comments, moreover, thanks for noticing the additional misallocation. I have to test it.

After the corrections, merge the corrections by linter related to the standard python type.

davide-f:
Good job @Hazem-IEG ! I am reviewing the request, yet they are easy comments, moreover, thanks for noticing the additional misallocation. I have to test it.

After the corrections, merge the corrections by linter related to the standard python type.

scripts/base_network.py
hazemakhalek: It's updated, no?

scripts/build_bus_regions.py
hazemakhalek: I like the suggestion, and I think it will be useful in some use cases. I can separate it into 2 different options, but I think it can also be added at a later stage.

PR #163, opened 2021-11-23
CI integration

pz-max:
A problem is that package updates and contributions could sometimes lead to breaking the basic workflow. This can be discouraging for newbies and should be avoided. This PR will include a continuous integration (CI) test via GitHub Actions. It will run before each pull request is merged and as a result, breaking code will be detected much earlier.


scripts/osm_pbf_power_data_extractor.py
```diff
@@ -429,7 +429,8 @@ def output_csv_geojson(output_files, country_code, df_all_feature, columns_featu
     # Generate Files
 
     if df_all_feature.empty:
-        _logger.warning(f"All feature data frame empty for {feature}")
+        _logger.warning(f"Store empty Dataframe for {feature}.")
+        gdf_feature = []
```
pz-max, line 433: Ups this is also a small fix. Countries with empty data frames were causing an error. This code addition resolves that.

test/config.test1.yaml
```diff
@@ -0,0 +1,337 @@
...
+
+electricity:
+  voltages: [220., 300., 380.]
+  co2limit: 7.75e+7 # 0.05 * 3.1e9*0.5
+  co2base: 1.487e9
+  agg_p_nom_limits: data/agg_p_nom_minmax.csv
+
+  extendable_carriers:
+    Generator: []
+    StorageUnit: [] # battery, H2
```
pz-max, line 86: maybe worth checking the other features like generator, link etc. at a later stage

pz-max:
Oh I just realize, I need to add the download tutorial part first :)

euronion:
Let me know when you want me to have a look!

davide-f:
@pz-max the issue that occurs can be removed by removing lines 37-41 of snakemake; I have just pushed the commit now.
By doing that, we remove the conflicts between the two rules, though download_osm_data is not skipped.

pz-max:
Data was not correctly uploaded. Will redo it! Almost there

davide-f:
The last issue about snakemake should be solved in the latest commits, by using directory("data/raw/landcover") in line 40 of snakemake file in new-main.
Exactly, the line:
`"data/raw/landcover/world_protected_area/WDPA_WDOECM_Oct2021_Public_AF_shp_2/WDPA_WDOECM_Oct2021_Public_AF_shp-polygons.shp",`
should become
`directory("data/raw/landcover"),`

PR #169, opened 2021-12-08
clean cluster network and modify CI config

pz-max:


davide-f:
Good job, yesterday I didn't want to leave with the CI breaking...
I think however, that these issues appeared after the change in the default parameters used for the kmeans: too low tolerances and high iterations may not perform well for small countries and may have led to this issues.
Increasing the number of clusters now works, but as standard parameters, I think it would be better to restore the previous values.
Anyway, some testing should be performed on that.

@Hazem-IEG chose those values after some tests, in fact. What is your idea @Hazem-IEG ?

P.S. there is the restyling pull request open with few linter changes

pz-max:
Seems still to break. 😕 Think it's not a problem to leave the CI in this "broken" status.
One just needs to debug the CI config and find out what's happening to fix it permanently.
So far we have just assumptions of what's happening... It's maybe the cluster config OR a data issue.

I could do that debugging part tomorrow between 13-14:00. Let me know if someone wants to tackle that earlier

davide-f:
I've been investigating a bit the problem and pushed an update that works locally on my computer; fingers crossed.
Anyway, while doing it, I've noticed some weird behavior that is added in the script osm_built_network:
some nodes are shifted with respect to their actual position. I will investigate this issue tomorrow

PR #172, opened 2021-12-11
Distribute clusters according to load, population or gdp

pz-max:
**Context:**
The `distribute_cluster()` function distributes the number of "cluster nodes" each country receives. In PyPSA-Eur we distributed the cluster_nodes according to the "load". While this is ok for Europe and a rather homogenous demand/capita factor, in Africa distributing according to load lead to the case that the sub-Saharan area almost gets no clusters.

**PR:**
This PR gives the option to distribute the clusters according to "load", "pop"(population) and  "gdp" to provide alternatives to existing methods.

### DIFF
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -29,7 +29,9 @@ countries: ["NG", "NE"]
   #["ZA"]  # South Africa
 
 # if True clusters to GADM shapes, if False Voronoi cells will be clustered
-alternative_clustering: False
+cluster_options:
+  alternative_clustering: False
+  distribute_cluster: ['load'] # 'pop'
 
 # options for build_shapes
 build_shape_options:
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -31,7 +31,8 @@ countries: ["NG", "NE"]
 # if True clusters to GADM shapes, if False Voronoi cells will be clustered
 cluster_options:
   alternative_clustering: False
-  distribute_cluster: ['load'] # 'pop'
+  distribute_cluster: ['load'] # ['load'],['pop'] or ['gdp']
+  out_logging: True  # When true, logging is printed to console
 
 # options for build_shapes
 build_shape_options:
--- a/scripts/build_shapes.py
+++ b/scripts/build_shapes.py
@@ -142,7 +142,8 @@ def get_GADM_layer(country_list, layer_id, update=False, outlogging=False):
 
         # convert country name representation of the main country (GID_0 column)
         geodf_temp["GID_0"] = [
-            _three_2_two_digits_country(twoD_c) for twoD_c in geodf_temp["GID_0"]
+            _three_2_two_digits_country(twoD_c)
+            for twoD_c in geodf_temp["GID_0"]
         ]
 
         # create a subindex column that is useful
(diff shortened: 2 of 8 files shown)


pz-max:
@davide-f after the review, please feel free to merge the PR.

@Hazem-IEG, you might be also interested in these changes

config.default.yaml
```diff
@@ -29,7 +29,10 @@ countries: ["NG", "NE"]
   #["ZA"]  # South Africa
 
 # if True clusters to GADM shapes, if False Voronoi cells will be clustered
-alternative_clustering: False
+cluster_options:
+  alternative_clustering: False
+  distribute_cluster: ['load'] # ['load'],['pop'] or ['gdp']
```
davide-f, line 34: Why are you proposing a list notation? maybe we could use as codes simply load, pop or gdp, like:
distribute_cluster: load # load, pop or gdp

yaml interprets them as strings

pz-max:
To keep it more flexible for future integrations. It is easier to iterate/handle lists than strings.

Get Outlook for iOS<https://aka.ms/o0ukef>

scripts/build_shapes.py
```diff
@@ -640,25 +647,41 @@ def _process_func_pop(c_code):
...
-                        year=2020,
-                        update=False,
-                        out_logging=False,
-                        nprocesses=2,
-                        disable_progressbar=False
-    ):
-    """Function to add the population info for each country shape in the gadm dataset"""
+def add_population_data(
+    df_gadm,
+    country_codes,
```
davide-f, line 663: To answer your question, country_codes are needed to download the appropriate worldpop data by country, whereas in gdp only a file is considered, thus they are not needed.
We could get rid of the mandadory column, but the same information must be got in other ways. Options:
- by a predefined dataframe column name [e.g. country]
- by a column name specified as an optional parameter of the function [maybe preferred option as more general]

The above can be addressed also in other moments, not necessarily in this pull requested, or never if not considered useful

scripts/cluster_network.py
```diff
@@ -173,15 +175,52 @@ def weighting_for_country(n, x):
...
+        L = (n.loads_t.p_set.mean().groupby(n.loads.bus).sum().groupby(
+            [n.buses.country]).sum().pipe(normed))
+        assert len(L.index) == len(n.buses.country.unique()), (
+            "The following countries have no load: "
+            f"{list(set(L.index).symmetric_difference(set(n.buses.country.unique())))}"
+        )
+        distribution_factor = L
+
+    if distribution_cluster == ["pop"]:
+        df_pop = gpd.read_file(snakemake.input.fundamental_onshore_region)
```
davide-f: why fundamental_*?

pz-max:
Thought it might be good to have a preclustered/ fundamental shape containing all static data (pop,gdp, etc.). But yeah probably that needs some proper discussion.

We could also do the add_gdp/pop stuff at the post-clustered shapes. These were already inputs and maybe avoid the introduction of a fundamental_shape

Get Outlook for iOS<https://aka.ms/o0ukef>

scripts/cluster_network.py
pz-max: Renamed it now to raw_onshore_busregion

scripts/build_shapes.py
pz-max: Yeah makes sense. Agree, let's discuss that another time

PR #175, opened 2021-12-13
Add augmented line connections

pz-max:
Problem:
The African network is not properly meshed. However, one might be interested to see how a well connected African network might look like.

Solution:
This PR aims to take the clustered network and mesh it according to the k-augmentation algorithm (i.e. make sure that every bus has at least two connections). The `augmented_line_connection.py` should thereby host in future also alternative algorithm to mesh/connect the network. This can include Steiner-tree or minimum spanning tree solutions (see initial discussions [here](https://github.com/pypsa-meets-africa/pypsa-africa/discussions/160))

Examples of k-augementation:
Pointed out by @fneum:
- https://github.com/PyPSA/pypsa-eur-sec/blob/82096f92e883fa508631e34a250249e29ad45a60/scripts/prepare_sector_network.py#L1155-L1157
- https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.connectivity.edge_augmentation.k_edge_augmentation.html

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -276,6 +276,22 @@ rule cluster_network:
     script: "scripts/cluster_network.py"
 
 
+rule augmented_line_connection:
+    input:
+        network='networks/elec_s{simpl}_{clusters}.nc',
+        regions_onshore="resources/regions_onshore_elec_s{simpl}_{clusters}.geojson",
+        regions_offshore="resources/regions_offshore_elec_s{simpl}_{clusters}.geojson",
+    output:
+        network='networks/elec_s{simpl}_{clusters}.nc',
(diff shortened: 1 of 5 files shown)


pz-max:
@fneum, looks kind of cool and does it's job. Thanks Fabian! :+1: 

meshed_vs_unmeshed

pz-max:
FYI, a feature that might interest you (see description and picture above). The method works on any kind of network - GADM or Voronoi cell or whatever. The grade of minimal connections per node can be adjusted by changing the config.yaml

```python
augmented_line_connection:
  connectivity_upgrade: 3  # min. lines connection per node
  new_line_type: "HVDC"  # or "HVAC"
  min_expansion: 1  # [MW]
```

@davide-f 
@hazemakhalek

PR #189, opened 2021-12-20
Build ppl feature -- Fix replace and drop

ekatef:
The `pd.replace( )` and `pd.drop( )` default to `inplace=False` which was missed in the previous OSM-PPM cleaning patch.


pz-max:
Hi @ekatef ,
Any idea why the CI is failing for Ubuntu? We know it failed for Windows as in PyPSA-Eur but the Ubuntu error is new. The failed CI run seems not related to the changes but didn't happen before (@davide-f ).

Regarding the restyler, you could merge the restyler changes into your branch by typing:

```console
git remote add upstream https://github.com/pypsa-meets-africa/pypsa-africa.git
git fetch upstream pull/<Restyler PR number>/head
git merge --ff-only FETCH_HEAD
git push
```

ekatef:
Hi @pz-max,

Many thanks for looking into it. 

It seems that both Ubuntu and MacOS IC checks are failing with similar errors ("TimeoutError: [Errno 110] Connection timed out" and "TimeoutError: [Errno 60] Operation timed out") while checks on Windows were stopped earlier ("Error: The operation was canceled").

I suspect that may be related to some troubles with connection to https://biogeo.ucdavis.edu/data. At least, the error is triggered when trying to obtain a GADM gpkg file and when I try to open the site it's not responding.

pz-max:
> 

Usually, ubuntu is checked first than MacOS, Windows. When one of the CI's fails the other will stop too. That's why in our case windows and macos is cancelled. Let's rerun the CI and see if it works. Yeah maybe GADM is down...

pz-max:
Oh CI works now @ekatef :)
Can you just add the restyler stuff?
I will hang around the discord channel if you need help on this

ekatef:
@pz-max, many thanks for your support. 
That's very interesting to see how IC works in real-life conditions :)

PR #194, opened 2021-12-22
Add and fix hydro

pz-max:
The "draft PR" means that the PR is not ready for review. However, it shows that the feature is currently under development.
Saying that, I am working now on hydro.


pz-max:
The hydro inflow is zero for shapes that probably have powerplants. A fix is required. One needs to investigate if all shapefiles with Run of River powerplant have inflows. A busmap with the average inflow might be best to assess the situation.

https://github.com/pypsa-meets-africa/pypsa-africa/blob/4e1f8c650b92e52364cfab790791a2cef49d4d63/scripts/add_electricity.py#L451-L455

We have the feeling that there might be an issue with indexing. I.e. regions have a index = 90 and name = 92. So when associating inflows with shapes this might be a reason for issues.

PR #211, opened 2021-12-29
Add plotting and summary features

davide-f:
This branch enables the use of plot_network and make_summary features:
- plot_network: enables plotting the nominal power of the generation assets [p_nom attribute]
- make_summary: creates csv fles and useful statistics for the results

Other than plot_network and make_summary rules, the following rules have also been introduced:
- plot_all_p_nom: generates the plots for all countries and extensions in formats pdf and png
- make_country_summaries: automatically generates summaries for each country
- make_all_summaries: generates summary for the countries as a whole (the aggregation of the countries)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ networks/
 resources/
 scripts/temp
 scripts/test.py
+results/summaries/*
 
 # Untrack some Jupyter changes
 .ipynb_checkpoints
--- a/Snakefile
+++ b/Snakefile
@@ -31,6 +31,14 @@ rule solve_all_networks:
     input: expand("results/networks/elec_s{simpl}_{clusters}_ec_l{ll}_{opts}.nc", **config['scenario'])
 
 
+rule plot_all_p_nom_to_pdf:
+    input: expand("results/plots/elec_s{simpl}_{clusters}_ec_l{ll}_{opts}_p_nom.pdf", **config['scenario'])
+
+
+rule plot_all_p_nom_to_png:
+    input: expand("results/plots/elec_s{simpl}_{clusters}_ec_l{ll}_{opts}_p_nom.png", **config['scenario'])
+
+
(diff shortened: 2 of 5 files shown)


davide-f:
@pz-max any comments? otherwise I merge (first this then the restyler)
As a little comment, now we have rules make_all_summaries and make_all_country_summaries, one focusing on all countries as a whole and the latter that iterate on all countries; we may merge them, however, for the case NG+BJ it leads to errors due to missing values, this shall be adjusted.

pz-max:
Hi Davide,
If you merge the restyler to this PR then the restyle PR will be closed anyways (maybe less git history?)
You can do this as it is suggested at the Restyler PR (212) by simply typing this in the console of your current #211 PR
```console
git remote add upstream https://github.com/pypsa-meets-africa/pypsa-africa.git
git fetch upstream pull/<PR number of Restyler>/head                   
git merge --ff-only FETCH_HEAD
git push
```

pz-max:
> @pz-max any comments? otherwise I merge (first this then the restyler) As a little comment, now we have rules make_all_summaries and make_all_country_summaries, one focusing on all countries as a whole and the latter that iterate on all countries; we may merge them, however, for the case NG+BJ it leads to errors due to missing values, this shall be adjusted.

I know that some stuff was previously breaking with plot and summaries. Hope you tested it a bit :) 
Some of the PyPSA-Eur-Sec colleagues were also mentioning that the sector coupled plotting library is more updated. Maybe good to add and adjust some of their features that they work here as well. Do you want to create an issue?

davide-f:
Yes, obviously I tested it and on the NG+BJ plot worked and make_summary as well, with the exception of BJ. In fact, there is the need to improve it, however, the problems on BJ may be due to problems inherited by he workflow. However, when no data are available no error shall be generated, rather zeros shall be shown. I add an issue

Issue #182, opened 2021-12-16
Set world dictionary in osm_data_config only for osm and create a new one for managing regions


davide-f:

ekatef:
@davide-f Thank you for creating the issue!

The `world` dictionary structure in the osm_data_config.py currently corresponds to ISO, right?
If I understand it properly, I could continue checking how does`world` dictionary  corresponds to the Geofabrik data structure and create an additional  `geofk_word` intended for data access.

pz-max:
> @davide-f Thank you for creating the issue!
> 
> The `world` dictionary structure in the osm_data_config.py currently corresponds to ISO, right? If I understand it properly, I could continue checking how does`world` dictionary corresponds to the Geofabrik data structure and create an additional `geofk_word` intended for data access.

Yep you are right @ekatef :) The `geofk_world` would be highly appreciated

ekatef:
Great, thank you!

davide-f:
#236 solves this issue. Some refinings and checks are still needed.
Once @ekatef has finished the checks, this issue can be closed

davide-f:
Closed as #249 has been merged

PR #220, opened 2022-01-07
Add world dictionary for managing regions

ekatef:
Relates to #182 

Adds a dictionary to be used for extraction of OSM data.

##  Basic checks

- [x] Fix territories which are merged in OSM
- [x] Check and comment island countries
- [ ] Check completeness of the regions taken for functionality testing

## Functionality tests

Let's see how the workflow runs if OSM data are loaded for different regions of the world. The macro region shortcuts for testing are defined according to the config_osm_data.continent_regions

- [ ] Africa
    - [x] Western African region
    - [x] Central African region
    - [x] Eastern African region
    - [x] Southern African region
- [ ] Asia
    - [x]  Western Asia
    - [x]  Central Asia
    - [x]  Far East
    - [x]  South-Eastern Asia 
    - [x]  Southern Asia
    - [x]  Middle East
- [x] Australia-Oceania
- [ ] Europe
    - [x]  Scandinavian region
    - [x] Eastern region (except for "RU")
    - [x] Central region
    - [x] Balkan region
    - [x] Western region
    - [x] Southern region
- [ ] Russia
- [ ] North-America
- [ ] South-America
- [ ] Central-America

## Checklist

- [ ] Changes in configuration options are added in all of `config.default.yaml`, `config.tutorial.yaml`, and `test/config.test1.yaml`.
- [x] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes.


davide-f:
Great job that you are doing @ekatef!
I add few comments to add little suggestions in the code

scripts/config_osm_data.py
```diff
@@ -484,3 +484,241 @@
...
+        "AM": "armenia",
+        "AZ": "azerbaijan",
+        "BD": "bangladesh",
+        "BT": "bhutan",
+        # 'IO': 'british indian ocean territory', # Island
+        "KH": "cambodia",
+        "CN": "china",
+        # 'CX': 'christmas island', # Island
+        # 'CC': 'cocos (keeling) islands', # Island
+        "GCC":
```
davide-f: this is missing, just noticed

scripts/config_osm_data.py
```diff
@@ -484,3 +484,241 @@
     "SN": "SNGM",  # senegal -> senegal-gambia
     "GM": "SNGM",  # gambia -> senegal-gambia
 }
+
+# Geofabrik and iso norm deviate for the following countries
+
+# Western Sahara -> Morocco ("EH": "MA")
```
davide-f: Love to track all these changes, that's good. Moreover, a step beyond is modifying the dictionary "iso_to_geofk_dict " that has been built exactly with that purpose and I'll shortly explain why.
Let's see the dictionary:
``` python
iso_to_geofk_dict = {
    "EH": "MA",  # western sahara -> Morocco
    "SN": "SNGM",  # senegal -> senegal-gambia
    "GM": "SNGM",  # gambia -> senegal-gambia
}
```
The scope of this dictionary is to instruct the workflow to modify the list of countries that the dataextractor shall download when countries in the keys of the dictionary are found.
In particular, as you specified, data of Western Sahara (ISO) are stored in the Morocco country in geofabrik .
Similarly, the data of Senegal (SN) are stored in the country coded "SNGM" in geofabrik , as for the gambia case.
That is the reason why there is the dictionary "iso_to_geofk_dict ": wherever in the list of countries in the config one of "EH", "SN" or "GM" is found, then the workflow substitute such iso coding with the geofabrik one.

For example, in the case of Hong Kong, a new line should be added into the dictionary, like:
``` python
iso_to_geofk_dict = {
    "EH": "MA",  # western sahara -> Morocco
    "SN": "SNGM",  # senegal -> senegal-gambia
    "GM": "SNGM",  # gambia -> senegal-gambia
    "HK": "CN",  # Hong Kong -> China
}
```

If you could also implement these kind of changes would be awesome

scripts/config_osm_data.py
ekatef: Thank you, @davide-f! Actually, it's a good idea to check completeness of the whole geofabric-related dictionary. I'll add a test for it. As far, the `world_geofk` dictionary is tested only to ensure that the Geofabric urls are reachable.

By the way, could you please advice if there is any proper place in the project to store such testing scripts? The scripts folder doesn't look a right place for such purely supplementary kind of things...

scripts/config_osm_data.py
ekatef: Okay, the ISO codes are the main input which should be transformed to Geofabrik codes... I have nearly missed that concentrating on logic behind the Geofabrick :) Of course, I do!

@davide-f Many thanks for clarification and suggestions.

scripts/config_osm_data.py
davide-f: Sorry, GCC stands for "gulf cooperaton council" at the following line; I suggest to format it similarly to the other and put them in line.
Moreover, note that any geofk shortcut or agglomeration or alike may need to have the proper list of codes, each for iso coding in the iso_to_geofk dictionary.
Amazing job!

Regarding the script, instead of the script, may you use a notebook instead with some comments and description?
It could be put in the notebook folder and be useful for other people for better understanding geofk

scripts/config_osm_data.py
```diff
@@ -484,3 +484,241 @@
...
+        "SI": "slovenia",
+        "ES": "spain",
+        # 'SJ': 'svalbard-and-jan-mayen', # Islands
+        "SE": "sweden",
+        "CH": "switzerland",
+        "UA": "ukraine",
+        "GB": "great-britain",
+        "TR": "turkey",
+    },
+    "russia": {
```
davide-f: I see the problem with russia.
Russia is the only nation that is classified as a continent basically, thus the typical downloading format we use does not work.
https://github.com/pypsa-meets-africa/pypsa-africa/blob/8d6a3adc79ab37da00bdfb13ddb67247ea157ff6/scripts/download_osm_data.py#L80-L83

Probably, we may hard code a change here so that if the continent is "russia" and the requested country_name code is also russia then we use the link `f"https://download.geofabrik.de/{continent}-latest.osm.pbf"` instead
Actually, the above concept may be generalized to whenever continent==country_name but may need tests; the hard coding of russia solve that special case

We discussed with @pz-max and he seemed prone to this solution, he proposed the hard fix for russia. Personally, I would like a generalization but looking at geofk it is really the only case...

scripts/config_osm_data.py
ekatef: @davide-f Absolutely agree regarding geofk shortcuts. And love the idea of an explaining notebook. Thank you!

scripts/config_osm_data.py
ekatef: @davide-f, @pz-max many thanks for suggestions. Yes, exactly. The condition `continent ==country_name` in Geofabrik seems to hold for Russia only.

To make things more complicated, it could be more convenient having an opportunity to extract the data bunch related to a single district according to Geofabrik logic. E.g. defining `"russia"` as a `{continent}` and `"far-eastern-fed-district"` as a `{country}`. But hard fix `https://download.geofabrik.de/russia` looks much more elegant.

davide-f:
Hey @ekatef , what is the current status of this draft pull request in your opinion?
I would like to work on it to test the possible downloading of all countries by the end of the week, not sure if that's feasible.

May you create a personalized check list to verify what regions have yet to be verified (like Africa, Europe, South America, etc.)? This may help the future developments.

ekatef:
﻿Hey @davide-f, at the moment the part with urls is almost completed if we'll keep the original world dictionary untouched. The only major change I would like to add is to create a notebook showing urls testing.

Do you mean region verifying as general checking that download_osm_data works properly? I would be happy to add such a check-list with regions to be checked.

scripts/config_osm_data.py
davide-f: Ok! Hard fix is fine :)
Thank you, great job!

davide-f:
Yeah, since you were checking the different region I was wondering the current status.
In general, it is possible to add "task lists": using the checkbox on top of any writing area or you can use "- [ ]". Image below.
image
The output is like:
- [ ] Option 1
- [ ] Option 2

ekatef:
@davide-f Great, many thanks for supporting these experimenting activities. Absolutely agree that it makes sense to check in a systematic way how data loading works.

I have added a check list with the regions and would like to go through Asia and Russia first.

ekatef:
Proposed changes were incorporated as part of #236 .
@davide-f many thanks for making it all work!

davide-f:
@ekatef thank you for the great work, yet I don't think it was complete.
I just gave the structure but I didn't complete the validation that you were doing.
You may finish the validation of the regions to check that all countries are properly managed.
Most data seem to be downloaded but there may be need to check the conversion dictionary between iso and geofk. I completely rely on your good work.

If you like you can keep this pull request alive and continue the validation process you were doing. If you fell like the job is done, feel free to close the issue again

ekatef:
@davide-f I was just not sure if it would not be better to create a new PR for checking/testing/tutorial purposes. If you think this one is not too messy, I'll be happy to continue working on it :)

davide-f:
I like that, if you feel like you don't lose anything in starting from scratch, then probably it is a good idea.
Anyway, copy and paste the list and description of this pull request

ekatef:
A cleaned version of this RP is reopened as #249

PR #275, opened 2022-02-17
Add load changes - dynamic file choice based on user parameters

pz-max:
# Closes # (if applicable).

## Checklist

- [x] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes.


pz-max:
based on PR https://github.com/pypsa-meets-africa/pypsa-africa/pull/218

config.default.yaml
```diff
@@ -66,10 +66,11 @@ base_network:
...
 
 load_options:
-  weather_year: 2014  # Not active, but load scenarios available with different weather year (different renewable potentials)
-  prediction_year: 2030  # Not active, but load scenarios available with different prediction year (GDP, population)
+  ssp: "ssp2-2.6" # shared socio-economic pathway (GDP and population growth) scenario to consider
+  weather_year: 2013  # Load scenarios available with different weather year (different renewable potentials)
+  prediction_year: 2030  # Load scenarios available with different prediction year (GDP, population)
   scale: 1  # scales all load time-series, i.e. 2 = doubles load
-
+  region_load: "Africa" # Load demand scenarios based on a specified reference region
```
davide-f: In general, as discussed, I don't like this fixed "region_load"; this should be automatized depending on the specific country list.
Basically, we need (another) configuration file where we state for every name file (e.g. Africa, Asia etc) what are the countries that are contained and the workflow load the file according to that.
I think this PR is fine as is once the workflow works; however, most of these loading features may be incorporated into a specific script (and rule) named build_load, for example, whose goal is to merge different files and data to produce an nc file containing the demand for every bus of the workflow.
This nc file would be now an input for add_electricity.

That proposed change may help the workflow also giving users a clear understanding about what is the input file they should change if they want to bypass our procedure, for example.

Any thoughts?

pz-max: Agree. To enable linkers the add_electrcity script should just upload a .nc or csv file.

The other options for the load will end up in a `build_load` script. The "region_load" idea is excellent! I would probably orient at the ISO code instead of using something else. Soon we should have the GEGIS power and build it also according to ISO naming.

Regarding the current code, I will just debug it and merge the PR. We need to write more clearly that the IDE is essential (i.e. VS code) and should link to the hackathon to get introduced to debugging (most important).

pz-max:
@davide-f implemented your suggestion, please have a look. Just some weird CI issues remaining... Any idea how to deal with them?

pz-max:
@giacfalk FYI, just about to finalize your contribution. Thanks for adding this feature :+1:

scripts/_helpers.py
```diff
@@ -391,6 +391,38 @@ def _get_country(target, **keys):
         return np.nan
 
 
+def _getContinent(code):
```
davide-f: Very good that the loaded files depend upon the country_list, that's a very good feature!

As a comment, however, I think we may need a different dictionary, we may not use the world_iso as most likely the naming of the countries would be different: I am not 100% sure that current GEGIS is using the same ISO standard to store data in the different nc files.

It could be helpful to generalize the function like _getContinent(code, regions_dict=world_iso) to give the option to use a generic dictionary to be used. This may be in handy, and in download_osm_data something similar may be already in place.

Anyway, the fine-tuning could be done in a subsequent PR focusing on the implementation of build_loads.

scripts/download_osm_data.py
```diff
@@ -55,8 +55,10 @@
 # logger.setLevel(logging.WARNING)
 
 
-# get continent according to ISO conventions
 def getContinentCountry(code):
```
davide-f, line 58: I see some duplication of the _getContinent here; with the generalization stated in the previous section, this function may be moved into the helpers (basically the one you added) but using world_geo as optional argument

scripts/_helpers.py
euronion: **Could be addressed**

* Most function names do not use CamelCase naming convention, but follow PEP8 recommendations.
* Leading `_` underscores by convention indicate internal functions. This is a helper function to be used outside the script, i.e. the leading `_` should be removed.

For more info, see: https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

```suggestion
def get_continent(code):
```

scripts/_helpers.py
```diff
@@ -391,6 +391,38 @@ def _get_country(target, **keys):
...
+        List of continent names
+
+    Example
+    -------
+    from helpers import _getContinent
+    code = ["DE", "GB", "NG", "ZA"]
+    _getContinent(code)
+    >>> ["africa", "europe"]
+    """
+    from config_osm_data import world_iso
```
euronion, line 417: **Needs attention**

The repository for creating synthetic demand (https://github.com/euronion/synde/) does not follow the continent <-> country mapping and continent naming conventions from `config_osm_data`. As such this function will fail in some cases, e.g.:

* "Australia" in the demand data is part of the region `Oceania`, but the function would return `australia` as continent.
* "Mexico" is part of the region called "NorthAmerica", while the OSM data would return "north_america". Since this string is used to construct a file path, this does not match.

I recommend to use a dedicated function specifically for mapping between countries and regions (in the file names) from `synde`. The mapping can be accessed and downloaded from here https://github.com/euronion/synde/blob/master/config/regions.yaml or we can put it into the databundle the with demand data. @pz-max @davide-f can assist with the databundle?

scripts/_helpers.py
```diff
@@ -391,6 +391,38 @@ def _get_country(target, **keys):
...
+
+    continent_list = []
+    code_set = set(code)
+    for continent in world_iso:
+        single_continent_set = set(world_iso[continent])
+        if code_set.intersection(single_continent_set):
+            continent_list.append(continent)
+    return continent_list
```
euronion, line 425: **Should be addressed**

I recommend you throw an exception in case the country is not in the list of continents, i.e. in case no match is found.

This code could work (haven't tested it)

```suggestion
        matching_countries = code_set.intersection(single_continent_set)
        if matching_countries:
            continent_list.append(continent)
            code_set = code_set - matching_countries
    
    if code_set:
        raise KeyError(f"Unable to map countries '{code_set}' to a demand data region.")
    return continent_list
```

scripts/add_electricity.py
```diff
@@ -246,14 +270,26 @@ def attach_load(n, regions, load, admin_shapes, countries, scale):
...
+    for continent in region_load:
+        load_path = os.path.join(
+            cwd_path,
+            "resources",
+            str(ssp),
+            str(prediction_year),
+            "era5_" + str(weather_year),
+            str(continent).capitalize() + ".nc",
+        )
+        load_paths.append(load_path)
```
euronion: **Should be addressed**

I highly recommend that this happens outside the script and in the `Snakefile`; file dependencies should all be made explicit there.

There all the information to construct the file paths are already available (via the `config`), so you could probably use this function in a very similar manner in the `Snakefile`.

The script would then simply get the path or paths via `snakemake.input["loads"]` (like it did before) and then read through `xr.open_dataset(snakemake.input["loads"]`

euronion (changes requested):
Hi there!

Thanks for your PR. @pz-max asked me to review it as I was also responsible for creating the demand datasets in the first place ;-).

There are a few things that **need** attention, some that **should be given** attention, and some that **could be given** attention. See the specific comments.

pz-max:
Great thanks @euronion for the concise feedback. We will address all of the points very soon 👍🏽

scripts/_helpers.py
pz-max: Looking at the data of [synde](https://github.com/euronion/synde/blob/master/config/regions.yaml), [our_world_iso](https://github.com/pypsa-meets-africa/pypsa-africa/blob/44b6236b641640b6502632df60773b30e4aa2f74/scripts/config_osm_data.py), and [wikipedia](https://en.wikipedia.org/wiki/List_of_sovereign_states_and_dependent_territories_by_continent_(data_file)), it seems our world_iso is not complete... We should update our world_iso to be complete and also adapt the correct continent naming

scripts/_helpers.py
pz-max: Thanks, I removed the underscore in the helpers

scripts/_helpers.py
pz-max: Quick fix is to change our world_iso continent naming according to synde. Will be done in the next commit

scripts/add_electricity.py
pz-max: Done with https://github.com/pypsa-meets-africa/pypsa-africa/pull/275/commits/f593c9db5de8b9cf9d2627bc1d370a1b713eae8a

scripts/_helpers.py
pz-max: I am not sure about this. Which country list do you want to compare?
If the country is not listed it already leads to an error such as load for "X" not found.
I think the exception is maybe not necessary?

pz-max:
@euronion just adressed all requested changes whcih are passing the tests. Let me know if I am missing something important or if you have any further recommendations. :)

pz-max:
> Is this function redundant now?

I guess no. It's just required for the Snakefile. I see this function as well as the demand stuff moving to a new .py script in future. Let's talk about that maybe at the next workflow meeting?

scripts/download_osm_data.py
davide-f: Not a duplication

PR #334, opened 2022-05-10
add augmented line connection notebook

pz-max:


pz-max:
Hey @davide-f , does the contributor-readme action lead in your fork also to issues occasionally (unnecessary commit/ merge conflict etc.)? Maybe a fix is given in the video but I need to check it better out: https://www.youtube.com/watch?v=Ww5gjAm-2GY .. Having a look into the documentation beforehand is probably also not a bad idea :+1:

PR #348, opened 2022-05-27
Implement HVDC lines into the model

### LINKED ISSUE #286: Verify HVDC lines
## Verify and validate HVDC lines

HVDC lines should be addressed and verified along the code.
For simplicity assumptions, lines have been assumed to be mostly AC, but this assumption can no longer be considered.

The code shall be revised to remove assumptions of AC lines, especially when we want to go global.

Such main assumptions are made in the [build_osm_network script](https://github.com/pypsa-meets-africa/pypsa-africa/blob/main/scripts/build_osm_network.py); generally the "DC" status is set to false by default but that shall be revised
DC nodes shall be distinguished from AC nodes.

In particular, in the function merge_stations_lines_by_station_id_and_voltage, AC nodes shall be distinguished from DC nodes
https://github.com/pypsa-meets-africa/pypsa-africa/blob/dff3930706a967e842e9e7427f7c01b0532bcf0d/scripts/build_osm_network.py#L422

ekatef:
The PR is intended to resolve #286 using the [following approach ](https://github.com/pypsa-meets-africa/pypsa-africa/issues/286#issuecomment-1137434530)proposed by @davide-f (thank you for the brilliant tutorial!)

## TODOs
- [x] Check if `merge_stations_lines_by_station_id_and_voltage( )` works properly for links (`build_osm_network` script)
- [x] Correct a value of the `"dc"` column of the buses csv (`build_osm_network` script)
- [x] Include converters into the model (`base_network` script)
- [x] Check assumptions on `p_nom_max` (`base_network` script)
- [x] Fix an issue with an empty links file

## Checklist

- [x] I tested my contribution locally and it seems to work fine:
     - [x] on regional scale
         - [x] with DC lines (CD, NA)
         - [x] without DC lines  (NG, CD on main)
     - [x] on the whole model


scripts/clean_osm_data.py
```diff
@@ -192,6 +194,28 @@ def filter_voltage(df, threshold_voltage=35000):
...
+    df.loc[:, "tag_frequency"] = df["tag_frequency"].astype(int)
+
+    # TODO Fix SettingWithCopyWarning, details via https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
+    # keep only lines with zero frequency as the most useful indication of DC lines
+    df = df[df.tag_frequency == 0]
+
+    return df
+
+
+def filter_ac(df):
```
davide-f: @ekatef Great job! I'd recommend to use "df.tag_frequency != 0" to find ac lines, which is more general (e.g. US).
Anyway, before you define the tag "dc" (line 101), may that column be used instead of checking again?

ekatef: @davide-f, thank you!
Yes, I have also thought that 50Hz default used for AC could be generalised. Not sure however if  `df.tag_frequency != 0` would be a good AC-indicator as frequency for AC lines is missed in OSM data rather often. Probably, it would make sense to extract the AC-frequency values for each country to be used instead 50Hz across the given country?

As for avoiding duplication, have tried to improve :)

davide-f: AC lines at 50Hz or 60Hz or other values are still AC lines and at first approximation, we can consider them as they were exactly the same (in practice there are obviously differences but negligible for our needs).
So where tag_frequency is specified and it is 0, I'd say they are DC lines, all the rest are AC regardless of the value of tag_frequency (nan, 50, 60, etc).

ekatef: That absolutely makes sense. 
I have changed filtering here and added TODOs to the points where 50Hz values are used.

davide-f:
@ekatef thank you for your updates :)
As a first comment, regarding problems related to the specific PR, it is best to discuss them here in the PR. This eases the reviewing process and keeping tracks of the updates.

I think that we need some clarity. In pypsa-eur, as you have seen, links do have a dataframe structure that is different from the lines; see for example the [links.csv](https://github.com/PyPSA/pypsa-eur/blob/master/data/entsoegridkit/links.csv) that is the input eg_links of base_network in pypsa-eur. So to use the pypsa-eur functions for links, that structure should be reproduced.

Currently, we are doing something in the between, as the links.csv seems to keep the original "lines" format, which may lead to conflicts. That error that you are experiencing is suggesting that the configuration of the link is not well executed for some reason.

For the moment, given what you mentioned, I would suggest you to focus on build_osm_network first and base_network later.
If it may be easier for you, as first step, we could introduce model everything as lines, which should be easier and would mean only changes into build_osm_network and limited into base_network as follows:

- with respect to what you have already implemented, I think that the only missing piece is the concept of converters, which has been currently proposed with the get_converters function that is alike the get_transformers, which models the converters as lines. Those fake converters can be added to the links.csv. That is needed to have a complete connected network and avoid problems into simplify_network.
- the build_osm_network then shall have as outputs the usual outputs, including lines.csv + converters.csv, where lines.csv may include (a) AC lines, (b) DC lines and (c) fake transformers, whereas converters.csv contains all fake converters (modelled as dc lines similarly to the current trasformer implementation and the draft get_converters function)
- with respect to what you have already implemented, I think that the only missing piece is the concept of converters, which has been currently proposed with the get_converters function that is alike the get_transformers, which models the converters as lines. Those fake converters can be added to the converters.csv
- In base network, then the lines and converters csvs could be read and merged. As a hard fix, we then may need to hardfix all buses and lines to "ac" to avoid issues with carrier types.

Next steps and PRs may involve to implement converters and transformers as separate converters.
Personally, I'd love to keep DC lines modelled as lines, instead of being links, but I think it is safe to avoid that issue now.

What do you think?

scripts/build_osm_network.py
```diff
@@ -753,7 +882,7 @@ def built_network(inputs, outputs, geo_crs, distance_crs):
         logger.info(
             f"Stage 4/5: Aggregate close substations: enabled with tolerance {tol} m"
         )
-        lines, buses = merge_stations_lines_by_station_id_and_voltage(
+        lines, links, buses = merge_stations_lines_by_station_id_and_voltage(
```
davide-f: There is no need to have the links processing and output into merge_stations_lines_by_station_id_and_voltage, I'd suggest to remove that

scripts/build_osm_network.py
```diff
@@ -395,11 +425,89 @@ def get_transformers(buses, lines):
...
+        by=["station_id"]
+    ):
+
+        # note: by construction there cannot be more that two buses with the same station_id and same voltage
+        n_voltages = len(g_value)
+
+        # Assuming that a converter stations should have both AC and DC parts
+        if g_value["dc"].any() & ~g_value["dc"].all():
+
+            for id in range(0, n_voltages - 1):
```
davide-f: With this procedure, you add a covnerter among multiple buses in the same substation, but we should add only a single converter for every dc bus between the two nodes with the closest voltage with respect to the DC voltage

ekatef:
@davide-f, thanks you so much for the detailed suggestions. And sorry for making some mess with the PR :)

Absolutely agree with you that it's important to obtain a working PR version with minimal functionality. That's always a relief when the code begins to work even with hard fixes. I'll try to implement the strategy you have drafted and hope that it'll not take long. Many thanks for you support!

scripts/build_osm_network.py
ekatef: Yes, I see: `merge_stations_lines_by_station_id_and_voltage( )` is executed only under a condition and may by bypassed. Moved a lines/links split into a higher-level function.

scripts/build_osm_network.py
ekatef: Should be fixed now along with a similar correction for transformers. A direct csvs check has been again very helpful :)

ekatef:
Hey @davide-f, the next iteration of this PR is completed. I'll be grateful for your review :)

Snakefile
```diff
@@ -157,6 +157,7 @@ rule clean_osm_data:
         generators="data/clean/africa_all_generators.geojson",
         generators_csv="data/clean/africa_all_generators.csv",
         lines="data/clean/africa_all_lines.geojson",
+        links="data/clean/africa_all_links.geojson",
```
davide-f, line 160: Thinking again, maybe we can use the keyword "converters" instead of links, this complies also with pypsa-eur, given the intended content of these outputs

doc/release_notes.rst
```diff
@@ -13,6 +13,8 @@ Upcoming Release
 
 **New Features and major Changes**
 
+* Identify DC lines but temporary transform them back into AC `PR #348 <https://github.com/pypsa-meets-africa/pypsa-africa/pull/348>`
```
davide-f: Missing the ending "__" to make that an url

scripts/_helpers.py
```diff
@@ -595,7 +595,11 @@ def read_csv_nafix(file, **kwargs):
 def to_csv_nafix(obj, path, **kwargs):
     if "na_rep" in kwargs:
         del kwargs["na_rep"]
-    return obj.to_csv(path, **kwargs, na_rep=NA_VALUE)
+    if len(obj) > 0:
```
davide-f: better to use the property obj.empty probably, good check!

scripts/base_network.py
```diff
@@ -121,7 +121,9 @@ def _load_buses_from_osm():
 
     buses = buses.loc[:, ~buses.columns.str.contains("^Unnamed")]
     buses["v_nom"] /= 1e3
-    buses["carrier"] = buses.pop("dc").map({True: "DC", False: "AC"})
+    # All carriers are temporary set to "AC" to avoid the mixed-carries problem
+    # buses["carrier"] = buses.pop("dc").map({True: "DC", False: "AC"})
+    buses["carrier"] = "AC"
```
davide-f, line 127: Add a TODO: remove to enable HVDC buses and lines, once probably debugged. It is good to have TODOs when known problems are known

scripts/base_network.py
```diff
@@ -317,8 +378,16 @@ def _rebase_voltage_to_config(component):
 
 def base_network():
     buses = _load_buses_from_osm().reset_index()
+
     lines = _load_lines_from_osm(buses)
+    lines_dc = _load_links_from_osm(buses)
+
     lines = _set_electrical_parameters_lines(lines)
```
davide-f: Should this electrical parameters be applied on lines_ac_dc? Rather than removing this line I would comment it and duplicated after adaptation below

scripts/build_osm_network.py
```diff
@@ -726,6 +852,14 @@ def built_network(inputs, outputs, geo_crs, distance_crs):
 
     substations = gpd.read_file(inputs["substations"])
     lines = gpd.read_file(inputs["lines"])
+    links = read_geojson(inputs["links"])
```
davide-f: I'd use the "converters" word instead of "links" also in the code, where they are intended to be converters

scripts/build_osm_network.py
```diff
@@ -639,6 +761,10 @@ def fix_overpassing_lines(lines, buses, distance_crs, tol=1):
         below which the line will be splitted
     """
 
+    # in case of @lines corresponding to links it may be an empty data frame
+    if len(lines) == 0:
```
davide-f: maybe lines.empty is more appropriate?

davide-f (changes requested):
@ekatef thanks for the contribution.
I added few minor recommendations.

Before merging, I think it is appropriate if you could also share the plots of the comparison between the networks (raw osm data, clean osm_data and simplify network) with the current version for countries where AC/DC lines.
After clean osm, it would be nice to differentiate between links and lines and see what happens.
A zoom may be in handy close to a station to check how components are connected.
What do you think? Those plots could be done using the already available validation plots or also using the code you have already used for the past images

doc/release_notes.rst
ekatef: Corrected. Thanks! :)

scripts/_helpers.py
ekatef: Agree that it would be great to have a more elegant emptiness check. But I was not able to found a way. The `empty` method belongs to the pandas dataframe and obj.empty throws an error if applied to a list while it's convenient to use an empty list for initialisation... 

According to a (quite hot) discussion on SO, an implicit boolean conversion  `if obj:` is recommended as the most pytonic way to test that an `obj` is non-empty. But that would throw an error if applied for a dataframe.

Or would it be probably a good idea to keep in mind that the `obj` should be a data frame when applying this function? What do you think?

scripts/base_network.py
ekatef: Yes, that's absolutely makes sense. Done.

scripts/base_network.py
ekatef: Thank you! Corrected.

scripts/build_osm_network.py
ekatef: I think it's definitely make sense to replace links with converters in the outputs of `base_network` and inputs of `build_osm_network`. 

It seems to be a little misleading to have the `links` in `clean_osm_data` as well, while we are going to model them as lines. But it feels a bit too early to use there a `converters` name as the converters will introduced only further (in `build_osm_network`). Probably, `lines_dc` would be more appropriate?

scripts/build_osm_network.py
ekatef: I think we can safely remove this check as `lines` here is a merged ac/dc dataframe which shouldn't be empty.

ekatef:
@davide-f, thanks for checking!
Revised and fixed.

I love your idea regarding the visual checks. Actually, a combination of a static big-picture plot and hvplot zooms gives excellent options to explore the grid structure. I have been playing with such visualisations which was quite funny :) Then I'll repeat these tests on the the revised code and share the results. Do you think it's appropriate to attach them to a PR or would it be better to create a notebook for that?

scripts/_helpers.py
davide-f: If obj is an array somewhere, then it is fine as it is right now.
I expected we all used dataframes, could you highlight when that is not the case?

scripts/build_osm_network.py
davide-f: I agree, in fact, I'd keep in lines all lines (AC and DC), while using the converters.csv for only the converters, be them fake, like now, or not (in the future).
What do you think?

scripts/build_osm_network.py
davide-f: I agree, this check could be removed

davide-f:
> @davide-f, thanks for checking! Revised and fixed.
> 
> I love your idea regarding the visual checks. Actually, a combination of a static big-picture plot and hvplot zooms gives excellent options to explore the grid structure. I have been playing with such visualisations which was quite funny :) Then I'll repeat these tests on the the revised code and share the results. Do you think it's appropriate to attach them to a PR or would it be better to create a notebook for that?

I crosschecked the comments again.
Regarding the images, you can post them here

scripts/_helpers.py
ekatef: Checked and found that there is always an explicit transformation to a dataframe befor to output even if the data structure was initialised as a list in the beginning. Have replaced the length condition with using the `empty` method and a bit modified an argument name to make more clear that the function is intended to be used on dataframes.

scripts/build_osm_network.py
ekatef: I absolutely agree that it's better to have a clean data structure. Removed links both from snakemake inputs and the code.

ekatef:
The PR implies some changes in  the network topology which makes it necessary to compare the resulted network  with ones from the current main branch where DC lines are treated as AC. 

Results of the graphical topology check on CD look as follows.

# on raw data

Both approaches give the identical results 
_Blue = AC, red = DC, thickness is set according to the voltage_

africa_all_lines_cables_raw_freq

# on clean data
_Blue = AC, red = DC, thickness is set according to the voltage_

## original approach

africa_all_lines_clean_from_main

All DC lines were transformed to AC ones.

## tested aproach

africa_all_lines_clean

# on `build_network`-csv data

That's a key point where OSM data are being prepared to be further used as inputs of the PyPSA network.

_Lines are in blue, buses are in red_

A general outline of the grid is looks the same for both approaches

general picture

A closer look on the fragments reveals some differences (results of the original approach from the left, of the tested one -- on the right)

## fragment **_a_**

fragment_a

## fragment **_b_**

fragment_b

## fragment **_c_**

fragment_c

# simplified network

## original approach

elec_s nc_from_main

## tested approach

elec_s nc

# final network

## original approach

elec_s_6_ec_lcopt_Co2L-3H nc_from_main

## tested approach

elec_s_6_ec_lcopt_Co2L-3H nc

ekatef:
@davide-f, could you please have a look? I think most of the issues should be fixed now. Although, I have broken the CI (sorry for that!) but it seems to be related to some minor troubles in access to `https://data.open-power-system-data.org`.

scripts/_helpers.py
davide-f: Perfect, closing this

davide-f:
> @davide-f, could you please have a look? I think most of the issues should be fixed now. Although, I have broken the CI (sorry for that!) but it seems to be related to some minor troubles in access to `https://data.open-power-system-data.org`.

Don't be sorry for the CI, it happens often, the CI works is not broken with that error :) super!

To me, the images look very fine; I think we are super close to merge this PR.

Few questions:
- in the fragments, what are the red lines? Are they the fake converters?
- Still in the fragments, it seems like there is a weird additional bus that pops up but it is not connected to anything. How do you justify that?

ekatef:
Thank you, @davide-f :)

> * in the fragments, what are the red lines? Are they the fake converters?

Yes, the converters and buses are red while both AC and DC lines are blue. I'll work to improve the colour scheme for further use as hvplot seems to provide fantastic options for grid visualisations.

> * Still in the fragments, it seems like there is a weird additional bus that pops up but it is not connected to anything. How do you justify that?

I think that's a switching station at Kananga city and the bus should be connected to a distribution network. There are a few similar structures along the Inga–Shaba line.

davide-f:
> Yes, the converters and buses are red while both AC and DC lines are blue. I'll work to improve the colour scheme for further use as hvplot seems to provide fantastic options for grid visualisations.

No need, I was just asking to crosschecking, perfect! no need to have more images :)

> I think that's a switching station at Kananga city and the bus should be connected to a distribution network. There are a few similar structures along the Inga–Shaba line.

Ok! My thought was why in the other approach there was no such a bus, but from the point of view of the network the network seems to keep the connections. Can you confirm that?
Then I think this PR is ready, well done!! :)

ekatef:
> Ok! My thought was why in the other approach there was no such a bus, but from the point of view of the network the network seems to keep the connections. Can you confirm that? Then I think this PR is ready, well done!! :)

In the original approach there is a bus also (close to the center of this X-shape) but it's poorly visible on the picture... Have checked connectivity manually and not found any problems. It seems that PR is really working. Thank you so much for your support :)

davide-f:
@ekatef Great work! This PR has been merged.
As next steps, would you feel like to implement converters as Link models? This is the last major task left before v0.1 :)

ekatef:
@davide-f, very glad to know that it works! It was an amazing problem to solve :)
Yes, I'ld be happy to implement a proper converter model. I have seen that you have drafted the strategy in #385?

davide-f:
> @davide-f, very glad to know that it works! It was an amazing problem to solve :) Yes, I'ld be happy to implement a proper converter model. I have seen that you have drafted the strategy in #385?

Yes, I'll add some comments about it :)

ekatef:
> Yes, I'll add some comments about it :)

Wonderful, thank you! Now it looks for me absolutely doable :)

PR #356, opened 2022-05-30
Harmonize CRS in the workflow

### LINKED ISSUE #297: Harmonize Coordinate Reference Systems (CRS) and set them as configuration options
## Checklist

- [x] I am using the current [`main`](https://github.com/pypsa-meets-africa/pypsa-africa/tree/main) branch or the latest [release](https://github.com/pypsa-meets-africa/pypsa-africa/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-africa` environment](https://github.com/pypsa-meets-africa/pypsa-africa/blob/main/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

Across all scripts, different Coordinate Reference Systems (CRS) have been used and there is confusion; they should be harmonized with criterion.
In particular, we need CRS for GIS analysis. Since we are interested into distances, EPSG reference type seems the most appropriate, [reference](https://docs.qgis.org/2.8/en/docs/gentle_gis_introduction/coordinate_reference_systems.html) and [reference 2](https://www.nceas.ucsb.edu/sites/default/files/2020-04/OverviewCoordinateReferenceSystems.pdf).
In particular, when it comes to CRS valid worldwide, EPSR:4326 may be the recommended and we are mainly using it, however it is in cylindrical reference (curved surface), thus it is not suitable when we need to measure distances in km.
For metric distances, we can use EPSG:3857, which seems to be the most appropriate.

Across the code, there are also different CRS that have been used and they shall be cleansed (refuse from PyPSA-Eur).
Moreover, I would like to have them as options, such as crs (EPSG:4326) and crs_metric (EPSG:3857) that can be customized.
Note that this does not mean to blindly substitute the optional value everywhere in the code, but where it is needed: when we set a crs on a loaded geodataframe from an external source (e.g. gadm) the crs may be specific to that external resource.
We shall specify the crs for our processed data.
That's why I some crs may be still hard coded, but it shall be discussed.

davide-f:
# Closes #297 

This PR aims at harmonizing the CRSs used in the workflow as follows:
- Avoid having "magic" epsg specifications along the text: constant values are used in the scripts according to the file that are downloaded
- In the config files default and metric crs is proposed

## Checklist

### DIFF
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -37,8 +37,9 @@ enable:
 
 # definition of the Coordinate Reference Systems
 crs:
-  default_crs: EPSG:4326  # general geographic projection, not used for metric measures. "EPSG:4326" is the standard used by OSM and google maps
-  metric_crs: EPSG:3857  # projection for metric measures only. Possible recommended values are "EPSG:3857" (used by OSM and Google Maps) or also Global Mollweide "ESRI:54009"
+  geo_crs: EPSG:4326  # general geographic projection, not used for metric measures. "EPSG:4326" is the standard used by OSM and google maps
+  distance_crs: EPSG:3857  # projection for distance measurements only. Possible recommended values are "EPSG:3857" (used by OSM and Google Maps)
+  area_crs: ESRI:54009  # projection for area measurements only. Possible recommended values are Global Mollweide "ESRI:54009"
 
 # download_osm_data_nprocesses: 10  # (optional) number of threads used to download osm data
 
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -24,7 +24,7 @@ summary_dir: results
 snapshots:
   start: "2013-01-01"
   end: "2014-01-01"
-  closed: "left" # end is not inclusive
+  inclusive: "left" # end is not inclusive
 
 enable:
   retrieve_databundle: true  #  Recommended 'true', for the first run. Otherwise data might be missing.
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -35,6 +35,11 @@ enable:
   # requires cds API key https://cds.climate.copernicus.eu/api-how-to
   # More information https://atlite.readthedocs.io/en/latest/introduction.html#datasets
 
+# definition of the Coordinate Reference Systems
+crs:
+  default_crs: EPSG:4326  # general geographic projection, not used for metric measures
+  metric_crs: EPSG:3085  # projection for metric measures only
+
 # download_osm_data_nprocesses: 10  # (optional) number of threads used to download osm data
 
 augmented_line_connection:
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -37,8 +37,8 @@ enable:
 
 # definition of the Coordinate Reference Systems
 crs:
-  default_crs: EPSG:4326  # general geographic projection, not used for metric measures
-  metric_crs: EPSG:3085  # projection for metric measures only
+  default_crs: EPSG:4326  # general geographic projection, not used for metric measures. "EPSG:4326" is the standard used by OSM and google maps
+  metric_crs: EPSG:3857  # projection for metric measures only. Possible recommended values are "EPSG:3857" (used by OSM and Google Maps) or also Global Mollweide "ESRI:54009"
 
 # download_osm_data_nprocesses: 10  # (optional) number of threads used to download osm data
 
--- a/config.tutorial.yaml
+++ b/config.tutorial.yaml
@@ -93,8 +93,9 @@ enable:
 
 # definition of the Coordinate Reference Systems
 crs:
-  default_crs: EPSG:4326  # general geographic projection, not used for metric measures. "EPSG:4326" is the standard used by OSM and google maps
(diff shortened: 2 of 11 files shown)


config.tutorial.yaml
```diff
@@ -91,6 +91,11 @@ enable:
   build_cutout: false
   build_natura_raster: true  # If True, then build_natura_raster can be run
 
+# definition of the Coordinate Reference Systems
+crs:
+  default_crs: EPSG:4326  # general geographic projection, not used for metric measures
+  metric_crs: EPSG:3085  # projection for metric measures only
```
pz-max: REALLY COOL and well needed! One note:

There are projection that preserve angles, preserve areas and distances. Usually, there is not a one solution fits all case: https://learn.arcgis.com/en/projects/choose-the-right-projection/

Good for **preserving and calculating angles**:
- https://epsg.io/4326

Suitable for **preserving and calculating distances**:
- https://epsg.io/3857 aka. wgs 84 is the OpenStreetMap and Google Maps standard

Ok, for **preserving and caluclating areas**:
- we can go for https://epsg.io/3085 , but it's scope is for AMERICA. Maybe good

pz-max: So maybe we should define this three options in the config.yaml?

davide-f: Yeah, the 3085 was supposed to be 3857! A refuse, thanks!
Anyway, I wanted to add also some suggestions of alternative values

davide-f: > So maybe we should define this three options in the config.yaml?

The 3085 should not be used unless the user focuses on America only.
The default values shall have global scope

scripts/build_renewable_profiles.py
```diff
@@ -340,7 +345,7 @@
         else:
             availability = cutout.availabilitymatrix(regions, excluder, **kwargs)
 
-        area = cutout.grid.to_crs(3035).area / 1e6
+        area = cutout.grid.to_crs(metric_crs).area / 1e6
```
pz-max: Atlite uses for areas 3035.
 https://github.com/PyPSA/atlite/blob/de20027a477b9ae79825fc2b9c5073a6f5a5f376/atlite/gis.py#L191-L206

scripts/build_renewable_profiles.py
```diff
@@ -292,7 +297,7 @@
 
         capacity_per_sqkm = config["capacity_per_sqkm"]
 
-        excluder = atlite.ExclusionContainer(crs=3035, res=100)
+        excluder = atlite.ExclusionContainer(crs=metric_crs, res=100)
```
pz-max: Should stay 3035. Atlite uses 3035 for exclusion stuff:  https://github.com/PyPSA/atlite/blob/de20027a477b9ae79825fc2b9c5073a6f5a5f376/atlite/gis.py#L191-L206

scripts/build_osm_network.py
```diff
@@ -700,26 +702,29 @@ def fix_overpassing_lines(lines, buses, tol=1):
     df_to_add.set_index(lines.index[-1] + df_to_add.index, inplace=True)
 
     # update length
-    df_to_add["length"] = df_to_add.to_crs(epsg=3857).geometry.length
+    df_to_add["length"] = df_to_add.to_crs(metric_crs).geometry.length
```
pz-max: Yes for line length (=distances) epsg 3857 is good

config.default.yaml
```diff
@@ -35,6 +35,11 @@ enable:
   # requires cds API key https://cds.climate.copernicus.eu/api-how-to
   # More information https://atlite.readthedocs.io/en/latest/introduction.html#datasets
 
+# definition of the Coordinate Reference Systems
+crs:
+  default_crs: EPSG:4326  # general geographic projection, not used for metric measures. "EPSG:4326" is the standard used by OSM and google maps
```
pz-max: maybe something like:
```
crs:
   degree_crs : EPSG 4326  # (e.g. default for copernicus raster imports)
   distance_crs : EPSG:3857  # (e.g. OSM line lengths, distance from offshore..)
   area_crs : EPSG:3035  # (e.g. Atlite exclusion area container stuff)
```

scripts/build_renewable_profiles.py
davide-f: Yes, but epsg:3035 is valid for europe: https://epsg.io/3035
That had to be modified

scripts/build_renewable_profiles.py
davide-f: As above

scripts/build_osm_network.py
davide-f: The idea of metric_crs is to use a crs having as a unit meters (that's why metric but can be revised)

scripts/build_renewable_profiles.py
pz-max: EPSG 3035 is however an EU crs... Let's maybe decide on an area crs & add something to the bug list for now? https://www.google.com/search?channel=fs&client=ubuntu&q=epsg+3035

Would be good to have an area crs depending on the region of interest

config.default.yaml
davide-f: Revised, see next version

pz-max (approved):
Looks great. Ready to merge? @davide-f

davide-f:
> Looks great. Ready to merge? @davide-f

Yes :)

PR #392, opened 2022-06-22
Add converters as links and enable DC

### LINKED ISSUE #385: Add converters as links and enable DC lines

HVDC lines and converters are not properly modelled yet. PR #348 will focus on the topology side to make sure that we are able to properly identify DC lines and have a complete network, yet all lines,buses and (fake-)converters will be treated as AC lines.
This issue shall focus on moving beyond that and model converters as links and HVDC lines as dc lines.

Note: In PyPSA-Eur DC lines are also modelled as links, though it may be better to model them as dc lines if it will be possible. With respect to the application to Africa, given the current status of the lines, no differences should occur.

ekatef:
# Closes #385 

The aim of the PR is to enable DC carrier in the network model and include converters as Link component

## Checklist


ekatef:
This PR follows the general strategy drafted #385. Network connectivity seems to be alright but there are some questions related to the modeling concepts.

Currently, the `DC` carrier is enabled for the buses and the DC lines are implemented as the PyPSA `Line` component and complying with [the Line specification](https://pypsa.readthedocs.io/en/latest/components.html#line). Which means that the nominal frequency `f_nom` is set to zero and the active resistance `r` has a non-zero value (_at the moment a rather arbitrary_) while a default zero value is used for the reactance `x`. But the `carrier` field of the DC lines is set to `AC` as according to the Line documentation “AC” is the only valid value.

It would be great to define DC lines by the line type as it is done for AC ones. But all the standard PyPSA line types seem to have `f_nom` == 50 Hz and the type specification should overwrite any custom setting for `r` and `x`. So, the only way to use `type` seems to define a custom DC lines one. Should we opt for that?

Another question relates to the testing strategy: there are both AC and DC buses as well as Link-converters in the `base.nc` network but that difference disappear after network simplification and a ready-to-solve network doesn't has any visible DC features... It seems to be a good idea to look a bit deeper into the resulted PyPSA network structure and/or to directly compare optimisation results for the model version with and without taking DC into account.

@davide-f, could you please let me know if that above sounds reasonable for you? :)

ekatef:
The network outline at the moment looks like that

## base network

base nc_test

## simplified network

elec_s nc_test

## network to the solver

elec_s_6_ec_lcopt_Co2L-3H nc_test

davide-f:
Super! we are moving closer and closer! :D

> Currently, the `DC` carrier is enabled for the buses and the DC lines are implemented as the PyPSA `Line` component and complying with [the Line specification](https://pypsa.readthedocs.io/en/latest/components.html#line). Which means that the nominal frequency `f_nom` is set to zero and the active resistance `r` has a non-zero value (_at the moment a rather arbitrary_) while a default zero value is used for the reactance `x`. But the `carrier` field of the DC lines is set to `AC` as according to the Line documentation “AC” is the only valid value.

Leave the "AC" flag, though that has no implication into the model. A line is a line whatever the frequency of the system.
In DC lines, I'd specify x=0.0 and the model of that line collapses into being equivalent to a DC line.

> It would be great to define DC lines by the line type as it is done for AC ones. But all the standard PyPSA line types seem to have `f_nom` == 50 Hz and the type specification should overwrite any custom setting for `r` and `x`. So, the only way to use `type` seems to define a custom DC lines one. Should we opt for that?

This is interesting, but for the moment as a fix we may set X=0.0 for the DC lines; if an error shows up because it doesn't like 0.0, we may set a small value.
If errors show up (it may happen), we address the corresponding power flow; worst case: we keep the x as is

> Another question relates to the testing strategy: there are both AC and DC buses as well as Link-converters in the `base.nc` network but that difference disappear after network simplification and a ready-to-solve network doesn't has any visible DC features... It seems to be a good idea to look a bit deeper into the resulted PyPSA network structure and/or to directly compare optimisation results for the model version with and without taking DC into account.

I agree and I think @fneum may help here. This question follows a discussion we had recently on the implications of setting DC lines as DC lines instead of links.
During the discussion, we agreed that *most likely* the change (at least in pypsa-eur) would be without harm.

davide-f:
> The network outline at the moment looks like that
> ...

Here it seems that something is not working, unfortunately.
My comments are below:
- It may be interesting to highlight in purple also DC lines, so it may be easier to understand where the DC line goes
- while DC buses are available in base_network, they disappear later, have you investigated why? (You mentioned it and I think it may be good to check; it may be wanted but we shall check if in the importing from pypsa-eur something on that regard got lost)
- Why are there links also in the disconnected buses on the topright corner? In base network there were no dc buses there
- Several nodes get disconnected in simplify network: is the network properly meshed? why does that happen? Did this happen also before the past PR?

ekatef:
@davide-f, many thanks for looking into it! As always, your comments are very helpful to move forward :)

> Leave the "AC" flag, though that has no implication into the model. A line is a line whatever the frequency of the system.
> In DC lines, I'd specify x=0.0 and the model of that line collapses into being equivalent to a DC line.

Thank you! Good to know that there is no pitfalls in this regard.

> This is interesting, but for the moment as a fix we may set X=0.0 for the DC lines; if an error shows up because it doesn't like 0.0, we may set a small value.
If errors show up (it may happen), we address the corresponding power flow; worst case: we keep the x as is

You are absolutely right: the solver seems to be not happy by current parameter values and throws a numpy error `LinAlgError("SVD did not converge")`. I'll try to fix it by playing with `x` and `r`. Currently, there seems to be some mess in the network parameters values probably because `type` and `r`-`x` specifications are mixed for AC and DC lines.

> I agree and I think @fneum may help here. This question follows a discussion we had recently on the implications of setting DC lines as DC lines instead of links.

Great, then I'll look into the PyPSA-networks deeper and check if anything get lost along the workflow.

> During the discussion, we agreed that _most likely_ the change (at least in pypsa-eur) would be without harm.

In my experience, self-respecting models rarely miss an opportunity to crash, preferably in the least suitable moment... And I absolutely that the step-by-step approach is here life-saving :D

ekatef:
Thank you for checking!

> * It may be interesting to highlight in purple also DC lines, so it may be easier to understand where the DC line goes

Yes, absolutely. If the DC lines are highlighted it's evident that they are present in the network up to the solving step (the updated pictures are bellow).

> * while DC buses are available in base_network, they disappear later, have you investigated why? (You mentioned it and I think it may be good to check; it may be wanted but we shall check if in the importing from pypsa-eur something on that regard got lost)

At the moment I see that the DC buses get lost after `simplify_network`. Will investigate what exactly happens there.

> * Why are there links also in the disconnected buses on the topright corner? In base network there were no dc buses there

The links appear in the network after `add_extra_components` has been applied. I suspect that may relate to addition of extra storage buses via two links and will check it.

> * Several nodes get disconnected in simplify network: is the network properly meshed? why does that happen? Did this happen also before the past PR?

Yes, application of the `simplify_network` has lead to the loss of some buses before the past PR. That seems to be related to the data quality. As if the meshing is proper I'm nor sure that understand your idea... Could you please clarify a bit?

ekatef:
## base network after generators added (elec.nc)
Output of `add_electricity`

elec nc

## simplified network (elec_s.nc)
Output of `simplify_network`: DC buses were dropped

elec_s nc

## network with extendable components added (elec_s_6_ec.nc)

Output of `add_extra_components`: the links (in red on the picture) were added

elec_s_6_ec nc

ekatef:
To summarise, the plan is:

1) check is something is missed in the workflow, in particular
    - look how the DC buses are handled by the `simplify_network`;
    - check how does `add_extra_components` work and what is exact meaning of the links which it appends;

2) check values of `x` and `r` to avoid singularity when solving.

davide-f:
As a first attempt: set carrier to dc in the DC lines.
That should enable the loadflow to be properly handled as by implementation in pypsa.
I fear errors may appear in the clustering part, but let's see I'd say. Worst case, we use dc lines coded as ac with the x equal to the r; in terms of load flow the results will be equivalent.

Regarding the links in the other buses yeah, add extra components add links and stores to account for h2 and batteries.

ekatef:
> As a first attempt: set carrier to dc in the DC lines. That should enable the loadflow to be properly handled as by implementation in pypsa. I fear errors may appear in the clustering part, but let's see I'd say. Worst case, we use dc lines coded as ac with the x equal to the r; in terms of load flow the results will be equivalent.

Surprisingly, there clustering runs fine but the general impression is that setting DC-carrier is being silently ignored. At least, I don't see any differences in the structures of the resulted networks and the convergence problem persists...

> Regarding the links in the other buses yeah, add extra components add links and stores to account for h2 and batteries.

Great, thank you for the confirmation!

Upd: Convergence problems seems to relate to the switch from specification of the line type ('Al/St 240/40 2-bundle 220.0' etc.) to explicit definition of the line parameters (x, r, s_max_pu). It looks like some required line parameters are missed.

ekatef:
After some playing with the line parameters, the problem looks like that:

1) In case the DC-lines are defined via type definition like it's done in `_set_electrical_parameters_lines`, everything works even if r or x are additionally specified for DC-lines (however, explicit parameters definitions should be overwritten by the type definition, if I have properly understood PyPSA documentation);

2) If type definition in `_set_electrical_parameters_dc_lines` is replaced by parameters definition in different combinations (r only; r and x; r and x and f_nom; s_max_pu is set for all the cases as it was before) the solution crashes with a numpy `LinAlgError("SVD did not converge")`. The full listing is bellow. 

3) The only difference in line parameters between in the network which causes the error and the network is being optimised without any numerical troubles, were s_nom and s_nom_min values as `_set_lines_s_nom_from_linetypes` uses the line type to define s_nom. Modifications were added to set s_nom according to the defaults and now the lines parameters for the network looks were similar to the initial version (this one from point 1 above), but the error persists.

The csv's with lines parameters for converged and diverged cases:

### solution diverges
[lines_parameters_diverges_v2.csv](https://github.com/pypsa-meets-africa/pypsa-africa/files/9024079/lines_parameters_diverges_v2.csv)

### solution converges
[lines_parameters_converged.csv](https://github.com/pypsa-meets-africa/pypsa-africa/files/9024083/lines_parameters_converged.csv)

I'm not able to find any substantial differences but the troubles disappear if the types definition for the DC lines is put back. So, probably the troubles are caused by singularities in some other part of the problem which is effected by specification of the line types indirectly? Is there probably a way to extract the system of equations which is being solved?

Setting both min_iterations and max_iterations to 1 make the solving process faster but in that case doesn't resolve convergence issues.

The error looks like that:

INFO:snakemake.logging:
[Fri Jul  1 02:58:52 2022]
INFO:snakemake.logging:[Fri Jul  1 02:58:52 2022]
rule solve_network:
    input: networks/elec_s_6_ec_lcopt_Co2L-3H.nc
    output: results/networks/elec_s_6_ec_lcopt_Co2L-3H.nc
    log: logs/solve_network/elec_s_6_ec_lcopt_Co2L-3H_solver.log, logs/solve_network/elec_s_6_ec_lcopt_Co2L-3H_python.log, logs/solve_network/elec_s_6_ec_lcopt_Co2L-3H_memory.log
    jobid: 1
    benchmark: benchmarks/solve_network/elec_s_6_ec_lcopt_Co2L-3H
    reason: Input files updated by another job: networks/elec_s_6_ec_lcopt_Co2L-3H.nc
    wildcards: simpl=, clusters=6, ll=copt, opts=Co2L-3H
    resources: tmpdir=/var/folders/qn/vpndfm21795ckkq89np1ckp40000gn/T, mem=11170
INFO:snakemake.logging:rule solve_network:
    input: networks/elec_s_6_ec_lcopt_Co2L-3H.nc
    output: results/networks/elec_s_6_ec_lcopt_Co2L-3H.nc
    log: logs/solve_network/elec_s_6_ec_lcopt_Co2L-3H_solver.log, logs/solve_network/elec_s_6_ec_lcopt_Co2L-3H_python.log, logs/solve_network/elec_s_6_ec_lcopt_Co2L-3H_memory.log
    jobid: 1
    benchmark: benchmarks/solve_network/elec_s_6_ec_lcopt_Co2L-3H
    reason: Input files updated by another job: networks/elec_s_6_ec_lcopt_Co2L-3H.nc
    wildcards: simpl=, clusters=6, ll=copt, opts=Co2L-3H
    resources: tmpdir=/var/folders/qn/vpndfm21795ckkq89np1ckp40000gn/T, mem=11170

```
INFO:snakemake.logging:
INFO:pypsa.io:Imported network elec_s_6_ec_lcopt_Co2L-3H.nc has buses, carriers, generators, global_constraints, lines, links, loads, storage_units, stores
INFO:pypsa.linopf:Prepare linear problem
INFO:pypsa.linopf:Total preparation time: 0.33s
INFO:pypsa.linopf:Solve linear problem using Glpk solver
INFO:pypsa.linopf:Optimization successful. Objective value: 8.99e+08
INFO:pypsa.linopf:Mean square difference after iteration 1 is 1.0411104755130154
INFO:pypsa.linopf:Iteration 2 beyond max_iterations 1. Stopping ...
INFO:pypsa.linopf:Running last lopf with fixed branches (HVDC links and HVAC lines)
~/opt/miniconda3/envs/pypsa-africa/lib/python3.10/site-packages/numpy/core/fromnumeric.py:3199: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
  return asarray(a).ndim
~/opt/miniconda3/envs/pypsa-africa/lib/python3.10/site-packages/numpy/core/fromnumeric.py:3199: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
  return asarray(a).ndim
INFO:pypsa.linopf:Prepare linear problem
INFO:pypsa.linopf:Total preparation time: 0.3s
INFO:pypsa.linopf:Solve linear problem using Glpk solver
INFO:pypsa.linopf:Optimization successful. Objective value: 8.99e+08
~/opt/miniconda3/envs/pypsa-africa/lib/python3.10/site-packages/pypsa/pf.py:1190: RuntimeWarning: divide by zero encountered in true_divide
  b = 1.0 / np.concatenate(
~/opt/miniconda3/envs/pypsa-africa/lib/python3.10/site-packages/pypsa/pf.py:1211: RuntimeWarning: invalid value encountered in multiply
  sub_network.p_branch_shift = -b * np.concatenate(
  File "~/Documents/_github_/pypsa-africa/.snakemake/shadow/tmp8o1_w0f9/.snakemake/scripts/tmpupnr9cou.solve_network.py", line 352, in <module>
    n = solve_network(
  File "~/Documents/_github_/pypsa-africa/.snakemake/shadow/tmp8o1_w0f9/.snakemake/scripts/tmpupnr9cou.solve_network.py", line 313, in solve_network
    ilopf(
  File "~/opt/miniconda3/envs/pypsa-africa/lib/python3.10/site-packages/pypsa/linopf.py", line 1637, in ilopf
    network_lopf(n, snapshots, **kwargs)
  File "~/opt/miniconda3/envs/pypsa-africa/lib/python3.10/site-packages/pypsa/linopf.py", line 1503, in network_lopf
    assign_solution(
  File "~/opt/miniconda3/envs/pypsa-africa/lib/python3.10/site-packages/pypsa/linopf.py", line 1343, in assign_solution
    [v_ang_for_(sub) for sub in n.sub_networks.obj], axis=1
  File "~/opt/miniconda3/envs/pypsa-africa/lib/python3.10/site-packages/pypsa/linopf.py", line 1343, in <listcomp>
    [v_ang_for_(sub) for sub in n.sub_networks.obj], axis=1
  File "~/opt/miniconda3/envs/pypsa-africa/lib/python3.10/site-packages/pypsa/linopf.py", line 1338, in v_ang_for_
    Z = pd.DataFrame(np.linalg.pinv((sub.B).todense()), buses_i, buses_i)
  File "<__array_function__ internals>", line 180, in pinv
  File "~/opt/miniconda3/envs/pypsa-africa/lib/python3.10/site-packages/numpy/linalg/linalg.py", line 1990, in pinv
    u, s, vt = svd(a, full_matrices=False, hermitian=hermitian)
  File "<__array_function__ internals>", line 180, in svd
  File "~/opt/miniconda3/envs/pypsa-africa/lib/python3.10/site-packages/numpy/linalg/linalg.py", line 1648, in svd
    u, s, vh = gufunc(a, signature=signature, extobj=extobj)
  File "~/opt/miniconda3/envs/pypsa-africa/lib/python3.10/site-packages/numpy/linalg/linalg.py", line 97, in _raise_linalgerror_svd_nonconvergence
    raise LinAlgError("SVD did not converge")
numpy.linalg.LinAlgError: SVD did not converge
```

@davide-f , @pz-max, would be very grateful for your comments :)

Update: Singularity during solving appears when B and H are being evaluated for sub-networks by `calculate_B_H( )`. The `r_pu_eff` (in case DC sub-networks) or `x_pu_eff` (for AC ones) values are being used there. At the moment DC network is treated in that sense as an AC. In case when line parameters are defined explicitly rather than via a linetype, the `x_pu_eff` becomes `inf` for DC lines which seems to result in the error above.

Update: It works now but with a quite brute force approach. Namely, if an additional row for a DC linetype is added to the "line_types.csv" on "~/opt/miniconda3/envs/pypsa-africa/lib/python3.10/site-packages/pypsa/standard_types", the workflow runs properly. While specifying the same parameters directly results in divergence. Probably, the matter is that under the current solving approach the solver is updating all the parameters on each step and should have an access to the parameters values which can be provided via a csv-file.

ekatef:
A final (hopefully :) task relates to removal of DC buses by `simplify_network`. The DC-buses were removed by the `remove_stubs()` function which has treated them as a kind of "dead-ends". That can be changed by setting the `matching_attrs` arguments to "carrier". Then the DC buses remain in the network but the mixed-carrier error appears then during clustering as it was expected. 

The strategy is needed to decouple DC and AC buses. Would it be reasonable to cut AC buses along the DC lines? And then to hope that the cluster boundary will be set across the converters that connect DC part to AC one (the picture is attached bellow).

@davide-f Could you please have a look?

ekatef:
The network before clustering now looks like that

elec_s nc_dc_by_freq

pz-max:
@ekatef regarding the singularity issues that you experienced. Reading about the linear power flow: https://pypsa.readthedocs.io/en/latest/power_flow.html#linear-power-flow :
`Note that for lines and transformers you MUST make sure that is non-zero, otherwise the bus admittance matrix will be singular.`
meaning missing values could cause singularity problems. This is also what you describe in the "update".

It seems you are on the way to solving the problem. I am now a bit confused by the status & the exact questions. Exchanging tomorrow on Discord if necessary might help. We meet anyways tomorrow. Looking forward to it

ekatef:
> @ekatef regarding the singularity issues that you experienced. Reading about the linear power flow: https://pypsa.readthedocs.io/en/latest/power_flow.html#linear-power-flow : `Note that for lines and transformers you MUST make sure that is non-zero, otherwise the bus admittance matrix will be singular.` meaning missing values could cause singularity problems. This is also what you describe in the "update".
> 
> It seems you are on the way to solving the problem. I am now a bit confused by the status & the exact questions. Exchanging tomorrow on Discord if necessary might help. We meet anyways tomorrow. Looking forward to it

Thank you @pz-max! It has helped :) Now the singularity issue is resolved. Although, it would be great to add a custom linetype in a little more civilised way than modifying a source datafile :) I have tried `n.import_components_from_dataframe(buses, "Linetype")` but was not able to get it work. Do you probably know what is the proper approach here?

pz-max:
> import_components_from_dataframe

@ekatef which components were missing the information?


Electrical parameters are assigned in PyPSA-Eur here:
https://github.com/PyPSA/pypsa-eur/blob/611cd383da6318ab04253d5f4b7c68dc0a677bb4/scripts/base_network.py#L552-L571

Comparing that to your implementation it looks fine:
https://github.com/ekatef/pypsa-africa/blob/63d17419f5496f46832160246540ae14068fce35/scripts/base_network.py#L462-L479

So electrical parameters should be already assigned to all components or not? If not, we need to find out where and why these `_set_electrical_parameters....` do not work

ekatef:
@pz-max, thank you for looking into it! Yes, we are following the PyPSA-Europe logic of in setting lines parameters but our implementation for DC lines inside differs a bit.

PyPSA-Europe code relies on the standard linetypes:

https://github.com/PyPSA/pypsa-eur/blob/611cd383da6318ab04253d5f4b7c68dc0a677bb4/scripts/base_network.py#L293-L296

While we need a custom linetype (I was not able to get work an approach with direct setting of the line parameters (`r`, `x` etc) as it has caused some convergency problems):

https://github.com/ekatef/pypsa-africa/blob/a22199b29da4f9c56eca9f5ca169809f54108af4/scripts/base_network.py#L263

The question is how to set this custom linetype properly. At the moment I have modified the PyPSA source code to have one more "standard" linetype but it doesn't look as an optimal implementation :) My thought was: probably we can add the `LineType` component to the PyPSA network with `import_components_from_dataframe( )`. According to PyPSA [docs](https://pypsa.readthedocs.io/en/latest/components.html#components) the LineType is just one of the network components  as Bus, Line, Transformers...

davide-f:
> @ekatef regarding the singularity issues that you experienced. Reading about the linear power flow: https://pypsa.readthedocs.io/en/latest/power_flow.html#linear-power-flow : `Note that for lines and transformers you MUST make sure that is non-zero, otherwise the bus admittance matrix will be singular.` meaning missing values could cause singularity problems. This is also what you describe in the "update".

Trues for AC lines; in the case of "DC" lines the resistance value should be used instead.
https://github.com/PyPSA/PyPSA/blob/9aa1be3441e2f71bb659ad32621a1a30790b4703/pypsa/opf.py#L1432

davide-f:
> @pz-max, thank you for looking into it! Yes, we are following the PyPSA-Europe logic of in setting lines parameters but our implementation for DC lines inside differs a bit.
> 
> PyPSA-Europe code relies on the standard linetypes:
> 
> https://github.com/PyPSA/pypsa-eur/blob/611cd383da6318ab04253d5f4b7c68dc0a677bb4/scripts/base_network.py#L293-L296
> 
> While we need a custom linetype (I was not able to get work an approach with direct setting of the line parameters (`r`, `x` etc) as it has caused some convergency problems):

I agree that now we need something that is reasonable and works; we could also propose a PR in PyPSA to add some reference DC cases but they shall be realistic, thus it may take some time, but it is feasible.
Anyway, on my side currently a hard fix and specifying the parameters by ourself would be a reasonable approach for the time beeing


My major concern is regarding the clustering procedure and we need to better investigating how that should be treated.
As mixed AC/DC buses cannot be mixed, this implies that we may need different nodes for AC and DC networks, which is possible but it increases the size of the problem.
As a "trick" we can specify that the "DC" nodes are "AC" instead, but by doing so, we need to check how the link between 2 AC nodes belonging to the same node are threated and if the link may simply disappear,which would change the mathematical problem.

[consideration] Anyway, personally, I don't understand why in pypsa-eur the dc lines treated as links work and in our case the dc lines treated as lines do not. Topologically, the two problems are identical. We may be doing something in the simplify_network and clustering_network that may not be fully equivalent to theirs.
Part of the reason is that in simplify, for example simplify_links is the same as in pypsa-eur, hence it expects the DC lines to be links.
I see two options:
- keep the DC line representation, but investigate and solve also the corresponding issues in simplify and cluster network
- implement DC lines as links by option: in the config we may add a flag to specify whether we want DC lines as links or lines. When they shall be lines, we use the current approach, otherwise, the lines are added to the converter.csv file. In this case, we are closer to the pypsa-eur implementation and the errors that may occur would be related to a probably imperfect porting from pypsa-eur, hence we may need to compare our implementation to theirs.

@ekatef This is a very delicate task but you are managing it :) how do you feel the two options?

davide-f:
> A final (hopefully :) task relates to removal of DC buses by `simplify_network`. The DC-buses were removed by the `remove_stubs()` function which has treated them as a kind of "dead-ends". That can be changed by setting the `matching_attrs` arguments to "carrier". Then the DC buses remain in the network but the mixed-carrier error appears then during clustering as it was expected.
> 
> The strategy is needed to decouple DC and AC buses. Would it be reasonable to cut AC buses along the DC lines? And then to hope that the cluster boundary will be set across the converters that connect DC part to AC one (the picture is attached bellow).
> 
> @davide-f Could you please have a look?

If here you refer to the buses that get isolated, I think that remove_stubs may perform well but because the different nodes are actually disconnected, and this latter is an issue.
This may happen for example when two interconnecting lines end in two GIS points that are close but not so close. When they are beyond the tolerance (parameter build_osm_network.group_tolerance_buses initialized to 500m) they are not recognized. It may be worthy to try a larger number, you may estimate it by the GIS location and or using google maps (no need to do sophisticated analysis on this). I think it may be possible to use max 5-10 km for this parameter

davide-f:
@ekatef Ah, please, make sure that the "carrier" is "DC" for dc lines (it should be by checking the changes) and not "dc". The former is used as reference for dc lines in pypsa.

ekatef:
> My major concern is regarding the clustering procedure and we need to better investigating how that should be treated. As mixed AC/DC buses cannot be mixed, this implies that we may need different nodes for AC and DC networks, which is possible but it increases the size of the problem. As a "trick" we can specify that the "DC" nodes are "AC" instead, but by doing so, we need to check how the link between 2 AC nodes belonging to the same node are threated and if the link may simply disappear,which would change the mathematical problem.
> 
> [consideration] Anyway, personally, I don't understand why in pypsa-eur the dc lines treated as links work and in our case the dc lines treated as lines do not. Topologically, the two problems are identical. We may be doing something in the simplify_network and clustering_network that may not be fully equivalent to theirs. Part of the reason is that in simplify, for example simplify_links is the same as in pypsa-eur, hence it expects the DC lines to be links. I see two options:
> 
> * keep the DC line representation, but investigate and solve also the corresponding issues in simplify and cluster network
> * implement DC lines as links by option: in the config we may add a flag to specify whether we want DC lines as links or lines. When they shall be lines, we use the current approach, otherwise, the lines are added to the converter.csv file. In this case, we are closer to the pypsa-eur implementation and the errors that may occur would be related to a probably imperfect porting from pypsa-eur, hence we may need to compare our implementation to theirs.
> 
> @ekatef This is a very delicate task but you are managing it :) how do you feel the two options?

@davide-f, many thanks for the detailed comment! It was incredibly helpful. At the moment both the mixed AC/DC nodes issue and the troubles with links settings seem to be resolved. Tests for the single countries were successful, now I'm waiting for the results for the whole continent.

The key to a proper AC/DC nodes representation was to take an advantage of PyPSA in-built clustering procedure by un-commenting a couple of sub-networks fixes (@pz-max, many thanks for the hint!) and making some minor adjustments. Although, I'm not sure why these fixes were needed and suspect that there may be some pitfalls :) As for the links porting, you were absolutely right: the appeared error was caused by a couple of columns with the names mis-interpreted during pre-solving. 

Regarding the modeling options, I love your idea to have an option for a switch between different representation of the HVDC lines and will be happy to introduce a "hvcd_as_links"-flag into the config.

Generally, I think it's a good idea to test the result not only from a development perspective but also from a modeling one to ensure that the effect of the changes introduced into the model is a reasonable one. What do you think?

ekatef:
> > A final (hopefully :) task relates to removal of DC buses by `simplify_network`. The DC-buses were removed by the `remove_stubs()` function which has treated them as a kind of "dead-ends". That can be changed by setting the `matching_attrs` arguments to "carrier". Then the DC buses remain in the network but the mixed-carrier error appears then during clustering as it was expected.
> > The strategy is needed to decouple DC and AC buses. Would it be reasonable to cut AC buses along the DC lines? And then to hope that the cluster boundary will be set across the converters that connect DC part to AC one (the picture is attached bellow).
> > @davide-f Could you please have a look?
> 
> If here you refer to the buses that get isolated, I think that remove_stubs may perform well but because the different nodes are actually disconnected, and this latter is an issue. This may happen for example when two interconnecting lines end in two GIS points that are close but not so close. When they are beyond the tolerance (parameter build_osm_network.group_tolerance_buses initialized to 500m) they are not recognized. It may be worthy to try a larger number, you may estimate it by the GIS location and or using google maps (no need to do sophisticated analysis on this). I think it may be possible to use max 5-10 km for this parameter

Great, thank you for the hint!

davide-f:
@ekatef wonderful!

Again, I kindly ask you to post the network images equivalent to the pictures mentioned above to understand the changes.

The test for Africa would also be welcome.

ekatef:
The network topology looks now like this (hvdc lines as lines):

## base network

base nc_dc_by_freq

## simplified network

elec_s nc_dc_by_freq

## network to the solver

elec_s_6_ec_lcopt_Co2L-3H nc_dc_by_freq

Spatial tolerance issue was not fixed yet and DC buses kept by an explicit restriction on removal of "dead-ends" in case the carriers don't match.

ekatef:
If `group_tolerance_buses` is set to 10 km (it should be enough to capture any spatial buses displacement according to the maps) and the matching-carrier restriction is lifted the picture looks like that:

## base network

base nc_dc_by_freq

## simplified network

elec_s nc_dc_by_freq

## network to the solver

elec_s_6_ec_lcopt_Co2L-3H nc_dc_by_freq

The DC buses here get lost. And I do not completely understand why does disappear the DC-AC converter with connects the DC part with the AC grid in the south-east. It looks like our representation of the converters is not quite compatible with the PyPSA-Europe approach as the converter branch is being treated by `remove_stubs` as a "dead-end" and we need explicitly set `matching_attrs="carrier"`.

ekatef:
@davide-f, I'ld be very grateful for your suggestions on the pictures above :)

Apart of that, I suspect that we could need to modify application of the Loads/Generators which currently seem to be applied to AC nodes only:

https://github.com/pypsa-meets-africa/pypsa-africa/blob/7bc769e9ac9d1878a6cba8188ecf8d155443cdf6/scripts/solve_network.py#L107-L122

ekatef:
Test for the whole continent runs but there are some troubles during solving if `max_iterations` is more than 1. 

It looks like it was not a perfect choice to set the number of clusters to 150 for the test run as there is a warning during clustering `ConvergenceWarning: Number of distinct clusters (7) found smaller than n_clusters (9). Possibly due to duplicate points in X. kmeans.fit(points)` and the clustered network looks a bit strange.

## base network

base nc_dc_by_freq

## simplified network

elec_s nc_dc_by_freq

## network to the solver

elec_s_150_ec_lcopt_Co2L-3H nc_dc_by_freq

davide-f:
> If `group_tolerance_buses` is set to 10 km (it should be enough to capture any spatial buses displacement according to the maps) and the matching-carrier restriction is lifted the picture looks like that:

This is interesting: what do you mean with matching carrier? you allow AC buses to be connected to DC ones?
I'm ok with 10km, but I think we should keep the carrier

> The DC buses here get lost. And I do not completely understand why does disappear the DC-AC converter with connects the DC part with the AC grid in the south-east. It looks like our representation of the converters is not quite compatible with the PyPSA-Europe approach as the converter branch is being treated by `remove_stubs` as a "dead-end" and we need explicitly set `matching_attrs="carrier"`.

This is quite weird and shall be investigated.
Does that problem occurs also in the link implementation of this PR?

davide-f:
> Test for the whole continent runs but there are some troubles during solving if `max_iterations` is more than 1.

Nice! What troubles?
This PR seems in very good shape, provided that the issue with remove_stubs is investigated a bit better I think

> It looks like it was not a perfect choice to set the number of clusters to 150 for the test run as there is a warning during clustering `ConvergenceWarning: Number of distinct clusters (7) found smaller than n_clusters (9). Possibly due to duplicate points in X. kmeans.fit(points)` and the clustered network looks a bit strange.

I usually use 55 for testing large scales. That may also depend on the tolerance used for the kmeans

ekatef:
@davide-f, thank you for looking into it! I also feel stubs management as a bit weird and would be very grateful for your help with understanding how it works.

> This is interesting: what do you mean with matching carrier? you allow AC buses to be connected to DC ones?

Sorry for being not quite clear: `matching_carriers` is an argument of the `busmap_by_stubs()`, a function that defines which buses should be removed from the network as "dead-ends":

https://github.com/pypsa-meets-africa/pypsa-africa/blob/7bc769e9ac9d1878a6cba8188ecf8d155443cdf6/scripts/simplify_network.py#L415-L418

https://github.com/pypsa-meets-africa/pypsa-africa/blob/7bc769e9ac9d1878a6cba8188ecf8d155443cdf6/scripts/simplify_network.py#L370

Setting `matching_attributes= "carrier"` allows to keep an apparently "dead-end" bus if its carries is not the same as at a neighbouring bus. 

> Does that problem occurs also in the link implementation of this PR? 

Yes, in the link implementation DC buses are being cut as well during the stubs removal. I do not quite understand why exactly the DC buses are cut not AC ones as `busmap_by_stubs()` is governed by graphs topology only disregarding nature of the nodes.

ekatef:
> > Test for the whole continent runs but there are some troubles during solving if `max_iterations` is more than 1.
> 
> Nice! What troubles?

The whole listing looks like that

```bash
rule solve_network:
    input: networks/elec_s_150_ec_lcopt_Co2L-3H.nc
    output: results/networks/elec_s_150_ec_lcopt_Co2L-3H.nc
    log: logs/solve_network/elec_s_150_ec_lcopt_Co2L-3H_solver.log, logs/solve_network/elec_s_150_ec_lcopt_Co2L-3H_python.log, logs/solve_network/elec_s_150_ec_lcopt_Co2L-3H_memory.log
    jobid: 1
    benchmark: benchmarks/solve_network/elec_s_150_ec_lcopt_Co2L-3H
    reason: Missing output files: results/networks/elec_s_150_ec_lcopt_Co2L-3H.nc
    wildcards: simpl=, clusters=150, ll=copt, opts=Co2L-3H
    resources: tmpdir=/var/folders/qn/vpndfm21795ckkq89np1ckp40000gn/T, mem=39250
INFO:snakemake.logging:rule solve_network:
    input: networks/elec_s_150_ec_lcopt_Co2L-3H.nc
    output: results/networks/elec_s_150_ec_lcopt_Co2L-3H.nc
    log: logs/solve_network/elec_s_150_ec_lcopt_Co2L-3H_solver.log, logs/solve_network/elec_s_150_ec_lcopt_Co2L-3H_python.log, logs/solve_network/elec_s_150_ec_lcopt_Co2L-3H_memory.log
    jobid: 1
    benchmark: benchmarks/solve_network/elec_s_150_ec_lcopt_Co2L-3H
    reason: Missing output files: results/networks/elec_s_150_ec_lcopt_Co2L-3H.nc
    wildcards: simpl=, clusters=150, ll=copt, opts=Co2L-3H
    resources: tmpdir=/var/folders/qn/vpndfm21795ckkq89np1ckp40000gn/T, mem=39250

INFO:snakemake.logging:
Restricted license - for non-production use only - expires 2023-10-25
No parameters matching '_test' found
INFO:pypsa.io:Imported network elec_s_150_ec_lcopt_Co2L-3H.nc has buses, carriers, generators, global_constraints, lines, links, loads, storage_units, stores
INFO:pypsa.linopf:Prepare linear problem
INFO:pypsa.linopf:Total preparation time: 7.82s
INFO:pypsa.linopf:Solve linear problem using Glpk solver
INFO:pypsa.linopf:Optimization successful. Objective value: 1.20e+15
INFO:pypsa.linopf:Mean square difference after iteration 1 is 1.4235588465238396
INFO:pypsa.linopf:Prepare linear problem
INFO:pypsa.linopf:Total preparation time: 7.83s
INFO:pypsa.linopf:Solve linear problem using warmstart
WARNING:pypsa.linopf:Optimization failed with status warning and termination condition infeasible (final)
  File "~/pypsa-africa/.snakemake/shadow/tmpwj0ulvkm/.snakemake/scripts/tmpzlyuzjxw.solve_network.py", line 432, in <module>
    n = solve_network(
  File "~/pypsa-africa/.snakemake/shadow/tmpwj0ulvkm/.snakemake/scripts/tmpzlyuzjxw.solve_network.py", line 393, in solve_network
    ilopf(
  File "~/opt/miniconda3/envs/pypsa-africa/lib/python3.9/site-packages/pypsa/linopf.py", line 1615, in ilopf
    assert status == "ok", (
AssertionError: Optimization failed with status warningand termination infeasible (final)
```

ekatef:
> > It looks like it was not a perfect choice to set the number of clusters to 150 for the test run as there is a warning during clustering `ConvergenceWarning: Number of distinct clusters (7) found smaller than n_clusters (9). Possibly due to duplicate points in X. kmeans.fit(points)` and the clustered network looks a bit strange.
> 
> I usually use 55 for testing large scales. That may also depend on the tolerance used for the kmeans

Thank you for the hint! I suspect introducing DC nodes may also has some influence as AC and DC sub-networks should be kept separately.

davide-f:
> @davide-f, thank you for looking into it! I also feel stubs management as a bit weird and would be very grateful for your help with understanding how it works.
> 
> > This is interesting: what do you mean with matching carrier? you allow AC buses to be connected to DC ones?
> 
> Sorry for being not quite clear: `matching_carriers` is an argument of the `busmap_by_stubs()`, a function that defines which buses should be removed from the network as "dead-ends":
> 
> https://github.com/pypsa-meets-africa/pypsa-africa/blob/7bc769e9ac9d1878a6cba8188ecf8d155443cdf6/scripts/simplify_network.py#L415-L418
> 
> https://github.com/pypsa-meets-africa/pypsa-africa/blob/7bc769e9ac9d1878a6cba8188ecf8d155443cdf6/scripts/simplify_network.py#L370
> 
> Setting `matching_attributes= "carrier"` allows to keep an apparently "dead-end" bus if its carries is not the same as at a neighbouring bus.
> 
> > Does that problem occurs also in the link implementation of this PR?
> 
> Yes, in the link implementation DC buses are being cut as well during the stubs removal. I do not quite understand why exactly the DC buses are cut not AC ones as `busmap_by_stubs()` is governed by graphs topology only disregarding nature of the nodes.

The remove stubs is based on the network description: when a node has only one branch linked to it, it is recognized as a dead end and prioritized for the simplification.
I'm wondering whether the network and links is complete: is the node linking matched properly? I'd investigate that starting from a small single-country example and manually checking it, starting from csv and the base_network output.
The graph of the network can be calculated using: `network.graph()` see https://github.com/PyPSA/PyPSA/blob/8432bd3e8fb173d424e7f272c9daf56e7355befc/pypsa/networkclustering.py#L1003




> > > Test for the whole continent runs but there are some troubles during solving if `max_iterations` is more than 1.
> > 
> > 
> > Nice! What troubles?
> 
> The whole listing looks like that
> ...

Probably the infeasibility is due to numerical issues related with the super large numbers that we obtain (not a fan of them) with obj values exceeding 10e10.
A fix is to increase the feasibility tolerance `FeasibilityTol` like 1e-5 (hopefully should be enough). In my case it was

https://github.com/pypsa-meets-africa/pypsa-africa/blob/7bc769e9ac9d1878a6cba8188ecf8d155443cdf6/config.default.yaml#L300

ekatef:
> The remove stubs is based on the network description: when a node has only one branch linked to it, it is recognized as a dead end and prioritized for the simplification.
> I'm wondering whether the network and links is complete: is the node linking matched properly? I'd investigate that starting from a small single-country example and manually checking it, starting from csv and the base_network output.
> The graph of the network can be calculated using: `network.graph()` see https://github.com/PyPSA/PyPSA/blob/8432bd3e8fb173d424e7f272c9daf56e7355befc/pypsa/networkclustering.py#L1003

Agree, it definitely makes sense to ensure that the network topology is correct. Will check. Thanks you for the hint regarding the graph calculation!

ekatef:
> Probably the infeasibility is due to numerical issues related with the super large numbers that we obtain (not a fan of them) with obj values exceeding 10e10. A fix is to increase the feasibility tolerance `FeasibilityTol` like 1e-5 (hopefully should be enough). In my case it was
> 
> https://github.com/pypsa-meets-africa/pypsa-africa/blob/7bc769e9ac9d1878a6cba8188ecf8d155443cdf6/config.default.yaml#L300

Many thanks for the idea! I'll try to play with the solver settings. Currently the continental model is being re-run with a different clusters number.

ekatef:
The network topology now looks like that:
## base network csv data

general view

### fragment a

fragment_a

### fragment b

fragment_b

### fragment c

fragment_c

### fragment d

fragment_d

ekatef:
## base _network

Links in red, transformers in orange

### HVDC as lines

base nc_dc_by_freq

### HVDC as links

base nc_dc_by_freq

ekatef:
## simplified network

### HVDC as lines

elec_s nc_dc_by_freq

### HVDC as links

elec_s nc_dc_by_freq

ekatef:
Test on the continental scale runs but I was not able to achieve any improvements regarding the infeasibility. Gurobi refuses to operate without a proper license for that size of problem (understandable) while working with the glpk interface was found to be not quite obvious :))

An additional problem found during testing for the continent were troubles for HVDC line in Namibia. The weights both from the generator and the load parts are by some reason zero which leads to division by zero:

https://github.com/pypsa-meets-africa/pypsa-africa/blob/66a162573b2cee9bf0c8622a4fb5e3c15193e70f/scripts/cluster_network.py#L173-L178

Have introduced a quick fix but I think it makes sense to look a bit more careful into that. Not necessarily as a part of this PR :)

ekatef:
_**Note to the network pictures above**_: now connections of the DC buses to the lines were improved. The result is that the DC buses are kept in the network model without any additional settings (like a requirement to match the carrier during the stubs removal).

davide-f:
@ekatef Very nice, few comments:
- why in fragment a there are 4 AC nodes? Do we have 4 different voltages + DC?
- Between the lines and link implementation, the location of buses change, how do you think it can be justified?
- The Namibia problem is related for sure to the fact that "NA" is interpreted as NaN for some reason... We already so that issue and we will solve it.

ekatef:
@davide-f, many thanks for checking! That's is always a bit of a challenge to finalise properly something more or less big  :) And your support is a huge help in that. 

To answer point-by-point:

- Yes, exactly. There are the voltage levels 132, 220, 400, 500 kV while the converter connects AC and DC nodes corresponding to the HV.
- I can confirm that this transformation is the result of `simplify_links()` application and plan to look into it a bit deeper.
- Good to know! Have removed the temporary fix.

config.default.yaml
```diff
@@ -91,6 +91,7 @@ electricity:
   co2limit: 7.75e+7  # European default, 0.05 * 3.1e9*0.5, needs to be adjusted for Africa
   co2base: 1.487e9  # European default, adjustment to Africa necessary
   agg_p_nom_limits: data/agg_p_nom_minmax.csv
+  hvdc_as_lines: true
```
davide-f: Please add a description of the setting

config.default.yaml
```diff
@@ -128,6 +129,7 @@ lines:
     220.: "Al/St 240/40 2-bundle 220.0"
     300.: "Al/St 240/40 3-bundle 300.0"
     380.: "Al/St 240/40 4-bundle 380.0"
+  dc_type: "DC_custom_linetype"
```
davide-f, line 132: Also here, please, add a description and options for the values

config.default.yaml
davide-f: This applies also to the other config*.yaml files

doc/release_notes.rst
```diff
@@ -13,6 +13,8 @@ Upcoming Release
 
 **New Features and major Changes**
 
+* Enable DC carrier in the network model and include converters into the model `PR #392 <https://github.com/pypsa-meets-africa/pypsa-africa/pull/392>`__
```
davide-f: This line should go at the end; most likely there are merge conflicts, please check

envs/environment.yaml
```diff
@@ -78,6 +78,7 @@ dependencies:
   # - git+https://github.com/pypsa/pypsa.git#egg=pypsa
   - git+https://github.com/davide-f/powerplantmatching.git@new_pypsa_africa#egg=powerplantmatching
   - git+https://github.com/davide-f/atlite.git@master
+  - git+https://github.com/ekatef/PyPSA.git@master
```
davide-f, line 81: Do we need this?

scripts/base_network.py
```diff
@@ -280,6 +291,25 @@ def _set_electrical_parameters_transformers(transformers):
...
+
+    # Converters are combined with links
+    converters["under_construction"] = False
+    converters["underground"] = False
+
+    return converters
+
+
+# TODO Do we need such a cross-border function?
+# def _replace_b2b_converter_at_country_border_by_link(n):
```
davide-f: Aren't converters already links?

scripts/base_network.py
```diff
@@ -405,15 +441,22 @@ def _rebase_voltage_to_config(component):
...
     buses = _load_buses_from_osm().reset_index()
     lines = _load_lines_from_osm(buses)
-    lines_dc = _load_links_from_osm(buses)
     transformers = _load_transformers_from_osm(buses)
+    converters = _load_converters_from_osm(buses)
+
+    lines_ac = lines[lines.tag_frequency.astype(float) != 0]
+    lines_dc = lines[lines.tag_frequency.astype(float) == 0]
+
+    # TODO Find a way to load a custom linetype. Currently the custom type definition seems to be ignored
```
davide-f: Should this todo remain?
If so please add an issue to track this point

scripts/cluster_network.py
```diff
@@ -366,8 +367,11 @@ def reduce_network(n, buses):
         return nr
 
     def busmap_for_country(x):
+
+        # A number of the countries in the clustering can be > 1
         if isinstance(n_clusters, pd.Series):
-            n_cluster_c = n_clusters[x.name]
+            # TODO Verify why n_clusters[x.name] is not working properly
+            n_cluster_c = n_clusters[x.country][0]
```
davide-f: Is this change needed?

scripts/simplify_network.py
```diff
@@ -391,10 +391,11 @@ def busmap_by_stubs(network, matching_attrs=None):
     def attrs_match(u, v):
         return (
             matching_attrs is None
+            # TODO account for a group length of more than just one pair
             or (
                 network.buses.loc[u, matching_attrs]
                 == network.buses.loc[v, matching_attrs]
-            ).all()
```
davide-f, line 397: Do we need this change?
I feel like if we include this change, in the case of multiple inputs we loose the original flow

davide-f (changes requested):
@ekatef Awesome! Few comments before merging :)

config.default.yaml
ekatef: Done!

envs/environment.yaml
ekatef: Unfortunately, yes. We need a custom linetype for HVDC lines. The only proper way to do this seems to be a PR to the main PyPSA repository. But I have first to find the reference data on electrical properties of the HVDC cables which wasn't done yet

scripts/base_network.py
ekatef: The function cited in the TODO is a single cross-border converted that by some reasons had to be transformed into a link in the PyPSA-Eur model. That looks like quite a specific case, and I think we can remove this TODO safely.

scripts/base_network.py
ekatef: It seems that there is currently an option to load a custom line type. Removed this TODO

scripts/cluster_network.py
ekatef: Yes, that's needed because we have switched-on the sub-networks. Have changed subsetting back from `[x.country][0]` to `[x.name[0]]` for better match with the initial implementation but `[x.name]` is still not working. Our implementation differs from PyPSA-Eur one with some details. Probably this needs to be investigated better as a warning `ConvergenceWarning: Number of distinct clusters (...) found smaller than n_clusters (...). Possibly due to duplicate points in X. kmeans.fit(points)` appears for some clusters values if a group of countries is considered.

scripts/simplify_network.py
ekatef: Have revised to make it work both for a single and for a multiple match.

doc/release_notes.rst
ekatef: Have moved the note to the end of the release changes but not sure that have understood you properly. If I haven't let me know please :)

envs/environment.yaml
davide-f: Ok for this PR.

Please also open a PR on PyPSA (they would also love that) so that once the PR is merged we can move the dependency to the original one.
Please, link it as a comment and track it so that once merged, it is possible to change the dependency.

scripts/base_network.py
davide-f: Ok

scripts/cluster_network.py
davide-f: Did you activate subnetworks in this branch? because I have never experienced this issue

scripts/simplify_network.py
davide-f: Perfect :)

doc/release_notes.rst
davide-f: Perfect! :)

config.default.yaml
davide-f: @ekatef I think this comment was missing, all the rest seems to be already covered except one comment below

scripts/cluster_network.py
ekatef: Yes, it was needed to get proper management of the AC and DC nodes. I have commented this row

https://github.com/pypsa-meets-africa/pypsa-africa/blob/66a162573b2cee9bf0c8622a4fb5e3c15193e70f/scripts/cluster_network.py#L350

and uncommented this one

https://github.com/pypsa-meets-africa/pypsa-africa/blob/66a162573b2cee9bf0c8622a4fb5e3c15193e70f/scripts/cluster_network.py#L396

envs/environment.yaml
ekatef: Thank you :) Ok!

scripts/cluster_network.py
davide-f: Ok!

davide-f (approved):

davide-f:
@ekatef super! PR merged, thank you very much for the great work!

ekatef:
@davide-f thank you for supporting it :) It was a pleasure to contribute!

PR #429, opened 2022-08-02
Allowing for arbitrary shapes that are not default GADM

hazemakhalek:
In this pull request, I tried to add different shapefiles to cluster with. In this case, `MAR2.geojson` represents the updated GADM level 2 regions for Morocco.

The file `MAR2.geojson` needs to be added to the folder `/resources/shapes` and the corresponding lines in the snakefile uncommented. 

Now the 3 following cases work well.
1) Alternative clustering = False
2) Alternative clustering = True (gadm_shapes in snakefile fetches from normal location)
3) Alternative clustering = True (shapes fetched from custom GADM file `MAR2.geojson`) 

The main change introduced is to always fetch the shape file as a snakemake input and not to use get_gadm_layer func. This is a first step towards generalizing the clustering algorithm with shapefiles.

## Checklist


hazemakhalek:
need to revert name environment.yaml to pypsa-africa. Waiting for CI

envs/environment.yaml
```diff
@@ -8,7 +8,6 @@ channels:
 - bioconda
 - gurobi
 dependencies:
-- python>=3.8
```
davide-f: Please, revert this change

Snakefile
```diff
@@ -240,6 +240,7 @@ rule build_bus_regions:
         country_shapes="resources/shapes/country_shapes.geojson",
         offshore_shapes="resources/shapes/offshore_shapes.geojson",
         base_network="networks/base.nc",
+        #gadm_shapes="resources/shapes/MAR2.geojson",
```
davide-f, line 243: Where applicable, instead of leaving this comment, in the same line of gadm_shapes below, may we add a comment on how to change these shapes with custom shapes? [This applies also for the following rules]
I was wondering also if the name "gadm" may be misleading in general. That may be especially true for cluster network, maybe cluster_shapes may be more appropriate? what do you think also?

davide-f (commented):
Good @Hazem!

Additional comments:
- Please remove MAR2.geojson
- First we merge the other PR (I see it includes such changes)
- Add the release_notes (it may be a change on the line of the previous PR)

hazemakhalek:
All the comments should be addressed now. There was a small version conflict in build_power_plants but is solved now too. @davide-f

Snakefile
davide-f: @Hazem-IEG Very good! May you add the direct link to the file to download? I think it would be way faster to use

davide-f (approved):

PR #432, opened 2022-08-05
Add seperate test for pypsa-earth-sec

energyLS:
This PR includes a new test config required to test pypsa-earth-sec. The country list is narrowed down to `MA`.

## Checklist

### DIFF
--- a/test/config.test1_sec.yaml
+++ b/test/config.test1_sec.yaml
@@ -1,0 +1,402 @@
+# SPDX-FileCopyrightText: : 2021 The PyPSA-Africa Authors
+#
+# SPDX-License-Identifier: CC0-1.0
+
+
+version: 0.0.2
+tutorial: true
+
+logging:
+  level: INFO
+  format: "%(levelname)s:%(name)s:%(message)s"
+
+scenario:
+  simpl: ['']
+  ll: ['copt']
+  clusters: [10]
+  opts: [Co2L-24H]
+
+countries: ["MA"]
+
+  # ['DZ', 'AO', 'BJ', 'BW', 'BF', 'BI', 'CM', 'CF', 'TD', 'CG', 'CD',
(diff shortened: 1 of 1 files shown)


pz-max (approved):
Looks good. Just realize though that this is not a sustainable solution. New developments require updating more and more config files. I think we could use in future one of the default configs & add/replace some values required for the test... a shell script or so can then merge the configs leading to the config which will be tested.

Something to worry about in future

davide-f:
I added some comments in discord, I copy them here:

My recommendation to add a MA-tutorial package is the following:
- Create a new cutout only for MA and store it in google drive; the new cutout shall be created with the same parameters as for the NG-test case but only for the MA case (the region is automatically selected by the list of countries)
- create a new bundle_data for MA cutouts: in configs/bundle_config add a new entry similarly to bundle_cutouts_tutorial_MA: but specific for MA
- Verify the coverage of the current data tutorial files: the tutorial data in the bundle_data_tutorial of config/config_bundle.yaml must be verified to be compatible with MA. In particular, the file eez_v11.gpkg shall contain the shape for MA, GEBCO_2021 shall contain data for MA, similarly to the PROBAV_LC100_global_v3.0.1_2019-nrt_Discrete-Classification-map_EPSG-4326.tif file and the Africa.nc file. If MA data are not available (they may have been removed to reduce the size of the package), they need to be created starting from the non-tutorial data and a new bundle_data shall be added.
Moreover, be aware that in the -sec version the testing yaml file is test/config.test1.yaml and it has not the countries feature set

The proposed new config file has the countries changed, but the databundles for MA are not available for the tutorial case.
The easy solution is to use the full non-tutorial bundle, but that takes a long time and >30Gb hard drive for the CI to run.
The cleanest approach would be with what defined above.
In the -sec version, if the CI is meant to be scalable and we want to use MA, the tutorial-bundle for MA is required

@pz-max I disabled the automerge ; if you feel like to merge anyway ok, but this test may fail or may not test the workflow as it should if used

pz-max:
Closed by https://github.com/pypsa-meets-africa/pypsa-earth-sec/pull/135

PR #439, opened 2022-08-12
Add NG optimization plots for validation

pz-max:
## Checklist


davide-f (approved):
@pz-max Are the plotting scripts compatible with the plotting here?
I think it would be awesome and advisable to integrate also such features into plotting and make summaries that we have basically disregarded till now.

I approve the changes, yet I'd love to have these plots being easy to obtain with the workflow

pz-max:
> @pz-max Are the plotting scripts compatible with the plotting here? I think it would be awesome and advisable to integrate also such features into plotting and make summaries that we have basically disregarded till now.
> 
> I approve the changes, yet I'd love to have these plots being easy to obtain with the workflow

At the moment, the plots sometimes require custom implementations. Generalizing them is possible and would take a bit more time, but is not a priority. I think PyPSA and PyPSA-Eur have some neat basic plots that we inherited already. Definitely an endeavour for the future.

PR #465, opened 2022-10-01
landlock countries fix

DeniseGiub:
## Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,6 +6,6 @@ jobs:
     name: A job to automate contrib in readme
     steps:
     - name: Contribute List
-      uses: akhilmhdh/contributors-readme-action@v2.2
+      uses: akhilmhdh/contributors-readme-action@v2.3.6
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/README.md
+++ b/README.md
@@ -289,6 +289,13 @@ The documentation is available here: [documentation](https://pypsa-meets-earth.r
         </a>
     </td>
     <td align="center">
+        <a href="https://github.com/DeniseGiub">
+            <img src="https://avatars.githubusercontent.com/u/113139589?v=4" width="100;" alt="DeniseGiub"/>
+            <br />
+            <sub><b>Null</b></sub>
+        </a>
+    </td>
+    <td align="center">
         <a href="https://github.com/Tooblippe">
             <img src="https://avatars.githubusercontent.com/u/805313?v=4" width="100;" alt="Tooblippe"/>
             <br />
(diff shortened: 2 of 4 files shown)


scripts/build_bus_regions.py
```diff
@@ -210,87 +140,50 @@ def get_id(coords):
...
 
         c_b = n.buses.country == country
-        if n.buses.loc[c_b & n.buses.substation_lv, ["x", "y"]].empty:
-            _logger.warning(f"No low voltage buses found for {country}!")
-            continue
 
         onshore_shape = country_shapes[country]
         onshore_locs = n.buses.loc[c_b & n.buses.substation_lv, ["x", "y"]]
-        if snakemake.config["cluster_options"]["alternative_clustering"]:
-            onshore_geometry = get_gadm_shape(onshore_locs, gadm_shapes)[0]
```
pz-max: This deletes the GADM clustering functionalities (see our preprint)
We could create a test for checking `["alternative_clustering"]: True` to detect such breaking changes in future.

pz-max (changes requested):
Changes look good. However, the code would break if the config.yaml clusters GADM shapes. To do: 
- [x] create a test for alternative_clustering: True
- [ ] Make the whole workflow working with alternative_clusterin: True (undo a couple of changes)

davide-f:
> Changes look good. However, the code would break if the config.yaml clusters GADM shapes. To do:
> 
> * [ ]  create a test for alternative_clustering: True
> * [ ]  Make the whole workflow working with alternative_clusterin: True (undo a couple of changes)

@DeniseGiub Great Job!

@pz-max for flexibility purposes, rather than creating a new file, which may lead to duplication, we could do some hacking to create it dynamically with the CI.
The other test config would be copy and paste of the other one, with few parameters changed; if we change the config file then we have to change 2 files instead of one.
This is more "hacking", we can also start with a duplicate of the test, that should be easy for Denise to do.
@DeniseGiub to do that, you would need to add a test config file in the test folder and modify the .github/workflow/* config files to run also the additional script, see https://github.com/pypsa-meets-earth/pypsa-earth/blob/e35545c3bcfa14b6eef714f28d95ef807db450cb/.github/workflows/ci-linux.yaml#L64-L68

pz-max:
> @pz-max for flexibility purposes, rather than creating a new file, which may lead to duplication, we could do some hacking to create it dynamically with the CI. The other test config would be copy and paste of the other one, with few parameters changed; if we change the config file then we have to change 2 files instead of one.

We redesigned the testing procedure in this PR: https://github.com/pypsa-meets-earth/pypsa-earth/pull/466
It will be merged after a successful CI run.

PR #466, opened 2022-10-03
New test config design

pz-max:
##### Aim: easy testing of multiple functionalities without copying the config 

Feature described now in the documentation:
Add a new test if you want to contribute new functionality to the config.
We perform currently *multiple* integration tests which means various workflows need to work.
All test configs are build by updating the ``config.tutorial.yaml`` with the configs in ``pypysa-earth/test/*.yaml``.

  * You can test your contribution locally with ``snakemake --cores 4 run_tests``. This will build test configs and executes them.
  * Run ``snakemake -j1 build_test_configs`` to build and analyse locally the test configs.

To contribute a test:

1. Provide a new test in ``test/<new test>.yaml``, or adjust one of the existing ones. These tests update the config.tutorial.yaml to test other options e.g. landlock countries. 
2. Add a new test config path to the ``rule build_all_test`` in the ``Snakefile``.
3. If your functionality should be tested in the CI for every pull request, add a respective code in ``.github/workflows/ci-linux.yaml``. We test all functionalities only for Linux while providing a general test for windows and mac.

## Checklist


scripts/build_renewable_profiles.py
```diff
@@ -240,7 +240,7 @@ def get_hydro_capacity_annual_hydro_generation(
             fn,
             comment="#",
             keep_default_na=False,
-            na_values=["-"],
+            na_values=["-",""],
```
pz-max: @davide-f small bug fix that caused issues. 
Scaling on "hydro_capcities" did not work before.

Will add in a moment a CI with the option 'hydro_capcity' and one with 'eia'

pz-max:
As discussed with @davide-f , we need to modify the run_test rule

Snakefile
```diff
@@ -681,15 +695,17 @@ rule plot_summary:
         "scripts/plot_summary.py"
 
 
-rule plot_network:
+rule build_test_configs:
```
davide-f, line 712: This rule may be generalized.
As input we may provide only two elements:
1. `base_config="config.tutorial.yaml"`: it represents base configuration file, hence the one called tutorial now
2. `file_changes_list=[...]`: a list of strings, each of them being the changes with respect to the tutorial. With respect to what already done I would mean: `["test/config.standard.yaml", ..., "test/tmp/config.landlock_tmp.yaml"]`

Then, in the build_test_configs, we can preload the base configuration as done now and loop over the `file_changes_list`.
For every file, we merge the changes into a brand new base configuration file and we can create the config file in the test folder as you are doing now.

I think this procedure is flexible and easy to extend with multiple test configurations beyond the existing one

scripts/build_test_configs.py
```diff
@@ -0,0 +1,76 @@
...
+if __name__ == "__main__":
+    if "snakemake" not in globals():
+        from _helpers import mock_snakemake
+
+        os.chdir(os.path.dirname(os.path.abspath(__file__)))
+        snakemake = mock_snakemake("build_test_configs")
+
+    # Input paths
+    fptutorial = Path(Path.cwd(), snakemake.input.tutorial)
+    fp0 = Path(Path.cwd(), snakemake.input.test_standard)
```
davide-f: See the previous comment.
In this file, the same code for fp0-3 is basically repeated. I think we could rewrite it using a loop.

davide-f (changes requested):

pz-max:
@davide-f requested changes are built in. Thanks :) 
I quite like the change because the user needs to do one step less:

1. provide a new `test\test.yaml`
2. add to the `Snakefile` in the `build_all_test` the new test config path
3. add the test to the CI

We could get rid of step 2 with some additions (probably not necessary atm) and also of step 3 if we test all cases in the CI (we said we don't want to do that atm -- to many data request from gadm and co... Maybe after we host our own cases we could automate that as well)

davide-f:
Ah, please check the plot_network rule: it may have been deleted.

Agree. If you check the plot network, then the pr is ready I think

pz-max:
Will fix it 🙂 thanks for the awesome review

Sent from Outlook for iOS<https://aka.ms/o0ukef>

pz-max:
@davide-f done + documentation added on how people add tests & how it works :100:

Snakefile
```diff
@@ -693,3 +707,23 @@ rule plot_network:
         "logs/plot_network/elec_s{simpl}_{clusters}_ec_l{ll}_{opts}_{attr}_{ext}.log",
     script:
         "scripts/plot_network.py"
+
+
+rule build_test_configs:
+    input:
+        base_config="config.tutorial.yaml",
+        update_file_list=[
+            "test/config.standard.yaml",
```
davide-f: Nice code! I think now it is almost done :) I'd have a last question.

Why do we need the test/config.standard.yaml?
I see in the code that the tutorial being merged with the config.standard.yaml is the standard for all files and there is the special case update_file_list[0] that is treated differently than the rest.
Could we make the config.tutorial.yaml the base configuration and get rid of the test/config..standard.yaml?
That may remove the special case of the [0] and make the code cleaner in my opinion

pz-max: Well spotted. Yes the standard was not necessary in case we test based on the tutorial immediately :)

davide-f (approved):
Well done! :D

davide-f:
Merging after CI completes

PR #509, opened 2022-11-04
Fixing generation of natura raster

### LINKED ISSUE #442: Checking correspondence between natura.tiff and cutout areas

ekatef:
# Closes #442

Discrepancy between the cutout and natura.tiff extents can lead to errors when generating renewable profiles. This PR is intended to prevent this by

* fixing a coordinates flip in natura.tiff calculation;
* adding a check that natura.tiff was generated using a proper CRS and fully covers the cutout area.

## Checklist


ekatef:
@davide-f, that is a version of the initial natura.tiff PR recreated after some git troubles. Would be very grateful for the review (although, the content hopefully should be the same as when you have seen it previously...)

davide-f:
@ekatef there are some conflicts with the main branch, may you revise them?
I'll add comments in the code

ekatef:
@davide-f done

scripts/build_renewable_profiles.py
```diff
@@ -420,6 +424,27 @@ def get_hydro_capacity_annual_hydro_generation(
         if "natura" in config and config["natura"]:
             excluder.add_raster(paths.natura, nodata=0, allow_no_overlap=True)
 
+            natura = rio.open(paths.natura)
```
davide-f, line 435: The more I think to it, the more I think that at user-level, the check should be done between the africa_shape and the cutout and natura separately.
An error shall be thrown if the africa_shape is not contained in both cutout and natura.

If the two shapes do not overlap it's not an issue

ekatef: @davide-f Agree. From the user's perspective we should care about the area of interest. Apart of that, if the provided natura.tiff does not match the africa_shape, we should recomment to use the default global natura.tiff

ekatef: If you don't have any objections, I'd open a new PR with a check of a proper match between the region and natura.tiff & cutout

davide-f: I'm ok with that; the comparison between natura and cutout is more for developers and it may actually harm the user:
if the two do not overlap, but the countries are contained in both, it's ok.

ekatef:
Closed to be re-structured for better match with the use cases requirements

PR #520, opened 2022-11-17
Fix geo-geometry issues

ekatef:
It has been found during workflow testing on data around the world that underlaying geographical shapes sometimes can be sometimes problematic, namely:

1) it looks like the `GID_0` feature in the gadm file may sometimes have values which do not exactly correspond to the country name. As a result, the`GADM_ID` in the `gadm_shapes.geojson` in may have some unexpected values like `"Z02.28_1"` or `"Z03.28_1"` (taken from data for China), and  leads in `country_shapes.geojson` file to zeros for the population value and `"name": null`. This problem may be reproduced by running `build_shapes` for China or for India and leads to numerous troubles along the workflow.

2) sometimes the country area is a multiply connected region (an area with holes) which is treated by shapely as invalid geometry. In particular, that is the case for Dahagram–Angarpota Bangladeshi enclave in India.

This pull request is intended to fix such regional-specific issues.

## Checklist

- [ ] I tested my contribution locally and it seems to work fine
    - [ ] for India
    - [ ] for China


pz-max:
Great spot. Maybe if this is an inconsistency in GADM we should report it to UC Berkley/ the GADM data provider.

ekatef:
@pz-max despite it looks a bit strange, yet there is method in it 😄 

Actually, the context looks like:

```
{ "GADM_ID": "CHN.27_1", ...}
{ "GADM_ID": "CHN.28_1", ...}
{ "GADM_ID": "Z02.28_1", ...}
{ "GADM_ID": "Z03.28_1", ...}
{ "GADM_ID": "CHN.29_1", ...}
{ "GADM_ID": "Z03.29_1", ...}
{ "GADM_ID": "Z08.29_1", ...}
{ "GADM_ID": "CHN.30_1", ...}
{ "GADM_ID": "CHN.31_1", ...}
```

My think it may be that these unusual indices are parts of the regions with the corresponding number splitted from the main region area for some reason. So, there may be just no "appropriate" name available. 

Agree that it's worth to investigate a bit deeper!

davide-f:
As discussed, the filtering may better occur in the build_shapes script.
Since these changes would be moved to the build_shape script, would it be better to start from a new PR? just to have a cleaner history

ekatef:
Agree. Probably, it makes also sense to split it on two PRs as currently it address two independent issues (a real geometry with holes and missed geometry fields)

ekatef:
Will be reopened in a cleaned version

Issue #531, opened 2022-12-02
Verify the effect of using/disregarding clustering by subnetworks in cluster_network


davide-f:
## Describe the Bug

In cluster_network [L243](https://github.com/pypsa-meets-earth/pypsa-earth/blob/5f2a13ab0b1d54c42d0ded80edd6143443f69f9b/scripts/cluster_network.py#L243) and [L403](https://github.com/pypsa-meets-earth/pypsa-earth/blob/5f2a13ab0b1d54c42d0ded80edd6143443f69f9b/scripts/cluster_network.py#L403), the clustering may be by country AND sub_network, as done in PyPSA-Eur.
In PyPSA-Earth we disregard this for simplicity, yet it now needs some study.
Moreover, the clustering shall address the improper matching of buses of different carrier types.

Consequences shall be investigated at least on:
- a sample test case such as the tutorial NG+BJ
- another country example with DC lines (es DRC or Brazil)

The tests that shall be done are:
1. verifying the total capacities of generation and networks after add_electricity (hence elec.nc), after the simplify (elec_s{}.nc} and after the clustering
2. verifying what happens when DC lines belong to the clustered nodes. Most likely, there will be the need for some fix here.

This issue is a remark that leverages on #506 and #530

ekatef:
There is an issue for a run considering China under a conservative assumption "all AC -> DC" due to mismatch of the sub-networks with `AssertionError: In Bus cluster sub_network the values of attribute sub_network do not agree`. The issue may be resolved by an increase of the clusters number from 10 to 20.

The network looks like that:

<img width="521" alt="image" src="https://user-images.githubusercontent.com/30229437/206310352-9368cf38-39da-4aa7-8a5b-537458d96e4a.png">

The full error listing:
```
INFO:snakemake.logging:
This is the repository path:  ~pypsa-earth
Had to go 0 folder(s) up.
INFO:pypsa.io:Imported network elec_s.nc has buses, carriers, generators, lines, loads, storage_units
Traceback (most recent call last):
  File "~pypsa-earth/.snakemake/scripts/tmp85yv8yyb.cluster_network.py", line 581, in <module>
    clustering = clustering_for_n_clusters(
  File "~pypsa-earth/.snakemake/scripts/tmp85yv8yyb.cluster_network.py", line 448, in clustering_for_n_clusters
    clustering = get_clustering_from_busmap(
[...]

Issue #528, opened 2022-12-01
Add a testing procedure to ensure consistency of the supplied inputs


ekatef:
## Describe the feature you'd like to see

It appears that the workflow may currently fail in a not very gracious way due to inconsistencies in the supplied inputs. It may be worth to add detection of such data flaws and communicate it to a user in a clear way and/or to propose a fix.

In particular:
- [ ] the provided country list corresponds to a single connected region;
- [ ] the requested time frames match each other in snapshot, cutout and load weather year settings;
- [ ] a proper natura.tiff should be used (recommend to use the global one);
- [ ] a supplied cutout correspond to the area requested;
- [ ] possible data loss in the cutout archive should be handled properly;
- [ ] hydro_capacities.csv should be used for Africa only, for all other regions the EIA data should be used; 
- [ ] DC-lines representation may be not consistent with alternative [administrative-based] clustering.

ekatef:
An example of the error which appears when executing `build_renewable_profiles` if a provided cutout doesn't match to a modeling domain (defined by the `countries` list):

```
INFO:snakemake.logging:[Fri Dec  2 11:16:46 2022]
rule build_renewable_profiles:
    input: networks/base.nc, resources/natura.tiff, data/copernicus/PROBAV_LC100_global_v3.0.1_2019-nrt_Discrete-Classification-map_EPSG-4326.tif, data/gebco/GEBCO_2021_TID.nc, resources/shapes/country_shapes.geojson, resources/shapes/offshore_shapes.geojson, data/hydro_capacities.csv, data/eia_hydro_annual_generation.csv, resources/powerplants.csv, resources/bus_regions/regions_offshore.geojson, cutouts/ear-2013-era5.nc
    output: resources/renewable_profiles/profile_offwind-dc.nc
    log: logs/build_renewable_profile_offwind-dc.log
    jobid: 16
    benchmark: benchmarks/build_renewable_profiles_offwind-dc
    reason: Input files updated by another job: resources/bus_regions/regions_offshore.geojson, resources/powerplants.csv
    wildcards: technology=offwind-dc
    resources: tmpdir=/var/folders/qn/vpndfm21795ckkq89np1ckp40000gn/T, mem_mb=20000
INFO:snakemake.logging:rule build_renewable_profiles:
[...]

PR #532, opened 2022-12-03
Fix geometry issues

ekatef:
It has been found during workflow testing on data around the world that underlaying geographical shapes sometimes can be sometimes problematic, namely:

1) it looks like the `GID_0` feature in the gadm file may sometimes have values which do not exactly correspond to the country name. As a result, the`GADM_ID` in the `gadm_shapes.geojson` in may have some unexpected values like `"Z02.28_1"` or `"Z03.28_1"` (taken from data for China), and  leads in `country_shapes.geojson` file to zeros for the population value and `"name": null`. This problem may be reproduced by running `build_shapes` for China or for India and leads to numerous troubles along the workflow.

2) sometimes the country area is a multiply connected region (an area with holes) which is treated by shapely as invalid geometry. In particular, that is the case for Dahagram–Angarpota Bangladeshi enclave in India.

3) it may happen that Voronoi partition leads to empty polygons which are not filtered by `dropna()`, are being written as `none` geometry into `"regions_onshore.geojson"` and cause some troubles propagating along the workflow. In particular, that is the case for Kazakhstan shapes (OSM data on 4.12.2022, 10 clusters).

This pull request is intended to fix such regional-specific issues.

## Checklist

- [x] Clean-up the code
- [x] I tested my contribution locally and it seems to work fine
    - [x] for India (the GADM issue & multply connected regions)
    - [x] for China (the GADM issue)
    - [x] for Kazakhstan (the empty geometry issue)
    - [x] for Malaysia (no issues, a complex shape)

### DIFF
--- a/scripts/build_bus_regions.py
+++ b/scripts/build_bus_regions.py
@@ -262,6 +262,7 @@ if __name__ == "__main__":
         crs=country_shapes.crs,
     ).dropna(axis="index", subset=["geometry"])
 
+    # sometimes Voronoi partition may give empty polygons
     onshore_regions = pd.concat([onshore_regions[~onshore_regions.is_empty]], ignore_index=True).to_file(
         snakemake.output.regions_onshore
     )
--- a/scripts/build_bus_regions.py
+++ b/scripts/build_bus_regions.py
@@ -76,6 +76,9 @@ def custom_voronoi_partition_pts(points, outline, add_bounds_shape=True, multipl
     import numpy as np
     from scipy.spatial import Voronoi
     from shapely.geometry import Point, Polygon
+    from shapely.ops import unary_union
+
+    outline = unary_union(outline)
 
     points = np.asarray(points)
 
--- a/scripts/build_bus_regions.py
+++ b/scripts/build_bus_regions.py
@@ -263,9 +263,9 @@ if __name__ == "__main__":
     ).dropna(axis="index", subset=["geometry"])
 
     # sometimes Voronoi partition may give empty polygons
-    onshore_regions = pd.concat([onshore_regions[~onshore_regions.is_empty]], ignore_index=True).to_file(
-        snakemake.output.regions_onshore
-    )
+    onshore_regions = pd.concat(
+        [onshore_regions[~onshore_regions.is_empty]], ignore_index=True
+    ).to_file(snakemake.output.regions_onshore)
 
     if offshore_regions:
         # if a offshore_regions exists excute below
--- a/scripts/build_bus_regions.py
+++ b/scripts/build_bus_regions.py
@@ -262,7 +262,7 @@ if __name__ == "__main__":
         crs=country_shapes.crs,
     ).dropna(axis="index", subset=["geometry"])
 
-    onshore_regions = pd.concat([onshore_regions], ignore_index=True).to_file(
+    onshore_regions = pd.concat([onshore_regions[~onshore_regions.is_empty]], ignore_index=True).to_file(
         snakemake.output.regions_onshore
     )
 
--- a/scripts/build_bus_regions.py
+++ b/scripts/build_bus_regions.py
@@ -270,12 +270,14 @@ if __name__ == "__main__":
             offshore_regions.append(offshore_regions_c)
 
     # create geodataframe and remove nan shapes
-    onshore_regions = gpd.GeoDataFrame(
-        pd.concat(onshore_regions, ignore_index=True),
-        crs=country_shapes.crs,
-    ).dropna(
-    axis="index", subset=["geometry"]
-    ).to_file(snakemake.output.regions_onshore)
+    onshore_regions = (
+        gpd.GeoDataFrame(
+            pd.concat(onshore_regions, ignore_index=True),
+            crs=country_shapes.crs,
+        )
+        .dropna(axis="index", subset=["geometry"])
+        .to_file(snakemake.output.regions_onshore)
+    )
 
     if offshore_regions:
         # if a offshore_regions exists excute below
--- a/scripts/build_bus_regions.py
+++ b/scripts/build_bus_regions.py
@@ -273,11 +273,8 @@ if __name__ == "__main__":
     onshore_regions = gpd.GeoDataFrame(
         pd.concat(onshore_regions, ignore_index=True),
         crs=country_shapes.crs,
-    ).dropna(axis="index", subset=["geometry"])
-
-    # sometimes Voronoi partition may give empty polygons
-    onshore_regions = pd.concat(
-        [onshore_regions[~onshore_regions.is_empty]], ignore_index=True
+    ).dropna(
+    axis="index", subset=["geometry"]
     ).to_file(snakemake.output.regions_onshore)
 
     if offshore_regions:
--- a/scripts/build_osm_network.py
+++ b/scripts/build_osm_network.py
@@ -831,8 +831,12 @@ def built_network(inputs, outputs, geo_crs, distance_crs):
     country_list = input
     bus_country_list = buses["country"].unique().tolist()
 
-    if len(bus_country_list) != len(country_list):
-        no_data_countries = set(country_list).difference(set(bus_country_list))
+    # it may happen that bus_country_list contains entries not relevant as a country name (e.g. "not found")
+    # difference can't give negative values; the following will return only releant country names
+    no_data_countries = set(country_list).difference(set(bus_country_list))
+    
+    if len(no_data_countries) > 0:
+        
         no_data_countries_shape = country_shapes[
             country_shapes.index.isin(no_data_countries) == True
         ].reset_index()
--- a/scripts/build_osm_network.py
+++ b/scripts/build_osm_network.py
@@ -834,9 +834,9 @@ def built_network(inputs, outputs, geo_crs, distance_crs):
     # it may happen that bus_country_list contains entries not relevant as a country name (e.g. "not found")
     # difference can't give negative values; the following will return only releant country names
     no_data_countries = set(country_list).difference(set(bus_country_list))
-    
+
     if len(no_data_countries) > 0:
-        
+
         no_data_countries_shape = country_shapes[
             country_shapes.index.isin(no_data_countries) == True
         ].reset_index()
--- a/scripts/build_shapes.py
+++ b/scripts/build_shapes.py
@@ -24,6 +24,7 @@ from _helpers import (
     three_2_two_digits_country,
     two_2_three_digits_country,
     two_digits_2_name_country,
+    country_name_2_two_digits,
 )
 from rasterio.mask import mask
 from shapely.geometry import LineString, MultiPolygon, Point, Polygon
--- a/scripts/build_shapes.py
+++ b/scripts/build_shapes.py
@@ -812,7 +812,11 @@ if __name__ == "__main__":
 
     country_shapes = countries(countries_list, geo_crs, update, out_logging)
 
-    country_shapes.reset_index().to_file(snakemake.output.country_shapes)
+    # there may be holes in geometry which cause a lot of troubles along the workflow
+    country_shapes_valid = country_shapes.apply(
+        lambda x: make_valid(x) if not x.is_valid else x
+    )
+    country_shapes_valid.reset_index().to_file(snakemake.output.country_shapes)
 
     offshore_shapes = eez(
         countries_list, geo_crs, country_shapes, EEZ_gpkg, out_logging
@@ -836,4 +840,10 @@ if __name__ == "__main__":
         year,
         nprocesses=nprocesses,
     )
+    gadm_shapes.geometry = gadm_shapes.apply(
+        lambda row: make_valid(row.geometry)
+        if not row.geometry.is_valid
+        else row.geometry,
+        axis=1,
+    )    
     save_to_geojson(gadm_shapes, out.gadm_shapes)
--- a/scripts/build_shapes.py
+++ b/scripts/build_shapes.py
@@ -87,6 +87,64 @@ def restore_country_code_by_name(row, country_code):
     else:
         return row["GID_0"]
 
+#  file=file_gpkg, layer=layer_id, cc=country_code 
+def build_gadm_df(file=file_gpkg, layer=layer_id, cc=country_code): 
+    # read gpkg file
+    geodf = gpd.read_file(file, layer="ADM_ADM_" + str(layer)).to_crs(
+        geo_crs
+    )
+
+    # convert country name representation of the main country (GID_0 column)
+    geodf["GID_0"] = [
+        three_2_two_digits_country(twoD_c) for twoD_c in geodf["GID_0"]
+    ]
+
+    # GID_0 may have some exotic values, "COUNTRY" column may be used instead
+    geodf["GID_0"] = geodf.apply(
+        lambda x: restore_country_code_by_name(x, country_code=cc), axis=1
+    )
+
+    # "not found" hardcoded according to country_converter conventions
+    geodf.drop(
+        geodf[geodf["GID_0"] == "not found"].index, inplace=True
+    )
+
+    # create a subindex column that is useful
+    # in the GADM processing of sub-national zones
+    geodf["GADM_ID"] = geodf[f"GID_{layer_id}"]
+
+    if layer_id >= 1:
+        available_gadm_codes = geodf["GADM_ID"].unique()
+        code_three_digits = two_2_three_digits_country(cc)
+
+        # normally the GADM code starts the ISO3
+        non_std_gadm_codes = [
+            w for w in available_gadm_codes if not w.startswith(code_three_digits)
+        ]
+
+        if len(non_std_gadm_codes) > 0:
+
+            # a "minimalistic" approach outputs a single file (~200kb for CN + IN)
+            d = {
+                "country_code": cc,
+                "non_standard_gadm_codes": non_std_gadm_codes,
+            }
+            df = pd.DataFrame(data=d)
+            output_path = "non_standard_gadm_codes.csv"
+            df.to_csv(
+                output_path,
+                mode="a",
+                header=not os.path.exists(output_path),
+                index=False,
+            )
+
+            # a (more useful) approach outputs a file for each of the country (~200kb each for CN and IN)
+            df_filtered = geodf[geodf["GADM_ID"].isin(non_std_gadm_codes)]
+            df_filtered.to_csv(
+                "non_standard_gadm_" + cc + "_raw.csv", index=False
+            )
+        return geodf
+
 
 def get_GADM_layer(country_list, layer_id, geo_crs, update=False, outlogging=False):
     """
@@ -117,61 +175,8 @@ def get_GADM_layer(country_list, layer_id, geo_crs, update=False, outlogging=Fal
             # when layer id is negative or larger than the number of layers, select the last layer
             layer_id = len(list_layers) - 1
 
-        # read gpkg file
-        geodf_temp = gpd.read_file(file_gpkg, layer="ADM_ADM_" + str(layer_id)).to_crs(
-            geo_crs
-        )
-
-        # convert country name representation of the main country (GID_0 column)
-        geodf_temp["GID_0"] = [
-            three_2_two_digits_country(twoD_c) for twoD_c in geodf_temp["GID_0"]
-        ]
-
-        # GID_0 may have some exotic values, "COUNTRY" column may be used instead
-        geodf_temp["GID_0"] = geodf_temp.apply(
-            lambda x: restore_country_code_by_name(x, country_code=country_code), axis=1
-        )
-
-        # "not found" hardcoded according to country_converter conventions
-        geodf_temp.drop(
-            geodf_temp[geodf_temp["GID_0"] == "not found"].index, inplace=True
-        )
-
-        # create a subindex column that is useful
-        # in the GADM processing of sub-national zones
-        geodf_temp["GADM_ID"] = geodf_temp[f"GID_{layer_id}"]
-
-        if layer_id >= 1:
-            available_gadm_codes = geodf_temp["GADM_ID"].unique()
-            code_three_digits = two_2_three_digits_country(country_code)
-
-            # normally the GADM code starts the ISO3
-            non_std_gadm_codes = [
-                w for w in available_gadm_codes if not w.startswith(code_three_digits)
-            ]
-
-            if len(non_std_gadm_codes) > 0:
-
-                # a "minimalistic" approach outputs a single file (~200kb for CN + IN)
-                d = {
-                    "country_code": country_code,
-                    "non_standard_gadm_codes": non_std_gadm_codes,
-                }
-                df = pd.DataFrame(data=d)
-                output_path = "non_standard_gadm_codes.csv"
-                df.to_csv(
-                    output_path,
-                    mode="a",
-                    header=not os.path.exists(output_path),
-                    index=False,
-                )
-
-                # a (more useful) approach outputs a file for each of the country (~200kb each for CN and IN)
-                df_filtered = geodf_temp[geodf_temp["GADM_ID"].isin(non_std_gadm_codes)]
-                df_filtered.to_csv(
-                    "non_standard_gadm_" + country_code + "_raw.csv", index=False
-                )
-
+        geodf_temp = build_gadm_df()
+                    
         # append geodataframes
         geodf_list.append(geodf_temp)
 
--- a/scripts/build_shapes.py
+++ b/scripts/build_shapes.py
@@ -126,6 +126,12 @@ def get_GADM_layer(country_list, layer_id, geo_crs, update=False, outlogging=Fal
             three_2_two_digits_country(twoD_c) for twoD_c in geodf_temp["GID_0"]
         ]
 
+        # GID_0 may have some exotic values, "COUNTRY" column may be used instead
+        geodf_temp["GID_0"] = geodf_temp.apply(
+            lambda x: restore_country_code_by_name(x, country_code=country_code), 
+            axis=1
+        )          
+        
         # create a subindex column that is useful
         # in the GADM processing of sub-national zones
         geodf_temp["GADM_ID"] = geodf_temp[f"GID_{layer_id}"]
--- a/scripts/build_shapes.py
+++ b/scripts/build_shapes.py
@@ -846,5 +846,5 @@ if __name__ == "__main__":
         if not row.geometry.is_valid
         else row.geometry,
         axis=1,
-    )    
+    )
     save_to_geojson(gadm_shapes, out.gadm_shapes)
--- a/scripts/build_shapes.py
+++ b/scripts/build_shapes.py
@@ -868,8 +868,7 @@ if __name__ == "__main__":
     # e.g. that is the case for enclaves like Dahagram–Angarpota for IN/BD
     country_shapes_valid = country_shapes.apply(
         lambda x: make_valid(x) if not x.is_valid else x
-    )
-    country_shapes_valid.reset_index().to_file(snakemake.output.country_shapes)
+    ).reset_index().to_file(snakemake.output.country_shapes)
 
     offshore_shapes = eez(
         countries_list, geo_crs, country_shapes, EEZ_gpkg, out_logging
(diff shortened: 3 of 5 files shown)


ekatef:
The GADM issue means that some GADM_ID may have non-standard GADM_ID, e.g. for India (part of `gadm_shapes`):

```
"GADM_ID": "Z07.3_1", "country": "not found", "pop": 0.0, 
"GADM_ID": "Z04.13_1", "country": "not found", "pop": 0.0,
"GADM_ID": "Z09.13_1", "country": "not found", "pop": 0.0,
"GADM_ID": "Z01.14_1", "country": "not found", "pop": 0.0,
"GADM_ID": "Z05.35_1", "country": "not found", "pop": 0.0,
"GADM_ID": "Z09.35_1", "country": "not found", "pop": 0.0,
```

A visual check gives the following picture (which actually doen't clarify the issue):

india_gadm_id

Regions with non-standard GADM_IDs are highlighted by violet.

ekatef:
It appears that non-standard GADM_IDs cause troubles when being interpreted as country names (e.g. 'not found') and propagating into the cleaned osm datasets. E.g. that is currently the case for India while for China non-standard GADM_IDs seem to remain in the country_shape and gadm_shape only do not causing any harm downstream. Although, such a difference may be a random effect of a particular workflow realisation.

ekatef:
A picture for non-standard GADM_IDs for China looks as follows

image

```
"GADM_ID": "Z02.28_1", "country": null, "pop": 0.0, "gdp": 13450274.0
"GADM_ID": "Z03.28_1", "country": null, "pop": 0.0, "gdp": 45227044.0
"GADM_ID": "Z03.29_1", "country": null, "pop": 0.0, "gdp": 5931376.0 
"GADM_ID": "Z08.29_1", "country": null, "pop": 0.0, "gdp": 3446499.0
```

ekatef:
It looks like a fix is needed in `build_shapes` to get rid of GADM_IDs non-valid as countries names. A current work-around with `len(no_data_countries) > 0` works for `build_osm_network` but it's likely that non-relevant values of the country name will lead to the problems downstream. E.g. if non-standard names are kept into shapes files, `cluster_networks` gives `AssertionError: The following countries have no load: ['not found']`  for `distribute_cluster: ['load']`.

The most strait-forward solution would be to filter by GID_0 (aka country name) values with something like `df_countries = df_countries.query("name in @countries")`. However, it looks like these areas represent meaningful geographical regions and a more data-saving approach may be preferable.

ekatef:
In a current implementation the onshore shape consists of a some shapes each corresponding to a special administrative area (those ones notated with the non-standard GADM_IDs)

image

```
"geometry": { "type": "MultiPolygon", "coordinates": [ [ [ [ 69.
"geometry": { "type": "Polygon", "coordinates": [ [ [ 75.0716094
"geometry": { "type": "Polygon", "coordinates": [ [ [ 78.6513452
"geometry": { "type": "Polygon", "coordinates": [ [ [ 80.0879363
"geometry": { "type": "Polygon", "coordinates": [ [ [ 94.1912460
"geometry": { "type": "Polygon", "coordinates": [ [ [ 78.9089126
```

That leads to some troubles along the workflow. In particular, `build_bus_regions` terminates with an error:

```
rule build_bus_regions:
    input: resources/shapes/country_shapes.geojson, resources/shapes/offshore_shapes.geojson, networks/base.nc, resources/shapes/gadm_shapes.geojson
    output: resources/bus_regions/regions_onshore.geojson, resources/bus_regions/regions_offshore.geojson
    log: logs/build_bus_regions.log
    jobid: 0
    reason: Input files updated by another job: networks/base.nc
    resources: tmpdir=/var/folders/qn/vpndfm21795ckkq89np1ckp40000gn/T, mem_mb=1000
INFO:snakemake.logging:rule build_bus_regions:
    input: resources/shapes/country_shapes.geojson, resources/shapes/offshore_shapes.geojson, networks/base.nc, resources/shapes/gadm_shapes.geojson
    output: resources/bus_regions/regions_onshore.geojson, resources/bus_regions/regions_offshore.geojson
    log: logs/build_bus_regions.log
    jobid: 0
    reason: Input files updated by another job: networks/base.nc
    resources: tmpdir=/var/folders/qn/vpndfm21795ckkq89np1ckp40000gn/T, mem_mb=1000

INFO:snakemake.logging:
INFO:pypsa.io:Imported network base.nc has buses, lines, links, transformers
  File "~pypsa-earth/.snakemake/scripts/tmp0qppxgve.build_bus_regions.py", line 221, in <module>
    onshore_geometry = custom_voronoi_partition_pts(
  File "~pypsa-earth/.snakemake/scripts/tmp0qppxgve.build_bus_regions.py", line 105, in custom_voronoi_partition_pts
    xmin = min(xmin, minx_o)
TypeError: '>' not supported between instances of 'numpy.ndarray' and 'str'
[Thu Dec  8 20:21:24 2022]
INFO:snakemake.logging:[Thu Dec  8 20:21:24 2022]
Error in rule build_bus_regions:
    jobid: 0
    output: resources/bus_regions/regions_onshore.geojson, resources/bus_regions/regions_offshore.geojson
    log: logs/build_bus_regions.log (check log file(s) for error message)

ERROR:snakemake.logging:Error in rule build_bus_regions:
    jobid: 0
    output: resources/bus_regions/regions_onshore.geojson, resources/bus_regions/regions_offshore.geojson
    log: logs/build_bus_regions.log (check log file(s) for error message)

RuleException:
CalledProcessError in line 279 of ~pypsa-earth/Snakefile:
Command 'set -euo pipefail;  /Users/ekaterina/opt/miniconda3/envs/pypsa-earth-pdfix/bin/python3.10 ~pypsa-earth/.snakemake/scripts/tmp0qppxgve.build_bus_regions.py' returned non-zero exit status 1.
  File "~pypsa-earth/Snakefile", line 279, in __rule_build_bus_regions
  File "/Users/ekaterina/opt/miniconda3/envs/pypsa-earth-pdfix/lib/python3.10/concurrent/futures/thread.py", line 58, in run
CalledProcessError in line 279 of ~pypsa-earth/Snakefile:
Command 'set -euo pipefail;  /Users/ekaterina/opt/miniconda3/envs/pypsa-earth-pdfix/bin/python3.10 ~pypsa-earth/.snakemake/scripts/tmp0qppxgve.build_bus_regions.py' returned non-zero exit status 1.
  File "~pypsa-earth/Snakefile", line 279, in __rule_build_bus_regions
  File "/Users/ekaterina/opt/miniconda3/envs/pypsa-earth-pdfix/lib/python3.10/concurrent/futures/thread.py", line 58, in run
```
That issue is caused by a fact that `bounds` method returns a dataframe instead of a single boundary vector. That problem may be easily resolved by unifying outline shapes before applying Voronoi partition.

ekatef:
It could be a good idea to unify all the polygons when generating onshore_shape. The problem is to merge all the geometries simultaneously keeping other features (like `GID_0`).

It's worth to investigate if the geometry consisting of different polygons would lead to any problems along the workflow apart of the mentioned one in `build_bus_regions`.

Apart of that, it may be worth to test the workflow on the countries which consist of different areas (e.g. Malaysia)

ekatef:
It looks like onshore_shape composed of different geometries doesn't lead to other problems along the workflow apart the known one in `build_bus_regions` which can be fixed by a direct application of `unary_union()` just before bound calculation.

However, it appears that for some reasons geometries were duplicated in offshore_shape. Currently offshore_shape consists of five identical geometries which needs to be investigated.

ekatef:
The issue with duplicated offshore geometries was fixed for India. 

However, during tests with `alternative_clustering: true` another error appeared (an error listing is bellow). However, this error persisted after changing the region, e.g. to ["CD"] or ["NA"] but doesn't appears for ["KZ"]. So, it looks like that is connected rather to some specific topology of the power grid (HVDC lines??) than to the geometry issues addressed by this PR. The issue #537 has been created to track this error

ekatef:
There are some weird issues during renewable profiles generation:

* one error appears during generation on a hydro profile: `AttributeError: 'numpy.int64' object has no attribute 'intersects'` when calculating indicator_matrix in `inflow = correction_factor * func(capacity_factor=True, **resource)` (it can be reproduced, e.g. for ["IN"] using `add_to_snakefile: true` for `augmented_line_connection`)
* another error appears during generation of a solar profile `TypeError: ufunc 'linestrings' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''` when forming a linestring `line = LineString([p, regions.loc[bus, ["x", "y"]]]` (that is the case for "KZ" or "NG" even for runs from the main branch which previously worked perfectly)

That is quite weird and most likely linked to some environment issues. #538 attempts to fix it

ekatef:
Update: in the CI run there is currently one of the errors listed above, namely `AttributeError: 'numpy.int64' object has no attribute 'intersects'`. The full listing:

```bash
INFO:snakemake.logging:
[Thu Dec 15 17:07:34 2022]
INFO:snakemake.logging:[Thu Dec 15 17:07:34 2022]
rule build_renewable_profiles:
    input: networks/base.nc, resources/natura.tiff, data/copernicus/PROBAV_LC100_global_v3.0.1_2019-nrt_Discrete-Classification-map_EPSG-4326.tif, data/gebco/GEBCO_2021_TID.nc, resources/shapes/country_shapes.geojson, resources/shapes/offshore_shapes.geojson, data/hydro_capacities.csv, data/eia_hydro_annual_generation.csv, resources/powerplants.csv, resources/bus_regions/regions_onshore.geojson, cutouts/africa-2013-era5-tutorial.nc
    output: resources/renewable_profiles/profile_hydro.nc
    log: logs/build_renewable_profile_hydro.log
    jobid: 20
    benchmark: benchmarks/build_renewable_profiles_hydro
    reason: Missing output files: resources/renewable_profiles/profile_hydro.nc; Input files updated by another job: networks/base.nc, resources/shapes/offshore_shapes.geojson, data/copernicus/PROBAV_LC100_global_v3.0.1_2019-nrt_Discrete-Classification-map_EPSG-4326.tif, resources/powerplants.csv, data/hydro_capacities.csv, cutouts/africa-2013-era5-tutorial.nc, resources/shapes/country_shapes.geojson, data/gebco/GEBCO_2021_TID.nc, resources/bus_regions/regions_onshore.geojson, resources/natura.tiff
    wildcards: technology=hydro
    threads: 2
    resources: tmpdir=/tmp, mem_mb=20000, mem_mib=19074
INFO:snakemake.logging:rule build_renewable_profiles:
    input: networks/base.nc, resources/natura.tiff, data/copernicus/PROBAV_LC100_global_v3.0.1_2019-nrt_Discrete-Classification-map_EPSG-4326.tif, data/gebco/GEBCO_2021_TID.nc, resources/shapes/country_shapes.geojson, resources/shapes/offshore_shapes.geojson, data/hydro_capacities.csv, data/eia_hydro_annual_generation.csv, resources/powerplants.csv, resources/bus_regions/regions_onshore.geojson, cutouts/africa-2013-era5-tutorial.nc
    output: resources/renewable_profiles/profile_hydro.nc
    log: logs/build_renewable_profile_hydro.log
    jobid: 20
    benchmark: benchmarks/build_renewable_profiles_hydro
    reason: Missing output files: resources/renewable_profiles/profile_hydro.nc; Input files updated by another job: networks/base.nc, resources/shapes/offshore_shapes.geojson, data/copernicus/PROBAV_LC100_global_v3.0.1_2019-nrt_Discrete-Classification-map_EPSG-4326.tif, resources/powerplants.csv, data/hydro_capacities.csv, cutouts/africa-2013-era5-tutorial.nc, resources/shapes/country_shapes.geojson, data/gebco/GEBCO_2021_TID.nc, resources/bus_regions/regions_onshore.geojson, resources/natura.tiff
    wildcards: technology=hydro
    threads: 2
    resources: tmpdir=/tmp, mem_mb=20000, mem_mib=19074

INFO:snakemake.logging:
INFO:__main__:Hydro normalization mode hydro_capacities

Determine upstream basins per plant: 0it [00:00, ?it/s]
Determine upstream basins per plant: 3it [00:00, 201.72it/s]
  File "/home/runner/work/pypsa-earth/pypsa-earth/.snakemake/scripts/tmpndxdiqdd.build_renewable_profiles.py", line 455, in <module>
    inflow = correction_factor * func(capacity_factor=True, **resource)
  File "/usr/share/miniconda3/envs/pypsa-earth/lib/python3.10/site-packages/atlite/convert.py", line 805, in hydro
    matrix = cutout.indicatormatrix(basins.shapes)
  File "/usr/share/miniconda3/envs/pypsa-earth/lib/python3.10/site-packages/atlite/cutout.py", line 536, in indicatormatrix
    return compute_indicatormatrix(self.grid, shapes, self.crs, shapes_crs)
  File "/usr/share/miniconda3/envs/pypsa-earth/lib/python3.10/site-packages/atlite/gis.py", line 148, in compute_indicatormatrix
    if o.intersects(d):
AttributeError: 'numpy.int64' object has no attribute 'intersects'
[Thu Dec 15 17:07:40 2022]
INFO:snakemake.logging:[Thu Dec 15 17:07:40 2022]
Error in rule build_renewable_profiles:
    jobid: 20
    input: networks/base.nc, resources/natura.tiff, data/copernicus/PROBAV_LC100_global_v3.0.1_2019-nrt_Discrete-Classification-map_EPSG-4326.tif, data/gebco/GEBCO_2021_TID.nc, resources/shapes/country_shapes.geojson, resources/shapes/offshore_shapes.geojson, data/hydro_capacities.csv, data/eia_hydro_annual_generation.csv, resources/powerplants.csv, resources/bus_regions/regions_onshore.geojson, cutouts/africa-2013-era5-tutorial.nc
    output: resources/renewable_profiles/profile_hydro.nc
    log: logs/build_renewable_profile_hydro.log (check log file(s) for error details)

ERROR:snakemake.logging:Error in rule build_renewable_profiles:
    jobid: 20
    input: networks/base.nc, resources/natura.tiff, data/copernicus/PROBAV_LC100_global_v3.0.1_2019-nrt_Discrete-Classification-map_EPSG-4326.tif, data/gebco/GEBCO_2021_TID.nc, resources/shapes/country_shapes.geojson, resources/shapes/offshore_shapes.geojson, data/hydro_capacities.csv, data/eia_hydro_annual_generation.csv, resources/powerplants.csv, resources/bus_regions/regions_onshore.geojson, cutouts/africa-2013-era5-tutorial.nc
    output: resources/renewable_profiles/profile_hydro.nc
    log: logs/build_renewable_profile_hydro.log (check log file(s) for error details)

RuleException:
CalledProcessError in file /home/runner/work/pypsa-earth/pypsa-earth/Snakefile, line 325:
Command 'set -euo pipefail;  /usr/share/miniconda3/envs/pypsa-earth/bin/python3.10 /home/runner/work/pypsa-earth/pypsa-earth/.snakemake/scripts/tmpndxdiqdd.build_renewable_profiles.py' returned non-zero exit status 1.
  File "/home/runner/work/pypsa-earth/pypsa-earth/Snakefile", line 325, in __rule_build_renewable_profiles
  File "/usr/share/miniconda3/envs/pypsa-earth/lib/python3.10/concurrent/futures/thread.py", line 58, in run
CalledProcessError in file /home/runner/work/pypsa-earth/pypsa-earth/Snakefile, line 325:
Command 'set -euo pipefail;  /usr/share/miniconda3/envs/pypsa-earth/bin/python3.10 /home/runner/work/pypsa-earth/pypsa-earth/.snakemake/scripts/tmpndxdiqdd.build_renewable_profiles.py' returned non-zero exit status 1.
  File "/home/runner/work/pypsa-earth/pypsa-earth/Snakefile", line 325, in __rule_build_renewable_profiles
  File "/usr/share/miniconda3/envs/pypsa-earth/lib/python3.10/concurrent/futures/thread.py", line 58, in run
Complete log: .snakemake/log/2022-12-15T165417.221370.snakemake.log
WARNING:snakemake.logging:Complete log: .snakemake/log/2022-12-15T165417.221370.snakemake.log
Error: Process completed with exit code 1.
```

davide-f:
@ekatef I think that issue is related to the new shapely that has updated to version 2.0

davide-f:
a fix shapely < 2.0 may be enough. Testing.

Confirm: it works

ekatef:
@davide-f Amazing :) Thank you so much Davide!

ekatef:
All the troubles mentioned above were resolved by the environment fix. The issue with powerplantmatching persists but isn't related to the geometry issue. It will be tracked with #537.

Currently testing for India is about to be completed. The next (hopefully final) step will be testing for China.

ekatef:
Testing for China done. There are some troubles related to #537 and #531 but all geometry-linked troubles seem to be fixed

@davide-f I think this PR is ready for review

ekatef:
A workflow for Pakistan is running to obtain the "non-found" shapes for it. 

Once it'll be finished, a workflow for Pakistan will be run from this PR to check if the proposed fixes are functional

ekatef:
A map for Pakistan looks as follows:

image

Overlaid with "special status" areas for India:

image

davide-f:
Nice! thanks!
does a shape in pakistan (or a combination of shapes) match the orange shape for India?
Could be nice to see the borders of the gadm shapes in the pakistan shape

scripts/build_bus_regions.py
```diff
@@ -265,9 +268,10 @@ def get_id(coords):
         crs=country_shapes.crs,
     ).dropna(axis="index", subset=["geometry"])
 
-    onshore_regions = pd.concat([onshore_regions], ignore_index=True).to_file(
-        snakemake.output.regions_onshore
-    )
+    # sometimes Voronoi partition may give empty polygons
+    onshore_regions = pd.concat(
```
davide-f: Not sure why, but changes do not reflect the current main: in the current main, the is_empty check is performed within the for loop before.
I think it may be nice to close and reopen this PR including the updated main and to clean the changes. I see >29 commits and it may be nice to clean a bit the history considering also the additional comments below.
What do you think?

davide-f: I think in the next revision we may check the need for this doubled pd.concat.
For historical reasons we had this: lines 263-266 and 268-270 with old line references
Probably in the new PR we may merge the two block of lines. I don't see the need for pd.concat([onshore_regions], ignore_index=True)

scripts/build_osm_network.py
```diff
@@ -831,8 +831,12 @@ def built_network(inputs, outputs, geo_crs, distance_crs):
     country_list = input
     bus_country_list = buses["country"].unique().tolist()
 
-    if len(bus_country_list) != len(country_list):
-        no_data_countries = set(country_list).difference(set(bus_country_list))
+    # it may happen that bus_country_list contains entries not relevant as a country name (e.g. "not found")
+    # difference can't give negative values; the following will return only releant country names
+    no_data_countries = set(country_list).difference(set(bus_country_list))
+
+    if len(no_data_countries) > 0:
```
davide-f, line 837: Very interesting work with sets!
Maybe we could do the check if set(country_list) != set(bus_country_list):
This may be more general and if that does not apply, then within the if we calculate the difference as you propose.
in such a case; it may be better to convert no_data_countries as a list; maybe something like `list(set(country_list).difference(set(bus_country_list)))`

We may use symmetric_difference instead of simple difference to generalize.
Most likely the difference would be negligible but by checking as proposed we miss whether some countries are included into bus_country_list and they are missing in country_list.
With simmetric_difference we keep this feature

Opinions?

scripts/build_shapes.py
```diff
@@ -120,10 +128,51 @@ def get_GADM_layer(country_list, layer_id, geo_crs, update=False, outlogging=Fal
             three_2_two_digits_country(twoD_c) for twoD_c in geodf_temp["GID_0"]
         ]
 
+        # GID_0 may have some exotic values, "COUNTRY" column may be used instead
```
davide-f: I think this GID fixing may be moved into a separate function and we should plan for an option to fix it.
I was thinking something like:
``` yaml
build_shapes:
    contended_zones: [values]
```
Where values may be a string e.g. "drop" and we have a function that drops them.
Otherwise, we could have a structured block that explains how to address zones by country.
Some options may be as follow:
``` yaml
build_shapes:
    contended_zones:
        IN:
           GIDXXXXX: [country-code]
        CH: [country code]
```

In the former example, we may give the option to force specific GID zones to be allocated to a country.
Conversely in the `CH: [country code]` option we give the option to force the contended zones of a country to be associated to the specified country code.

Opinions?

scripts/build_shapes.py
```diff
@@ -120,10 +128,51 @@ def get_GADM_layer(country_list, layer_id, geo_crs, update=False, outlogging=Fal
...
+            ]
+
+            if len(non_std_gadm_codes) > 0:
+
+                # a "minimalistic" approach outputs a single file (~200kb for CN + IN)
+                d = {
+                    "country_code": country_code,
+                    "non_standard_gadm_codes": non_std_gadm_codes,
+                }
+                df = pd.DataFrame(data=d)
```
davide-f: This code is super useful for debugging, but for more than 1 country it would need some changes.
For example, we would need to pack all non-contended zones in a single file and add the path to the output file by snakemake to make it snakemake-compliant.
I think it is super useful for debugging, but for general use may create noise.
As this part may be moved into another function, we can add an argument to the function to state whether to save this output or not.
By default, I wouldn't like to have this output.

What do you think?

scripts/build_shapes.py
```diff
@@ -120,10 +128,51 @@ def get_GADM_layer(country_list, layer_id, geo_crs, update=False, outlogging=Fal
...
+                output_path = "non_standard_gadm_codes.csv"
+                df.to_csv(
+                    output_path,
```
davide-f: It would be better to keep only the non-appending mode below.
For consecutive runs the appending approach may add redundant lines.

scripts/build_shapes.py
```diff
@@ -268,7 +317,9 @@ def eez(countries, geo_crs, country_shapes, EEZ_gpkg, out_logging=False, distanc
     )
 
     ret_df = ret_df.apply(lambda x: make_valid(x))
-    country_shapes = country_shapes.apply(lambda x: make_valid(x))
+
+    # country_shapes may consist of different geometries which need to be united
+    country_shapes = country_shapes.apply(lambda x: make_valid(x)).unary_union
```
davide-f, line 304: By doing this don't we merge all country shapes into one?
Does this work for multi-country regions?

scripts/build_shapes.py
```diff
@@ -812,7 +863,12 @@ def gadm(
 
     country_shapes = countries(countries_list, geo_crs, update, out_logging)
 
-    country_shapes.reset_index().to_file(snakemake.output.country_shapes)
+    # there may be "holes" in the countries geometry which cause troubles along the workflow
+    # e.g. that is the case for enclaves like Dahagram–Angarpota for IN/BD
+    country_shapes_valid = country_shapes.apply(
```
davide-f: I'd overwrite the country_shapes variable or pipe the function to clean the code

scripts/build_shapes.py
```diff
@@ -836,4 +892,10 @@ def gadm(
         year,
         nprocesses=nprocesses,
     )
+    gadm_shapes.geometry = gadm_shapes.apply(
```
davide-f: In the main some checks are already added; see: https://github.com/pypsa-meets-earth/pypsa-earth/blob/b745abe78090b634c0840bf9c289c35eb51e05ca/scripts/build_shapes.py#L787

It may be better to review this according to the new main.
This code may be moved there: the make_valid is not done but it should as you did.
Maybe apply is not the best function to be used on the entire dataframe.
We could do that directly on the series.
Something like this may do: `gadm_shapes.geometry.apply(lambda r: make_valid(r) if not r.is_valid else r)`?
However, I'd like this to be merged into the gadm function rather than here.

davide-f (commented):
Very nice bugfixing!
It works and we may generalize this approach.
The comments do not need to be all implemented. I think just implementing the "drop" option and/or fix to country may be enough.

Comments are welcomen; this task is a tricky one...

ekatef:
> Nice! thanks!
> does a shape in pakistan (or a combination of shapes) match the orange shape for India?
> Could be nice to see the borders of the gadm shapes in the pakistan shape

Pictures updated. The orange area belongs only to India (according to gadm database). The violet shape matches the gadm shapes for Pakistan (tried to highlight it using different line widths)

davide-f:
> > Nice! thanks!
> > does a shape in pakistan (or a combination of shapes) match the orange shape for India?
> > Could be nice to see the borders of the gadm shapes in the pakistan shape
> 
> Pictures updated. The orange area belongs only to India (according to gadm database). The violet shape matches the gadm shapes for Pakistan (tried to highlight it using different line widths)

Very interesting.
It seems like there is no Overlapping between the Indian and Pakistan shape, then.
Does it apply also to the other Indian regions to your knowledge?

Because if that's the case just fixing the not found equivalent to the country should do the trick.
Anyway better to keep an option "discard"/"bycountry"/"throwerror", but the solution based on gid id may not be needed now.
The solution id based could be interesting when the same shapes belong to multiple countries

ekatef:
> Very interesting. It seems like there is no Overlapping between the Indian and Pakistan shape, then. Does it apply also to the other Indian regions to your knowledge?

> Because if that's the case just fixing the not found equivalent to the country should do the trick. Anyway better to keep an option "discard"/"bycountry"/"throwerror", but the solution based on gid id may not be needed now. The solution id based could be interesting when the same shapes belong to multiple countries

Yes, the picture for India and China is similar:

image

It looks like GADM database prescribes each of the contented zones to one of the countries. Agree that it simplifies the issue.

Do I understand properly that all other areas with some troubles in build_shapes belong to single countries only? So, there is no need to implement a general countries solution for multiple countries?

davide-f:
> Do I understand properly that all other areas with some troubles in build_shapes belong to single countries only? So, there is no need to implement a general countries solution for multiple countries?

I am unsure, but for what concerns PK and IN, that seems the case: in the pictures I haven't seen overlap. Could you confirm that?
If that's the case, and similar behaviors is seen for the other cases under consideration, we can definitely avoid doing the general approach.

What I've seen would greatly goes in favour to avoid that general case and so implement only the options (names to be potentially revised):
``` yaml
build_shapes:
    contended_zones: [drop/gadm_country]
```
In the future, this structure may be improved if necessary, but no need for the general approach I think

scripts/build_osm_network.py
ekatef: Thank you :)

A very beautiful idea regarding `symmetric_difference()` and I'll definitely keep in mind this approach. But not sure if we may generalise this particular case. As far as I understand,  we have to treat a situation when there are some unexpected `country` entries in the buses dataframe in a different way as we do the countries without buses

Actually, if there are some unexpected countries in the buses dataframe, I'd rather raise an error as that is likely that something went wrong during OSM data cleaning (e.g. the data folders need some clean-up). Probably, it should be added to #528 

Do you agree?

scripts/build_shapes.py
ekatef: Yes! That is applied under a loop, country-by-country. Have tested on a two-countries list (["PK", "KG"]) and it seems to work properly

scripts/build_shapes.py
ekatef: Revised

scripts/build_shapes.py
ekatef: Done!

scripts/build_shapes.py
ekatef: Agree, revised

As discussed, there is also a need to add a config parameter and a snakemake output if we'll decide to keep it in a cleaned version. I'd suggest to think if it's really worth to have this output in the mainstream workflow: wouldn't it be a bit of over-complication?

scripts/build_shapes.py
ekatef: Absolutely agree :) Probably, we could move this part to a second step

scripts/build_shapes.py
ekatef: Refactoring with a wrap to a function done. 

I very much like the idea to have a customisation option and suggest to work on it on the next step

ekatef:
@davide-f, thank you so much for the review and the discussion! My feeling is that we are close to finalise the first bug-fixing keeping advanced features for the second step.

The only point left is replacing a `if else` block with `pandas.where()`. Have to look into it in some more details: not sure if the replacing function has a vectorised version in Python.

If you don't have major objections regarding the introduced changes, I'll move the changes into a clean PR

davide-f:
@ekatef May we close this PR since we worked in the others?

ekatef:
Closed to be finalised in a cleaned version

PR #547, opened 2022-12-28
add earth-osm package, remove esy-osm

pz-max:
Remove `esy-osm` as pip dependency by `earth-osm` package from @mnm-matin  

Still missing:
- [x] conda-forge & pypi recipe for `earth-osm`
- [x] earth-osm needs to be more mature e.g. add CI tests, deploy docs etc
- [x] some improvements that would reduce the code on our side e.g. provide concated outputs instead of single files
- [x] testing if the earth-osm integration is stable for more than a dozen countries around the world. TESTED for EG, NG, MA

## Checklist

- [x] **No config** changes required


pz-max:
Windows breaks @mnm-matin 

```
  File "D:\a\pypsa-earth\pypsa-earth\.snakemake\scripts\tmpycprlere.download_osm_data.py", line 14, in <module>
    from earth_osm import eo
  File "C:\Miniconda3\envs\pypsa-earth\lib\site-packages\earth_osm\eo.py", line 16, in <module>
    from earth_osm.gfk_data import get_region_tuple
  File "C:\Miniconda3\envs\pypsa-earth\lib\site-packages\earth_osm\gfk_data.py", line 30, in <module>
    d = json.load(f)
  File "C:\Miniconda3\envs\pypsa-earth\lib\json\__init__.py", line 293, in load
    return loads(fp.read(),
  File "C:\Miniconda3\envs\pypsa-earth\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8d in position 157545: character maps to <undefined>
```

scripts/download_osm_data.py
```diff
@@ -425,338 +70,44 @@ def convert_iso_to_geofk(iso_code, iso_coding=True, convert_dict=iso_to_geofk_di
...
+            filename = Path.joinpath(out_path, f"all_{name}.{f}")
+            if not Path.exists(filename):
+                _logger.info(f"{filename} does not exist, create empty file")
+                open(filename, "w").close()
+
+            # Move and rename
+            old_path = Path.joinpath(out_path, f"all_{name}.{f}")
+            new_path = Path.joinpath(store_path_resources, f"africa_all_{name}.{f}")
+            _logger.info(f"Create {old_path} and move to {new_path}")
+            shutil.move(old_path, new_path)
```
pz-max: We could remove these lines in case we:
- Read directly from the `pypsa-earth/data/osm/out` folder
- Rename the files in the Snakefile e.g. from `africa_all_generators.csv` to `all_generators.csv`

pz-max: Any opinion here? @davide-f 
I think the rename is due to anyway (there should be no 'africa' in the name anymore), but I am not sure about reading the files form `data/osm/out` instead from `resources/osm/`

pz-max:
Missing. Need to compare optimization objectives of countries:

| Country      | earth-osm (new integration) | esy-osm (old integration)  |
| :---:       |    :----:   |          :---: |
| Nigeria      |  1.67039499e+09      | 1.67039396e+09   |
| Morroco   |  1.70002264e+09       |      |
| Egypt   | 1.15049500e+10        |  1.15048809e+10  |
| South Africa   |   |   |


**Nigeria, I am using the following config adjustments:   clusters: [5], opts: [Co2L-12H], co2limit: 1.e+10,  co2base: 2.e+10 (otherwise I experience numerical issues)

scripts/download_osm_data.py
davide-f: @pz-max I am ok with removing the africa_ because we want to be general.
Regarding whether to move the raw outputs into data, I think we need to adopt a policy according to the others as well.

At the moment, I'd keep things as is as otherwise there may conflicts everywhere in the PRs and this is a detail I'd say.

Regarding earth-osm, I have few questions:
- can multiple executions of earth-osm work in parallel? before because of esy-* that was impossible: all parallel threads would have worked on a common temp file; what now?
- Linked to above, the download_osm_data may be done in a snakemake-like way, where the type of resource may be a wildcard. That could be really useful for the -sec and distribution packages if done well

Can be left to further PRs though

scripts/download_osm_data.py
```diff
@@ -2,399 +2,43 @@
 # SPDX-FileCopyrightText: : 2021 PyPSA-Africa Authors
```
davide-f: I like the PR!

I think that now that it's getting stable, it could be nice to add a docstring at the beginning to state the link to earth-osm and give a description of this rule that is missing.
Not necessarily something long

scripts/download_osm_data.py
```diff
@@ -425,338 +69,48 @@ def convert_iso_to_geofk(iso_code, iso_coding=True, convert_dict=iso_to_geofk_di
...
-    process_data(
-        feature_list,
-        country_list,
-        output_files,
-        geo_crs,
-        distance_crs,
-        iso_coding=True,
+    # Python interface to download OpenStreetMap data
+    # Documented at https://github.com/pypsa-meets-earth/earth-osm
+    eo.get_osm_data(
```
davide-f, line 106: may it be worthy to encapsulate this function in an additional function that does the filtering by the iso_to_geofk function?
This may be handy for other rules e.g. -sec or -distribution

davide-f (commented):
Few comments; this looks good!
Validation with more countries standard vs earth-osm may be recommended as mentioned.
Very nice! :)

scripts/download_osm_data.py
pz-max: @mnm-matin feel free to correct me. The current implementation includes an option `mp=bool` which stands for multiprocessing. It works exactly like esy-osm. It will use all available computational resources. (e.g. set by snakemake)
- When you speak about resources, you mean the type of infrastructure? Sure, once we have a nice interface to also extract other infrastructure, we might only need to define a config and can automate stuff

davide-f: So, having parallel computing within the tool is cool.
My concern is what if two parallel processes (e.g. the same rule with different wildcard) operate using the same data_dir?
Are there differences if they work on the same or different countries?

Depending these points, it may or may not be possible to use wildcards easily with earth-osm.
That is not necessarily a bad thing, but just to understand.

With esy filter no parallelization at all was possible at all

scripts/download_osm_data.py
pz-max: Great idea. Let's do that when needed? (probably takes a while until it is necessary -- we also don't know yet how the interface will handle other infrastructure)

scripts/download_osm_data.py
pz-max: Done

pz-max:
> Few comments; this looks good! Validation with more countries standard vs earth-osm may be recommended as mentioned. Very nice! :)

@davide-f The validation was successful for two countries. The results suggest that there is no issue. Probably no more tests needed. PR is ready for review. :)

scripts/download_osm_data.py
pz-max: Not sure about this. We can also deactivate parallel processes within earth-osm (it's an option `mp=False`)

pz-max:
As discussed with @davide-f. This will be merged after https://github.com/pypsa-meets-earth/pypsa-earth/pull/543 is DONE

pz-max:
@davide-f ready to merge.
Thanks for reviewing.

Snakefile
```diff
@@ -143,18 +143,18 @@ if config["enable"].get("download_osm_data", True):
...
+        substations="resources/osm/raw/all_raw_substations.geojson",
         country_shapes="resources/shapes/country_shapes.geojson",
         offshore_shapes="resources/shapes/offshore_shapes.geojson",
         africa_shape="resources/shapes/africa_shape.geojson",
     output:
-        generators="resources/osm/clean/africa_all_generators.geojson",
-        generators_csv="resources/osm/clean/africa_all_generators.csv",
-        lines="resources/osm/clean/africa_all_lines.geojson",
-        substations="resources/osm/clean/africa_all_substations.geojson",
+        generators="resources/osm/clean/all_generators.geojson",
```
davide-f: I'd be prone to rename the outputs as "all_clean_{feature}.{ext}" alike the raw version.
What do you think?

davide-f (approved):
I leave a comment only; I'd be personally prone to revise as specified in the comment, but it won't be a blocking feature.

I approve the merge but please check the message above first

Snakefile
pz-max: Let's keep it as is? I like your suggestion as it might reduce a bit code in the Snakefile. Small disadvantages might arise. E.g. code is less explicit, loosing left side assigned parameter which requires different read in in the scripts (.py). E.g. snakemake.inputs[1] instead of snakemake.inputs.generators ... I am sure some can be tackled.

Probably not necessary atm?

davide-f: Ah wait my suggestion was very stupid, keep everything as is but simply rename "all_generators" into "all_clean_generators", "all_lines" into "all_clean_lines" and so on.
No wildcards or anything like that.

When running some codes, having the raw/clean duality also in the name may be handy rather than raw/none, both for symmetries than coding (I had to do some special ifs in the add_scenario PR indeed)

pz-max: Good point! Done

davide-f:
@Hazem-IEG I point you here as these changes may have implications for pypsa-earth-sec.
Since you work with the pypsa files, that should not be an issue, but worth notifying

davide-f:
@pz-max once the conflicts are solved and verified, feel free to merge

pz-max:
Path fix is required due to RDIR in `download_osm....py`

davide-f:
> Path fix is required due to RDIR in `download_osm....py`

I see two options:
1. implement the snakemake outputs as a named list by feature name and within download_osm_data instead of hard coding the feature name, we loop over the feature names according to snakemake.output.keys().
   For every key, the value of that key is goign to be the new path where the final file is saved. This is snakemake-compliant, extendable for new features (just adding a new output key in the snakefile would be sufficient), but requires little work
2. hard-code the scenario folder within download osm data. This works but it is less clean

The problem with the previous and actual implementation is that the output paths of the snakefiles are not used in the script and that is the reason we experience this problem unfortunately.
Just saying for people may be reading this PR

pz-max:
> this PR

Too late :D seems I went for option 2.
The issue is that we store the outputs in `resources` which has this RDIR path stuff
I think point 1 can be ideally touched when we need to extract more OSM data

davide-f (approved):

davide-f:
Merged :)

PR #548, opened 2022-12-28
Fix cleaning of OSM data

### LINKED ISSUE #469: 0 column and frequency is not properly cleaned in 'clean_osm_data.py'
## Checklist

- [x] I am using the current [`main`](https://github.com/pypsa-meets-earth/pypsa-earth/tree/main) branch.

## Describe the Bug

I am extracting and cleaning the OSM data for the United States. However, this works not smoothly and requires fixes.

## Error Message 1 - 0 column
Executing this lines:
https://github.com/pypsa-meets-earth/pypsa-earth/blob/e35545c3bcfa14b6eef714f28d95ef807db450cb/scripts/clean_osm_data.py#L716
leads to the error message:
```bash
AttributeError: 'int' object has no attribute 'encode' 
```
 because `df_all_lines` has a 0 column. Using the US as test case, we should figure out where the 0 columns is created and remove the bug. Adding this line `df_all_lines.drop(labels=0, axis=1, inplace=True)` before the geojson can be used as temporary fix.

## Error Message 2 - frequency is not properly cleaned

This line:
https://github.com/pypsa-meets-earth/pypsa-earth/blob/e35545c3bcfa14b6eef714f28d95ef807db450cb/scripts/clean_osm_data.py#L80
Is raising the following error message:
```
Exception has occurred: ValueError
could not convert string to float: '60;60'
```
Analysing the `lines` by `lines.tag_frequency.unique()` shows quickly that the frequency is not properly cleaned: 
```bash
array([60.0, '60', '60;60', '0', '25', '60;0', '6', '50'], dtype=object)
```

ekatef:

Improves OSM data cleaning workflow

## Checklist

### DIFF
--- a/scripts/clean_osm_data.py
+++ b/scripts/clean_osm_data.py
@@ -290,9 +290,11 @@ def split_cells_multiple(df, list_col=["cables", "circuits", "voltage"]):
             if [";" in s for s in sub].count(True) == len(list_col):
                 d = [s.split(";") for s in sub]  # split them
                 r = df.loc[i, :].copy()
-                df.loc[i, list_col[0]] = d[0][0]  # first split  [0]
+                # first split  [0]
+                df.loc[i, list_col[0]] = d[0][0]
                 df.loc[i, list_col[1]] = d[1][0]
-                r[list_col[0]] = d[0][1]  # second split [1]
+                # second split [1]
+                r[list_col[0]] = d[0][1]
                 r[list_col[1]] = d[1][1]
                 df_list.append(r)
 
--- a/scripts/clean_osm_data.py
+++ b/scripts/clean_osm_data.py
@@ -691,7 +691,7 @@ def clean_data(
         df_cables = finalize_lines_type(df_cables)
 
         # concatenate lines and cables in a single dataframe
-        df_all_lines = pd.concat([df_lines, df_cables])
+        df_all_lines = pd.concat([df_lines, df_cables], ignore_index=True)
 
     # Add underground, under_construction, frequency and circuits columns to the dataframe
     # and drop corresponding unused columns
--- a/scripts/clean_osm_data.py
+++ b/scripts/clean_osm_data.py
@@ -403,6 +403,7 @@ def integrate_lines_df(df_all_lines, distance_crs):
     else:
         df_all_lines["tag_frequency"] = ac_freq_default
 
+    # it may happen that the frequency column also contains ";"    
     df_all_lines["tag_frequency"] = df_all_lines["tag_frequency"].astype(str)
     df_all_lines = split_cells(df_all_lines, lst_col="tag_frequency")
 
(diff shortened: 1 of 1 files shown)


scripts/clean_osm_data.py
```diff
@@ -399,6 +403,10 @@ def integrate_lines_df(df_all_lines, distance_crs):
     else:
         df_all_lines["tag_frequency"] = ac_freq_default
 
+    # it may happen that the frequency column also contains ";"
+    df_all_lines["tag_frequency"] = df_all_lines["tag_frequency"].astype(str)
+    df_all_lines = split_cells(df_all_lines, lst_col="tag_frequency")
```
davide-f, line 408: I think that the split_cells used here is improper as we duplicate line entries that may have the ";" in also the cables and circuits.
The tag_frequency should be added into the split_cells_multiple, but that function should be heavily improved to account for semicolumn partially missining in some columns.

In the case where semicolumn is found in tag_frequency, are the values consistent (e.i. "50;50") or not (e.i. ("50;0")?

A quick fix may be to just put the first entry where semicolumn is found, yet the right way is discussed above.
What country experiences this problem?

davide-f (changes requested):

davide-f:
@ekatef regarding this PR, in the case multiple entries are found for frequency may make sense to just consider the first value?
May you recall an example of values that needed this fix?
I'd be prone to do a simple fix and when the ";" is found to just get the first value of the list, provided that the elements are simple duplicated values.

ekatef:
@davide-f I have encountered multiple entries for frequency as `50;50`. So, taking the first entry should work fine

davide-f:
> @davide-f I have encountered multiple entries for frequency as `50;50`. So, taking the first entry should work fine

Thanks :)
Do you think you could propose a quick fix for this? forcepushing on this branch or opening a new PR may work fine

ekatef:
> > @davide-f I have encountered multiple entries for frequency as `50;50`. So, taking the first entry should work fine
> 
> Thanks :) Do you think you could propose a quick fix for this? forcepushing on this branch or opening a new PR may work fine

@davide-f, yes absolutely. Will be happy to add a fix for this. Although, not sure how quick is quick enough :) Do you see this issue as a blocking one?

davide-f:
> > > @davide-f I have encountered multiple entries for frequency as `50;50`. So, taking the first entry should work fine
> > 
> > 
> > Thanks :) Do you think you could propose a quick fix for this? forcepushing on this branch or opening a new PR may work fine
> 
> @davide-f, yes absolutely. Will be happy to add a fix for this. Although, not sure how quick is quick enough :) Do you see this issue as a blocking one?

I don't see it a blocking issue but since we have it and you already scouted it I think we could find a fix and then move to the AC/DC problem.
The fix I'd expect for this problem may be 1-2 lines basically. Something like:
``` python
df_all_lines["tag_frequency"] = df_all_lines["tag_frequency"].astype(str).str.split(pat=";",n=1)
```
Or something like that. The pat may be specified also with a regular expression to capture ; or , or alike but not sure if it makes sense to make it that complex

PR #555, opened 2023-01-03
Save logging outputs

### LINKED ISSUE #535: Check saving of logger outputs
It looks like that logger outputs are not saved in some cases. E.g. there is some output in `build_shapes.log` but any content is missed in logs for `build_renewable_profiles`.

As suggested by @davide-f, there is a need to be check it together with the name of the logging which currently is always “__name__” .

ekatef:
# Closes #535

It looks like the warnings of our workflow are not saved as the root logger still has some handlers configured when `basicConfig( )` is called. In that case `basicConfig( )` function does nothing, according to [logging documentation](https://github.com/python/cpython/blob/3.8/Lib/logging/__init__.py#L1912). This issue may be resolved just by setting `force` to `True`.

Another improvement probably worth of considering is to add a timestamp to logger output. It could be handy if looking into the log files.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,27 +148,6 @@ The documentation is available here: [documentation](https://pypsa-earth.readthe
 <table>
 <tr>
     <td align="center">
-        <a href="https://github.com/hazemakhalek">
-            <img src="https://avatars.githubusercontent.com/u/26235356?v=4" width="100;" alt="hazemakhalek"/>
-            <br />
-            <sub><b>Hazemakhalek</b></sub>
-        </a>
-    </td>
-    <td align="center">
-        <a href="https://github.com/jarry7">
-            <img src="https://avatars.githubusercontent.com/u/27745389?v=4" width="100;" alt="jarry7"/>
-            <br />
-            <sub><b>Jarrad Wright</b></sub>
-        </a>
-    </td>
-    <td align="center">
-        <a href="https://github.com/fneum">
-            <img src="https://avatars.githubusercontent.com/u/29101152?v=4" width="100;" alt="fneum"/>
-            <br />
-            <sub><b>Fabian Neumann</b></sub>
-        </a>
-    </td>
-    <td align="center">
         <a href="https://github.com/ekatef">
             <img src="https://avatars.githubusercontent.com/u/30229437?v=4" width="100;" alt="ekatef"/>
             <br />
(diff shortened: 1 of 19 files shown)


ekatef:
Change of `basicConfig()` does not resolve an issue with the error logs not being saved. Currently I see two ways to output error logs to a file:

1. We could capture the exceptions with `try ... except` in the `__main__` and using     `logger.exception( )` 
2. It is also possible to use [Snakemake way](https://snakemake.readthedocs.io/en/stable/tutorial/advanced.html#step-5-logging) directing the console output to a file. That would allow to keep not only workflow logging but also Snakemake one

ekatef:
@davide-f may be of interest to handle errors and warnings in a more convenient way

davide-f:
Thanks! this is very interesting; we should do that: copying all the output to the log file should be the way to go

ekatef:
The current modification allows to have logging output both in a terminal and in the logs. Handlers for both terminal and log files are defined with 

https://github.com/pypsa-meets-earth/pypsa-earth/blob/dd805b7bd542039ace82575cee29139c875be8b6/scripts/_helpers.py#L88-L89

The problem was that the logger in the main still had some handlers before `configure_logging( )` was applied. Existence of these handlers instances lead to the fact that logging configuration has no effect according to the conventions of `basicConfig( )`.

Adding `try...catch` statement captures the error message into the logs and as a side effect prevents the workflow from crashing. Snakemake terminates the process in some seconds after exception has hapened. To make the termination faster execution termination with `sys.exit()` has been added

ekatef:
PR was tested to check if the error messages along with some traceback are being saved into the log files and it works fine. However, when testing locally there was a weird error for Nigeria in `simplify_network`:

```bash
in aggregategenerators
weighting = generators.weight.groupby(grouper, axis=0).transform(normed_or_uniform)
AttributeError: 'DataFrame' object has no attribute 'weight'
```

The error was captured properly both in the console and in the log file; most likely it doesn't relate to the changes introduced in the PR but I'd like to resolve it before finalising the PR.

davide-f:
I encountered this problem as well and I think I fixed it in add_scenario PR but I don't remember...

The new changes involve all files; I think it is quite heavy now.
Maybe there was a misunderstanding.
I like the approach especially where it is clean. Some rules like the cleaning and alike adhere to that: the try catch is across a single function; but other scripts do not have such nature and it becomes a big messy.
Moreover, this PR would conflict significantly with the other PRs.

I'd be prone to merge basically immediately the simple fix `basicConfig(**kwargs, force=True)`, but I'd wait for the try-catch across all rules.
We may merge it into clean_osm_data for testing, but for all rules I think could be messy for the current coding.

What do you think?

To save the history, other options may be viable as well.
For the moment, I'm usually working with screens and there is the flag -L that enables storing everything to file.
The normal user would not heavily check the log files and the console may be enough. Developers instead may be interested, yet there are also alternatives to that.

Given the status of the code, i'd say that if you create PR with changing the basicConfig(...) [commit e13242bb893d320c3c631128f27a2a7f591f4a8f], I'd merge it right away.

If you create a new branch from main and you do:
git pull
git cherry-pick e13242bb893d320c3c631128f27a2a7f591f4a8f

and you publish the new PR, that could be merged and this PR would not be affected.

ekatef:
> I encountered this problem as well and I think I fixed it in add_scenario PR but I don't remember...
> 
> The new changes involve all files; I think it is quite heavy now.
> Maybe there was a misunderstanding.
> I like the approach especially where it is clean. Some rules like the cleaning and alike adhere to that: the try catch is across a single function; but other scripts do not have such nature and it becomes a big messy.
> Moreover, this PR would conflict significantly with the other PRs.
> 
> I'd be prone to merge basically immediately the simple fix basicConfig(**kwargs, force=True), but I'd wait for the try-catch across all rules.
> We may merge it into clean_osm_data for testing, but for all rules I think could be messy for the current coding.
> 
> What do you think?
> 
> To save the history, other options may be viable as well.
> For the moment, I'm usually working with screens and there is the flag -L that enables storing everything to file.
> The normal user would not heavily check the log files and the console may be enough. Developers instead may be interested, yet there are also alternatives to that.
> 
> Given the status of the code, i'd say that if you create PR with changing the basicConfig(...) [commit https://github.com/pypsa-meets-earth/pypsa-earth/commit/e13242bb893d320c3c631128f27a2a7f591f4a8f], I'd merge it right away.
> 
> If you create a new branch from main and you do:
> git pull
> git cherry-pick https://github.com/pypsa-meets-earth/pypsa-earth/commit/e13242bb893d320c3c631128f27a2a7f591f4a8f
> 
> and you publish the new PR, that could be merged and this PR would not be affected.

Absolutely agree with all the points.

Thanks a lot for the detailed analysis and sorry for alarming you :) Actually, that is still a draft and I'm not quite satisfied with it exactly due to the reasons you mention. Not sure that using `try-except` construct in some scripts only would be a really good solution as it'd break consistency of the code...

What I do not quite understand is why `StreamHandler` directs traceback to the console when an error happens, while `FileHandler` does nothing in such a case and needs an explicit call of `logger.exception()`.It looks like there is `[sys.excepthook()](sys.excepthook)` in play and the only way to deal with that is to [overwrite it's default behaviour](https://stackoverflow.com/a/6234491/8465924). According to [this discussion](https://stackoverflow.com/q/73133462/8465924) it seems to be the only way to have non-captured exception written to a log. But it sounds too complicated especially keeping in mind the option you found to store output in a file .

So, I'd close this PR and open a new one as you suggest. Currently I don't see the changes in `basicConfig()` in a version of `run_scenarios` pushed to github. But I'd keep this branch with some comments added from our discussion in case if we'd need to go back to the loggers. Would it work?

ekatef:
Re-opened in a cleaned version #559

Issue #262, opened 2022-02-08
Move from 2-code to 3-code convention for countries; examples North Cyprus does not have 2-code representation


EmreYorat:
In the <https://download.geofabrik.de/europe/cyprus.html> data for Cyprus is represented for both North and South Cyprus. But in the build_shapes notebook only South Cyprus is being filled. North Cyprus is looking empty. Also gadm server has divided its data for Cyprus to two one for North and the other is South. What would be your suggestion(s) for this case? I attached the notebook output below. 

![cyprus_build_shapes](https://user-images.githubusercontent.com/93644024/153053280-e36c7bf4-76a4-44d0-a93d-65fa3ba1b90a.png)

![gadmXNC](https://user-images.githubusercontent.com/93644024/153163591-fb71eef6-8a46-4a67-b18e-0fd07487c0b6.png)

![southcyprus](https://user-images.githubusercontent.com/93644024/153164459-773b8c6e-396f-4e41-9fbb-e7baddde820c.png)

davide-f:
Thanks @EmreYorat for your comment.
As discussed, for the moment I suggest you and your team to focus on the other regions and without northern Cyprus, at least temporarily.
The main problem is that there is no 2-code country for Cyprus.
My main concern is: how often may this occur at global scale? Expected answer: few times but for the moment we don't know exactly

To tackle this issue I see two options:
- create a hard-fix for northern cyprus (and eventually few more countries), which means creating a non-standard 2-code name for it
- move from 2-digit code to 3-digit. Personally I think this may be more scalable in long-term perspective but it means larger work

pz-max:
**Decided today in the weekly developer meeting:** 
Preference of 3-letter code over 2-letter code
- More intuitive to have 3 letters (we have 180 countries)
- We avoid the ‘NA’ issue where a NA string was converted to np.nan
- Some regions only have 3 letter code representation e.g. Cyprus

pz-max:
Some discussions related to this: https://github.com/pypsa-meets-earth/pypsa-earth/discussions/398

davide-f:
Interestingly, we could have a preliminary pre-processing of the countries list to translate whatever list is passed to a standard output, e.g. 3-digit.
the code may also be revised to have that in mind and the type of convention, may be a variable hard-coded in helpers or a config.

Emre-Yorat89:
Hello,
[...]

PR #563, opened 2023-01-10
Fix geometry issues - clean version

ekatef:

That is a cleaned version of #532 

## Checklist


ekatef:
A test on ["PK", "KG"] works

ekatef:
Tests both on ["PK"] and ["PK", "KG"] was successful. I think it's ready for review now.
There are a couple of questions left.

1) do we really need to keep output of the non-standard regions to a file? Currently it's done with writing a csv without proper Snakemake declarations:

https://github.com/ekatef/pypsa-earth/blob/53c2c04b7555ad054d66742bc2375f7d1d956a19/scripts/build_shapes.py#L114-L121

If we are intended to keep this output, it may be better to add a proper snakemake output. But I'm not sure if it's really necessary as it looks like such situations are quite rare.

2) when implementing `where` instead of `apply` it was found that to do so there may be a need to vectorise the name-transforming functions:

https://github.com/pypsa-meets-earth/pypsa-earth/commit/75a581c712ea76d60311b5e646590be35a8cf7e8

Here changes in a way of applying `country_name_2_two_digits` required to add `any` to the function definition to make it work. Obviously, more throughout revision of `country_name_2_two_digits` code is needed to make if work properly both on a single string and on a list of strings. But again, I'm not sure if it's worth it :)

@davide-f, would be grateful for your opinion on this and the review

README.md
```diff
@@ -147,27 +147,6 @@ The documentation is available here: [documentation](https://pypsa-earth.readthe
 <!-- readme: collaborators,contributors -start -->
 <table>
 <tr>
-    <td align="center">
```
davide-f, line 150: I've noticed this also last time.
Not sure if the cherry-picking intended to include this too

scripts/_helpers.py
```diff
@@ -551,7 +551,7 @@ def country_name_2_two_digits(country_name):
         2-digit country name
     """
     if (
-        country_name
+        country_name.any()
```
davide-f: This is no the intended behavior. It should be removed

scripts/build_bus_regions.py
```diff
@@ -77,6 +78,8 @@ def custom_voronoi_partition_pts(points, outline, add_bounds_shape=True, multipl
     from scipy.spatial import Voronoi
     from shapely.geometry import Point, Polygon
 
+    outline = unary_union(outline)
```
davide-f, line 79: By looking at the docstring there is a refuse in the docstring with the option no_multipolygons that is not used.
In pypsa-eur there have been some changes.

What are the implications of keeping the multipolygon here?

scripts/build_osm_network.py
```diff
@@ -830,8 +830,12 @@ def built_network(inputs, outputs, geo_crs, distance_crs):
     country_list = input
     bus_country_list = buses["country"].unique().tolist()
 
-    if len(bus_country_list) != len(country_list):
-        no_data_countries = set(country_list).difference(set(bus_country_list))
+    # it may happen that bus_country_list contains entries not relevant as a country name (e.g. "not found")
+    # difference can't give negative values; the following will return only releant country names
+    no_data_countries = set(country_list).difference(set(bus_country_list))
```
davide-f: may be worth using symmetric_difference here

scripts/build_shapes.py
```diff
@@ -80,6 +81,48 @@ def download_GADM(country_code, update=False, out_logging=False):
...
 
 
+def restore_country_code_by_name(row, country_code):
+    if row["GID_0"] != country_code:
+        return country_name_2_two_digits(row["COUNTRY"])
+    else:
+        return row["GID_0"]
+
+
+def build_gadm_df(file, layer, cc):
```
davide-f: Maybe there has been a misunderstanding; I was proposing to move to a separate function just the filtering of the GID_0 component.

Basically, keeping the get_GADM_layer as it was and there only adding a line or two like these:

```
if {check config option drop}:
  geodf.drop(list of indices not matching, axis=0, inplace=True);
elif {check config option set to country}:
  geodf["COUNTRY"] = country_code
```

if more lines are needed, we may define a function that contains the rows above and beyond ( e.g. the output of the non-standard zones). By default, the output file shall not be saved
```
def filter_gadm_flag(geodf, config, save_non_standard_geo=False):
   .... stuff
```

scripts/build_shapes.py
```diff
@@ -267,8 +298,10 @@ def eez(countries, geo_crs, country_shapes, EEZ_gpkg, out_logging=False, distanc
     )
 
     ret_df = ret_df.apply(lambda x: make_valid(x))
-    country_shapes = country_shapes.apply(lambda x: make_valid(x))
+    # country_shapes may consist of different geometries which need to be united
+    country_shapes = country_shapes.apply(lambda x: make_valid(x)).unary_union
```
davide-f: why unary_union?

davide-f (commented):
I've added some comments, I fear there have been some misunderstanding; hope not

README.md
ekatef: Agree that it is annoying :) However, a contributor currently doesn't have an option to avoid fetching this git-creativity. I may look into settings of git actions if you feel it's necessary

scripts/_helpers.py
ekatef: Okay! So, your answer to my second questions (attached to the review notification) is "no". Agree, the `where` modification doesn't look great. Commit reverted

scripts/build_bus_regions.py
ekatef: Well, it looks like our docstring is simply outdated (see commit https://github.com/PyPSA/pypsa-eur/commit/97fbf77ff8ddb7da8bb75602b73853e319854cb5 in PyPSA-Europe). Fixed

scripts/build_osm_network.py
ekatef: Because we can't handle two difference cases as symmetrical:

1) `set(country_list).difference(set(bus_country_list))` = countries from the `countries` parameter of the config which don't have any data to restore a buses dataframe [meaning that we need to generate some data for such areas]

2) `set(bus_country_list).difference(set(country_list)`) = countries from the buses dataframe which are not in the `countries` from the config and hence were not requested by the user to be included into the model [which basically means that something went wrong in the workflow before]

Initially the following code chunk was intended to address the first situation but the condition captured both of them being `len(bus_country_list) != len(country_list)` ( see https://github.com/pypsa-meets-earth/pypsa-earth/blob/fa799bf0723496a3ac5ec50feb9a2ffe2876d758/scripts/build_osm_network.py#L833 ). That lead to troubles when a non-standard code appeared in the buses list. So, I had to introduce a fix exactly to avoid mixing two situations. (Sorry for not being clear enough when explaining it in our discussion!)

scripts/build_shapes.py
ekatef: Mmmm... Agree that the code could be better structured but don't quite get what your idea

Besides, I'm not sure that `drop` is a good default option. Apart of ethical concerns, it'd require some additional changes in the code. (I'll attach a picture to the main PR conversation to explain what is the matter)

So, I suggest to focus in this PR on following GADM conventions with introducing `custom_prescribe` and (probably) `drop` option as the next step

scripts/build_shapes.py
ekatef: to avoid having multiple offshore shapes which is the case if the onshore dataframe for a country contains multiple geometries

Each non-standard GADM code leads to an additional geometry entry in the countries_shape. For some reasons this results is duplication of the offshore shapes for this country when calculating the difference in `ret_df.difference(country_shapes_with_buffer)`

scripts/build_osm_network.py
davide-f: My proposal would be to use:
`difference_countries = set(country_list).symmetric_difference(set(bus_country_list))`
difference_countries should be empty if the two sets match, othewise difference_countries contains the items that are missing in one set or in the other one

if country list is ["AG"]and bus_country_list is ["AG", "Something"], the current revision doesn't catch the difference while the previousone using len does

scripts/build_shapes.py
davide-f: My question was because the unary_union merges all shapes of all countries and I was wondering if that introduces (a) additional computational time that may not be required and (b) alter the results as a single eez of one country would be compared to a merged shape by the unary_union.
Have you tested it with multiple countries and check that the output shapes are ok?

scripts/build_shapes.py
davide-f: I think too that by default shapes shouldn't be dropped, it was just a simple proposal to provide 2 options

ekatef:
> I've added some comments, I fear there have been some misunderstanding; hope not

@davide-f, thanks a lot for the review. I'd say it's an iterative work process. Which seems to be converging :)

Regarding `drop` option: the point is that some of the contended areas contain substations and/or power plants:

image

image

points in greens are buses, points in reds are powerplants extracted from PPM 

So, it looks like when dropping the non-standard areas we may need to add further modifications along the workflow: e.g. we may need to fix generators which belong to the requested area but can't be located there

scripts/build_osm_network.py
ekatef: Agree, if the case you mention arise, it should be fixed

But the approach should be different. Would it be probably a good idea to have an additional check on `set(bus_country_list).difference(set(country_list))` and throw an error if it happens? Because this would mean that our attempt to fix it in `build_shapes` has failed

scripts/build_shapes.py
ekatef: Yes, I have tested it and it looks ok. But I'd be happy to understand the underlying behaviour of `difference` and `unary_union` better. It looks like there is some implicit grouping by country name and looping inside it. Regarding performance, would be very interested in your opinion. This aspect I haven't (yet) considered properly

davide-f:
> > I've added some comments, I fear there have been some misunderstanding; hope not
> 
> @davide-f, thanks a lot for the review. I'd say it's an iterative work process. Which seems to be converging :)
> 
> Regarding `drop` option: the point is that some of the contended areas contain substations and/or power plants:
> 
> image
> 
> image
> 
> points in greens are buses, points in reds are powerplants extracted from PPM
> 
> So, it looks like when dropping the non-standard areas we may need to add further modifications along the workflow: e.g. we may need to fix generators which belong to the requested area but can't be located there

As discussed, don't worry too much about the drop option: I'd like to add it only for "marketing" purposes than actual need.
The default option would be to use the set_by_country option or alike.
I don't expect users to actually use the drop option, they usually use the standard option.
If the user chooses to change that option and something fails, it may be the opportunity to investigate that issue with them.
Those users may also be more sensible than us on the concerns on contended areas and may propose a better approach on how to address it.

We could also not implement the drop option (though it is just 2 lines of code) but we need to set the option in the config for marketing purposes, as discussed: we need to give users that may be sensible to the contended issue the vision that we are sensible to that and where they may included that.

ekatef:
> As discussed, don't worry too much about the drop option: I'd like to add it only for "marketing" purposes than actual need.
> 
> The default option would be to use the set_by_country option or alike.
> 
> I don't expect users to actually use the drop option, they usually use the standard option.
> 
> If the user chooses to change that option and something fails, it may be the opportunity to investigate that issue with them.
> 
> Those users may also be more sensible than us on the concerns on contended areas and may propose a better approach on how to address it.
> 
> 
> 
> We could also not implement the drop option (though it is just 2 lines of code) but we need to set the option in the config for marketing purposes, as discussed: we need to give users that may be sensible to the contended issue the vision that we are sensible to that and where they may included that.

Love your attitude :) My feeling is that the approach you propose can work very well.

Have introduced some improvements to address your comments. History looks indeed terrible but as promised I'll fix it once the content will be finalized.

Would be grateful if you could look into the changes

davide-f:
I like the new version, few minor comments and we are close to finalize it I think :D

scripts/build_shapes.py
```diff
@@ -80,7 +81,64 @@ def download_GADM(country_code, update=False, out_logging=False):
...
+    # convert country name representation of the main country (GID_0 column)
+    geodf["GID_0"] = geodf["GID_0"].map(three_2_two_digits_country)
+
+    if keep_all_areas:
+        # in case GID_0 have any exotic values, "COUNTRY" column may be used instead
+        geodf["GID_0"] = geodf.apply(
+            lambda x: replace_nonstd_codes(x, col="GID_0", country_code=cc), axis=1
+        )
+        # "not found" hardcoded according to country_converter conventions
+        geodf.drop(geodf[geodf["GID_0"] == "not found"].index, inplace=True)
```
davide-f: I think lines from 98 to 108 may be replaced with the following if GID_0 supposed to be the 3 country letter code?
``` python

geodf.drop(geodf[geodf["GID_0"] != two_2_three_digits_country(cc)].index, inplace=True)
```
Calling the conversion function may take some time that may be avoided, unless necessary.

What do you think?

scripts/build_shapes.py
```diff
@@ -113,11 +171,17 @@ def get_GADM_layer(country_list, layer_id, geo_crs, update=False, outlogging=Fal
...
 
-        # convert country name representation of the main country (GID_0 column)
-        geodf_temp["GID_0"] = [
-            three_2_two_digits_country(twoD_c) for twoD_c in geodf_temp["GID_0"]
-        ]
+        geodf_temp = filter_gadm(
+            geodf=geodf_temp,
+            layer=layer_id,
+            cc=country_code,
+            output_nonstd_to_csv=True,
```
davide-f: The option output_nonstd_to_csv should be set to false in my opinion.
I'd avoid to keep debug files for the normal user

scripts/build_shapes.py
```diff
@@ -805,15 +875,26 @@ def gadm(
...
     nprocesses = snakemake.config["build_shape_options"]["nprocesses"]
+    contended_areas = snakemake.config["build_shape_options"]["contended_areas"]
     EEZ_gpkg = snakemake.input["eez"]
     worldpop_method = snakemake.config["build_shape_options"]["worldpop_method"]
     gdp_method = snakemake.config["build_shape_options"]["gdp_method"]
     geo_crs = snakemake.config["crs"]["geo_crs"]
     distance_crs = snakemake.config["crs"]["distance_crs"]
 
-    country_shapes = countries(countries_list, geo_crs, update, out_logging)
+    keep_all_areas = False if contended_areas == "drop" else True
```
davide-f: I'd pass the contended_areas option directly to the function instead of parsing it to boolean; it gives more flexibility as discussed above.

scripts/build_shapes.py
```diff
@@ -805,15 +875,26 @@ def gadm(
...
 
-    country_shapes.reset_index().to_file(snakemake.output.country_shapes)
+    country_shapes = countries(
+        countries_list, geo_crs, update, out_logging, keep_all_areas=True
+    )
+
+    # there may be "holes" in the countries geometry which cause troubles along the workflow
+    # e.g. that is the case for enclaves like Dahagram–Angarpota for IN/BD
+    country_shapes_valid = (
+        country_shapes.apply(lambda x: make_valid(x) if not x.is_valid else x)
```
davide-f: This apply and filter may be moved into the countries function so to clean the main+

scripts/build_shapes.py
```diff
@@ -80,7 +81,64 @@ def download_GADM(country_code, update=False, out_logging=False):
...
+    else:
+        return row["GID_0"]
+
+
+def filter_gadm(geodf, layer, cc, output_nonstd_to_csv=True, keep_all_areas=True):
+
+    # convert country name representation of the main country (GID_0 column)
+    geodf["GID_0"] = geodf["GID_0"].map(three_2_two_digits_country)
+
+    if keep_all_areas:
```
davide-f: Instead of keep_all_areas, as discussed below, we can pass directly the option fix_contented_areas and revise this if as:
```python
if fix_contented_areas == "drop":
...
else:
   if fix_contended_areas != "set_by_country":
      logger.warning("Option contented areas not implemented; fallback to set_by_country")  (message to be improved)
    ...
```

scripts/build_shapes.py
```diff
@@ -80,7 +81,64 @@ def download_GADM(country_code, update=False, out_logging=False):
...
+    if (layer == 0) & (geodf.shape[0] > 1):
+        # take the first row only
+        geodf_union = geodf.iloc[[0]]
+        geodf_union = geodf_union.set_geometry(
+            # GeoSeries transformation is neded as Polygon type is not hashable
+            gpd.GeoSeries(unary_union(geodf["geometry"]))  # ,
+            # drop=True,
+            # inplace=True
+        )
+        geodf = geodf_union
```
davide-f: May you provide some examples of countries where multiple rows are found here for layer=0?

Eventually, these lines may be fused as
```
geodf = geodf.iloc[[0]].set_geometry([geodf.unary_union])
```
To be tested, but it should work

For compatibility purposes, I think it may be better to return a dataframe anyway.

Moreover, a warning needs to be added to advise that this is happening

scripts/build_shapes.py
```diff
@@ -80,7 +81,64 @@ def download_GADM(country_code, update=False, out_logging=False):
...
+        )
+        # "not found" hardcoded according to country_converter conventions
+        geodf.drop(geodf[geodf["GID_0"] == "not found"].index, inplace=True)
+    else:
+        geodf_non_std = geodf[geodf["GID_0"] != cc]
+        geodf.drop(geodf[geodf["GID_0"] != cc].index, inplace=True)
+        logger.warning("Regions with non-standard codes dropped:{geodf_non_std}")
+
+    # country shape should have a single geomerty
+    if (layer == 0) & (geodf.shape[0] > 1):
```
davide-f: Instead of & I'd use and

davide-f (commented):
I like it a lot more, few minor changes :)

scripts/build_shapes.py
```diff
@@ -80,7 +81,57 @@ def download_GADM(country_code, update=False, out_logging=False):
...
+    if row["GID_0"] != country_code:
+        return country_name_2_two_digits(row["COUNTRY"])
+    else:
+        return row["GID_0"]
+
+
+def filter_gadm(geodf, layer, cc, output_nonstd_to_csv=True, keep_all_areas=True):
+
+    # convert country name representation of the main country (GID_0 column)
+    geodf["GID_0"] = geodf["GID_0"].map(three_2_two_digits_country)
```
davide-f, line 100: I'd remove this line and having the filtering being done by time below in the ifs; it may be redundant.

However, we need to move the block to save the nonstandard codes here otherwise the ifs overwrite the data.
Please, revise the check of the if with the appropriate parameter (output_nonstd_to_csv), whose default value shall be false.

scripts/build_shapes.py
```diff
@@ -80,7 +81,57 @@ def download_GADM(country_code, update=False, out_logging=False):
...
+        code_three_digits = two_2_three_digits_country(cc)
+        non_std_gadm_codes = [
+            w for w in available_gadm_codes if not w.startswith(code_three_digits)
+        ]
+
+        if len(non_std_gadm_codes) > 0:
+            df_filtered = geodf[geodf["GID_0"].isin(non_std_gadm_codes)]
+            df_filtered.to_csv(
+                "resources/non_standard_gadm_" + cc + "_raw.csv", index=False
+            )
```
davide-f: Re-using code above, maybe non standard ids may be obtained with the more compact
`geodf[geodf["GID_0"] != two_2_three_digits_country(cc)].index`

Check that the if layer >= 1 shall also have the `and `with output_nonstd_csv, otherwise the output is always saved

scripts/build_shapes.py
ekatef: Thanks! Fixed

scripts/build_shapes.py
ekatef: Revised

scripts/build_shapes.py
ekatef: Agree and done

scripts/build_shapes.py
ekatef: Revised. I think it's better to assume that the all the areas should be kept by default unless "drop" is specified explicitly

scripts/build_shapes.py
ekatef: I'm testing on data for Pakistan where there are non-standard data both for layers 0 and 1. The output was GeoDataFrame, transformation to GeoSeries was needed to be able feed the geometry column with the new data due to some hashable magic. However, the resulting block is not really elegant :/

The solution you suggest is working brilliantly. Thanks a lot!

Regarding a warning not sure I get the point: do you mean to issue a warning when non-standard areas are being merged? Is it relevant for a user? Probably it would be a good idea to inform that the non-standard codes are present in the data.

Would be grateful on your opinion in this regard

scripts/build_shapes.py
ekatef: Done

scripts/build_shapes.py
ekatef: Not quite sure I understand :)
That is not filtering but ISO2 -> ISO3 transformation

scripts/build_shapes.py
ekatef: But it would drop all the non-standard codes from the very beginning instead trying to restore them. What do you mean as a conversion function?

scripts/build_shapes.py
ekatef: Fixed :) Thank you!

davide-f:
Thanks @ekatef ! :D
I think it's a good base, I think I can finalize the details starting from here.
I can take over for little changes, but first would you mind to do some squashing to reduce the history?
Once that is done, I'd like to add few commits, then we comment the PR and finalize

scripts/build_shapes.py
davide-f: Yeah, and that could would do that in a reversed order:
- if the user specifies the option drop, the drop code applies (the first ...)
- otherwise the option set_by_country applies.
   However, if the option specified by the user does not match "set_by_country" a message must be thrown to notify that the desired option does not exist.
   An error may be thrown instead

scripts/build_shapes.py
davide-f: The comment was linked to the revision described below in which the first if would take care of the option drop.

In programming a typical approach is to manage the cases from the most particular (in our case drop) with ifs-elifs to the default value (in our case set_by_country). That is a clean way to do that.

Think about the first case: what if a user adds an additional case to the if list?
We should revise the first if into something like option not in list, where list contains the options that follow. That is not so clean.
Often these conditions are infact managed like:
if a [first particular]
elif b [second particular]
else c [the default case]

If a user is willing to add an additional option, simply an additional elif is added and anything else

scripts/build_shapes.py
davide-f: I think that the user shall be notified that we are merging the shapes as they may not be with our standard approach.
Anyway, there is no need to notify the nonstandard ids, but just notifying like "contended areas have been found. Shapes are being merged"

Probably, the contended_areas option may be handy. In the future we may apply the same approach for these areas: with set_by country we merge them and with drop we drop them

scripts/build_shapes.py
davide-f: My feeling is that this first line is not necessary.
The scope of that line is to preprocess the column to identify what areas are contended zones by checking those that are converted into "not found".
The content of GID_0, indeed, is replaced into the ifs according to the flag contended zones

However, since(please confirm) the content of GID_0 is ISO3 code, then the code `geodf["GID_0"] != two_2_three_digits_country(cc)` should do the same without performing that preprocessing.

Applying the `three_2_two_digits_country` may only slow down the computational time.

scripts/build_shapes.py
ekatef: That absolutely makes sense. Thanks a lot for the explanation!

scripts/build_shapes.py
ekatef: Done

scripts/build_shapes.py
ekatef: Agree. Warnings revised

What I'm not sure about is how will `unary_union` process countries which consist of multiple connected regions. That seems to be a good idea to track it

scripts/build_shapes.py
ekatef: Thanks for the explanations!

If the workflow will work properly using directly GID_0 as ISO3 code instead transforming it into ISO2, then agree that the first line is not needed. I was following the original code when keeping this transformation:

https://github.com/pypsa-meets-earth/pypsa-earth/blob/7f6eb6ad0af1fe31dd37d7cee6a642e190bd1893/scripts/build_shapes.py#L117-L120

My understanding was that #262 issue means that we need ISO2 code everywhere but I can be wrong

ekatef:
Thank you for your guidance @davide-f :)
Have tried to implement your suggestions

Result of squashing is in #570. That is the result of `git merge --squash country_geom_fixes_clean` Not sure if it's the best approach to have a new branch and open a new PR for that. But I wanted to avoid using `reset` to guarantee that the history will be kept. What I was not able to do is to get rid of README modifications :( Again, some git spells probably could be used to modify the history but not sure if it's really a good idea

Could you please have a look?

davide-f:
Can we close this PR?

ekatef:
@davide-f, yes absolutely! :)

ekatef:
Closed to be finalized in #572

PR #564, opened 2023-01-13
Add version tagged technology data inteface

pz-max:
Adapted from https://github.com/PyPSA/pypsa-eur/pull/184: Make a new rule for retrieving the cost data from a specified version. Adjust code in load_cost function.

`retrieve_cost_data`: True, will retrieve data from technology_data
`retrieve_cost_data`: False, will use data/costs.csv  (we could potentially drop this local csv in future)

## Checklist


pz-max:
Ready to merge in case tests are successful

pz-max:
Ah test is failing. I renamed `hydrogen storage` to `hydrogen storage underground`
Problem is that the tutorial will download and overwrite the costs.csv with the old version.

Two solutions:
- Fix the costs.csv in the tutorial (not sure if this is complex?)
- Rename based on case distinction (not so nice)

Snakefile
```diff
@@ -293,6 +296,24 @@ if not config["enable"].get("build_natura_raster", False):
             shutil.copyfile(input[0], output[0])
 
 
+if config["enable"].get("retrieve_cost_data", True):
```
davide-f, line 299: This would fail with the default config file: the costs.csv file shall may be provided also by the databundles.
For this to work, thre is the need to remove in config/bundle_config the "/costs.csv" output.
Have you tried to run a default run locally also?

Snakefile
```diff
@@ -38,7 +38,10 @@ config["scenario"]["unc"] = [
 ]
 
 load_data_paths = get_load_paths_gegis("data", config)
-COSTS = "data/costs.csv"
```
davide-f, line 41: We need to improve the snakefile; it's getting superlong.... but not now...
Also in the feature, we need to understand and give priorities to the features we want

davide-f (approved):

PR #572, opened 2023-01-22
Finalize clean_build_shapes

davide-f:
# Closes # (if applicable).

This is a finalized version of the PR mainly edited by @ekatef in #563 and #570.
Big credits to @ekatef 

Tested as follows:
- PK with drop/set_by_country options
- [IN, PK]

## Checklist

### DIFF
--- a/scripts/build_shapes.py
+++ b/scripts/build_shapes.py
@@ -95,19 +95,18 @@ def filter_gadm(
 
     if not geodf_non_std.empty:
         logger.info(
-            f"Contended areas have been found for gadm layer {layer}. They will be dropped according to {contended_flag} option"
+            f"Contended areas have been found for gadm layer {layer}. They will be treated according to {contended_flag} option"
         )
 
         # NOTE: in these options GID_0 is not changed because it is modified below
         if contended_flag == "drop":
             geodf.drop(geodf_non_std.index, inplace=True)
-        else:
-            # this is the "set_by_country" option that is also the fallback for unexpected values
-            if contended_flag != "set_by_country":
-                logger.warning(
-                    f"Value '{contended_flag}' for option contented_flag is not recognized.\n"
-                    + "Fallback to 'set_by_country'"
-                )
+        elif contended_flag != "set_by_country":
+            # "set_by_country" option is the default; if this elif applies, the desired option falls back to the default
+            logger.warning(
+                f"Value '{contended_flag}' for option contented_flag is not recognized.\n"
+                + "Fallback to 'set_by_country'"
+            )
 
     # force GID_0 to be the country code for the relevant countries
     geodf["GID_0"] = cc
@@ -115,7 +114,7 @@ def filter_gadm(
     # country shape should have a single geomerty
     if (layer == 0) and (geodf.shape[0] > 1):
         logger.warning(
-            f"Country shape contains multiple contended areas that are being merged in agreement to contented_flag option '{contended_flag}'"
+            f"Country shape is composed by multiple shapes that are being merged in agreement to contented_flag option '{contended_flag}'"
         )
         # take the first row only to re-define geometry keeping other columns
         geodf = geodf.iloc[[0]].set_geometry([geodf.unary_union])
@@ -175,7 +174,7 @@ def get_GADM_layer(
             layer=layer_id,
             cc=country_code,
             contended_flag=contended_flag,
-            output_nonstd_to_csv=True,
+            output_nonstd_to_csv=False,
         )
 
         # create a subindex column that is useful
@@ -318,12 +317,13 @@ def eez(countries, geo_crs, country_shapes, EEZ_gpkg, out_logging=False, distanc
     # load data
     df_eez = load_EEZ(countries, geo_crs, EEZ_gpkg)
 
+    eez_countries = [cc for cc in countries if cc in df_eez.name]
     ret_df = gpd.GeoDataFrame(
         {
-            "name": df_eez.name.copy(),
+            "name": eez_countries,
             "geometry": [
                 df_eez.geometry.loc[df_eez.name == cc].geometry.unary_union
-                for cc in df_eez.name
+                for cc in eez_countries
             ],
         }
     ).set_index("name")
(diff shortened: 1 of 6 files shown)


davide-f:
@ekatef This is what I think may be a finalized version of the PR.
Please, add your comments so that we may finalize it.

Once agreed the code, we need to finalize the documentation e.g. release_notes and config tables.
I can quickly do that but as you mainly on it, you could do that as well.
Let me know how to handle it :)

scripts/build_shapes.py
```diff
@@ -80,7 +82,61 @@ def download_GADM(country_code, update=False, out_logging=False):
...
+    contended_flag,
+    output_nonstd_to_csv=False,
+):
+
+    # identify non standard geodf rows
+    geodf_non_std = geodf[geodf["GID_0"] != two_2_three_digits_country(cc)].copy()
+
+    if not geodf_non_std.empty:
+        logger.info(
+            f"Contended areas have been found for gadm layer {layer}. They will be dropped according to {contended_flag} option"
```
ekatef: Not necessarily dropped I think :) Can we replace "dropped" with "treated/processed/handled..."?

scripts/build_shapes.py
```diff
@@ -80,7 +82,61 @@ def download_GADM(country_code, update=False, out_logging=False):
...
+        )
+
+        # NOTE: in these options GID_0 is not changed because it is modified below
+        if contended_flag == "drop":
+            geodf.drop(geodf_non_std.index, inplace=True)
+        else:
+            # this is the "set_by_country" option that is also the fallback for unexpected values
+            if contended_flag != "set_by_country":
```
ekatef: Not sure that a three-level nested condition is easy to understand. Wouldn't be simpler to use if-elif-else?

scripts/build_shapes.py
```diff
@@ -80,7 +82,61 @@ def download_GADM(country_code, update=False, out_logging=False):
...
+                logger.warning(
+                    f"Value '{contended_flag}' for option contented_flag is not recognized.\n"
+                    + "Fallback to 'set_by_country'"
+                )
+
+    # force GID_0 to be the country code for the relevant countries
+    geodf["GID_0"] = cc
```
ekatef, line 112: That's brilliant :)

scripts/build_shapes.py
```diff
@@ -80,7 +82,61 @@ def download_GADM(country_code, update=False, out_logging=False):
...
+
+    # country shape should have a single geomerty
+    if (layer == 0) and (geodf.shape[0] > 1):
+        logger.warning(
+            f"Country shape contains multiple contended areas that are being merged in agreement to contented_flag option '{contended_flag}'"
```
ekatef: Can we change the level of the condition to be sure that merging is done if and only if there are non-standard codes?

scripts/build_shapes.py
```diff
@@ -114,10 +170,13 @@ def get_GADM_layer(country_list, layer_id, geo_crs, update=False, outlogging=Fal
...
-        # convert country name representation of the main country (GID_0 column)
-        geodf_temp["GID_0"] = [
-            three_2_two_digits_country(twoD_c) for twoD_c in geodf_temp["GID_0"]
-        ]
+        geodf_temp = filter_gadm(
+            geodf=geodf_temp,
+            layer=layer_id,
+            cc=country_code,
+            contended_flag=contended_flag,
+            output_nonstd_to_csv=True,
```
ekatef: Isn't it meant to be `output_nonstd_to_csv=False` in the final version? :)

scripts/build_shapes.py
```diff
@@ -248,26 +318,21 @@ def eez(countries, geo_crs, country_shapes, EEZ_gpkg, out_logging=False, distanc
...
-            ret_df.drop(ret_df[selection].index, inplace=True)
-            ret_df = ret_df.append(
-                {"name": c_code, "geometry": geom}, ignore_index=True
-            )
+    ret_df = gpd.GeoDataFrame(
+        {
+            "name": df_eez.name.copy(),
+            "geometry": [
+                df_eez.geometry.loc[df_eez.name == cc].geometry.unary_union
+                for cc in df_eez.name
```
ekatef: An elegant replacement. But wouldn't it lead to a mismatch when there are multiple rows for a single country?

scripts/build_shapes.py
davide-f: Agree :) moved the text improperly, many thanks!

scripts/build_shapes.py
davide-f: Revised slightly

scripts/build_shapes.py
davide-f: thanks, hope it is understandable too

scripts/build_shapes.py
davide-f: I feel like we want that to happen even when there are not standard zones.
However, it is important to notify that.
Anyway, this condition would only apply to country shapes so I think it is fine

scripts/build_shapes.py
davide-f: Definitely, thanks for spotting! I forgot to disable it after the debug

scripts/build_shapes.py
davide-f: Nice spotting! Revised

davide-f:
@ekatef Please revise also including a comment on how to proceed for the release_docs.
I realized that configtables are missing for build_shapes and created an issue for that. I won't be addressing that issue here so I'd proceed even without that option implemented.

Before proceeding, I'd merge all my commits into a single one by resetting and merging everything again.

scripts/build_shapes.py
ekatef: I think now it's as clear as possible

ekatef (approved):

davide-f:
Merging the PR once CI works.
I've cherry-picked your commit and the PR seems ready to go. Let me know your thoughts :)

ekatef:
@davide-f, thank you :) I think it looks great and looking forward to see how it works globally

Agree on that we both have to add configtables and can do it as a new PR. Touched with you suggestion on adding a release note :) But I'm also absolutely happy if you add a release note yourself as it could be much simpler

davide-f:
> @davide-f, thank you :) I think it looks great and looking forward to see how it works globally
> 
> Agree on that we both have to add configtables and can do it as a new PR. Touched with you suggestion on adding a release note :) But I'm also absolutely happy if you add a release note yourself as it could be much simpler

I added your release note :) commit e196ac41402acbca8ee48f72d0215e335b545f78

Issue #609, opened 2023-02-16
Update on documentation - configuration section


asv365:
Increased specifications and typo fixes in the documentation. The proposed changes in the documentation are:

1.	Specify a limit on snapshots/start and snapshots/end. Is 2023_01_01 a good upper bound? Is there a lower bound?

2.	There is no section in the documentation for load_options. There, we can specify that the load_options/weather_year has to be within the snapshot range. On load_options/prediction_year are 2030, 2040, 2050, and 2100 the only options for now?
 
3.	The picture in the lines section includes too many things, the “lines” only appear at the very end. Replace this.

4.	Create subsections in the electricity section, as in the renewables section.

5.	The section load from the documentation doesn’t seem to refer to anything on the config file. Shall we remove this?

ekatef:
Hello @asv365! Thanks a lot for your feedback. That is very helpful to improve the model usability.

Let's go point by point:

1. The snapshot should correspond to the timeframe for which cutout has been built.
2. Good point! The available `weather_year` and `prediction_year` values can be checked by looking into `pypsa-earth/data/ssp2-2.6` folder. Currently there are demand data for 2011, 2013, 2018 weather years and for 2030, 2040, 2050, and 2100 scenario prediction years

I think we may include some explanations on points 1 and 2 into an ongoing PR #603. Many thanks for your suggestions!

Regarding 3-5, absolutely agree that the section structure could be more clear. I'd also say that it would be helpful to supplement each subsection with a short description like it's done for `run`. However, such a revision should keep in sight the whole structure `Configuration` section and be consistent with the configuration files. If you want to work on this we would be happy to support you :)

asv365:
Thanks for your feedback, @ekatef! Shall I write a comment on PR #603 about points 1. and 2.?

Regarding point 1., is there a limit on the years for which the cutouts can be built?

I will start working on points 3-5 soon, creating a pull request once it is ready. Agreed on the short description like it's done for `run`.

ekatef:
> Thanks for your feedback, @ekatef! Shall I write a comment on PR #603 about points 1. and 2.?
> 
> Regarding point 1., is there a limit on the years for which the cutouts can be built?
> 
> I will start working on points 3-5 soon, creating a pull request once it is ready. Agreed on the short description like it's done for `run`.

Thanks, I have linked #603 with our discussion here :)

[...]

PR #612, opened 2023-02-20
Documentation improvements in the Configuration section 

asv365:
# Closes # 
This pull request aims to improve the documentation of the Configuration section. It addresses points 3, 4 and 5 of #609. Thanks @ekatef for the feedback.

1. The picture in the lines section includes too many things, the “lines” only appear at the very end.

2. Create subsections in the electricity section, as in the renewables section.

3. The section load from the documentation doesn’t seem to refer to anything on the config file. Shall we remove this?

4. Add a short description in each subsection, as it is done in `run`.

5. General consistency check accross the configuaration section.

## Checklist

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -97,7 +97,7 @@ Specifies the temporal range to build an energy system model for as arguments to
 .. _crs_cf:
 
 ``crs``
-===============
+=======
 
 Defines the coordinate reference systems (crs).
 
@@ -110,6 +110,23 @@ Defines the coordinate reference systems (crs).
    :header-rows: 1
    :widths: 25,7,22,30
    :file: configtables/crs.csv
+
+.. _augmented_line_connection_cf:
+
+``augmented_line_connection``
+=============================
+
+.. literalinclude:: ../config.default.yaml
+   :language: yaml
+   :start-at: augmented_line_connection:
+   :end-at: min_DC_length:
+
+.. csv-table::
+   :header-rows: 1
+   :widths: 25,7,22,30
+   :file: configtables/augmented_line_connection.csv
+
+
 
 .. _electricity_cf:
 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -244,7 +244,7 @@ Define and specify the ``atlite.Cutout`` used for calculating renewable potentia
    :widths: 25,7,22,30
    :file: configtables/transformers.csv
 
-.. _load_cf:
+.. _load_options_cf:
 
 ``load_options``
 =============
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -206,7 +206,7 @@ Define and specify the ``atlite.Cutout`` used for calculating renewable potentia
 
 .. literalinclude:: ../config.default.yaml
    :language: yaml
-   :start-at: lines:
+   :start-at: PV:
    :end-before: links:
 
 .. csv-table::
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -94,6 +94,23 @@ Specifies the temporal range to build an energy system model for as arguments to
    :widths: 25,7,22,30
    :file: configtables/snapshots.csv
 
+.. _crs_cf:
+
+``crs``
+===============
+
+Defines the coordinate reference systems (crs).
+
+.. literalinclude:: ../config.default.yaml
+   :language: yaml
+   :start-at: crs:
+   :end-at: area_crs:
+
+.. csv-table::
+   :header-rows: 1
+   :widths: 25,7,22,30
+   :file: configtables/crs.csv
+
 .. _electricity_cf:
 
 ``electricity``
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -206,7 +206,7 @@ Define and specify the ``atlite.Cutout`` used for calculating renewable potentia
 
 .. literalinclude:: ../config.default.yaml
    :language: yaml
-   :start-at: PV:
+   :start-after: PV:
    :end-before: links:
 
 .. csv-table::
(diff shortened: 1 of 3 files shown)


asv365:
Point 1 is caused by the appearance of the text "lines:" before the lines section header. The lines section header is where the image should start to be displayed. 

For context, the images are created using the instruction:

.. literalinclude:: ../config.default.yaml
   :language: yaml
   :start-at: lines:
   :end-before: links:

Since the text ''lines:'' appears before in the config.default.yaml file (namely in the instruction split_overpassing_**lines:** true), the image displayed is from that first appearance of "lines:" until the first appearance of the word "links:".

Haven't found a good fix yet, but to illustrate the problem the first commit changes "lines:" for "PV:", which is an improvement.

ekatef:
@asv365 A very good point regarding catching text from a wrong fragment. Thanks a lot for investigating it!

Probably, Sphinx docs on [literalinclude](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude) could be helpful to resolve this, eventually combined with other Shpinx magic. The most straightforward way seems to be use tag comments. But I suspect there might be a more elegant solution, as well.

asv365:
Thanks @ekatef, the suggestion on comments can be a good workaround if we can't find how to do it properly with Sphinx. If Sphinx proves tricky, we could add a header to every section of the config file, to be consistent.

asv365:
On point 3., the section "load" has been renamed "operational_reserve", and has been placed as a subsection of electricity. There is also the section "load_options", which doesn't exist in the documentation. This will be addressed as point 5, as part of the general consistency check.

asv365:
@ekatef  found a solution to the point 1 we were discussing earlier without the need to use comments in the config file. 

Literalinclude allows the command `start-after` as well as `start-at`. `start-after` ignores the line you are pointing to, and starts displaying in the next one. Since the previous line is `PV:`, which is unique, it works for this case.

ekatef:
> @ekatef found a solution to the point 1 we were discussing earlier without the need to use comments in the config file.
> 
> Literalinclude allows the command `start-after` as well as `start-at`. `start-after` ignores the line you are pointing to, and starts displaying in the next one. Since the previous line is `PV:`, which is unique, it works for this case.

@asv365 a very nice trick which looks absolutely in spirit of Sphynx! :) 

Probably, there is a little risk to catch an additional section in case it would be inserted between `estimate_renewable_capacities` and `lines`...  Could it make sense to add a comment to `configuration.rst` as a reminder on why do we refer exactly `PV`?

Although, the work you are doing is still a great improvement for clarity of the docs 🥇

asv365:
Good point @ekatef. I'll add that comment in `configuration.rst`

asv365:
@ekatef The top-level configuration section (which, according to the table in the docs includes "version", "tutorial", "logging", "countries", and "enable") has other sections inbetween. For example, "run" and "scenarios" are between "logging" and "countries" in the `config.default.yaml`. 

Do we want to bring all the items of the top-level configuration together at the top of the `config.default.yaml` and `config.tutorial.yaml`? Not doing that forces literalinclude to display using line numbers, which was creating trouble earlier. See, e.g., commit [d71a3a5](https://github.com/pypsa-meets-earth/pypsa-earth/pull/612/commits/d71a3a5cc73d60df5ca8da0214335c3ecc23cfee) above. 

Another option would be to include those intermediate sections as part of the top-level configuration. Let me know what you think :).

ekatef:
@asv365, thanks for rising this point! Our current `top-level configuration` is in fact rendered in not optimal way...

Completely agree that it might be worth to revise underlying assumptions for making the result more robust and the docs more convenient. I feel that it is a very good point for discussion during developers meeting :) Would you mind to introduce this topic tomorrow?

asv365:
Thank you @ekatef. I'll raise it in tomorrow's meeting :)

asv365:
In the pypsa-earth weekly meeting on 02/03/2023, it was decided that:

1. In literalinclude, preference of "start-at" like command instead of hard-coded line numbers.
2. The top-level configuration section should include the sections that are already there. Reshuffling in config.default.yaml allowed, but keep to the bare minimum and report it in the PR. Another option is to include several "literalincludes".
3. Make the Value column as short as possible, and very symbolic.

doc/configuration.rst
```diff
@@ -25,7 +25,7 @@ Top-level configuration
 
 .. literalinclude:: ../config.default.yaml
    :language: yaml
-   :lines: 5-12,20,31-38
+   :lines: 5-12,24,31-38
```
pz-max, line 28: I think we will change that by the `start-at` and `end-at` stuff?

pz-max (commented):
pre-commit needs fixes as it could not automatically get fixed.
How far are we from merging this  PR @asv365 ?

doc/configuration.rst
asv365: yes @pz-max, that is the plan. Just did this to motivate the `start-at` and `end-at` approach. Will work through this soon I hope, a lot of work has popped up recently.

ekatef:
Hey @asv365! Very nice work so far :)
Could you please give an update on the PR status? Don't hesitate to let us know if you need any additional inputs. Would be happy to support you if needed

asv365:
Hi @ekatef, everything is good, thanks for asking and for offering help. I don't think there are bottlenecks so far.

The reason for being so slow are job applications. Will come back to this in the next week or the one after hopefully. The dust is settling now. Apologies for the delay.

ekatef:
@asv365, no worries! Good luck with your applications and come back once you'd have more time ;)

asolavi:
This pull request is being addressed in PR #694 because a fresh restart was useful.

davide-f:
Hello :)
I'm closing this PR as it is continued into the other PR.
Thanks for the contribution!

asolavi:
> Hello :) I'm closing this PR as it is continued into the other PR. Thanks for the contribution!

Thank you, yes!

Issue #586, opened 2023-02-01
Error in simplify_network due to misinterpretation of links


davide-f:
## Checklist

- [x] I am using the current [`main`](https://github.com/pypsa-meets-earth/pypsa-earth/tree/main) branch or the latest [release](https://github.com/pypsa-meets-earth/pypsa-earth/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-earth` environment](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

When running the workflow branch [create_statistics](https://github.com/pypsa-meets-earth/pypsa-earth/pull/579) for China, an error on simplify_network appears that needs to be addressed.

This issue requires:
- investigating and fixing simplify_links as needed; could be worth to compare the script with the pypsa-eur developments to see if we missed something
- it would be worth investigating the effects of simplify_links with option hvdc_as_lines = true

## Error Message

```bash
Traceback (most recent call last):
  File "/data/davidef/git_world/pypsa-earth/.snakemake/scripts/tmp14nyount.simplify_network.py", line 532, in <module>
    n, simplify_links_map = simplify_links(
  File "/data/davidef/git_world/pypsa-earth/.snakemake/scripts/tmp14nyount.simplify_network.py", line 346, in simplify_links
    lengths = n.links.loc[all_links, "length"]
  File "/home/davidef/miniconda3/envs/pypsa-earth/lib/python3.8/site-packages/pandas/core/indexing.py", line 1067, in __getitem__
    return self._getitem_tuple(key)
  File "/home/davidef/miniconda3/envs/pypsa-earth/lib/python3.8/site-packages/pandas/core/indexing.py", line 1247, in _getitem_tuple
    return self._getitem_lowerdim(tup)
  File "/home/davidef/miniconda3/envs/pypsa-earth/lib/python3.8/site-packages/pandas/core/indexing.py", line 991, in _getitem_lowerdim
    return getattr(section, self.name)[new_key]
  File "/home/davidef/miniconda3/envs/pypsa-earth/lib/python3.8/site-packages/pandas/core/indexing.py", line 1073, in __getitem__
    return self._getitem_axis(maybe_callable, axis=axis)
  File "/home/davidef/miniconda3/envs/pypsa-earth/lib/python3.8/site-packages/pandas/core/indexing.py", line 1301, in _getitem_axis
[...]

PR #614, opened 2023-02-26
Fix islanded buses and add clustering by networks

### LINKED ISSUE #531: Verify the effect of using/disregarding clustering by subnetworks in cluster_network

ekatef:
A fix for islanded buses treatment is introduced into `simplify_network` to switch-on clustering by country and sub-network in `cluster_network`.

As discussed with @davide-f (thanks a lot for the discussions!), all the load and generation capacity is aggregated and attached to a single islanded bus.

Position of this representative islanded buses should calculated:
* disregarding buses having a load below a given threshold;
* using spatial-weighted average of the buses with the load higher a threshold.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,81 +155,10 @@ The documentation is available here: [documentation](https://pypsa-earth.readthe
 <table>
 <tr>
     <td align="center">
-        <a href="https://github.com/hazemakhalek">
-            <img src="https://avatars.githubusercontent.com/u/26235356?v=4" width="100;" alt="hazemakhalek"/>
-            <br />
-            <sub><b>Null</b></sub>
-        </a>
-    </td>
-    <td align="center">
-        <a href="https://github.com/jarry7">
-            <img src="https://avatars.githubusercontent.com/u/27745389?v=4" width="100;" alt="jarry7"/>
-            <br />
-            <sub><b>Jarrad Wright</b></sub>
-        </a>
-    </td>
-    <td align="center">
-        <a href="https://github.com/fneum">
-            <img src="https://avatars.githubusercontent.com/u/29101152?v=4" width="100;" alt="fneum"/>
-            <br />
-            <sub><b>Fabian Neumann</b></sub>
-        </a>
-    </td>
-    <td align="center">
         <a href="https://github.com/ekatef">
             <img src="https://avatars.githubusercontent.com/u/30229437?v=4" width="100;" alt="ekatef"/>
             <br />
             <sub><b>Ekaterina</b></sub>
-        </a>
-    </td>
-    <td align="center">
-        <a href="https://github.com/euronion">
-            <img src="https://avatars.githubusercontent.com/u/42553970?v=4" width="100;" alt="euronion"/>
-            <br />
-            <sub><b>Euronion</b></sub>
-        </a>
-    </td>
-    <td align="center">
-        <a href="https://github.com/Justus-coded">
-            <img src="https://avatars.githubusercontent.com/u/44394641?v=4" width="100;" alt="Justus-coded"/>
-            <br />
-            <sub><b>Justus Ilemobayo</b></sub>
-        </a>
-    </td></tr>
-<tr>
-    <td align="center">
-        <a href="https://github.com/mnm-matin">
-            <img src="https://avatars.githubusercontent.com/u/45293386?v=4" width="100;" alt="mnm-matin"/>
-            <br />
-            <sub><b>Mnm-matin</b></sub>
-        </a>
-    </td>
-    <td align="center">
-        <a href="https://github.com/desenk">
-            <img src="https://avatars.githubusercontent.com/u/48335263?v=4" width="100;" alt="desenk"/>
-            <br />
-            <sub><b>Desen Kirli</b></sub>
-        </a>
-    </td>
-    <td align="center">
-        <a href="https://github.com/LukasFrankenQ">
-            <img src="https://avatars.githubusercontent.com/u/55196140?v=4" width="100;" alt="LukasFrankenQ"/>
-            <br />
-            <sub><b>Lukas Franken</b></sub>
-        </a>
-    </td>
-    <td align="center">
-        <a href="https://github.com/pz-max">
-            <img src="https://avatars.githubusercontent.com/u/61968949?v=4" width="100;" alt="pz-max"/>
-            <br />
-            <sub><b>Max Parzen</b></sub>
-        </a>
-    </td>
-    <td align="center">
-        <a href="https://github.com/Cesare-Caputo">
-            <img src="https://avatars.githubusercontent.com/u/62548290?v=4" width="100;" alt="Cesare-Caputo"/>
-            <br />
-            <sub><b>Null</b></sub>
         </a>
     </td>
     <td align="center">
@@ -238,34 +167,12 @@ The documentation is available here: [documentation](https://pypsa-earth.readthe
             <br />
             <sub><b>Davide-f</b></sub>
         </a>
-    </td></tr>
-<tr>
-    <td align="center">
-        <a href="https://github.com/koen-vg">
-            <img src="https://avatars.githubusercontent.com/u/74298901?v=4" width="100;" alt="koen-vg"/>
-            <br />
-            <sub><b>Koen Van Greevenbroek</b></sub>
-        </a>
     </td>
     <td align="center">
-        <a href="https://github.com/Hazem-IEG">
-            <img src="https://avatars.githubusercontent.com/u/87850910?v=4" width="100;" alt="Hazem-IEG"/>
+        <a href="https://github.com/pz-max">
+            <img src="https://avatars.githubusercontent.com/u/61968949?v=4" width="100;" alt="pz-max"/>
             <br />
-            <sub><b>Hazem-IEG</b></sub>
-        </a>
-    </td>
-    <td align="center">
-        <a href="https://github.com/energyLS">
-            <img src="https://avatars.githubusercontent.com/u/89515385?v=4" width="100;" alt="energyLS"/>
-            <br />
-            <sub><b>EnergyLS</b></sub>
-        </a>
-    </td>
-    <td align="center">
-        <a href="https://github.com/AnasAlgarei">
-            <img src="https://avatars.githubusercontent.com/u/101210563?v=4" width="100;" alt="AnasAlgarei"/>
-            <br />
-            <sub><b>AnasAlgarei</b></sub>
+            <sub><b>Max Parzen</b></sub>
         </a>
     </td>
     <td align="center">
@@ -281,8 +188,29 @@ The documentation is available here: [documentation](https://pypsa-earth.readthe
             <br />
             <sub><b>Null</b></sub>
         </a>
+    </td>
+    <td align="center">
+        <a href="https://github.com/mnm-matin">
+            <img src="https://avatars.githubusercontent.com/u/45293386?v=4" width="100;" alt="mnm-matin"/>
+            <br />
+            <sub><b>Mnm-matin</b></sub>
+        </a>
     </td></tr>
 <tr>
+    <td align="center">
+        <a href="https://github.com/Hazem-IEG">
+            <img src="https://avatars.githubusercontent.com/u/87850910?v=4" width="100;" alt="Hazem-IEG"/>
+            <br />
+            <sub><b>Hazem-IEG</b></sub>
+        </a>
+    </td>
+    <td align="center">
+        <a href="https://github.com/energyLS">
+            <img src="https://avatars.githubusercontent.com/u/89515385?v=4" width="100;" alt="energyLS"/>
+            <br />
+            <sub><b>EnergyLS</b></sub>
+        </a>
+    </td>
     <td align="center">
         <a href="https://github.com/Tomkourou">
             <img src="https://avatars.githubusercontent.com/u/5240283?v=4" width="100;" alt="Tomkourou"/>
@@ -305,10 +233,39 @@ The documentation is available here: [documentation](https://pypsa-earth.readthe
         </a>
     </td>
     <td align="center">
+        <a href="https://github.com/euronion">
+            <img src="https://avatars.githubusercontent.com/u/42553970?v=4" width="100;" alt="euronion"/>
+            <br />
+            <sub><b>Euronion</b></sub>
+        </a>
+    </td></tr>
+<tr>
+    <td align="center">
+        <a href="https://github.com/AnasAlgarei">
+            <img src="https://avatars.githubusercontent.com/u/101210563?v=4" width="100;" alt="AnasAlgarei"/>
+            <br />
+            <sub><b>AnasAlgarei</b></sub>
+        </a>
+    </td>
+    <td align="center">
+        <a href="https://github.com/LukasFrankenQ">
+            <img src="https://avatars.githubusercontent.com/u/55196140?v=4" width="100;" alt="LukasFrankenQ"/>
+            <br />
+            <sub><b>Lukas Franken</b></sub>
+        </a>
+    </td>
+    <td align="center">
         <a href="https://github.com/Tooblippe">
             <img src="https://avatars.githubusercontent.com/u/805313?v=4" width="100;" alt="Tooblippe"/>
             <br />
             <sub><b>Jarrad Wright</b></sub>
+        </a>
+    </td>
+    <td align="center">
+        <a href="https://github.com/koen-vg">
+            <img src="https://avatars.githubusercontent.com/u/74298901?v=4" width="100;" alt="koen-vg"/>
+            <br />
+            <sub><b>Koen Van Greevenbroek</b></sub>
         </a>
     </td>
     <td align="center">
@@ -326,6 +283,13 @@ The documentation is available here: [documentation](https://pypsa-earth.readthe
         </a>
     </td></tr>
 <tr>
+    <td align="center">
+        <a href="https://github.com/jarry7">
+            <img src="https://avatars.githubusercontent.com/u/27745389?v=4" width="100;" alt="jarry7"/>
+            <br />
+            <sub><b>Jarrad Wright</b></sub>
+        </a>
+    </td>
     <td align="center">
         <a href="https://github.com/squoilin">
             <img src="https://avatars.githubusercontent.com/u/4547840?v=4" width="100;" alt="squoilin"/>
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -62,6 +62,8 @@ cluster_options:
     exclude_carriers: []
     remove_stubs: true
     remove_stubs_across_borders: true
+    p_threshold_drop_isolated: 10 # [MW] isolated buses are being discarded if bus mean power is below the specified threshold 
+    p_threshold_merge_isolated: 300 # [MW] isolated buses are being merged into a single isolated bus if bus mean power is above the specified threshold 
   cluster_network:
     algorithm: kmeans
     feature: solar+onwind-time
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -62,8 +62,8 @@ cluster_options:
     exclude_carriers: []
     remove_stubs: true
     remove_stubs_across_borders: true
-    p_threshold_drop_isolated: 10 # [MW] isolated buses are being discarded if bus mean power is below the specified threshold 
-    p_threshold_merge_isolated: 300 # [MW] isolated buses are being merged into a single isolated bus if bus mean power is above the specified threshold 
+    p_threshold_drop_isolated: 10 # [MW] isolated buses are being discarded if bus mean power is below the specified threshold
+    p_threshold_merge_isolated: 300 # [MW] isolated buses are being merged into a single isolated bus if bus mean power is above the specified threshold
   cluster_network:
     algorithm: kmeans
     feature: solar+onwind-time
--- a/config.tutorial.yaml
+++ b/config.tutorial.yaml
@@ -74,6 +74,8 @@ cluster_options:
     exclude_carriers: []
     remove_stubs: true
     remove_stubs_across_borders: true
+    p_threshold_drop_isolated: 10 # [MW] isolated buses are being discarded if bus mean power is below the specified threshold 
+    p_threshold_merge_isolated: 300 # [MW] isolated buses are being merged into a single isolated bus if bus mean power is above the specified threshold 
   cluster_network:
     algorithm: kmeans
     feature: solar+onwind-time
--- a/config.tutorial.yaml
+++ b/config.tutorial.yaml
@@ -74,8 +74,8 @@ cluster_options:
     exclude_carriers: []
     remove_stubs: true
     remove_stubs_across_borders: true
-    p_threshold_drop_isolated: 10 # [MW] isolated buses are being discarded if bus mean power is below the specified threshold 
-    p_threshold_merge_isolated: 300 # [MW] isolated buses are being merged into a single isolated bus if bus mean power is above the specified threshold 
+    p_threshold_drop_isolated: 10 # [MW] isolated buses are being discarded if bus mean power is below the specified threshold
+    p_threshold_merge_isolated: 300 # [MW] isolated buses are being merged into a single isolated bus if bus mean power is above the specified threshold
   cluster_network:
     algorithm: kmeans
     feature: solar+onwind-time
--- a/scripts/cluster_network.py
+++ b/scripts/cluster_network.py
@@ -197,7 +197,7 @@ def get_feature_for_hac(n, buses_i=None, feature=None):
     if "offwind" in carriers:
         carriers.remove("offwind")
         carriers = np.append(
-            carriers, network.generators.carrier.filter(like="offwind").unique()
+            carriers, n.generators.carrier.filter(like="offwind").unique()
         )
 
     if feature.split("-")[1] == "cap":
@@ -249,13 +249,14 @@ def distribute_clusters(
             n.loads_t.p_set.mean()
             .groupby(n.loads.bus)
             .sum()
-            .groupby([n.buses.country])
+            .groupby([n.buses.country, n.buses.sub_network])
             .sum()
             .pipe(normed)
         )
-        assert len(L.index) == len(n.buses.country.unique()), (
+        countries_in_L = pd.unique(L.index.get_level_values(0))
+        assert len(countries_in_L) == len(n.buses.country.unique()), (
             "The following countries have no load: "
-            f"{list(set(L.index).symmetric_difference(set(n.buses.country.unique())))}"
+            f"{list(set(countries_in_L).symmetric_difference(set(n.buses.country.unique())))}"
         )
         distribution_factor = L
 
@@ -264,11 +265,17 @@ def distribute_clusters(
             columns={"name": "country"}
         )
         add_population_data(
-            df_pop_c, country_list, year, update, out_logging, nprocesses=nprocesses
+            df_pop_c, country_list, "standard", year, update, out_logging
         )
         P = df_pop_c.loc[:, ("country", "pop")]
-        P = P.groupby(P["country"]).sum().pipe(normed).squeeze()
-        distribution_factor = P
+        n_df = n.buses.copy()[["country", "sub_network"]]
+
+        pop_dict = P.set_index("country")["pop"].to_dict()
+        n_df["pop"] = n_df["country"].map(pop_dict)
+
+        distribution_factor = (
+            n_df.groupby(["country", "sub_network"]).sum().pipe(normed).squeeze()
+        )
 
     if distribution_cluster == ["gdp"]:
         df_gdp_c = gpd.read_file(inputs.country_shapes).rename(
@@ -281,12 +288,19 @@ def distribute_clusters(
             out_logging,
             name_file_nc="GDP_PPP_1990_2015_5arcmin_v2.nc",
         )
+
         G = df_gdp_c.loc[:, ("country", "gdp")]
-        G = G.groupby(df_gdp_c["country"]).sum().pipe(normed).squeeze()
-        distribution_factor = G
+        n_df = n.buses.copy()[["country", "sub_network"]]
+
+        gdp_dict = G.set_index("country")["gdp"].to_dict()
+        n_df["gdp"] = n_df["country"].map(gdp_dict)
+
+        distribution_factor = (
+            n_df.groupby(["country", "sub_network"]).sum().pipe(normed).squeeze()
+        )
 
     # TODO: 1. Check if sub_networks can be added here i.e. ["country", "sub_network"]
-    N = n.buses.groupby(["country"]).size()
+    N = n.buses.groupby(["country", "sub_network"]).size()
 
     assert (
         n_clusters >= len(N) and n_clusters <= N.sum()
@@ -318,7 +332,7 @@ def distribute_clusters(
 
     m = po.ConcreteModel()
 
-    def n_bounds(model, n_id):
+    def n_bounds(model, *n_id):
         """
         Create a function that makes a bound pair for pyomo
 
@@ -379,7 +393,11 @@ def busmap_for_gadm_clusters(inputs, n, gadm_level, geo_crs, country_list):
     buses["gadm_{}".format(gadm_level)] = buses[["x", "y", "country"]].apply(
         lambda bus: locate_bus(bus[["x", "y"]], bus["country"]), axis=1
     )
-    busmap = buses["gadm_{}".format(gadm_level)]
+
+    buses["gadm_subnetwork"] = (
+        buses["gadm_{}".format(gadm_level)] + "_" + buses["carrier"].astype(str)
+    )
+    busmap = buses["gadm_subnetwork"]
 
     return busmap
 
@@ -476,11 +494,10 @@ def busmap_for_n_clusters(
     def busmap_for_country(x):
         # A number of the countries in the clustering can be > 1
         if isinstance(n_clusters, pd.Series):
+            n_cluster_c = n_clusters[x.name]
             if isinstance(x.name, tuple):
-                n_cluster_c = n_clusters[x.name[0]]
                 prefix = x.name[0] + x.name[1] + " "
             else:
-                n_cluster_c = n_clusters[x.name]
                 prefix = x.name + " "
         else:
             n_cluster_c = n_clusters
@@ -519,8 +536,8 @@ def busmap_for_n_clusters(
 
     return (
         n.buses.groupby(
-            ["country"],
-            # ["country", "sub_network"] # TODO: 2. Add sub_networks (see previous TODO)
+            # ["country"],
+            ["country", "sub_network"],  # TODO: 2. Add sub_networks (see previous TODO)
             group_keys=False,
         )
         .apply(busmap_for_country)
--- a/scripts/simplify_network.py
+++ b/scripts/simplify_network.py
@@ -423,7 +423,20 @@ def aggregate_to_substations(n, aggregation_strategies=dict(), buses_i=None):
         logger.info(
             "Aggregating buses that are no substations or have no valid offshore connection"
         )
-        buses_i = list(set(n.buses.index) - set(n.generators.bus) - set(n.loads.bus))
+        # isolated buses should be excluded from adjacency_matrix calculations
+        n.determine_network_topology()
+        # duplicated sub-networks mean that there is at least one interconnection between buses
+        i_islands = n.buses[
+            (~n.buses.duplicated(subset=["sub_network"], keep=False))
+            & (n.buses.carrier == "AC")
+        ].index
+
+        buses_i = list(
+            set(n.buses.index)
+            - set(i_islands)
+            - set(n.generators.bus)
+            - set(n.loads.bus)
+        )
 
     weight = pd.concat(
         {
@@ -446,6 +459,10 @@ def aggregate_to_substations(n, aggregation_strategies=dict(), buses_i=None):
     for c in n.buses.country.unique():
         incountry_b = n.buses.country == c
         dist.loc[incountry_b, ~incountry_b] = np.inf
+
+    for c in n.buses.carrier.unique():
+        incarrier_b = n.buses.carrier == c
+        dist.loc[incarrier_b, ~incarrier_b] = np.inf
 
     busmap = n.buses.index.to_series()
     busmap.loc[buses_i] = dist.idxmin(1)
@@ -522,6 +539,150 @@ def cluster(
     return clustering.network, clustering.busmap
 
 
+def drop_isolated_nodes(n, threshold):
+    """
+    Find isolated nodes in the network and drop those of them which don't have
+    load or have a load value lower than a threshold
+
+    Parameters
+    ----------
+    iso_code : PyPSA.Network
+        Original network
+    threshold: float
+        Load power used as a threshold to drop isolated nodes
+
+    Returns
+    -------
+    modified network
+    """
+    n.determine_network_topology()
+
+    # keep original values of the overall load and generation in the network
+    # to track changes due to drop of buses
+    generators_mean_origin = n.generators.p_nom.mean()
+    load_mean_origin = n.loads_t.p_set.mean().mean()
+
+    # duplicated sub-networks mean that there is at least one interconnection between buses
+    i_islands = n.buses[
+        (~n.buses.duplicated(subset=["sub_network"], keep=False))
+        & (n.buses.carrier == "AC")
+    ].index
+
+    i_load_islands = n.loads_t.p_set.columns.intersection(i_islands)
+
+    # isolated buses without load should be discarded
+    isl_no_load = i_islands.difference(i_load_islands)
+    n.mremove("Bus", isl_no_load)
+    n.determine_network_topology()
+
+    # isolated buses with load lower than a specified threshold should be discarded as well
+    i_small_load = i_load_islands[
+        n.loads_t.p_set[i_load_islands].mean(axis=0) <= threshold
+    ]
+
+    if (i_islands.empty) | (len(i_small_load) == 0):
+        return n
+
+    i_loads_drop = n.loads[n.loads.bus.isin(i_small_load)].index
+    i_generators_drop = n.generators[n.generators.bus.isin(i_small_load)].index
+
+    n.mremove("Bus", i_small_load)
+    n.mremove("Load", i_loads_drop)
+    n.mremove("Generator", i_generators_drop)
+
+    n.determine_network_topology()
+
+    load_mean_final = n.loads_t.p_set.mean().mean()
+    generators_mean_final = n.generators.p_nom.mean()
+
+    logger.info(
+        f"Dropped {len(i_small_load)} buses. A resulted load discrepancy is {(100 * ((load_mean_final - load_mean_origin)/load_mean_origin)):2.1}% and {(100 * ((generators_mean_final - generators_mean_origin)/generators_mean_origin)):2.1}% for average load and generation capacity, respectivelly"
+    )
+
+    return n
+
+
+def merge_isolated_nodes(n, threshold, aggregation_strategies=dict()):
+    """
+    Find isolated nodes in the network and merge those of them which have
+    load value below than a specified threshold into a single isolated node
+    which represents all the remote generation
+
+    Parameters
+    ----------
+    n : PyPSA.Network
+        Original network
+    threshold : float
+        Load power used as a threshold to merge isolated nodes
+
+    Returns
+    -------
+    modified network
+    """
+    # keep original values of the overall load and generation in the network
+    # to track changes due to drop of buses
+    generators_mean_origin = n.generators.p_nom.mean()
+    load_mean_origin = n.loads_t.p_set.mean().mean()
+
+    n.determine_network_topology()
+
+    # duplicated sub-networks mean that there is at least one interconnection between buses
+    i_islands = n.buses[
+        (~n.buses.duplicated(subset=["sub_network"], keep=False))
+        & (n.buses.carrier == "AC")
+    ].index
+
+    # isolated buses with load below than a specified threshold should be merged
+    i_load_islands = n.loads_t.p_set.columns.intersection(i_islands)
+    i_suffic_load = i_load_islands[
+        n.loads_t.p_set[i_load_islands].mean(axis=0) <= threshold
+    ]
+
+    if (i_islands.empty) | (len(i_suffic_load) == 0):
+        return n, n.buses.index.to_series()
+
+    # all the noded to be merged should be mapped into a single node
+    agg_buses_list = []
+    countries_list = n.buses.country.unique()
+    for c in countries_list:
+        buses_in_country = n.buses.loc[i_suffic_load].country == c
+        i_aggreg_bus = n.buses.loc[i_suffic_load][buses_in_country].index[0]
+        agg_buses_list.append(i_aggreg_bus)
+
+    country_to_buses_dict = dict(zip(countries_list, agg_buses_list))
+    n_buses_df = n.buses[["country"]].copy().assign(bus_id=n.buses.index)
+    n_buses_df["agg_bus"] = n_buses_df.loc[i_suffic_load, "country"].map(
+        country_to_buses_dict
+    )
+    n_buses_df["agg_bus"].fillna(n_buses_df.bus_id, inplace=True)
+    busmap = n_buses_df["agg_bus"]
+
+    bus_strategies, generator_strategies = get_aggregation_strategies(
+        aggregation_strategies
+    )
+
+    clustering = get_clustering_from_busmap(
+        n,
+        busmap,
+        bus_strategies=bus_strategies,
+        aggregate_generators_weighted=True,
+        aggregate_generators_carriers=None,
+        aggregate_one_ports=["Load", "StorageUnit"],
+        line_length_factor=1.0,
+        generator_strategies=generator_strategies,
+        scale_link_capital_costs=False,
+    )
+
+    load_mean_final = n.loads_t.p_set.mean().mean()
+    generators_mean_final = n.generators.p_nom.mean()
+
+    logger.info(
+        f"Merged {len(i_suffic_load)} buses. Load attached to a single bus with discrepancies of {(100 * ((load_mean_final - load_mean_origin)/load_mean_origin)):2.1E}% and {(100 * ((generators_mean_final - generators_mean_origin)/generators_mean_origin)):2.1E}% for load and generation capacity, respectivelly"
+    )
+
+    return clustering.network, busmap
+
+
 if __name__ == "__main__":
     if "snakemake" not in globals():
         from _helpers import mock_snakemake
@@ -577,7 +738,7 @@ if __name__ == "__main__":
     # treatment of outliers (nodes without a profile for considered carrier):
     # all nodes that have no profile of the given carrier are being aggregated to closest neighbor
     if (
-        snakemake.config.get("clustering", {})
+        snakemake.config.get("cluster_options", {})
         .get("cluster_network", {})
         .get("algorithm", "hac")
         == "hac"
@@ -587,8 +748,18 @@ if __name__ == "__main__":
             cluster_config.get("feature", "solar+onwind-time").split("-")[0].split("+")
         )
         for carrier in carriers:
+            # isolated buses should be excluded from adjacency_matrix calculations
+            n.determine_network_topology()
+            # duplicated sub-networks mean that there is at least one interconnection between buses
+            i_islands = n.buses[
+                (~n.buses.duplicated(subset=["sub_network"], keep=False))
+                & (n.buses.carrier == "AC")
+            ].index
+
             buses_i = list(
-                set(n.buses.index) - set(n.generators.query("carrier == @carrier").bus)
+                set(n.buses.index)
+                - set(i_islands)
+                - set(n.generators.query("carrier == @carrier").bus)
             )
             logger.info(
                 f"clustering preparaton (hac): aggregating {len(buses_i)} buses of type {carrier}."
@@ -619,6 +790,20 @@ if __name__ == "__main__":
     n.buses = n.buses.drop(buses_c, axis=1)
 
     update_p_nom_max(n)
+
+    p_threshold_drop_isolated = max(
+        0.0, cluster_config.get("p_threshold_drop_isolated", 0.0)
+    )
+    p_threshold_merge_isolated = cluster_config.get("p_threshold_merge_isolated", 0.0)
+
+    n = drop_isolated_nodes(n, threshold=p_threshold_drop_isolated)
+    if p_threshold_merge_isolated:
+        n, merged_nodes_map = merge_isolated_nodes(
+            n,
+            threshold=p_threshold_merge_isolated,
+            aggregation_strategies=aggregation_strategies,
+        )
+        busmaps.append(merged_nodes_map)
 
     n.meta = dict(snakemake.config, **dict(wildcards=dict(snakemake.wildcards)))
     n.export_to_netcdf(snakemake.output.network)
--- a/scripts/build_demand_profiles.py
+++ b/scripts/build_demand_profiles.py
@@ -182,7 +182,7 @@ def build_demand_profiles(
     demand_profiles = demand_profiles.loc[start_date:end_date]
     demand_profiles.to_csv(out_path, header=True)
 
-    logger.info(f"Demand_profiles csv file created for the corrisponding snapshots.")
+    logger.info(f"Demand_profiles csv file created for the corresponding snapshots.")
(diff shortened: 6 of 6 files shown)


ekatef:
This preliminary version of the fix resolves AC/DC issues for "CD" but leads to troubles in `cluster_networks` for the tutorial case. Interestingly, these troubles appear in a call of solver factory `opt.solve(m)` regardless of using `merge_isolated_buses(n)` in `simplify_network` or disregarding it. The trigger is switching-on grouping by sub-networks and country in `cluster_network`.

Topology of a simplified network seems to be alright:

elec_s nc_dc_by_carrier

Values or the distribution factors also don't look suspicious:

```bash
country
BJ    0.22056
NG    0.77944
```
The full error listing:

```bash
  File "~/pypsa-earth/.snakemake/scripts/tmp_yx5e94k.cluster_network.py", line 732, in <module>
    clustering = clustering_for_n_clusters(
  File "~/pypsa-earth/.snakemake/scripts/tmp_yx5e94k.cluster_network.py", line 578, in clustering_for_n_clusters
    busmap = busmap_for_n_clusters(
  File "~/pypsa-earth/.snakemake/scripts/tmp_yx5e94k.cluster_network.py", line 472, in busmap_for_n_clusters
    n_clusters = distribute_clusters(
  File "~/pypsa-earth/.snakemake/scripts/tmp_yx5e94k.cluster_network.py", line 369, in distribute_clusters
    results = opt.solve(m)
  File "~/miniconda3/envs/pypsa-earth-upd4/lib/python3.10/site-packages/pyomo/opt/base/solvers.py", line 570, in solve
    self._presolve(*args, **kwds)
  File "~/miniconda3/envs/pypsa-earth-upd4/lib/python3.10/site-packages/pyomo/opt/solver/shellcmd.py", line 209, in _presolve
    OptSolver._presolve(self, *args, **kwds)
  File "~/miniconda3/envs/pypsa-earth-upd4/lib/python3.10/site-packages/pyomo/opt/base/solvers.py", line 667, in _presolve
    self._convert_problem(args,
  File "~/miniconda3/envs/pypsa-earth-upd4/lib/python3.10/site-packages/pyomo/opt/base/solvers.py", line 718, in _convert_problem
    return convert_problem(args,
  File "~/miniconda3/envs/pypsa-earth-upd4/lib/python3.10/site-packages/pyomo/opt/base/convert.py", line 101, in convert_problem
    problem_files, symbol_map = converter.apply(*tmp, **tmpkw)
  File "~/miniconda3/envs/pypsa-earth-upd4/lib/python3.10/site-packages/pyomo/solvers/plugins/converter/model.py", line 181, in apply
    instance.write(
  File "~/miniconda3/envs/pypsa-earth-upd4/lib/python3.10/site-packages/pyomo/core/base/block.py", line 2015, in write
    (filename, smap) = problem_writer(self,
  File "pyomo/repn/plugins/ampl/ampl_.pyx", line 407, in pyomo.repn.plugins.ampl.ampl_.ProblemWriter_nl.__call__
  File "pyomo/repn/plugins/ampl/ampl_.pyx", line 408, in pyomo.repn.plugins.ampl.ampl_.ProblemWriter_nl.__call__
  File "pyomo/repn/plugins/ampl/ampl_.pyx", line 410, in pyomo.repn.plugins.ampl.ampl_.ProblemWriter_nl.__call__
  File "pyomo/repn/plugins/ampl/ampl_.pyx", line 1714, in pyomo.repn.plugins.ampl.ampl_.ProblemWriter_nl._print_model_NL
  File "~/miniconda3/envs/pypsa-earth-upd4/lib/python3.10/site-packages/pyomo/core/base/var.py", line 90, in has_ub
    return self.ub not in _no_upper_bound
  File "~/miniconda3/envs/pypsa-earth-upd4/lib/python3.10/site-packages/pyomo/core/base/var.py", line 455, in ub
    ub = ub()
TypeError: 'Series' object is not callable
```

scripts/simplify_network.py
```diff
@@ -522,6 +523,64 @@ def consense(x):
...
 
 
+def merge_isolated_buses(n, drop_p_threshold=0):
+    generators_sum_origin = n.generators.p_nom.sum()
+    load_sum_origin = n.loads_t.p_set.sum().sum()
+
+    # duplicated sub-networks mean that there is at least one interconnection between buses
+    i_islands = n.buses[
+        (~n.buses.duplicated(subset=["sub_network"], keep=False))
+        & (n.buses.carrier == "AC")
```
davide-f, line 568: Do we need this limitation to AC? I'd remove this limitation unless necessary for some reasons

scripts/simplify_network.py
```diff
@@ -522,6 +523,64 @@ def consense(x):
...
+    ].index
+
+    load_sum_isol_origin = n.loads_t.p_set[i_islands].sum().sum()
```
davide-f: unused variable

scripts/simplify_network.py
```diff
@@ -522,6 +523,64 @@ def consense(x):
...
+
+    # disregard buses having very small attached load when rearranging
+    loads_t_avr = n.loads_t.p_set.mean(axis=0)
```
davide-f: better to make the filter by yearly demand. we can calculate the mean and multiply by 8760 (better than sum as it is reliant to variable length of the dataset)

scripts/simplify_network.py
```diff
@@ -522,6 +523,64 @@ def consense(x):
...
+    i_islands_large = i_islands[loads_t_avr[i_islands] >= drop_p_threshold]
```
davide-f: does this work? the two elements may have different length

scripts/simplify_network.py
```diff
@@ -522,6 +523,64 @@ def consense(x):
     return clustering.network, clustering.busmap
 
 
+def merge_isolated_buses(n, drop_p_threshold=0):
```
davide-f: this function can be split into two:
1. a function that actually only drops buses below an energy threashold
2. a function that aggregates buses below another energy threshold

scripts/simplify_network.py
```diff
@@ -522,6 +523,64 @@ def consense(x):
...
+        & (n.buses.carrier == "AC")
+    ].index
+
+    load_sum_isol_origin = n.loads_t.p_set[i_islands].sum().sum()
+
+    # disregard buses having very small attached load when rearranging
+    loads_t_avr = n.loads_t.p_set.mean(axis=0)
+    i_islands_large = i_islands[loads_t_avr[i_islands] >= drop_p_threshold]
+
+    # agregate all the load to a single isolated bus
```
davide-f: Regarding the clustering approach of the isolated nodes we may exploit the clustering function get_clustering_from_busmap and test its use with a custom busmap that maps the isolated nodes to merge into a new unique bus.
The use of that function is already present into cluster_network.
That should also take care of the aggregation strategies in the same was as done for the normal clustering approach.
This may fix some problems the CI is facing

scripts/simplify_network.py
```diff
@@ -620,6 +679,8 @@ def consense(x):
 
     update_p_nom_max(n)
 
+    merge_isolated_buses(n)
```
davide-f: Recalling above, the proposal is to decompose the problem into two pieces: 1. dropping isolated nodes below a threashold, 2 merge isolated nodes below another threashold.
Note: when the value of a threshold is False, then that function is not abilited, it is true, that function is applied to all isolated nodes.

davide-f (changes requested):
Hello Katia :)

Nice draft!
I've added some comments, let me know your feeling.

Regarding the CI problem, it may be due to the clustering proposal you are doing.

It may be better to decompose the problem into two subproblems, each of them having its own function:
1. drop isolated nodes below a given energy threhold.
2. aggregate nodes below a given energy threshold

I think it may be better to start with problem 1 and implement the function that does that.
Once that is ready, 2 may be done :) Also 2 may be done using some clustering features by pypsa. Comments are added.
That may also improve the problems the CI is facing.

What do you think?

ekatef:
Hello Davide!

Thank you so much for checking the PR. I'm a bit puzzled by a way CI fails and your comments are very helpful to see a possible workaround for it :) Agree that it absolutely makes sense to decompose the problem

scripts/simplify_network.py
ekatef: That is a quick (and not the cleanest) fix for DC-links. If we are using unique sub-networks as the only criterium, such DC-link-buses are treated as islanded

scripts/simplify_network.py
ekatef: Fixed. Thank you!

scripts/simplify_network.py
ekatef: Agree. Fixed

scripts/simplify_network.py
ekatef: It works as `i_islands_large` is an index created right here. So, we can subset `i_islands ` with a boolean series and write into `i_islands_large` despite both `i_*` usually have different lengths. Although, agree that in case of dataframes such assignment could lead to troubles

scripts/simplify_network.py
ekatef: Absolutely agree that it would be much more consistent to keep the same approach with busmaps as in clustering. It was first I have tried but the results weren't very good :)

In particular, `import_components_from_dataframe` which is normally used in clustering to add the aggregated buses to the network, seems to add interconnections between these buses and the network. Probably, we just need a modified version of it

scripts/simplify_network.py
ekatef: Agree, it sounds absolutely reasonable. A question: can we assume that if all isolated buses are dropped, the whole workflow should work with the grouping by country and sub-network? I'd suppose yes as it's the limit case of the final approach

scripts/simplify_network.py
davide-f: ok! let's keep them for now

scripts/simplify_network.py
davide-f: with the i_islands check it should work, the [i_islands] does the trick :)

scripts/simplify_network.py
davide-f: mmm let's address the drop feature first and the grouping afterwards, also because the grouping should be done within each country. isolated buses belonging to different countries shouldn't be merged.

Basically, this aggregation of buses should replicate the cluster_network procedure that relies on the get_clustering_from_busmap, as done here: https://github.com/pypsa-meets-earth/pypsa-earth/blob/5da36819ba1873ebbf3d1f40558137fafd2b24ed/scripts/cluster_network.py#L574

The clustering object that it returns contains also the clustered network in clustering.network that would be the new "n"

scripts/simplify_network.py
davide-f: Obviously the determine_topology function should be repeated after each cleaning procedure to update the values of sub_network.
If we drop all isolated buses, the cluster_network procedure would still work with ["countries", "sub_network"] but it will be equivalent to "countries" as long as all country is completely interconnected.

Even when all isolated buses are dropped the clustering shall be done by "countries" and "sub_network"

scripts/simplify_network.py
davide-f: ok

> Hello Davide!
> 
> Thank you so much for checking the PR. I'm a bit puzzled by a way CI fails and your comments are very helpful to see a possible workaround for it :) Agree that it absolutely makes sense to decompose the problem

I completely understand.
My feeling is that this issues are the consequence of manually removing data from the dataframes to handle the merge of isolated buses. the clustering procedure does that automatically.

The first option, implying dropping the buses should involve less code and I'd recommend to implement a specific function for that and (hopefully) these issues may disappear

scripts/simplify_network.py
davide-f: Here I'd expect something like:

``` python

e_threshold_drop_isolated = config["simplify_network"].get("e_threshold_drop_isolated ", False)
if e_threshold_drop_isolated :
   drop_isolated_nodes(n, e_threshold_isolated_buses)

e_threshold_mesge_isolated = config["simplify_network"].get("e_threshold_mesge_isolated ", False)
if e_threshold_mesge_isolated:
   merge_isolated_buses(n, e_threshold_mesge_isolated)
```

naming of options may be improved and the config management adjusted; that's a concept.
Ideally we can start from the first function and see if the issue was due to the second block, which I think is better handled with the busmap clustering

ekatef: Thank you very much for the explanations!
Was not aware on `determine_topology` usage which seems to be very relevant

ekatef: A clarification question: do you think it makes sense to have two different thresholds? Would it be alright to leave isolated buses which have load between the thresholds? Just trying to understand the concept you are suggesting :)

scripts/simplify_network.py
ekatef: > ok
> 
> > Hello Davide!
> > Thank you so much for checking the PR. I'm a bit puzzled by a way CI fails and your comments are very helpful to see a possible workaround for it :) Agree that it absolutely makes sense to decompose the problem
> 
> I completely understand. My feeling is that this issues are the consequence of manually removing data from the dataframes to handle the merge of isolated buses. the clustering procedure does that automatically.
> 
> The first option, implying dropping the buses should involve less code and I'd recommend to implement a specific function for that and (hopefully) these issues may disappear

Have tried to implement the drop option and all the issues were even resolved locally for the tutorial case, but re-appeared after some additional changes. Which confirms that the decomposing approach is a very good idea ;) 

It'll be definitely simpler to debug the dropping part first as compared to the whole approach. Thanks a lot for you support!

scripts/simplify_network.py
```diff
@@ -522,6 +523,91 @@ def consense(x):
...
+    i_islands = n.buses[
+        (~n.buses.duplicated(subset=["sub_network"], keep=False))
+        & (n.buses.carrier == "AC")
+    ].index
+    i_islands_drop = i_islands[loads_t_ann[i_islands] <= drop_e_threshold]
+
+    i_generators_drop = n.generators[n.generators.bus.isin(i_islands_drop)].index
+
+    n.mremove("Bus", i_islands_drop)
+    n.mremove("Load", i_islands_drop)
```
davide-f: For loads please verify the name of the loads to be removed, the index may be different than the index of buses.
You can quickly check by looking at the demand profile csv

scripts/simplify_network.py
```diff
@@ -522,6 +523,91 @@ def consense(x):
...
+    n.mremove("Generator", i_generators_drop)
+
+    load_sum_final = n.loads_t.p_set.sum().sum()
+    generators_sum_final = n.generators.p_nom.sum()
+
+    logger.info(
```
davide-f, line 598: This log may be enabled only if buses are dropped

scripts/simplify_network.py
```diff
@@ -620,6 +706,11 @@ def consense(x):
 
     update_p_nom_max(n)
 
+    e_threshold_drop_isolated = cluster_config.get("e_threshold_drop_isolated ", False)
+    if e_threshold_drop_isolated:
+        drop_isolated_nodes(n, drop_e_threshold)
+        n.determine_network_topology()
```
davide-f: The update of the determine topology may be moved within ghr drop isolated function

scripts/simplify_network.py
```diff
@@ -522,6 +523,91 @@ def consense(x):
...
+    generators_sum_final = n.generators.p_nom.sum()
+
+    logger.info(
+        f"Dropped {len(i_islands)} buses. A resulted load discrepancy is {(100 * ((load_sum_final - load_sum_origin)/load_sum_origin)):2.1E}% and {(100 * ((generators_sum_final - generators_sum_origin)/generators_sum_origin)):2.1E}% for average load and generation capacity, respectivelly"
+    )
+
+    return n
+
+
+def merge_isolated_buses(n, drop_p_threshold=0, n_periods_in_year=8760):
```
davide-f: The core of this function may be modified similarly to aggregate_to_substations

That function is quite close to what we would like to achieve with this merge isolated buses

davide-f (changes requested):

scripts/simplify_network.py
davide-f: You got it :) : 2 thresholds, one to drop and the other one to merge.
I'd expect the merge to be larger than dropping but no check is needed for this: if the converse applies, no buses are merged as all the relevant isolated are dropped.

scripts/simplify_network.py
ekatef: Fixed. Thank you :)

scripts/simplify_network.py
ekatef: I'd even say, if there are no buses to drop we can just return the initial network. Do you agree?

scripts/simplify_network.py
ekatef: Done

ekatef: What I don't understand is why filtering isolated buses `n.buses[(~n.buses.duplicated(subset=["sub_network"], keep=False))]` gives an empty output unless `n.determine_network_topology()` is called before that

scripts/simplify_network.py
ekatef: Thank you!
A question: wouldn't it be more convenient to specify the thresholds as a share of the annual load? An absolute value might be not always obvious in advance...

ekatef:
The good news is that dropping isolated buses + using grouping by  country and sub-network works for CDR meaning that AC/DC issue for a single country is resolved. However, there are still some troubles if the domain consists of more that one country

scripts/simplify_network.py
davide-f: Yeah, that may be an implicit output of the procedure unless some pandas stuff goes wrong

scripts/simplify_network.py
davide-f: That function fills the sub_network information, that's why it is needed to have the value well calculated

scripts/simplify_network.py
davide-f: I'd say let's keep it simple for now.
Otherwise mathematically it becomes more complex. It is ok with a numeric threshold

scripts/simplify_network.py
```diff
@@ -620,6 +714,10 @@ def consense(x):
 
     update_p_nom_max(n)
 
+    e_threshold_drop_isolated = cluster_config.get("e_threshold_drop_isolated", False)
+    if e_threshold_drop_isolated:
+        drop_isolated_nodes(n, drop_e_threshold=e_threshold_drop_isolated)
```
davide-f: Ah little refuse: here we need n = drop_isolated_nodes(...)

scripts/simplify_network.py
ekatef: Ok!

scripts/simplify_network.py
ekatef: Let's try :)
Although, I have an impression that is has worked in the previous version, too
At least, the isolated buses disappeared from the network picture

davide-f: Oh yeah, there may not be the need for n = but later in the merge nodes there will be most likely.
Let's leep it :) it doesn't harm

config.tutorial.yaml
```diff
@@ -74,6 +74,8 @@ cluster_options:
     exclude_carriers: []
     remove_stubs: true
     remove_stubs_across_borders: true
+    e_threshold_drop_isolated: 10000000
+    e_threshold_merge_isolated: 100000000
```
davide-f: These numbers are a bit ackward
1MW unit would produce 8.76 GWh yearly.
I think that we could safely disregard anything below 5-10 MW -> about 50-100 GWh
The unit shall be added as a comment, and as default values we could use 100GWh?
What do the tests suggest in your cases?

davide-f:
The tutorial works :D

> The good news is that dropping isolated buses + using grouping by country and sub-network works for CDR meaning that AC/DC issue for a single country is resolved. However, there are still some troubles if the domain consists of more that one country

The CI also works :D I'd like a revision on the option values to a reasonable number but it's quite promising!
Next step would be the merge; please, check out the function `aggregate_to_substations`.
The merge_isolated_buses should do something highly similar to that function; the implementation of that feature may be easier with that :)

scripts/simplify_network.py
ekatef: I have an impression that it was a change that made CI work...

config.tutorial.yaml
ekatef: Agree, that is a good alternative to a dimensionless threshold :) Changed.

In the case CDR+Zambia the annual loads are of order 1e5..1e7 (which means I suppose MWh, right?) 

For testing purposes I have set 10 and 100 TWh so far. It should be revised further as it feels like the the considered testing case is the one when merging would be a more proper approach as compared with drop

ekatef:
> The tutorial works :D
> 
> > The good news is that dropping isolated buses + using grouping by country and sub-network works for CDR meaning that AC/DC issue for a single country is resolved. However, there are still some troubles if the domain consists of more that one country
> 
> The CI also works :D I'd like a revision on the option values to a reasonable number but it's quite promising! Next step would be the merge; please, check out the function `aggregate_to_substations`. The merge_isolated_buses should do something highly similar to that function; the implementation of that feature may be easier with that :)

Agree, that feels like very good news. But a testing case (CD + ZW) on my local machine still fails :( Will try to reproduce the tutorial locally.

Update: locally tutorial works as well 😊 🎉 🚀 
Thanks a lot for the hint regarding `aggregate_to_substations`! ;)

ekatef:
Testing with current threshold values leads to AssertionError due to having different countries in the same cluster (`AssertionError: In Bus cluster country the values of attribute country do not agree`) when running `get_clustering_from_busmap( )` at the end of `merge_isolated_nodes( )`. 

This discrepancy has to be prevented with 

https://github.com/ekatef/pypsa-earth/blob/96dc0cb0eb5c781948eb16234649026051384e7e/scripts/simplify_network.py#L601-L603

But it doesn't work as something gets wrong with distance matrix `dist`: all the elements are either infinite (np.inf is used to indicate pairs of the buses which should not be merged) or zero (when i == j). It looks like calculation of adjacency matrix works properly, while distance calculations `dijkstra` for the graph gives these infs. Indices of the buses should be checked

ekatef:
According to definition of `PyPSA::adjacency_matrix( )`, it calculates wieghts for each pair of buses connected by a branch. Which means that isolated buses are excluded from these calculations. [_It looks like this leads to infinite distances when calling `dijkstra` as it tries to calculate the distances between the nodes which don't belong to the processes graph_]

My feeling is that we need an analogue of `adjacency_matrix( )`&`dist` block which would:
1) utilise Euclidean distance instead of a branch weight;
2) distinguish between the interconnected nodes and isolated ones;
3) build a matrix of pairwise distances between the isolated nodes and the interconnected ones (something likes in [this SO answer ](https://stackoverflow.com/a/55410594/8465924) might be used).

The result should be `dist` matrix which can be processed further as in `aggregate_to_substations`. I'm concerned about is performance of points 1&2 as spatial calculations may be quite heavy. Probably, some graph tricks could be used

@davide-f would be very grateful for your suggestions

scripts/simplify_network.py
```diff
@@ -522,6 +523,115 @@ def consense(x):
...
+        (~n.buses.duplicated(subset=["sub_network"], keep=False))
+        & (n.buses.carrier == "AC")
+    ].index
+
+    # merge buses having sufficient attached load
+    loads_t_ann = n_periods_in_year * n.loads_t.p_set.mean(axis=0)
+    i_islands_large = i_islands[loads_t_ann[i_islands] >= merge_p_threshold]
+    buses_i = list(set(i_islands_large))
+
+    weight = pd.concat(
```
davide-f: Nice draft Katia:)

I saw that you kept the previous implementation using the adjacency matrix.
In practice, we don't need that.
In our case, we want to replace the calculation os busmap using our own way based on the isolated nodes.

Let's say we have 2 countries with 5 nodes total of which a1, a2 and b1 are isolated and to be merged
A1
A2
A3 (not isolated)
B1
B2 (not isolated)

The key is to create a busmap as follows:
a1->"newa1"
a2->"newa1"
a3->"newa2"
b1->"newb1"
b2->"newb2'

Please note that we cannot mix buses belonging to different countries.
The above does not require adjacency matrix calculation.

As a comment, in this testing, you have now the same setting for both the drop and merge options.
If the drop option is consolidated you can deactivate it while testing the merge.
Once also the latter works that the final version may be enabled

ekatef: > Nice draft Katia:)
> 
> I saw that you kept the previous implementation using the adjacency matrix. In practice, we don't need that. In our case, we want to replace the calculation os busmap using our own way based on the isolated nodes.
> 
> Let's say we have 2 countries with 5 nodes total of which a1, a2 and b1 are isolated and to be merged A1 A2 A3 (not isolated) B1 B2 (not isolated)
> 
> The key is to create a busmap as follows: a1->"newa1" a2->"newa1" a3->"newa2" b1->"newb1" b2->"newb2'
> 
> Please note that we cannot mix buses belonging to different countries. The above does not require adjacency matrix calculation.
> 
> As a comment, in this testing, you have now the same setting for both the drop and merge options. If the drop option is consolidated you can deactivate it while testing the merge. Once also the latter works that the final version may be enabled

Thanks a lot for explanations Davide!
Now it's much clearer with busmaps :)

As for replacement for adjacency matrix, my first thought was to build a matrix of the distances for each point and take the minimal values. But probably using KD-trees from scipy would be a more elegant approach...

Absolutely regarding the thresholds. Thanks for the hint!

davide-f: So, the use for the adjacency matrix in that function is justified by the fact that that function aimed to cluster substations to their closest substation, thus it needed the adjacency matrix.

In our case, we really don't need that.
By knowing for each country the buses that are isolated, you know that those buses will have to be mapped into the same bus.
I don't see where you would need to use the KD trees

ekatef: Ouch... Sorry, have mixed-up aggregation strategies :(
And thanks for pointing this out!

ekatef:
A version with merging isolated nodes works on a single country with AD/DC nodes (tested on CDR) but fails on a mutli-country (tutorial case) with the same pyomo-related error as before: `in has_ub return self.ub not in _no_upper_bound... in ub ub = ub() TypeError: 'Series' object is not callable`

config.tutorial.yaml
davide-f: Those thresholds are quite high.
Ok for testing but I'd expect lower values, values need checking with the full model later on

pz-max: Just FYI, you can use smt like this in Python 100_000_000

ekatef: @davide-f thanks! Let's go for threshold as a share of the overall load? We anyway calculate it for checks

@pz-max great hint :) thanks a lot!

davide-f: > @davide-f thanks! Let's go for threshold as a share of the overall load? We anyway calculate it for checks
> 
> @pz-max great hint :) thanks a lot!

I think having it as a share of the total load may makes things more complicated to understand what happens, I'd keep it as is.
Also because the threshold must be country-based. it cannot be related to the overall demand of all the region of interest.
Moreover, it would make things difficult to compare from big and small countries.
For big countries like china, a 0.1% of the total demand may make disappear even large cities, while for small countries may keep also small buses with few MW of installed capacity.
That is a complication that I would keep for later if needed.

davide-f:
> A version with merging isolated nodes works on a single country with AD/DC nodes (tested on CDR) but fails on a mutli-country (tutorial case) with the same pyomo-related error as before: `in has_ub return self.ub not in _no_upper_bound... in ub ub = ub() TypeError: 'Series' object is not callable`

Unfortunately, I don't know, needs debugging.
Most likely is something wrong in the aggregation and options provided to the clustering procedure, given that the buses we are trying to group together are isolated.
May be a good idea to understand better the options and the aggregation strategies we use.
Not sure if @fneum may have some suggestions here.

scripts/simplify_network.py
```diff
@@ -620,6 +724,18 @@ def consense(x):
...
     update_p_nom_max(n)
 
+    e_threshold_drop_isolated = cluster_config.get("e_threshold_drop_isolated", False)
+    e_threshold_merge_isolated = cluster_config.get("e_threshold_merge_isolated", False)
+    if e_threshold_drop_isolated:
+        n = drop_isolated_nodes(n, drop_e_threshold=e_threshold_drop_isolated)
+    if e_threshold_merge_isolated:
+        n, merged_nodes_map = merge_isolated_nodes(
+            n,
+            merge_p_threshold=e_threshold_merge_isolated,
```
davide-f: revise "merge_p_threshold" to "merge_e_threshold"

config.tutorial.yaml
ekatef: > I think having it as a share of the total load may makes things more complicated to understand what happens, I'd keep it as is. Also because the threshold must be country-based. it cannot be related to the overall demand of all the region of interest. Moreover, it would make things difficult to compare from big and small countries. For big countries like china, a 0.1% of the total demand may make disappear even large cities, while for small countries may keep also small buses with few MW of installed capacity. That is a complication that I would keep for later if needed.

Ah, I see. Agree that the absolute value also absolutely make sense. I'll revert it

ekatef:
> > A version with merging isolated nodes works on a single country with AD/DC nodes (tested on CDR) but fails on a mutli-country (tutorial case) with the same pyomo-related error as before: `in has_ub return self.ub not in _no_upper_bound... in ub ub = ub() TypeError: 'Series' object is not callable`
> 
> Unfortunately, I don't know, needs debugging. Most likely is something wrong in the aggregation and options provided to the clustering procedure, given that the buses we are trying to group together are isolated. May be a good idea to understand better the options and the aggregation strategies we use. Not sure if @fneum may have some suggestions here.

When looking into the clustering script, it looks like current troubles are caused by the line (when grouping by sub_networks is added):

https://github.com/pypsa-meets-earth/pypsa-earth/blob/1ca88eff8bf66579d486808487daeb71837e78da/scripts/cluster_network.py#L289

When clustering by sub-network and country, the results is like this:

```bash
country  sub_network
BJ       1               1
NG      0              83
           2               1
```
It seems there you are absolutely right, there is definitely something went wrong when assigning sub-networks... Thanks for the idea :)

Update: don't see any obvious errors in the network structure. This above is a perfectly normal formatting of a grouping object. But something is not working when calculating `n_bounds` for pyomo

ekatef:
@davide-f, do you probably know what is the meaning of bounds in context of this pyomo definition:

https://github.com/pypsa-meets-earth/pypsa-earth/blob/be8e715c337d57307459fb7a2771f85761a20a3f/scripts/cluster_network.py#L335

where
* `m` is an empty pyomo model
* `country_list` is a list of modelled countries

It looks like optimisation is used here to distribute cluster nodes between countries. Which means that `n_bounds` should be constrained with something simple like the total number or clusters. But currently definition of `n_bounds( )` leads to some complications

davide-f:
So bounds should be a tuple or a function that defines the boundaries of the variables; for more information, the [pyomo website](https://pyomo.readthedocs.io/en/stable/pyomo_modeling_components/Variables.html) may be handy.

Not sure what is the value of it now.

The groupby countries-only or countries-subnetwork shown above may makes sense if the last line still corresponds to NG.
We would have the groups (BJ, 1), (NG, 0) and (NG, 2).
This suggests that the BJ network is disconnected from NG and in NG there is a disconnected bus. By the content of the groupby (NG, 2) is the disconnected node as it counts only 1 bus.

ekatef:
> So bounds should be a tuple or a function that defines the boundaries of the variables; for more information, the [pyomo website](https://pyomo.readthedocs.io/en/stable/pyomo_modeling_components/Variables.html) may be handy.
> 
> Not sure what is the value of it now.
> 
> The groupby countries-only or countries-subnetwork shown above may makes sense if the last line still corresponds to NG. We would have the groups (BJ, 1), (NG, 0) and (NG, 2). This suggests that the BJ network is disconnected from NG and in NG there is a disconnected bus. By the content of the groupby (NG, 2) is the disconnected node as it counts only 1 bus.

@davide-f thanks a lot for clarification!

It seems that there is a slight types mismatch when calculating the bounds in case there are some countries and some have more than one sub-network. Then calling this line

https://github.com/pypsa-meets-earth/pypsa-earth/blob/be8e715c337d57307459fb7a2771f85761a20a3f/scripts/cluster_network.py#L333

results in a something series-like instead of a single value expected by pyomo. It could be fixed by using dimension reduction. My feeling is that summation would be reasonable here as it looks like we are optimizing distribution of nodes across clusters/countries and are constrained with a number of non-clustered buses available for each country. Would be grateful for double-check of this logic

Locally this approach works. Let's look what will be results of CI

ekatef:
> So bounds should be a tuple or a function that defines the boundaries of the variables; for more information, the [pyomo website](https://pyomo.readthedocs.io/en/stable/pyomo_modeling_components/Variables.html) may be handy.
> 
> Not sure what is the value of it now.
> 
> The groupby countries-only or countries-subnetwork shown above may makes sense if the last line still corresponds to NG. We would have the groups (BJ, 1), (NG, 0) and (NG, 2). This suggests that the BJ network is disconnected from NG and in NG there is a disconnected bus. By the content of the groupby (NG, 2) is the disconnected node as it counts only 1 bus.

Yes, there is a network for NG and one isolated bus for each of the countries (better visible in a clustered version):

elec_s_6 nc_dc_by_carrier

config.tutorial.yaml
ekatef: Revised to keep config parameters in a reasonable range. Would it be probably more convenient to use mean power [MW] instead of load [GWh]?

(I'll clean-up the history once we'll complete work! ;) )

davide-f: It's the same actually. My feeling is that it may messo sense an energy unit but if you feel average power is more convienent it is fine as well.
We just need clarify what we mean and obviously revise the variable names accordingly.
If we go for average power then let's keep MW as unit

ekatef:
@davide-f, thanks a lot for help! Have done a bit of testing and so far it looks like our solution is working properly. It has been tested on.

Blue = AC lines, magenta = DC lines, red = links

1) CDR

image

2) CD + ZM

image

3) ZM

image

Not sure why the DC lines get transformed into a link for CD + ZM during clustering

ekatef:
Double-check for CD+ZM:

image

It looks like the link is just a converter being zoomed as a result of nodes clustering :)

ekatef:
There is a plot for `base.nc` for CD+ZM:

image

It looks like there is no isolated nodes for Zambia... Testing of another region :)

ekatef:
Simplified network for Central Asia (KZ + TM + UZ):

image

But for clustering there is an error `KeyError: ('UZ', '3')` in `n_cluster_c = n_clusters[x.name]`. `3` for some reasons isn't in a sub-network index:

```bash
n_clusters
country  sub_network
KZ       0              24
         1               5
         6               1
TM       0               6
         4               2
         5               1
UZ       0              20
         2               1
dtype: int64
```

ekatef:
The problem was caused by an isolated which didn't have neither load, not generation attached. Adding an additional function to remove such "empty" buses has allowed to resolve this with the following result:

image

image

Note: the islands resulted from clustering are highlighted with red (in fact there are local networks reduced to buses), while "trully" isolated buses are in green circles

ekatef:
Functionality testing on IN, RU was successful. The AD/DC problem seems to be resolved. For CN there is still a problem due misinterpretation of links #586 but if DC are forced to DC, clustering is working

An example for India:

image

image

ekatef:
For AZ with 5 clusters the workflow works, as well with the clustered networks pretty similar to what we had before in discussion to #531 (https://user-images.githubusercontent.com/30229437/218205386-aa3849b9-aa69-4b1f-86f8-337843ee990d.jpg) with the only difference that we have only one isolated bus

image

ekatef:
@davide-f basic functionality seems to be working. Thanks a lot for guidance and helping with implementation! :)))

There are still some details which might need improvements. Ready to discuss them when you'd have time for it

davide-f:
FANTASTIC @ekatef !
I'll revise it soon :)
I think the china deserves an issue per-se; since you have the images and knowledge, may you add the issue?

scripts/simplify_network.py
```diff
@@ -522,6 +523,151 @@ def consense(x):
     return clustering.network, clustering.busmap
 
 
+def remove_empty_isolated_nodes(n):
```
davide-f: I recall the comment, could we do without this function or exploit the drop_isolated_nodes function instead?
My feeling is that the issue rises because some .sum() turns out to be empty or alike.
But if we set: drop_isolated_nodes(n, max(1kWh, {parameter})) we could save some code, provided that the drop_isolated_nodes drop those nodes

scripts/simplify_network.py
```diff
@@ -522,6 +523,151 @@ def consense(x):
...
+
+    # duplicated sub-networks mean that there is at least one interconnection between buses
+    i_islands = n.buses[
+        (~n.buses.duplicated(subset=["sub_network"], keep=False))
+        & (n.buses.carrier == "AC")
+    ].index
+
+    load_islands = loads_t_ann.loc[loads_t_ann.index.isin(i_islands)]
+    # drop_e_threshold, GWh
+    small_load_index = load_islands[load_islands <= (1e3 * drop_e_threshold)].index
```
davide-f: I'd remove the 1e3 here, and move it below when this function is called.

scripts/simplify_network.py
```diff
@@ -522,6 +523,151 @@ def consense(x):
...
+        (~n.buses.duplicated(subset=["sub_network"], keep=False))
+        & (n.buses.carrier == "AC")
+    ].index
+
+    # merge buses having sufficient attached load
+    loads_t_ann = n_periods_in_year * n.loads_t.p_set.mean(axis=0)
+    load_islands = loads_t_ann.loc[loads_t_ann.index.isin(i_islands)]
+
+    # merge_e_threshold, GWh
+    suffic_load_index = load_islands[load_islands >= (1e3 * merge_e_threshold)].index
```
davide-f: similarly to above, please move the 1e3 outside the function so to keep it without hard-coded values

scripts/simplify_network.py
```diff
@@ -522,6 +523,151 @@ def consense(x):
...
+    generators_sum_final = n.generators.p_nom.sum()
+
+    logger.info(
+        f"Dropped {len(isl_empty)} buses. A resulted load discrepancy is {(100 * ((load_sum_final - load_sum_origin)/load_sum_origin)):2.1}% and {(100 * ((generators_sum_final - generators_sum_origin)/generators_sum_origin)):2.1}% for average load and generation capacity, respectivelly"
+    )
+
+    return n
+
+
+def drop_isolated_nodes(n, drop_e_threshold, n_periods_in_year=8760):
```
davide-f: Please, add docstrings to clarify what all new functions do.

scripts/simplify_network.py
```diff
@@ -522,6 +523,151 @@ def consense(x):
...
+    generators_sum_final = n.generators.p_nom.sum()
+
+    logger.info(
+        f"Dropped {len(i_islands)} buses. A resulted load discrepancy is {(100 * ((load_sum_final - load_sum_origin)/load_sum_origin)):2.1}% and {(100 * ((generators_sum_final - generators_sum_origin)/generators_sum_origin)):2.1}% for average load and generation capacity, respectivelly"
+    )
+
+    return n
+
+
+def merge_isolated_nodes(
```
davide-f: no need for default value of merge_e

scripts/simplify_network.py
```diff
@@ -522,6 +523,151 @@ def consense(x):
...
+    load_islands = loads_t_ann.loc[loads_t_ann.index.isin(i_islands)]
+
+    # merge_e_threshold, GWh
+    suffic_load_index = load_islands[load_islands >= (1e3 * merge_e_threshold)].index
+    i_islands_large = i_islands[i_islands.isin(suffic_load_index)]
+
+    if (i_islands.empty) | (len(i_islands_large) == 0):
+        return n, n.buses.index.to_series()
+
+    agg_buses_list = []
```
davide-f, line 645: Some comments describing what happens may be advisable in general

scripts/simplify_network.py
```diff
@@ -620,6 +766,20 @@ def consense(x):
 
     update_p_nom_max(n)
 
+    n = remove_empty_isolated_nodes(n)
+
+    e_threshold_drop_isolated = cluster_config.get("e_threshold_drop_isolated", False)
+    e_threshold_merge_isolated = cluster_config.get("e_threshold_merge_isolated", False)
+    if e_threshold_drop_isolated:
+        n = drop_isolated_nodes(n, drop_e_threshold=e_threshold_drop_isolated)
```
davide-f: the 1e3 may be placed here and in the call below

davide-f (commented):
Some comments, very nice!
I need to leave, I may add minor comments later on, but very nice job :D

scripts/simplify_network.py
ekatef: Let's try MW? Then we'd not need not 10e3 neither 8760

scripts/simplify_network.py
ekatef: Removed

scripts/simplify_network.py
ekatef: Agree, removed

scripts/simplify_network.py
ekatef: Revised and removed

scripts/simplify_network.py
ekatef: I very much like the idea with replacing energy threshold with power one. Have introduced some changes

I also have an impression that a subnetwork which consists of empty buses leads to some unexpected zeros/None's somewhere. Haven't investigated it in depth but it seems reasonable to drop these nodes. Absolutely agree that it makes sense to place all the drops in the dropping function. Thinking how to do that (more or less) graciously :)

ekatef:
> FANTASTIC @ekatef ! I'll revise it soon :) I think the china deserves an issue per-se; since you have the images and knowledge, may you add the issue?

Thank you!! :)

Have started to address the comments and will come back tomorrow.
Indeed, will be happy to add an issue on China :)

scripts/simplify_network.py
davide-f: Ok let's go for MW but we need to clarify average hourly demand.

config.default.yaml
```diff
@@ -62,6 +62,8 @@ cluster_options:
     exclude_carriers: []
     remove_stubs: true
     remove_stubs_across_borders: true
+    e_threshold_drop_isolated: 10 # [GWh]
+    e_threshold_merge_isolated: 10 # [GWh]
```
davide-f: Improve the comment to specify  that is a demand threshold below which isolated buses are dropped and below are merged.
1 MW (10GWh) for drop is ok and I'd merge all below 10MW.

scripts/cluster_network.py
```diff
@@ -249,13 +249,14 @@ def distribute_clusters(
...
             .groupby(n.loads.bus)
             .sum()
-            .groupby([n.buses.country])
+            .groupby([n.buses.country, n.buses.sub_network])
             .sum()
             .pipe(normed)
         )
-        assert len(L.index) == len(n.buses.country.unique()), (
+        countries_in_L = pd.unique(L.index.get_level_values(0))
+        assert len(countries_in_L) == len(n.buses.country.unique()), (
```
davide-f, line 257: Have you tested the different clustering? by gdp/pop/load...?

scripts/simplify_network.py
davide-f: In general this block of code may be improved:
- load_sum_origin  does not account for the time length (if we move towards mean power that improves, only a mean is needed)
- In general, there is no need to compute loads_t_ann for all buses and then get the subset. you may do loads_t_ann[i_islands].mean() or alike.
- You make use of a lot of [].index. I think it can be cleaned. Interestingly, you may get the indices of those buses where demand is higher than the threshold and discard the difference.
  `load_islands = loads_t_ann.loc[loads_t_ann.index.intersection(i_islands)]`
  `small_load_index= i_islands.difference(load_islands[load_islands > threashold].index)`
  This approach should automatically discard those buses that do not have any load as load_islands won't contain in the index that bus
  
No need to check on generation: if an isolated bus does have generation but no load it is useless: no generation will be produced

scripts/simplify_network.py
davide-f: Above I added a suggestion to merge remove_empty isolated nodes and this function.
In that case, drop_isolated_nodes will always be executed by the drop_threshold would be max(0.0, config_threshold).
It is worth noticing that any scalar value, even negative would make empty nodes be dropped, see description above

scripts/simplify_network.py
```diff
@@ -620,6 +766,20 @@ def consense(x):
...
+    if e_threshold_merge_isolated:
+        n, merged_nodes_map = merge_isolated_nodes(
+            n,
+            merge_e_threshold=e_threshold_merge_isolated,
+            aggregation_strategies=aggregation_strategies,
+        )
+        busmaps.append(merged_nodes_map)
```
davide-f, line 806: do we store busmaps somewhere?
Should we drop this line?

davide-f:
Another round :) very nice!

scripts/cluster_network.py
ekatef: Have done a set of quick tests on AZ. It works for kmeans + clustering by load/pop/gdp. But fails for hac complaining that `Cannot extract more clusters than samples: 5 clusters where given for a tree with 3 leaves`. Apart of that, there are some troubles for alternative clustering when calculating hydro profiles. 

Probably, it makes sense for that PR to focus on making work k-means clustering keeping track of the issues appeared for `hac` and alternative clustering. What do you think?

It would like to test also more complicated geometries. My feeling is that it would be great to use scenario management you have developed but I was not able to get it work... Would be very grateful if you could briefly explain that. Ready to add the needed steps to the documentation ;)

scripts/simplify_network.py
ekatef: Revised

scripts/simplify_network.py
ekatef: Revised

config.default.yaml
ekatef: Added a clarification and revised the values

scripts/simplify_network.py
ekatef: Thanks a lot! Revised

scripts/simplify_network.py
ekatef: Done. Thanks :)

scripts/simplify_network.py
ekatef: Done :)

scripts/simplify_network.py
ekatef: As far as I understand, all the busmaps are being collected into the `busmap` list to be reduced and written into a file

scripts/cluster_network.py
davide-f: I agree that alternative clustering may be secondary if the CI passes.
However before damaging hac, I think it would be good to investigate little to make it compatible.
My feeling is that little should be needed since Eur uses the subnetwork.
What do you think?

scripts/simplify_network.py
davide-f: But maybe before?
I don't see any saving afterwards

scripts/cluster_network.py
davide-f: Mmm I'm ok to keep alternative clustering as a secondary priority, but I feel that little time spent on hac and gadm clustering may be worth.
The gadm should be used by @Hazem-IEG right?
And hac may require little changes as the eur version employs the clustering by sub network.
Maybe we miss little to fix it.
What do you think?

ekatef: Will be happy to play with hac clustering :)

Not sure I get the point with gadm/alternative clustering: isn't it the same? (just to clarify)

scripts/simplify_network.py
ekatef: Done

scripts/simplify_network.py
ekatef: Have added some. Does it look more clear now?

scripts/simplify_network.py
ekatef: I have tried to keep close to the approach used by `aggregate_to_substations( )`. We have `cluster_regions( )` in the very end:

https://github.com/ekatef/pypsa-earth/blob/b0f3f8a476a71a6b660406d987aad254030f2c96/scripts/simplify_network.py#L788

which is defined in `cluster_networks` and contains output to the file

https://github.com/ekatef/pypsa-earth/blob/b0f3f8a476a71a6b660406d987aad254030f2c96/scripts/cluster_network.py#L607-L618

ekatef:
After some initial testing with hierarchical clustering, it looks like the reason of troubles are micro-grids which present in the network. E.g. if there are tree nodes in such a grid, the clustering algorithm is happy only is a specified number of clusters is 3 or below. Clustering results doesn't look good in this case:

image

I think it's likely that such micro-grids in many cases are just a result of data loss and using of the recent `clean_osm_data` may improve the situation significantly

@davide-f, wouldn't it cause too much mess if I'll merge the current main into this branch? (It'll try to clean it up at the end! :) )

scripts/cluster_network.py
davide-f: yeah sorry mb

davide-f:
Could you better explain the above? Sorry but I quite didn't get the point

config.default.yaml
```diff
@@ -62,6 +62,8 @@ cluster_options:
     exclude_carriers: []
     remove_stubs: true
     remove_stubs_across_borders: true
+    p_threshold_drop_isolated: 10 # [MW] isolated buses are being discarded if bus mean power is below the specified threshold 
+    p_threshold_merge_isolated: 1 # [MW] isolated buses are being merged into a single isolated bus if bus mean power is above the specified threshold
```
davide-f: The values shall be the opposite, otherwise the merge does not work

scripts/simplify_network.py
```diff
@@ -522,6 +523,150 @@ def consense(x):
...
+
+    n.determine_network_topology()
+
+    # duplicated sub-networks mean that there is at least one interconnection between buses
+    i_islands = n.buses[
+        (~n.buses.duplicated(subset=["sub_network"], keep=False))
+        & (n.buses.carrier == "AC")
+    ].index
+
+    # isolated buses with load higher than a specified threshold should be merged
```
davide-f: isolated buses below the threshold should be merged

scripts/simplify_network.py
```diff
@@ -522,6 +523,150 @@ def consense(x):
...
+    )
+
+    return n
+
+
+def merge_isolated_nodes(n, threshold, aggregation_strategies=dict()):
+    """
+    Find isolated nodes in the network and merge those of them which have
+    load value higher than a specified threshold into a single isolated node
+    which represents all the remote generation
```
davide-f: Revise docstring:lower than threshold

scripts/simplify_network.py
```diff
@@ -620,6 +765,20 @@ def consense(x):
 
     update_p_nom_max(n)
 
+    p_threshold_drop_isolated = max(
+        0.0, cluster_config.get("p_threshold_drop_isolated", False)
```
davide-f: The default should be again 0.0 instead of False.
Probably it has not practical difference but it may be better

scripts/simplify_network.py
davide-f: Ok! Though my feeling is that the new busmap may not be compatible with the old one.
Maybe we should pipe them, otherwise we lose retrocompatibility.
if the first busmap maps 1 into 2 and then the second one 2 into 3.
Probably, the expected busmap to save is 1->3.
This enables compatibility with the current formulation I think

davide-f (commented):
Nice :)

A question: does the DC network get clustered?
By the images it doesn't like so, what is your feeling?

Especially the indian case seemed a bit particular

scripts/simplify_network.py
davide-f: better, thanks :)

ekatef:
> Could you better explain the above? Sorry but I quite didn't get the point

It looks like if we have isolated networks in the model, the hac algorithm works only if the clusters number is set equal or below the number of nodes in such disconnected network. The underlying `sklearn::AgglomerativeClustering` accounts for connectivity which in our case is determined by adjacency_matrix and thus obviously effected by disconnected networks

However, it looks like isolated networks are purely artifacts for some cases (e.g. Azerbajan) which should be treated earlier when cleaning the data or specifying the tolerance for building networks

Haven't yet merged the main with the recent improvements of `clean_osm_data` but it looks like might be worth :) I think it's better to be done on an experimental copy of the working branch

config.default.yaml
ekatef: Fixed

scripts/simplify_network.py
ekatef: Ah, that was the idea! :) Sorry, have misunderstood. Agree, it absolutely make sense. Revised

scripts/simplify_network.py
ekatef: Done

scripts/simplify_network.py
ekatef: Agree, done

ekatef:
> Ok! Though my feeling is that the new busmap may not be compatible with the old one.
> Maybe we should pipe them, otherwise we lose retrocompatibility.
if the first busmap maps 1 into 2 and then the second one 2 into 3.
> Probably, the expected busmap to save is 1->3.
> This enables compatibility with the current formulation I think

My feeling is that it may be better to discuss in a short exchange :) It sounds like a very good idea but not sure I fully get your point

ekatef:
> Nice :)
> 
> A question: does the DC network get clustered? By the images it doesn't like so, what is your feeling?
> 
> Especially the indian case seemed a bit particular

DC lines are here modelled as links, which seem to be treated by sub-network calculations in a bit specific way. Haven't investigated it in depth by it looks like each link might be treated by a workflow as a single network. Agree, the results of `hac` for India look a bit specific...

I have also found an issue when testing Central Asian domain for some specific  value of clusters number. Looking into this currently

scripts/cluster_network.py
ekatef: Thanks for clarification :)

ekatef:
Clustering for Central Asia testing domain works for all types of clustering. E.g. resulsts of hac-clustering with clusters distribution by GDP:

image

Interestingly, some troubles were caused by a fact that `pd.idxmin()` returns an arbitrary element if being applied to a row which consists of `np.inf.` To fix this, isolated buses were excluded from processing by `aggregate_to_substations()` as inf values are being used there as indicators to prevent improper buses assignment.

For India, kmeans clustering works for all the distribution factors with quite appropriate results. An example for distribution by GDP:

image

ekatef:
hac clustering for India (and generally networks with DC lines) should be tested

Not sure if it'll be possible to get work alternative clustering due to the mixed carries problem which arises for such topologies

README.md
```diff
@@ -154,27 +154,6 @@ The documentation is available here: [documentation](https://pypsa-earth.readthe
...
-            <sub><b>Jarrad Wright</b></sub>
-        </a>
-    </td>
-    <td align="center">
-        <a href="https://github.com/fneum">
-            <img src="https://avatars.githubusercontent.com/u/29101152?v=4" width="100;" alt="fneum"/>
-            <br />
-            <sub><b>Fabian Neumann</b></sub>
-        </a>
-    </td>
```
davide-f, line 177: This stuff re-appeared. I feel like your local fork needs cleaning.
When you do git merge main, you generally merge the local version of your github.
If your local main is not synced with the upstream main (hence your github fork not synced with the upstream and the local fork not synced with your github fork).

scripts/simplify_network.py
```diff
@@ -570,14 +728,14 @@ def consense(x):
         )
         busmaps.append(stub_map)
 
-    if cluster_config.get("to_substations", False):
+    if cluster_config.get("to_substations", 0.0):
```
davide-f: I think this False was ok, when false, we don't want to use the to_substations aggregations

ekatef: Fixed, thanks!

ekatef:
Finalized as #632

PR #618, opened 2023-03-06
emission_extractor() function is added

Emre-Yorat89:
# Added emission extraction function 

added the emission_extractor() to extract emission values and also the libraries it associates with.

- The scripts returns co2 emissions for given countries.

## Checklist


scripts/prepare_network.py
```diff
@@ -59,17 +59,45 @@
...
 
+def emission_extractor():
+    country_names = create_country_list(snakemake.config["countries"])
+    # data reading process
+    os.chdir(
+        os.path.join(
+            os.getcwd(),
+            "data",
+        )
+    )
```
pz-max: This will change the directory. I think you do this to read the excel file?
Let's read here only the path. Suggestion:

`data_path =  os.path.join(os.getcwd(), "data", "v60_CO2_excl_short-cycle_org_C_1970_2018.xls")`
which allows you to read in the data like this:
 `df = pd.ExcelFile(data_path)`

Probably this won't be necessary then:
`sets_path_to_root()`

scripts/prepare_network.py
```diff
@@ -59,17 +59,45 @@
...
+    emission_of_countries = []
+    df = pd.ExcelFile("v60_CO2_excl_short-cycle_org_C_1970_2018.xls")
```
pz-max: This looks pretty manual. Is there a way to retrieve the data from the web through e.g. Python directly? What you require for that is a download link which is not changing and to check google for examples of such autmated downloads.

As mentioned a function within this `emission_extractor()` that is called e.g. `download_emission_data()` doing this job would help

scripts/prepare_network.py
```diff
@@ -59,17 +59,45 @@
...
+    for j in country_names:
+        j = coco.convert(j, to="ISO3")
+        for i in range(9, 3918):
+            # 2 represents Country_Code_A3 column of the dataframe
+            three_digits = df.loc[i][2]
+            if j == three_digits:
+                # 27 represents Y_1990 column of the dataframe
+                if df.loc[i][5] == "Public electricity and heat production":
+                    emission_of_countries.append(df.loc[i][27])
+    sets_path_to_root("pypsa-earth")
```
pz-max: As mentioned above, this is probably not necessary

scripts/prepare_network.py
```diff
@@ -59,17 +59,45 @@
...
+            "data",
+        )
+    )
+    emission_of_countries = []
+    df = pd.ExcelFile("v60_CO2_excl_short-cycle_org_C_1970_2018.xls")
+    df = pd.read_excel(df, 1)
+
+    for j in country_names:
+        j = coco.convert(j, to="ISO3")
+        for i in range(9, 3918):
```
pz-max: You could add in the `read_excel()` some extra functions to skip the first 9lines. See https://pandas.pydata.org/docs/reference/api/pandas.read_excel.html.. Here the following is interesting for you:

- sheet_name, to only focus on the sheet name "Public electricity and heat production"
- skiprows, to skip the first nine rows

scripts/prepare_network.py
```diff
@@ -59,17 +59,45 @@
...
+            # 2 represents Country_Code_A3 column of the dataframe
+            three_digits = df.loc[i][2]
+            if j == three_digits:
+                # 27 represents Y_1990 column of the dataframe
```
pz-max, line 119: why not doing `df.loc[i, "Y_1990"]` ?

pz-max (changes requested):
Hi @Emre-Yorat89 , cool PR. Just addded some notes. If the automated web extraction of the data is possible, we could go together troough your code to see if we can improve something code wise

scripts/prepare_network.py
Emre-Yorat89: In the excel file "Y_1990" is not directly a column. The table is inserted after 6, 7 information rows. Therefore I could not use it as column name instead directly specified the index number

scripts/prepare_network.py
Emre-Yorat89: I could try skiprows parameter then use "Y_1990" as column name. Thank you :)

scripts/prepare_network.py
```diff
@@ -59,18 +59,68 @@
...
+    df = pd.ExcelFile(datapath)
+    df = pd.read_excel(df, 1)
+    # NEPAL
+    df.iloc[3293, 27] = df.iloc[3293, 28]
+    # GREENLAND
+    df.iloc[2045, 27] = df.iloc[2045, 41]
+    # GHANA
+    df.iloc[1107, 27] = df.iloc[1107, 30]
+    # Turks and Cacao Islands
+    df.iloc[499, 27] = df.iloc[499, 32]
```
ekatef: As a comment, would it be probably possible to generalize a proposed approach? For example, in case when no data are available in `Y_1990`, CO2 emissions are taken for the latest available year before 1990

scripts/prepare_network.py
```diff
@@ -280,8 +330,11 @@ def set_line_nom_max(n, s_nom_max_set=np.inf, p_nom_max_set=np.inf):
 
     for o in opts:
         if "Co2L" in o:
-            m = re.findall("[0-9]*\.?[0-9]+$", o)
```
ekatef, line 283: Was it really meant to be removed? It looks like we need this `m` value below

scripts/prepare_network.py
```diff
@@ -280,8 +330,11 @@ def set_line_nom_max(n, s_nom_max_set=np.inf, p_nom_max_set=np.inf):
...
-            if len(m) > 0:
+            if snakemake.config["electricity"]["automatic_emission_base_year"]:
```
ekatef: Shouldn't `len(m) > 0` be added here along with `"automatic_emission_base_year"` value? We need to load the emissions only if CO2 constrain is requested by the file name, right?

scripts/prepare_network.py
Emre-Yorat89: I have looked at [the documentation](https://pypsa-earth.readthedocs.io/en/latest/wildcards.html) and as far as I understand the m value is used to indicate whether to use co2limit or co2base as emission value. 
If m is lower than 0 then co2base is used and multiplied with m[0] which is used to take percentage of co2base (co2base*m[0]). Therefore, I think if we want the emission_extractor() function to work then we do not need the m value any longer. So the modification of the code in this part looks okay. What do you think?

scripts/prepare_network.py
Emre-Yorat89: I have modified it like in your suggestion

scripts/prepare_network.py
Emre-Yorat89: I have modified as you suggested, thank you for improving this.

scripts/prepare_network.py
Emre-Yorat89: I have added download_emission_data() function to download the data. Thanks for your feedback.

scripts/prepare_network.py
Emre-Yorat89: I have tried that now it is working as you can see in the last PR

scripts/prepare_network.py
Emre-Yorat89: sets_path_to_root(...) part is deleted. Thank you

scripts/prepare_network.py
ekatef: @Emre-Yorat89 you are absolutely right that `m` is used to switch between reading `co2limit` and `co2base` from the config. The first value `co2limit` means emissions threshold which shouldn't be crossed by the model. The second value `co2base` corresponds to some year in the past which is used as a reference state to compare the modelled emissions against. Like we want our power system to have 50% less emissions as compared with 1990 (a base year assumed in [Kyoto agreement](https://unfccc.int/process-and-meetings/transparency-and-reporting/reporting-and-review-under-the-kyoto-protocol/first-commitment-period/kyoto-protocol-base-year-data-for-the-first-commitment-period-of-the-kyoto-protocol) and quite often used until today).

As far as I understand, you are loading accurate values for a base year 1990, while a user still may want to modify them considering different goals for emissions reductions. So, I'd suggest to 
1) leave in use `m` as a user-defined parameter;
2) in case `len(m) > 0` calculate `co2limit` using this reduction coefficient like `co2limit = m * emission_extractor(filename)`

What is your feeling about that?

ekatef:
Hey @Emre-Yorat89! Very nice work and helpful result. Looking forward to see this PR merged, as it will facilitate a lot work with emissions data.

Currently CI is failing because it can't find `automatic_emission_base_year` in `config.tutorial.yaml` when running the tutorial workflow. Could you please fix this to make CI happy?

I have added some explanations on different definitions of CO2 constraint. It feels like quite a delicate moment. We may need to improve docs in this part, as currently `co2limit` and `co2base` are not defined in the documentation. As for the control logic of the workflow, I'd suggest to keep interfaces as stable as we can. Previously `CO20.5` was used to decrease CO2 emissions to 50% comparing to some abstract value of `1.487e+9` tonnes. As a user, I'd be happy to have `CO20.5` triggering same 50% emissions reduction but relative to a more accurate national-specific value, automatically extracted by the workflow. What do you think?

Generally, I very much like the changes you have introduced and think that your PR is close to be finalised. 

@pz-max would you like to have a look?

Emre-Yorat89:
Hey @ekatef ,
Thank you for solving CI issue, I will add automatic_emission_base_year line to config.tutorial.yaml file in the next PR. 

I think your logic gives more flexibility to the user. Therefore I will modify that part of code to include the future where user can decrease the co2 emission of the base year (1990).

Have a great day, thank you.

config.default.yaml
Emre-Yorat89, line 1: I accidentally changed the countries part as I was running the program. Can recommit to edit this to the old version?

ekatef: > I accidentally changed the countries part as I was running the program. Can recommit to edit this to the old version?

No worries, no harm done! You absolutely can fix and commit all what you feel is needed

ekatef:
> Hey @ekatef ,
> Thank you for solving CI issue, I will add automatic_emission_base_year line to config.tutorial.yaml file in the next PR.

My pleasure :)

> I think your logic gives more flexibility to the user. Therefore I will modify that part of code to include the future where user can decrease the co2 emission of the base year (1990).
> 
> Have a great day, thank you.

Super! I really like the changes you introduced

config.default.yaml
davide-f: This default file must be restored

scripts/prepare_network.py
```diff
@@ -59,18 +59,72 @@
...
+            )
+        os.remove(file_path)
+        return "v60_CO2_excl_short-cycle_org_C_1970_2018.xls"
+    except:
+        logger.warning(f"Failed download resource from '{url}'.")
+        return False
+
+
+def emission_extractor(filename):
+    country_names = create_country_list(snakemake.config["countries"])
```
davide-f, line 98: this function is not needed.
You can add a new argument called "country_names" to the function and you can pass the list that you get from the network.

scripts/prepare_network.py
```diff
@@ -59,18 +59,72 @@
...
+    # data reading process
+    datapath = os.path.join(os.getcwd(), "data", filename)
+    emission_of_countries = []
+    df = pd.ExcelFile(datapath)
```
davide-f, line 102: I don't think you need to do this preprocessing.
below, it should work `read_excel(datapath, ...`

scripts/prepare_network.py
```diff
@@ -59,18 +59,72 @@
...
+    df = pd.read_excel(df, sheet_name="v6.0_EM_CO2_fossil_IPCC1996", skiprows=8)
+    df.columns = df.iloc[0]
+    # NEPAL
+    df.loc[3285, "Y_1990"] = df.loc[3285, "Y_1988"]
+    # GREENLAND
+    df.loc[2037, "Y_1990"] = df.loc[2037, "Y_2004"]
+    # GHANA
+    df.loc[1099, "Y_1990"] = df.loc[1099, "Y_1986"]
+    # TURKS AND CAICOS ISLANDS
+    df.loc[491, "Y_1990"] = df.loc[491, "Y_1995"]
```
davide-f, line 112: Could you explain why we need this? Is the data complete?

scripts/prepare_network.py
```diff
@@ -59,18 +59,72 @@
...
+        for i in range(1, 3909):
+            # 2 represents Country_Code_A3 column of the dataframe
+            three_digits = df.loc[i]["Country_code_A3"]
+            if j == three_digits:
+                # 27 represents Y_1990 column of the dataframe
+                if (
+                    df.loc[i]["IPCC_for_std_report_desc"]
+                    == "Public electricity and heat production"
+                ):
+                    emission_of_countries.append(df.loc[i]["Y_1990"])
```
davide-f, line 124: There is a more pythonic way to do this:
1. you can use the function `df.set_index` to set "Country_code_A3" as the new index of the dataframe df
2. then, you can postprocess the country_names that are the 2-country code letters using `cc.pandas_convert(series=some_countries, to='ISO3')`
3. then, you can get the emissions by countries with df.loc[country_names_iso3, [column]]

What do you think? I think that may be nicer, but it requires little work and some problems may arise.

davide-f, line 124: Is there a reason why specifying "Y_1990"? Are there alternatives?
May that be an option?

davide-f (commented):
Thanks for the PR! I add some comments :)

ekatef:
> Thanks for the PR! I add some comments :)

Thanks a lot for the great review @davide-f! :)

We have agreed with @Emre-Yorat89 to present the PR during the tomorrow meeting. It'll be amazing to discuss also the ways to polish the code you suggest.

scripts/prepare_network.py
Emre-Yorat89: There are other years and I asked that question to @pz-max. Since all climate policies are based on 1990 therefore I used 1990.

Emre-Yorat89: Great points, I will try to cover them in the next PR :)

scripts/prepare_network.py
Emre-Yorat89: 1990 emission data is missing in those countries. Therefore I have filled by the values before 1990 (if any) otherwise used most recent year.

scripts/prepare_network.py
Emre-Yorat89: I will try that and I think it will make code look more clean. Thank you :)

scripts/prepare_network.py
Emre-Yorat89: I could not understand I will ask it in the developers' meeting :)

scripts/prepare_network.py
davide-f: Ok, then it may be interesting to have the year as an option loaded from config.
The `automatic_emission_base_year` could assume as values:
- false: when that function is disabled
- any numeric year (e.g. 1990) to specify the desired reference year.

In the config, we can add a comment specifying exactly that: use false to disable the function, otherwise a numeric value. We can suggest 1990 also.

Moreover, the additional config parameter shall be added to [electricity config table](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/doc/configtables/electricity.csv) to specify the option.

davide-f: Do you mean elaborate on next commits?

scripts/prepare_network.py
davide-f: Are there other columns that are fully filled?

scripts/prepare_network.py
davide-f: ok! the idea is to have `emission_extractor(country_names, filename)` and the country_names list is obtained from the network using `n.buses.country.unique()` for example

davide-f (commented):
Thanks @Emre-Yorat89 :) attached few suggestions

scripts/prepare_network.py
Emre-Yorat89: Yes, depending on the country, emission values in other years available and I have used them to fill missing value at 1990 (preferable before 1990).

scripts/prepare_network.py
Emre-Yorat89: In order to that functionality an intermediate preprocess is needed which obtains "Public Electricity and Heat Production" sector. Because there are more than one emission values depending on the sector interested.
The code I am thinking to use to obtain aforementioned sector is:  df = df.loc[df['IPCC_for_std_report_desc'] == "Public electricity and heat production"]
The code now looks more clean with your suggestions, thank you.

davide-f:
Hello @Emre-Yorat89  :)
This is a great contribution! :) I  was wondering what's the status on the comments.
We would love to have it integrated soon

davide-f:
Closing this because of #748

PR #620, opened 2023-03-06
Revise download and cleaning of world osm data

### LINKED ISSUE #504: Remove code duplication build_osm_network and clean_osm_network
### See more information in the discussion in https://github.com/pypsa-meets-earth/pypsa-earth/discussions/400

<div type='discussions-op-text'>

<sup>Originally posted by **pz-max** July  2, 2022</sup>
I was just checking a bit the osm code. It seems to me that we have some code duplications.

A. See build osm:
https://github.com/pypsa-meets-africa/pypsa-africa/blob/1d662a4c77e76d40a6ff36d383687614af6ab3ae/scripts/build_osm_network.py#L66-L105

B. And clean osm:
https://github.com/pypsa-meets-africa/pypsa-africa/blob/1d662a4c77e76d40a6ff36d383687614af6ab3ae/scripts/clean_osm_data.py#L75-L122

C. This function is also not used at any place:
https://github.com/pypsa-meets-africa/pypsa-africa/blob/1d662a4c77e76d40a6ff36d383687614af6ab3ae/scripts/build_osm_network.py#L37-L63

Can we remove C and merge A & B?   
Thoughts @davide-f & @ekatef </div>

### LINKED ISSUE #601: `split_cells_multiple` function shall be completely rewritten
## Description
The function `split_cells_multiple` in clean_osm_data need to be rewritten to improve both performances and data accuracy.

## Scope of the function

The function has the scope to manipulate the columns voltage, "circuits" and "cables" obtained from osm data.
In particular, some osm entries represent multiple circuits: a line in osm data may represent more than a typical 3-phase line stacked alltogether. In such case, entries of the columns contain ";", e.g. "220;380" hence specifying that two circuits of 220V and 380V are present.
When that applies circuits, voltage and cables column may or may not contain the ";" character.
The function has the goal to decouple these rows and create a unique entry for them.

## Expected improvements

In particular, in some cases the ";" is present only in the voltage column but not in the others and viceversa.
Currently, we address the case when the ";" is present in all columns; the new version of the function shall be more general.

To improve performances, the pandas function str.split function shall be wisely used so to avoid using python row-wise for loops that are highly inefficient

### LINKED ISSUE #619: Modify clean_osm_data in a non-breaking way
## Checklist

## Describe the Bug

When running PyPSA-Earth on some regions (e.g. US) the procedure breaks because of unexpected OSM entries, e.g. "24 kHz" in the frequency column.
The cleaning of the data shall be expeption-compliant

davide-f:
# Closes #601 , #619 and #504

This PR aims at restructuring the osm cleaning process to be more exception-resilient.
In particular:
- it fixes the interface with earth-osm for selected countries to be compliant with our "Earth" list
- it improves and normalizes the split_cells procedure to be length-independent, eventually using some heuristics to fix the mismatch
- it makes the cleaning process resilient to unexpected values: it drops them
- it decomposes the cleaning process by OSM column (voltage/frequency/circuits) in specific sub-functions
- it improves the logging
- add "force_ac" option in build osm network to force lines and buses to be AC-only
- clean duplication between clean_osm_data and build_osm_network

## Checklist

### DIFF
--- a/scripts/build_osm_network.py
+++ b/scripts/build_osm_network.py
@@ -725,6 +725,8 @@ def built_network(inputs, outputs, config, geo_crs, distance_crs, force_ac=False
     lines = gpd.read_file(inputs["lines"])
     generators = read_geojson(inputs["generators"])
 
+    lines = line_endings_to_bus_conversion(lines)
+
     if force_ac:
         logger.info(
             "Stage 2/5: AC and DC network: disabled, forced buses and lines to AC"
--- a/scripts/build_osm_network.py
+++ b/scripts/build_osm_network.py
@@ -705,8 +705,14 @@ def fix_overpassing_lines(lines, buses, distance_crs, tol=1):
 
 
 def force_ac_lines(df, col="tag_frequency"):
-    "Function to force AC for lines"
+    """
+    Function that forces all PyPSA lines to be AC lines.
 
+    A network can contain AC and DC power lines that are modelled as PyPSA "Line" component.
+    When DC lines are available, their power flow can be controlled by their converter.
+    When it is artificially converted into AC, this feature is lost.
+    However, for debugging and preliminary analysis, it can be useful to bypass problems.
+    """
     DC_freq = 0.0
     DC_lines = (df["tag_frequency"] - DC_freq).abs() <= 0.01
 
--- a/scripts/build_osm_network.py
+++ b/scripts/build_osm_network.py
@@ -33,47 +33,6 @@ def line_endings_to_bus_conversion(lines):
     return lines
 
 
-def add_line_endings_tosubstations(substations, lines):
-    # extract columns from substation df
-    bus_s = gpd.GeoDataFrame(columns=substations.columns)
-    bus_e = gpd.GeoDataFrame(columns=substations.columns)
-
-    is_ac = lines["tag_frequency"].astype(float) != 0
-
-    # Read information from line.csv
-    bus_s[["voltage", "country"]] = lines[["voltage", "country"]]  # line start points
-    bus_s["geometry"] = lines.geometry.boundary.map(lambda p: p.geoms[0])
-    bus_s["lon"] = bus_s["geometry"].x
-    bus_s["lat"] = bus_s["geometry"].y
-    bus_s["bus_id"] = lines["line_id"].astype(str) + "_s"
-    bus_s["dc"] = ~is_ac
-
-    bus_e[["voltage", "country"]] = lines[["voltage", "country"]]  # line start points
-    bus_e["geometry"] = lines.geometry.boundary.map(lambda p: p.geoms[1])
-    bus_e["lon"] = bus_s["geometry"].x
-    bus_e["lat"] = bus_s["geometry"].y
-    bus_e["bus_id"] = lines["line_id"].astype(str) + "_e"
-    bus_e["dc"] = ~is_ac
-
-    bus_all = pd.concat([bus_s, bus_e], ignore_index=True).set_crs(substations.crs)
-    # Assign index to bus_id
-    bus_all["bus_id"] = bus_all.index
-
-    # Add NaN as default
-    bus_all["station_id"] = np.nan
-    # bus_all["dc"] = False  # np.nan
-    # Assuming substations completed for installed lines
-    bus_all["under_construction"] = False
-    bus_all["tag_area"] = 0.0  # np.nan
-    bus_all["symbol"] = "substation"
-    # TODO: this tag may be improved, maybe depending on voltage levels
-    bus_all["tag_substation"] = "transmission"
-
-    buses = pd.concat([substations, bus_all], ignore_index=True)
-
-    return buses
-
-
 # tol in m
 def set_substations_ids(buses, distance_crs, tol=2000):
     """
@@ -745,18 +704,35 @@ def fix_overpassing_lines(lines, buses, distance_crs, tol=1):
     return lines, buses
 
 
-def built_network(inputs, outputs, config, geo_crs, distance_crs):
+def force_ac_lines(df, col="tag_frequency"):
+    "Function to force AC for lines"
+
+    DC_freq = 0.0
+    DC_lines = (df["tag_frequency"] - DC_freq).abs() <= 0.01
+
+    # TODO: default frequency may be by country
+    default_frequency = 50
+
+    df[df[DC_lines].index] = default_frequency
+
+    return df
+
+
+def built_network(inputs, outputs, config, geo_crs, distance_crs, force_ac=False):
     logger.info("Stage 1/5: Read input data")
 
-    substations = gpd.read_file(inputs["substations"])
+    buses = gpd.read_file(inputs["substations"])
     lines = gpd.read_file(inputs["lines"])
     generators = read_geojson(inputs["generators"])
 
-    logger.info("Stage 2/5: Add line endings to the substation datasets")
-
-    # Use lines and create bus/line df
-    lines = line_endings_to_bus_conversion(lines)
-    buses = add_line_endings_tosubstations(substations, lines)
+    if force_ac:
+        logger.info(
+            "Stage 2/5: AC and DC network: disabled, forced buses and lines to AC"
+        )
+        lines = force_ac_lines(lines)
+        buses["dc"] = False
+    else:
+        logger.info("Stage 2/5: AC and DC network: enabled")
 
     # Address the overpassing line issue Step 3/5
     if config.get("build_osm_network", {}).get("split_overpassing_lines", False):
@@ -867,9 +843,15 @@ if __name__ == "__main__":
     # load default crs
     geo_crs = snakemake.config["crs"]["geo_crs"]
     distance_crs = snakemake.config["crs"]["distance_crs"]
+    force_ac = snakemake.config["build_osm_network"].get("force_ac", False)
 
     sets_path_to_root("pypsa-earth")
 
     built_network(
-        snakemake.input, snakemake.output, snakemake.config, geo_crs, distance_crs
+        snakemake.input,
+        snakemake.output,
+        snakemake.config,
+        geo_crs,
+        distance_crs,
+        force_ac=force_ac,
     )
--- a/scripts/clean_osm_data.py
+++ b/scripts/clean_osm_data.py
@@ -770,15 +770,15 @@ def clean_data(
         df_all_lines.apply(lambda x: africa_shape.contains(x.geometry.boundary), axis=1)
     ]
 
-    # set unique line ids
-    df_all_lines = set_unique_id(df_all_lines, "line_id")
-
     df_all_lines = gpd.GeoDataFrame(df_all_lines, geometry="geometry")
 
     # set the country name by the shape
     if names_by_shapes:
         logger.info("Setting lines country name using the GADM shapes")
         df_all_lines = set_countryname_by_shape(df_all_lines, ext_country_shapes)
+
+    # set unique line ids
+    df_all_lines = set_unique_id(df_all_lines, "line_id")
 
     # save lines output
     logger.info("Saving lines output")
@@ -816,9 +816,6 @@ def clean_data(
     # finalize dataframe types
     df_all_substations = finalize_substation_types(df_all_substations)
 
-    # set unique bus ids
-    df_all_substations = set_unique_id(df_all_substations, "bus_id")
-
     # save to geojson file
     df_all_substations = gpd.GeoDataFrame(df_all_substations, geometry="geometry")
 
@@ -830,6 +827,9 @@ def clean_data(
             ext_country_shapes,
             col_country="Country",
         )
+
+    # set unique bus ids
+    df_all_substations = set_unique_id(df_all_substations, "bus_id")
 
     # save substations output
     logger.info("Saving substations output")
--- a/scripts/clean_osm_data.py
+++ b/scripts/clean_osm_data.py
@@ -43,8 +43,7 @@ def prepare_substation_df(df_all_substations):
     df_all_substations["lon"] = df_all_substations["geometry"].x
     df_all_substations["lat"] = df_all_substations["geometry"].y
 
-    # Add NaN as default
-    df_all_substations["station_id"] = np.nan
+    # Initialize columns to default value
     df_all_substations["dc"] = False
     df_all_substations["under_construction"] = False
 
@@ -105,11 +104,11 @@ def add_line_endings_tosubstations(substations, lines):
 
     bus_all = pd.concat([bus_s, bus_e], ignore_index=True)
 
-    # Add NaN as default
+    # Initialize default values
     bus_all["station_id"] = np.nan
     # Assuming substations completed for installed lines
     bus_all["under_construction"] = False
-    bus_all["tag_area"] = 0.0  # np.nan
+    bus_all["tag_area"] = 0.0
     bus_all["symbol"] = "substation"
     # TODO: this tag may be improved, maybe depending on voltage levels
     bus_all["tag_substation"] = "transmission"
@@ -162,6 +161,15 @@ def split_cells(df, cols=["voltage"]):
         Dataframe under analysis
     cols : list
         List of target columns over which to perform the analysis
+
+    Example
+    -------
+    Original data:
+    row 1: '66000;220000', '50'
+
+    After applying split_cells():
+    row 1, '66000', '50'
+    row 2, '220000', '50'
     """
     if df.empty:
         return df
@@ -172,6 +180,7 @@ def split_cells(df, cols=["voltage"]):
 
 
 def filter_voltage(df, threshold_voltage=35000):
+    """Filters df to contain only lines with voltage above threshold_voltage"""
     # Convert to numeric and drop any row with N/A voltage
     df["voltage"] = pd.to_numeric(df["voltage"], errors="coerce").astype(float)
     df.dropna(subset=["voltage"], inplace=True)
@@ -186,6 +195,7 @@ def filter_voltage(df, threshold_voltage=35000):
 
 
 def filter_frequency(df, accepted_values=[50, 60, 0]):
+    """Filters df to contain only lines with frequency with accepted_values"""
     df["tag_frequency"] = pd.to_numeric(df["tag_frequency"], errors="coerce").astype(
         float
     )
@@ -202,6 +212,7 @@ def filter_frequency(df, accepted_values=[50, 60, 0]):
 
 
 def filter_circuits(df, min_value_circuit=0.1):
+    """Filters df to contain only lines with circuit value above min_value_circuit."""
     df["circuits"] = pd.to_numeric(df["circuits"], errors="coerce").astype(float)
     df.dropna(subset=["circuits"], inplace=True)
 
@@ -246,12 +257,6 @@ def prepare_lines_df(df_lines):
         }
     )
 
-    # Add NaN as default
-    df_lines["bus0"] = np.nan
-    df_lines["bus1"] = np.nan
-    df_lines["underground"] = np.nan
-    df_lines["under_construction"] = np.nan
-
     # Rearrange columns
     clist = [
         "line_id",
@@ -288,13 +293,13 @@ def finalize_lines_type(df_lines):
     return df_lines
 
 
-def parse_frequency(df):
-    "Function to parse raw frequency column: manual fixing and drop undesired values"
-
+def clean_frequency(df):
+    """
+    Function to clean raw frequency column: manual fixing and fill nan values
+    """
     # replace raw values
     repl_freq = {
         "16.67": "16.7",
-        "50;50;16.716.7": "50;50;16.7;16.7",
         "50;50;16.716.7": "50;50;16.7;16.7",
         "50;16.7?": "50;16.7",
         # "24 kHz": "24000",
@@ -311,9 +316,10 @@ def parse_frequency(df):
     return df
 
 
-def parse_voltage(df):
-    "Function to parse raw voltage column: manual fixing and drop undesired values"
-
+def clean_voltage(df):
+    """
+    Function to clean the raw voltage column: manual fixing and drop nan values
+    """
     # replace raw values
     repl_voltage = {
         "medium": "33000",
@@ -332,7 +338,8 @@ def parse_voltage(df):
         .str.replace("_", "")
         .str.replace("kv", "000")
         .str.replace("v", "")
-        # .str.replace("/", ";")
+        # .str.replace("/", ";")  # few OSM entries are separated by / instead of ;
+        # this line can be a fix for that if relevant
     )
 
     # drop na values in voltage
@@ -341,13 +348,14 @@ def parse_voltage(df):
     return df
 
 
-def parse_circuits(df):
-    "Function to parse raw circuits column: manual fixing and drop undesired values"
-
+def clean_circuits(df):
+    """
+    Function to clean the raw circuits column: manual fixing and clean nan values
+    """
     # replace raw values
     repl_circuits = {
-        "1/3": "0",
-        "2/3": "0",
+        "1/3": "1",
+        "2/3": "2",
         # assumption of two lines and one grounding wire
         "2-1": "2",
         "single": "1",
@@ -366,9 +374,10 @@ def parse_circuits(df):
     return df
 
 
-def parse_cables(df):
-    "Function to parse raw cables column: manual fixing and drop undesired values"
-
+def clean_cables(df):
+    """
+    Function to clean the raw cables column: manual fixing and drop undesired values
+    """
     # replace raw values
     repl_cables = {
         "single": "1",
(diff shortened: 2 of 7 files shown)


davide-f:
This PR is quite heavy and two reviews may be better, including by @ekatef since you first developed the cleaning process.
If you agree, @ekatef you may have a first check to this PR, then @pz-max.

This PR has been tested on the whole world dataset (without the shape comparison for simplicity) and it works; hence I'd expect to be quite error resilient.
As comments, some options may be needed as well some warning for codes that may be dropped, but it is better to hear your opinion before further coding.

davide-f:
Note: the CI works and the objective function is different from before: now (1.18e+09) vs before (1.34e+09).
I have not investigated the specific reasons in the network but this result is not surprising and, on the contrary, very promising due to the following: we may capture a larger number of lines hence improving the overall efficiency of the network.

The problem being a minimization problem, hence with the proposed changes costs decrease: good if it makes sense.
This PR has the effect to capture a larger number of lines, hence improving the connectivity of the network that leads to less isolated nodes, more interconnection capacity and less load shedding.
Therefore, this result makes sense.

davide-f:
The CI doesn't work because in botswana some nodes are classified as DC and then again the AC/DC problem appears.

An additional improvement on top of this can be fixing all lines to ac.
To place this we have to options:
1. having it into clean_osm data (I'd have the PR ready basically https://github.com/davide-f/pypsa-earth/tree/clean_osm_with_ac_forced)
2. having it into build_osm_network

(2) I think may be more appropriate, however I am not sure if it is a good idea to include it here now.
In case of (2) the add_line_ending_to_substations in build_osm_network shall be removed too.

An additional PR can address that but first we need to agree on this PR, otherwise the revision becomes basically impossible.

ekatef:
@davide-f Fantastic work! 

Would be happy to review and will try to do so as fast as I can 🙂

davide-f:
> @davide-f Fantastic work!
> 
> Would be happy to review and will try to do so as fast as I can 🙂

Many thanks, actually this is a quite important PR also for the conference paper indeed

scripts/build_osm_network.py
```diff
@@ -745,18 +704,37 @@ def fix_overpassing_lines(lines, buses, distance_crs, tol=1):
     return lines, buses
 
 
-def built_network(inputs, outputs, config, geo_crs, distance_crs):
+def force_ac_lines(df, col="tag_frequency"):
+    "Function to force AC for lines"
```
pz-max: """Function that forces all PyPSA lines to be AC lines.

A network can contain AC and DC power lines that are modelled as PyPSA "Line" component.
Do debug and allow quick fixes, it can be useful to convert all lines to the type of AC lines,
even though, it creates a less realistic power flow. This function converts all lines to AC lines.
"""

scripts/clean_osm_data.py
```diff
@@ -45,8 +45,8 @@ def prepare_substation_df(df_all_substations):
 
     # Add NaN as default
```
pz-max: revise this comment

scripts/clean_osm_data.py
```diff
@@ -105,13 +101,12 @@ def add_line_endings_tosubstations(substations, lines):
     bus_e["lon"] = bus_e["geometry"].map(lambda p: p.x if p != None else None)
     bus_e["lat"] = bus_e["geometry"].map(lambda p: p.y if p != None else None)
     bus_e["bus_id"] = bus_s["bus_id"].max() + 1 + bus_e.index
-    bus_s["dc"] = ~is_ac
```
pz-max, line 108: Looks like this was a bug :D

scripts/clean_osm_data.py
```diff
@@ -105,13 +101,12 @@ def add_line_endings_tosubstations(substations, lines):
...
+    bus_e["dc"] = ~is_ac
 
     bus_all = pd.concat([bus_s, bus_e], ignore_index=True)
 
     # Add NaN as default
```
pz-max: revise this comment e.g. add it as line comment in the next line

scripts/clean_osm_data.py
```diff
@@ -157,42 +152,29 @@ def set_unique_id(df, col):
     return df
 
 
-def split_cells(df, lst_col="voltage"):
+def split_cells(df, cols=["voltage"]):
     """
     Split semicolon separated cells i.e. [66000;220000] and create new identical rows
```
pz-max, line 157: I think an example would be nice.
rows, voltage, frequency
row 1, [66000;220000], 50
after applying split_cells()
row 1, 66000, 50
row 2, 220000, 50

scripts/clean_osm_data.py
```diff
@@ -157,42 +152,29 @@ def set_unique_id(df, col):
...
-    # Split semicolon separated cells i.e. [66000;220000] and create new identical rows
-    df = split_cells(df)
 
-    # Convert voltage to float, if impossible, discard row
-    df["voltage"] = (
-        df["voltage"].apply(lambda x: pd.to_numeric(x, errors="coerce")).astype(float)
-    )
-    df = df.dropna(subset=["voltage"])  # Drop any row with Voltage = N/A
+def filter_voltage(df, threshold_voltage=35000):
+    # Convert to numeric and drop any row with N/A voltage
```
pz-max, line 185: """Filters df to contain only lines with voltage above threshold_voltage"""

scripts/clean_osm_data.py
```diff
@@ -203,6 +185,33 @@ def filter_voltage(df, threshold_voltage=35000):
     return df
 
 
+def filter_frequency(df, accepted_values=[50, 60, 0]):
+    df["tag_frequency"] = pd.to_numeric(df["tag_frequency"], errors="coerce").astype(
```
pz-max, line 200: """Filters df to contain only lines with frequency with accepted_values"""

scripts/clean_osm_data.py
```diff
@@ -203,6 +185,33 @@ def filter_voltage(df, threshold_voltage=35000):
...
+        axis=1,
+    ).any(axis="columns")
+
+    df.drop(df[~accepted_rows].index, inplace=True)
+
+    return df
+
+
+def filter_circuits(df, min_value_circuit=0.1):
+    df["circuits"] = pd.to_numeric(df["circuits"], errors="coerce").astype(float)
```
pz-max, line 217: """Filters df to contain only lines with circuit value above min_value_circuit."""

pz-max: Just a comment, I thought circuits are of type int?

scripts/clean_osm_data.py
```diff
@@ -279,279 +288,320 @@ def finalize_lines_type(df_lines):
...
-    x for x in cables_tag_to_n_cables.keys() if cables_tag_to_n_cables[x] == "0"
-]
-dropped_circuits_tags = [
-    x for x in circuits_tag_to_n_circuits.keys() if circuits_tag_to_n_circuits[x] == "0"
-]
+    # replace raw values
+    repl_freq = {
+        "16.67": "16.7",
+        "50;50;16.716.7": "50;50;16.7;16.7",
+        "50;50;16.716.7": "50;50;16.7;16.7",
```
pz-max: 297 & 298 seems duplicated?

scripts/clean_osm_data.py
```diff
@@ -279,279 +288,320 @@ def finalize_lines_type(df_lines):
     return df_lines
 
 
-def split_cells_multiple(df, list_col=["cables", "circuits", "voltage"]):
-    """
-    Split function for cables and voltage
+def parse_frequency(df):
+    "Function to parse raw frequency column: manual fixing and drop undesired values"
```
pz-max: """DocString."""

scripts/clean_osm_data.py
```diff
@@ -279,279 +288,320 @@ def finalize_lines_type(df_lines):
     return df_lines
 
 
-def split_cells_multiple(df, list_col=["cables", "circuits", "voltage"]):
-    """
-    Split function for cables and voltage
+def parse_frequency(df):
```
pz-max: The word parser is to me a bit confusing, knowing it is common in compiler stuff (see [here](https://www.techopedia.com/definition/3854/parser)). I would call it `clean_frequency` (fits also to clean_osm if same function have clean in the name). What we do here is to replace certain values & for every na value set a default.

scripts/clean_osm_data.py
```diff
@@ -279,279 +288,320 @@ def finalize_lines_type(df_lines):
...
+
+    df["voltage"] = (
+        df["voltage"]
+        .replace(repl_voltage)
+        .str.lower()
+        .str.replace(" ", "")
+        .str.replace("_", "")
+        .str.replace("kv", "000")
+        .str.replace("v", "")
+        # .str.replace("/", ";")
```
pz-max: can be removed?

scripts/clean_osm_data.py
```diff
@@ -279,279 +288,320 @@ def finalize_lines_type(df_lines):
...
-    # Simplified. If tag_type cable then underground is True
-    df_all_lines["underground"] = df_all_lines["tag_type"] == "cable"
+    len_freq = df["tag_frequency"].map(len)
+    len_voltage = df["voltage"].map(len)
 
-    # More information extractable for "underground" by looking at "tag_location".
-    if "tag_location" in df_all_lines:  # drop column if exist
-        df_all_lines.drop(columns="tag_location", inplace=True)
+    def _fill_by_last(row, col_to_fill, size_col):
+        size_to_fill = len(row[size_col])
```
pz-max, line 442: dostring would be ncie here

scripts/clean_osm_data.py
```diff
@@ -279,279 +288,320 @@ def finalize_lines_type(df_lines):
...
+
+    df["cables"] = df["cables"].replace(repl_cables).str.replace(" ", "")
+
+    return df
+
+
+def split_and_match_vf(df):
     """
-    Function to add underground, under_construction, frequency and circuits
+    Function to match the length of the columns in subset
```
pz-max, line 408: vf stands for voltage and frequency?

pz-max: Docstring is not clear. It talk about match but what about the split? Why does it exist? Regarding the 'match', see two comments further down what I suggest. Can you adapt it here as well?

Just checking the full function, this looks fairly complex

scripts/clean_osm_data.py
```diff
@@ -279,279 +288,320 @@ def finalize_lines_type(df_lines):
...
-        not_resolved_cables = (
-            ~pd.to_numeric(df_all_lines["cables"], errors="coerce")
-            .astype(float)
-            .apply(float.is_integer)
-        )
-        unknown_cables_tags = set(
-            df_all_lines.loc[not_resolved_cables]["cables"].values
+def fill_circuits(df):
+    """
+    This function fills the circuits column to match the length of v/f columns
```
pz-max: This function fills the circuits column to match the length of v/f columns OR
This function fills the rows in the circuits column to match the length of the rows in the voltage and frequency columns?

scripts/clean_osm_data.py
```diff
@@ -279,279 +288,320 @@ def finalize_lines_type(df_lines):
...
+        lambda x: ";".join(["1"] * x)
+    )
+
+    # explode column
+    df["circuits"] = df["circuits"].str.split(";")
+
+    return df
+
+
+def explode_columns(df, cols):
```
pz-max: explode rows or columns?

pz-max: docstring?

scripts/clean_osm_data.py
```diff
@@ -279,279 +288,320 @@ def finalize_lines_type(df_lines):
...
+
+def integrate_lines_df(df_all_lines, distance_crs):
+    """
+    Function to add underground, under_construction, frequency and circuits
+    """
+    # explode frequency and columns
+    df = pd.DataFrame(df_all_lines)
+
+    # preliminary raw parsing of raw columns
+    parse_frequency(df)
```
pz-max: as mentioned various commetns above, check about if renamig to clean_frequency etc...

scripts/clean_osm_data.py
```diff
@@ -279,279 +288,320 @@ def finalize_lines_type(df_lines):
...
+    parse_voltage(df)
+    parse_circuits(df)
+    parse_cables(df)
+
+    # fill voltage and frequency columns length for explode
+    split_and_match_vf(df)
+
+    # fill the circuits column for explode
```
pz-max, line 646: unclear

scripts/clean_osm_data.py
```diff
@@ -279,279 +288,320 @@ def finalize_lines_type(df_lines):
...
+    # explode frequency and columns
+    df = pd.DataFrame(df_all_lines)
+
+    # preliminary raw parsing of raw columns
+    parse_frequency(df)
+    parse_voltage(df)
+    parse_circuits(df)
+    parse_cables(df)
+
+    # fill voltage and frequency columns length for explode
```
pz-max: unclear description

scripts/clean_osm_data.py
pz-max: I think this could be tested by a unit test to check if it works (maybe overkill here_

pz-max (changes requested):
Looks good Davide. Massive changes & improvements. See my comments that can be addressed

scripts/build_osm_network.py
davide-f: Added

scripts/clean_osm_data.py
davide-f: Revised

scripts/clean_osm_data.py
davide-f: Yeah XD but I don't expect it would create much problems, maybe it simplified some XD

scripts/clean_osm_data.py
davide-f: Revised comment

scripts/clean_osm_data.py
davide-f: Done

davide-f: > I think this could be tested by a unit test to check if it works (maybe overkill here_

Yeah, we would need first the unit test by script first XD

scripts/clean_osm_data.py
davide-f: Done

scripts/clean_osm_data.py
davide-f: Done

scripts/clean_osm_data.py
davide-f: revised

scripts/clean_osm_data.py
davide-f: The "" works fine as well as long as it is a line, is there a specific reason to use the """"""?

davide-f: Anyway I'm doing it

scripts/clean_osm_data.py
davide-f: They are, my bad: removing

scripts/clean_osm_data.py
davide-f: Theoretically yes, but I'd like to keep it to point out that some lines in the dataset are actually separated by / instead of ;.
However, for some entries it works ok but others do not.
I add a comment here

scripts/clean_osm_data.py
davide-f: > vf stands for voltage and frequency?

Yes, revised

davide-f: > Docstring is not clear. It talk about match but what about the split? Why does it exist? Regarding the 'match', see two comments further down what I suggest. Can you adapt it here as well?
> 
> Just checking the full function, this looks fairly complex

Docstring revised; the algorithm may be changed to be more conservative in case.
We can talk about it

scripts/clean_osm_data.py
davide-f: done

scripts/clean_osm_data.py
davide-f: exactly.
I added a large docstring

scripts/clean_osm_data.py
davide-f: Added

scripts/clean_osm_data.py
davide-f: Done

scripts/clean_osm_data.py
davide-f: Revised

scripts/clean_osm_data.py
davide-f: Not necessarily, it works too.
If not necessary, I think it is worth considering them float as we can easily address the "1/3" circuits

davide-f: > """Filters df to contain only lines with circuit value above min_value_circuit."""

done

scripts/clean_osm_data.py
```diff
@@ -279,279 +293,382 @@ def finalize_lines_type(df_lines):
...
+       become:
+       row 1: ['50'], ['220000']
+       row 2: ['50','50','50'], ['220000','380000']
+
+    2. Then, it harmonize each row to match the length of the lists
+       by filling the missing values with the last elements of each list.
+       In agreement to the example of before, after the cleaning:
+
+       row 1: ['50'], ['220000']
+       row 2: ['50','50','50'], ['220000','380000','380000']
```
davide-f, line 426: An alternative cleaning fo this case may be:
row 2: ['50','50'], ['220000','380000']

Using the following: when the length in frequency is higher than in voltage, we stay conservative and we drop a frequency value.
However, this is not frequent

scripts/clean_osm_data.py
davide-f: Revised

scripts/clean_osm_data.py
pz-max: I will add GitHub issue. This is a great example for unit testing :)

davide-f: By the way, ok for the issue, but I don't think it is needed for this function rather others.
This function is pretty ok with respect to others I think.

pz-max: Agree. I just remember some nasty errors can appear for such code parts.
Testing more in the cleaning infrastructure would be good

scripts/clean_osm_data.py
```diff
@@ -279,279 +294,382 @@ def finalize_lines_type(df_lines):
...
+    clean_voltage(df)
+    clean_circuits(df)
+    clean_cables(df)
+
+    # analyse each row of voltage and requency and match their content
+    # Example:
+    #       tag_frequency   voltage
+    # row1: 50;50           33000
+    # Becomes
+    # row1: 50;50           33000;33000
```
pz-max: artifact

pz-max (approved):
Great contribution. Approve, but please remove the artefact.

scripts/clean_osm_data.py
davide-f: Yeah, that's why I added some warming checks in the explode_rows function.
That may help too

scripts/clean_osm_data.py
davide-f: What do you mean artifact?

scripts/clean_osm_data.py
```diff
@@ -279,279 +294,382 @@ def finalize_lines_type(df_lines):
...
+    repl_freq = {
+        "16.67": "16.7",
+        "50;50;16.716.7": "50;50;16.7;16.7",
+        "50;16.7?": "50;16.7",
+        # "24 kHz": "24000",
+    }
+
+    # TODO: default frequency may be by country
+    # default frequency
+    default_frequency = "50"
```
ekatef: Not sure it's an obvious place to look for the default frequency value. Wouldn't it be more convenient to have it in the arguments list?

scripts/clean_osm_data.py
```diff
@@ -279,279 +294,382 @@ def finalize_lines_type(df_lines):
...
+    repl_cables = {
+        "single": "1",
+        "triple": "3",
+        "partial": "1",
+        "1 disused": "0",
+        "3;3 disused": "3;0",
+        "1 (Looped - Haul & Return) + 1 power wire": "1",
+        "ground": "0",
+        "line": "1",
+        # assuming that in case of a typo there is at least one line
```
ekatef: I'm afraid the comment might slightly misleading (which is indeed inherited from the previous version). Would it be probably more clear to move the whole `*: "1"` block to the end of the dictionary?

scripts/clean_osm_data.py
```diff
@@ -279,279 +294,382 @@ def finalize_lines_type(df_lines):
...
+        .str.replace("kv", "000")
+        .str.replace("v", "")
+        # .str.replace("/", ";")  # few OSM entries are separated by / instead of ;
+        # this line can be a fix for that if relevant
+    )
 
-    # More information extractable for "underground" by looking at "tag_location".
-    if "tag_location" in df_all_lines:  # drop column if exist
-        df_all_lines.drop(columns="tag_location", inplace=True)
+    # drop na values in voltage
```
ekatef: Is there a special reason for this comment? It looks rather obvious from the code below

scripts/clean_osm_data.py
```diff
@@ -279,279 +294,382 @@ def finalize_lines_type(df_lines):
...
+    1. First, it splits voltage and frequency columns by semicolon
+       For example, the following lines
+       row 1: '50', '220000
+       row 2: '50;50;50', '220000;380000'
+
+       become:
+       row 1: ['50'], ['220000']
+       row 2: ['50','50','50'], ['220000','380000']
+
+    2. Then, it harmonize each row to match the length of the lists
```
ekatef, line 421: Just an understanding question: previously we have looked into cables and circuits numbers to restore a number of the lines. Now we focus on checking frequency-voltage correspondence but disregard looking into match between circuits-cables-lines, right? 

I suspect the current approach is more meaningful and it definitely looks much cleaner. But doesn't it mean some data loss? Or is it negligible?

scripts/clean_osm_data.py
davide-f: done

scripts/clean_osm_data.py
```diff
@@ -597,11 +716,11 @@ def prepare_generators_df(df_all_generators):
...
-    """Return the first country whose shape intersects the geometry"""
-    for c_name, c_geom in country_geoms.items():
-        if not geom.disjoint(c_geom):
-            return c_name
-    return default_name
+    """Return the first index whose shape intersects the geometry"""
+    return next(
+        (idx for idx, c_geom in country_geoms.items() if not geom.disjoint(c_geom)),
+        default_name,
+    )
```
ekatef: Not sure it became more clear :( Indeed, an initial nested loop isn't the most elegant approach. But I fear re-formulating if as a nested loop in functional programming style might be a bit harsh towards a code reader

Wouldn't it be more clear to keep one explicit loop?

ekatef (commented):
@davide-f, amazing work! It looks like a massive improvement, and my feeling is that the code is became much cleaner and clearer as before. 

Have added a few (rather minor) comments

scripts/clean_osm_data.py
davide-f: Moved to the argument, although future improvements may revise this significantly with a fix by country

scripts/clean_osm_data.py
davide-f: removed

scripts/clean_osm_data.py
davide-f: Done

scripts/clean_osm_data.py
davide-f: May you elaborate the question?

So, also in the previous approach, the important columns where frequency, voltage and circuits.
The cables column was used to understand the circuits value when missing.
Actually, also in the old code, the cables column was dropped at the end of the code.

The focus on matching the number of elements of voltage-frequency-circuits is needed to successfully use the explode function that explodes the rows similarly to how split_cells_multiple was doing.
However, by using the native explode function, the efficiency of calculation is significantly increased.

The current approach has instead increased the data availability, future validation shall verify if the data enhancement may be actually too large and if so understand what to revise

scripts/clean_osm_data.py
davide-f: This approach was proposed to increase the efficiency of the code and reduce the overhead.
The function next exactly replace the previous approach exploiting native python activities that are expected to be highly efficient.

However, after this check: https://realpython.com/python-first-match/
The increase in performances is negligible. I revert it back.
Thanks!

scripts/clean_osm_data.py
ekatef: Ok, thank you!

scripts/clean_osm_data.py
ekatef: Super :)

scripts/clean_osm_data.py
ekatef: Thanks a lot for the explanation! My feeling is that the current approach with exploding list-columns looks clean and consistent

Yes, I have meant restoring the circuits value by the cables data. It looks like we don't use it anymore and I was wondering if it was an intentional omission. But if you feel that the data availability might get even too much, than neglecting with cables-to-circuits transformation is obviously not a problem at all :)

scripts/clean_osm_data.py
ekatef: Thanks a lot for the explanation!

ekatef (approved):
Thanks a lot for the explanations, Davide!
Awesome work :)

scripts/clean_osm_data.py
davide-f: The cables&circuit management is addressed in a subsequent fuction indeed.
This function aims at fixing the frequency and voltage frequency first

davide-f:
Thanks a lot @ekatef @pz-max 
Merging

PR #632, opened 2023-03-13
Fix islanded buses and add clustering by networks - clean branch

ekatef:
A cleaned version of #614 

## Checklist

### DIFF
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -62,8 +62,8 @@ cluster_options:
     exclude_carriers: []
     remove_stubs: true
     remove_stubs_across_borders: true
-    p_threshold_drop_isolated: 10 # [MW] isolated buses are being discarded if bus mean power is below the specified threshold 
-    p_threshold_merge_isolated: 300 # [MW] isolated buses are being merged into a single isolated bus if bus mean power is above the specified threshold 
+    p_threshold_drop_isolated: 10 # [MW] isolated buses are being discarded if bus mean power is below the specified threshold
+    p_threshold_merge_isolated: 300 # [MW] isolated buses are being merged into a single isolated bus if bus mean power is above the specified threshold
   cluster_network:
     algorithm: kmeans
     feature: solar+onwind-time
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -62,6 +62,8 @@ cluster_options:
     exclude_carriers: []
     remove_stubs: true
     remove_stubs_across_borders: true
+    p_threshold_drop_isolated: 10 # [MW] isolated buses are being discarded if bus mean power is below the specified threshold 
+    p_threshold_merge_isolated: 300 # [MW] isolated buses are being merged into a single isolated bus if bus mean power is above the specified threshold 
   cluster_network:
     algorithm: kmeans
     feature: solar+onwind-time
@@ -91,6 +93,7 @@ build_shape_options:
   worldpop_method: "standard"  # "standard" pulls from web 1kmx1km raster, "api" pulls from API 100mx100m raster, false (not "false") no pop addition to shape which is useful when generating only cutout
   gdp_method: "standard"  # "standard" pulls from web 1x1km raster, false (not "false") no gdp addition to shape which useful when generating only cutout
   contended_flag: "set_by_country" # "set_by_country" assigns the contended areas to the countries according to the GADM database, "drop" drops these contended areas from the model
+
 clean_osm_data_options:  # osm = OpenStreetMap
   names_by_shapes: true  # Set the country name based on the extended country shapes
   threshold_voltage: 35000  # [V] assets below that voltage threshold will not be used (cable, line, generator, etc.)
@@ -103,6 +106,7 @@ build_osm_network:  # Options of the build_osm_network script; osm = OpenStreetM
   group_tolerance_buses: 500  # [m] (default 500) Tolerance in meters of the close buses to merge
   split_overpassing_lines: true  # When True, lines overpassing buses are splitted and connected to the bueses
   overpassing_lines_tolerance: 1  # [m] (default 1) Tolerance to identify lines overpassing buses
+  force_ac: false  # When true, it forces all components (lines and substation) to be AC-only. To be used if DC assets create problem.
 
 base_network:
   min_voltage_substation_offshore: 35000  # [V] minimum voltage of the offshore substations
@@ -168,7 +172,6 @@ lines:
 links:
   p_max_pu: 1.0
   p_nom_max: .inf
-  include_tyndp: true
   under_construction: "zero" # 'zero': set capacity to zero, 'remove': remove, 'keep': with full capacity
 
 transformers:
--- a/scripts/simplify_network.py
+++ b/scripts/simplify_network.py
@@ -521,6 +521,59 @@ def cluster(
 
     return clustering.network, clustering.busmap
 
+def merge_isolated_buses(n, drop_p_threshold=0):
+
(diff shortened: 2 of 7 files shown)


config.default.yaml
```diff
@@ -62,6 +62,8 @@ cluster_options:
     exclude_carriers: []
     remove_stubs: true
     remove_stubs_across_borders: true
+    p_threshold_drop_isolated: 10 # [MW] isolated buses are being discarded if bus mean power is below the specified threshold 
+    p_threshold_merge_isolated: 300 # [MW] isolated buses are being merged into a single isolated bus if bus mean power is above the specified threshold
```
davide-f: yamlint signals trailing spaces; please, few " " are left at the end of the lines. Please check all changes in the configs.

scripts/simplify_network.py
```diff
@@ -423,7 +423,20 @@ def aggregate_to_substations(n, aggregation_strategies=dict(), buses_i=None):
...
+            (~n.buses.duplicated(subset=["sub_network"], keep=False))
+            & (n.buses.carrier == "AC")
+        ].index
+
+        buses_i = list(
+            set(n.buses.index)
+            - set(i_islands)
+            - set(n.generators.bus)
+            - set(n.loads.bus)
+        )
```
davide-f, line 439: Did we have to modify this?

scripts/simplify_network.py
```diff
@@ -447,6 +460,10 @@ def aggregate_to_substations(n, aggregation_strategies=dict(), buses_i=None):
         incountry_b = n.buses.country == c
         dist.loc[incountry_b, ~incountry_b] = np.inf
 
+    for c in n.buses.carrier.unique():
+        incarrier_b = n.buses.carrier == c
+        dist.loc[incarrier_b, ~incarrier_b] = np.inf
```
davide-f, line 467: Please add a comment why this is needed

davide-f (commented):
Very nice PR :)
temporary comments in the meantime :)
Let us know when ready

config.default.yaml
ekatef: thanks a lot! fixed

scripts/simplify_network.py
ekatef: Yes: we had to exclude isolated buses from treating with `aggregate_to_substations()` to resolves troubles which arise due to using inf values. Setting `inf` is used in `aggregate_to_substations()` to properly assign buses in a busmap. But in case of isolated buses, the whole string consists of `inf` and `idxmin()` picks an arbitrary bus

ekatef:
Thanks a lot! :)

Wouldn't it be more convenient to move back to #614? I have been experimenting a bit with git history and #614 has been cleaned-up as well. Although, checking results are better here :) Let me know please your preferences

Currently there are some problems with reproducibility of the workflow. Not sure what exactly is the reason

ekatef:
Mmmm... 
It looks like we have a new task with better data: more connections between buses are captured now. Which leads to some troubles when clustering:

### simplified network

image

### clustered network
(with minimum 28 buses)

image

Note please that the each of the seemingly isolated buses in `elec_s.nc` now has some company. The first my idea is to drop all the isolated buses remained after clustering. Or, probably, to remove all the isolated pair/groups of buses earlier, in `simplify_network`?

ekatef:
Update: the above are effects of some specific config settings. The workflow and topology model seems to be quite sensitive to it. Probably, some kind of visual guide would be helpful to understand the model behaviour better.

By proper config setting, Testing results for Central Asia look quite appropriate:

### kmeans (by load)

image

### hac (by load)

image

### hac (by gdp)

image

### gadm

image
 
What is really surprising, hac algorithm doesn't require to decrease a number of clusters to a very few in this config setup. Not sure it's good news as I was not able to understand which exactly config changes had such an effect

scripts/simplify_network.py
ekatef: Done

ekatef:
@davide-f, I think this PR is ready. 

Testing on CD as well as CD + ZM was successful for all clustering types. I think we may assume that the introduced changes work properly.

However, there are some pitfalls along the workflow which can make difficult to reproduce clustering results. In particular, I'd keep in sight using `inf` values in `aggregate_to_substations()`: 

https://github.com/pypsa-meets-earth/pypsa-earth/blob/6f0192178917020c004bb7273326b2b1f158e116/scripts/simplify_network.py#L442-L451

It can lead to the errors which are difficult to catch as idxmin() being applied to the row of inf gives a pretty arbitrary result. Being overlayed with variations in output of clustering algorithms and the overall complexity of our workflow, it can give really fascinating consequences :D

ekatef:
Sorry for messing-up the branch history :( Haven't meant it, honestly!

scripts/simplify_network.py
```diff
@@ -620,6 +793,20 @@ def consense(x):
 
     update_p_nom_max(n)
 
+    p_threshold_drop_isolated = max(
+        0.0, cluster_config.get("p_threshold_drop_isolated", 0.0)
+    )
+    p_threshold_merge_isolated = cluster_config.get("p_threshold_merge_isolated", 0.0)
```
davide-f: In this case, let's have False as adefault: that means this merge is not executed.

davide-f (approved):
A little comment, please modify but this PR looks good to me :)

@pz-max do you want to take a look or shall we go?
This is a quite heavy PR.

scripts/simplify_network.py
```diff
@@ -577,7 +740,7 @@ def consense(x):
     # treatment of outliers (nodes without a profile for considered carrier):
     # all nodes that have no profile of the given carrier are being aggregated to closest neighbor
     if (
-        snakemake.config.get("clustering", {})
+        snakemake.config.get("cluster_options", {})
```
davide-f, line 743: Before the object cluster_config is already defined that is equivalent to this. May be worth using that variable

pz-max:
> A little comment, please modify but this PR looks good to me :)
> 
> @pz-max do you want to take a look or shall we go? This is a quite heavy PR.

@davide-f no need to review. Trusting you both to do the best :1st_place_medal:

scripts/simplify_network.py
ekatef: Done

scripts/simplify_network.py
ekatef: `cluster_config` includes only `simplify_network` part:

https://github.com/pypsa-meets-earth/pypsa-earth/blob/6f0192178917020c004bb7273326b2b1f158e116/scripts/simplify_network.py#L562

While we are interested in the another part, namely `cluster_network`:

https://github.com/pypsa-meets-earth/pypsa-earth/blob/6f0192178917020c004bb7273326b2b1f158e116/config.default.yaml#L57-L65

Although, agree that linking to the config could be done in a more clear way...

davide-f: No no it's ok, I was just noting that few lines earlier, there is the definition cluster_options = snakemake.config.get("cluster_options", {}) that matches that line.
So we could use directly the dictionary cluster_options.

ekatef: I'm able to find only `cluster_config = snakemake.config["cluster_options"]["simplify_network"]` :( But that is not exaclty what we need...
Do I miss something?

davide-f: Right! rtm :)

davide-f:
Merged, many thanks @ekatef !

ekatef:
Super :) Thank you so much, @davide-f!

Issue #634, opened 2023-03-15
Reduce computational resources needed by build_shapes: precompiled population and gdp data for gadm shapes


davide-f:
## Describe the feature you'd like to see

The computational resources needed for the calculation of population and gdp in our workflow is very high and we may need procedures to break it down.
The activity is very RAM intensive as it requires raster analysis on images that may be very large for big countries such as China.
Possible improvements are:
- [ ] Revision of the workflow with less RAM-/CPU-intensive procedures
- [ ] Use pre-compiled CSV files to be added in the pre-compiled bundle-configs, which stores population and gdp data for every gadm shape. The workflow, when finds this csv would use the information instead of executing the raster analysis. This may be the recommended approach.

ekatef:
Hello @davide-f, could you please check if the issue is still relevant? 🙏🏽

davide-f:
I'm fine with closing, it is super low priority until something new comes out

Issue #336, opened 2022-05-11
Explore speed benefit from replacing "rasterio" with "xagg" -> raster2polygon operations


pz-max:
[xagg](https://github.com/ks905383/xagg) promises to be faster than rasterio when summing raster data (eg. 10x10km) of polygon shapes. An interesting task would be to explore the performance benefits since these operations are quite time-consuming.

One example where this could be tested: https://github.com/pypsa-meets-africa/pypsa-africa/blob/1fd74e703d507a0105eaca89981efffa6ea021db/scripts/build_shapes.py#L454-L470

If the test is positive we might also enhance the performance of [Atlite](https://github.com/PyPSA/atlite) accordingly which uses this type of function to sum a renewable raster into polygons & also use it to exclude rasters (e.g. a protected area should reduce the sum)

davide-f:
With PR #650 this PR can be closed. Feel free to reopen it if deemed useful

PR #650, opened 2023-03-22
Optimize build shapes add_population_data

GridGrapher:

Introduce new methods to compute the population data
Improves runtime significantly of the `add_population_data` method

A quick overview of the proposed changes

#### Original
`add_population_data` -> (For each country) -> (For each geometry) -> Mask the geometry over the .tif file and calculate the population by summing the values.

#### This PR
`add_population_data` -> (For each country) -> [Iterate over windows if full image can't be loaded into memory] -> 
Get values with their coordinates from the .tif file -> 
Generate an array filled with GADM_ID for each coordinate ->
Sum values into "pop" using the corresponding GADM_ID

## TODO

- [x] Scale the method for large countries:
  - [x] Implement windows into process
  - [ ] ~~Change country_geomask to dict (Additional memory savings)~~
  - [x] Test rasterio.geomask for multipolygons out of bounds and slight overlap

## Checklist

- [x] I tested my contribution locally and it seems to work fine. (Completed testing on the tutorial workflow)


davide-f:
Hello @GridGrapher and nice tackling!
Sorry for the absense, I know you discussed with Max.
I'll add some comments to clarify.

This issue you are addressing is quite a nice have and there may be multiple ways to handle and let's be on the same page.

Please, in this PR on github (not in the code) add a description of what you are aiming to achieve, also to guide the revision.

scripts/build_shapes.py
```diff
@@ -705,6 +726,259 @@ def _process_func_download_pop(c_code):
...
+    # Set an empty numpy array with the dimensions of the country .tif file
+    # np_map_ID will contain a ID for each location (undefined is 0)
+    # ID corresponds to a specific geometry in country_rows
+    np_map_ID = np.zeros((y_axis_len, x_axis_len))
+
+    # List to contain the mappings of id to GADM_ID
+    id_to_GADM_ID = []
+
+    # Loop the country_rows geoDataFrame
+    for i in range(len(country_rows)):
```
davide-f, line 849: With this new revision we currently loose the paralleling capabilities that were available before.
If this new approach applied in large scale is really fast that may be ok, but my feeling is that we should preserve this feature for the final implementation

scripts/build_shapes.py
```diff
@@ -705,6 +726,259 @@ def _process_func_download_pop(c_code):
...
+        # Set the current geometry
+        cur_geometry = country_rows.iloc[i]["geometry"]
+        # Generate a mask for the specific geometry
+        temp_mask = rasterio.features.geometry_mask(
```
davide-f, line 866: I see that the main revision proposed by this PR is the use of geometry_mask instead of the mask function, used previously. Am I right?
A brief description of the goal of this PR in github would highly help.

I was wondering if there might be a more code-efficient way to implement the same features with relatively little changes with respect to the current status.

Given the work that you have done, my feeling (maybe wrong though) could be that a revision in _process_func_pop and _sum_raster_over_mask with this new geometry_mask-based approach could be more efficient and easier to maintain, what do you think?

scripts/build_shapes.py
```diff
@@ -281,7 +302,7 @@ def load_EEZ(countries_codes, geo_crs, EEZ_gpkg="./data/eez/eez_v11.gpkg"):
             f"File EEZ {EEZ_gpkg} not found, please download it from https://www.marineregions.org/download_file.php?name=World_EEZ_v11_20191118_gpkg.zip and copy it in {os.path.dirname(EEZ_gpkg)}"
         )
 
-    geodf_EEZ = gpd.read_file(EEZ_gpkg).to_crs(geo_crs)
+    geodf_EEZ = gpd.read_file(EEZ_gpkg, engine="pyogrio").to_crs(geo_crs)
```
davide-f, line 332: Could be nice to use the old engine while testing this feature.
By doing this, you can test the algorithm that you are developing and not the algorithm+a new engine.

davide-f (commented):
Hello @GridGrapher :) some comments as promised

davide-f:
To add on this; I think it is important to test these algorithms also on large countries, e.g. US and test the old vs new approach.
Please report here on github the status and a comparison of the computational times.

Good idea and we look forward for the final version :D

GridGrapher:
@davide-f I added more context to the PR description.

I've tested this method on 'CN' and the tutorial run with the following runtimes:
`python -m cProfile -o build_shapes.prof ./scripts/build_shapes.py`

Tutorial settings - 53.8 seconds
['CN'] as countries_list - 487 seconds (8 minutes and 7 seconds) [3 GB memory limit]

I'll just add that these times are with the files already downloaded.

davide-f:
These numbers are amazing! very interesting!
Do you think we can gain something by parallelizing some major for loops or it is not-needed overhead?
I fear that when we go earth-scale that 8minutes can easily become hours.

Could you also please answer to the comments above?

scripts/build_shapes.py
GridGrapher: The main revision is to not run rasterio mask for every geometry, rather sequentially window the image so its less memory intensive and computational faster.

I believe this approach to be the most code efficient and easier to maintain.

scripts/build_shapes.py
GridGrapher: Its possible to introduce parallel processing, however at a different stage then before, processing the countries in parallel rather then the image.

scripts/build_shapes.py
GridGrapher: Why would I wait 30 seconds more each time I run the file when I know of a solution?

GridGrapher:
I can try and profile the runtime for all countries, do you have a `config.yaml` file for this?
I would just need the `countries` and `build_shape_options` parts.

I'll give parallel processing a go, as it is one of the largest potential speedups remaining.

The only other one that I am considering is using `numba` to speed up a loop, which has the potential to save 10-60 seconds per country for the current version (from profiling 'US' and the tutorial ['NG','BJ'] ).

davide-f:
To run it for Earth, you just need to specify `countries: ["Earth"]` but be aware that it takes long.
A test on US or a large area like north America should be enough ["NorthAmerica"]

GridGrapher:
Running `python -m cProfile -o build_shapes.prof ./scripts/build_shapes.py` with `["NorthAmerica"]` takes 1520 seconds (25 minutes and 20 seconds) without implementing parallel processing for now.

I also saw that this PR would address the first task in issue https://github.com/pypsa-meets-earth/pypsa-earth/issues/634
Specifically `worldpop_byte_limit` https://github.com/pypsa-meets-earth/pypsa-earth/blob/2c9833810d0fc8ad5c39ef665acc4ce658323252/scripts/build_shapes.py#L894 determines the window size that will be read out by `src.read(1, window=current_window)` https://github.com/pypsa-meets-earth/pypsa-earth/blob/2c9833810d0fc8ad5c39ef665acc4ce658323252/scripts/build_shapes.py#L763 hence can be adjusted or set to a reasonable number to allow parallel processing.
For this parallel processing I'm thinking setting it to 3GB or 2GB which limits the readout per process (so nProcess=2 would be 2x(2GB)).
This specifically targets the window size, so there is still some overhead required per process.

GridGrapher:
I ran the model using ["earth"] without 'NF' and 'PN' due to worldpop files not existing for those.
`python -m cProfile -o build_shapes.prof ./scripts/build_shapes.py`

Which took 4900 seconds (~ 1h 22m)

Earth_ADM_1

The main task remaining is parallel processing, if required, which should speed up the task to sub 1h.

davide-f:
Amazing! This PR is very promising, I go for another review :)
How complex do you think paralleling may improve?

Going below 1h could be awesome.

scripts/build_shapes.py
```diff
@@ -153,33 +179,36 @@ def get_GADM_layer(
     geodf_list = []
 
     for country_code in country_list:
+        # Set the current layer id (cur_layer_id) to global layer_id
+        cur_layer_id = layer_id
+
```
davide-f: I've seen you implemented these changes also here; may the other PR be redundant?
Are the same commits with git cherry-pick/merge or did you reimplement them in the new PR?

scripts/build_shapes.py
davide-f: Sure, I'm fine with that, but I'd avoid addind a new package to the environment.
I think it would be needed if that reduces the computation by 20% for large instances

scripts/build_shapes.py
```diff
@@ -705,6 +734,340 @@ def _process_func_download_pop(c_code):
...
+    with rasterio.open(WorldPop_inputfile) as src:
+        return src.meta["transform"], src.shape
+
+
+def get_worldpop_val_xy(WorldPop_inputfile, window_dimensions):
+    """
+    Function
+
+
+
```
davide-f: Missing docstring

scripts/build_shapes.py
```diff
@@ -705,6 +734,340 @@ def _process_func_download_pop(c_code):
...
+def compute_geomask_region(
+    country_rows, affine_transform, window_dimensions, windowed=False
+):
+    """
+    Function
+
+
+
+    """
```
davide-f: docstring

scripts/build_shapes.py
davide-f: Using Earth where are the bottlenecks?
What fraction of the time is spent on the for loop below and how much on the reading?

scripts/build_shapes.py
davide-f: Agree

scripts/build_shapes.py
```diff
@@ -705,6 +734,340 @@ def _process_func_download_pop(c_code):
...
+
+    worldpop_y_dim, worldpop_x_dim = worldpop_dim
+
+    # Rasterio doesn't support lower than float32 readout
+    # Hence reading the file will take up: nbytes = 4 * y_dim * x_dim
+    expected_bytes_input_read = 4 * worldpop_y_dim * worldpop_x_dim
+
+    # Introduce a max byte size to avoid overfilling RAM
+    # Ensure worldpop_byte_limit > 883 * 10**6 (minimum memory for 'US')
+    worldpop_byte_limit = 3096 * 10**6
```
davide-f: This parameter can be read from the snakemake parameters. snakemake.resources["mem_mb"] and passed to this function as an optional parameter.
When using large computers, we may increase the available ram to speed up computations

scripts/build_shapes.py
```diff
@@ -705,6 +734,340 @@ def _process_func_download_pop(c_code):
...
+
+    return df_pop_count
+
+
+def windowed_compute_population(country_rows, WorldPop_inputfile, worldpop_byte_limit):
+    """
+    Function
+
+
+
```
davide-f: docstring

scripts/build_shapes.py
```diff
@@ -705,6 +734,340 @@ def _process_func_download_pop(c_code):
...
+    return df_pop_count
+
+
+def sum_values_using_geomask(np_pop_val, np_pop_xy, region_geomask, id_mapping):
+    """
+    Function
+
+
+
```
davide-f: docstring

scripts/build_shapes.py
```diff
@@ -705,6 +734,340 @@ def _process_func_download_pop(c_code):
...
+@njit
+def loop_and_extact_val_x_y(
+    np_pop_count, np_pop_val, np_pop_xy, region_geomask, dict_id
+):
+    """
+    Function
+
+
+
```
davide-f: docstirng

davide-f (commented):
Did you compare the total population obtained by the previous and new method?
Do they match?

scripts/build_shapes.py
GridGrapher: This is a visualization of the .prof file for the previous run as shown in the comment above (https://github.com/pypsa-meets-earth/pypsa-earth/pull/650#issuecomment-1493454118).

build_shapes_world

The speed up for file reading is clearly more noticeable if you are working with a few countries vs a continent or the world.

GridGrapher:
When I did some quick testing it was about 1% difference, mainly the code at the moment does not add values that aren't directly on a shape/geometry. The clearest case of this is the coastline, where the worldpop file has pixels over the water.

The solution I propose is to search nearby and take the ID with highest frequency, i.e. take a 3x3 grid around the point and count unique values, pick the most common one. This shouldn't add much overhead.

Another relevant setting in the code:
https://github.com/pypsa-meets-earth/pypsa-earth/blob/7524b98f8fe662af90325b063b154fa6112cff39/scripts/build_shapes.py#L857
I only mention it for clarity, as it may result in overlapping pixels, but I haven't done detailed analysis here.

Overall I assumed that within ~1% accuracy was worth it for the speed up, along with potential for fine tuning it once the runtime is where we want it.

scripts/build_shapes.py
GridGrapher: I implemented them here as it was prudent to test the model, the other PR (https://github.com/pypsa-meets-earth/pypsa-earth/pull/663) has been adjusted to refer to this commit.

scripts/build_shapes.py
GridGrapher: I agree it makes sense to take this from the snakemake file, just one consideration is how to divide the allocated memory effectively once we introduce more cores.

There is a minimum amount of memory required for loading the largest window. The largest window in our dataset is the 'US' as it spans the entire globe and as such has a minimum of 883 MB.
In addition, the `worldpop_byte_limit` targets the most memory intensive operation (src.read(1, ...)) but the whole code has overhead for storing the other arrays, hence I don't want to dilute `mem_mb` if its intended as a concrete ceiling.

It's possible to test and find the size of the overhead, but conservatively I estimated 1GB, so theoretically 4GB per process with `worldpop_byte_limit = 3096 * 10**6`.

A simple approach would be:

`worldpop_byte_limit = max(883, (mem_mb / nProcess) - 1000)`

Obviously this would cause problems when nProcess is large as it would likely result in memory overflow (assuming the computer doesn't have 2GB * nProcess of RAM available). What do you think @davide-f ?

davide-f: Thanks for the comment!
So, I agree that the memory in the snakefile shall be the top required by all tasks.
That number can be changed anyway. 800mb can be increased.
Moreover for simplicity, I would span them uniformly across different process. No need for the -1000 overhead.

why does the 883Mb shall be set?
Can't we simply state: worldpop_byte_limit = mem_mb / nProcess
?

davide-f:
Thanks @GridGrapher for the update and nice comparison.
<1% error is acceptable I think; some dots are counted twice and some are dropped, but it is acceptable as the borders are small and population beyond the shapes is limited.
Uncertainties in the modelling are larger than that, so I believe it is acceptable, good testing! :)

I did a world comparison using the old version and it will be nice to compare it later on with this PR incorporated.

scripts/build_shapes.py
davide-f: @GridGrapher I've seen you are finalizing the PR :) Do you have any comment here?

davide-f:
@GridGrapher I was wondering the status of the PR :)
As comments I'd have:
- the docstring of loop_and_extact_val_x_y is partial
- this PR is very nice, but also very long, I will review it further as it would be nice to shorten the code to make it more maintenable.

Do you have ideas as well? Feel free to comment.
At first hand, on this point of view, I'm wondering:
- Do we need get_worldpop_features as a function? It is used only once, maybe a comment on it is enough
- There is the double option windowed/non-windowed approach: what would be the RAM requirement to run Russia? (it seems the largest country).
   I was wondering if we can keep only one approach instead of two

davide-f:
Moreover, @GridGrapher please, resolve the conflicts in build_shapes

scripts/build_shapes.py
```diff
@@ -698,13 +709,394 @@ def _process_func_pop(gadm_idxs):
...
+    worldpop_y_dim, worldpop_x_dim = worldpop_dim
+
+    # Rasterio doesn't support lower than float32 readout
+    # Hence reading the file will take up: nbytes = 4 * y_dim * x_dim
+    expected_bytes_input_read = 4 * worldpop_y_dim * worldpop_x_dim
+
+    # Introduce a max byte size to avoid overfilling RAM
+    # Ensure worldpop_byte_limit > 883 * 10**6 (minimum memory for 'US')
+    worldpop_byte_limit = 1024 * 10**6
```
davide-f: This value shall be a config parameter or an argument of the function

davide-f:
@GridGrapher please, could you fix the conflicts in build_shapes? I think this PR is ready to be tested with the CI as well

GridGrapher:
Just a short update, 
parallel processing has sped up the program but there is at least one country of the 191 that errors out on `compute_geomask_region`, specifically `return ... pd.DataFrame(id_to_GADM_ID).set_index(0)` with `set_index(0)` being the issue (during ADM_1 run).

So I don't know the entire runtime yet, I'll try and find the country to reproduce the bug but multiprocessing makes it trickier to track down.

Once this has been resolved the PR is in a good state functionally, I'll then go over and refactor for readability.

Note: when trying to run `-m cProfile` with multiprocessing it gives a picklable error, hence I now use viztracer for profiling

scripts/build_shapes.py
GridGrapher: Now incorporated with 418207e28d023382afe3c21460e08c9595e2d822

scripts/build_shapes.py
GridGrapher: Implemented in 418207e28d023382afe3c21460e08c9595e2d822

GridGrapher:
The rerunning the previous workflow for 'Earth' took 50 minutes and 11seconds. with the following settings:
`mem_mb=3096, nprocess=2, with ADM_1`.

I've restructured the methods slightly so it now parallelizes computing windows instead of countries, this produced notable improvements when running `'US' with ADM_2` as my benchmark.

All that's left in my opinion is updating the documentation then it should be good to merge.

I would also propose that this PR closes these issues related to performance of `build_shapes`
https://github.com/pypsa-meets-earth/pypsa-earth/issues/634
https://github.com/pypsa-meets-earth/pypsa-earth/issues/336

scripts/build_shapes.py
GridGrapher: Parallel capabilities are re-implemented in add_population_data

scripts/build_shapes.py
```diff
@@ -737,72 +1092,85 @@ def add_population_data(
...
+            for df_pop_count in pool.imap_unordered(
+                process_function_population, range(len(df_tasks))
+            ):
+                # Acquire the lock before accessing df_gadm
+                with lock:
+                    # Loop the regions and write population to df_gadm
+                    for i in range(len(df_pop_count)):
+                        gadm_id, pop_count = df_pop_count.iloc[i]
+                        # Select the row with the same "GADM_ID" and set the population count
+                        df_gadm.loc[df_gadm["GADM_ID"] == gadm_id, "pop"] += pop_count
```
davide-f, line 1227: Maybe no need for this loop:
if the index of df_pop_count and df_gadm match (I'd expect that if df_pop_count is a subset of df_gadm as it should be originally), we could do something like: df_gadm.loc[df_pop_count.index, "pop"] = df_pop_count["pop"]
What do you think?
The lock may be not necessary as the threads should operate in different memory cells, but it doesn't harm to keep it.
Did you see the need to use it?

davide-f: By revising it further I understand the above it is not possible as the parallelization is not by country.
With this implementation pyomrio is necessary asalso the tiff are opened twice.

More or less, for every thread, what is the average time spent on loading the file and how much on doing the computation for your experience (rough number, no need to repeat any simulation)?

davide-f:
I think that the PR is ready, I'm still concerned by the complexity and length, but you have worked a lot, the improvement is significant and further restyling may be left to other PRs, if needed.

Just few consideration on (1) the comment above and (2) if you see potential improvements of "style" by basically reducing the code and making the whole approach more compact for potential future PRs (if any)

davide-f:
Moreover, please, could you do a validation before/after the PR?
What's the status on that?

davide-f:
Hello @GridGrapher :) May you have any updates? :)

davide-f:
Hello @GridGrapher :)
This PR is highly valuable and I was wondering if we could plan to finalize it.
According to you availability, how do you think we should proceed? Available also for a quick chat if needed.

Tomkourou:
@davide-f what is missing here to merge? I am happy to familiarise myself with the PR and wrap it up.

davide-f:
> @davide-f what is missing here to merge? I am happy to familiarise myself with the PR and wrap it up.

Thanks for the message!
So, here it is missing strong validation by country: checking if the new implementation leads to the same results as before, as well as going through the open comments.

This PR is pretty big and I wanted to validate it as well, but I had scarse time, if you wish to help me that would be amazing!

This PR is needed for the v1.0. It would be merged in a new major release for sure.
I'm considering creating a minor release very soon, yet it wont be joining that.

davide-f:
So, I've been working here a bit for testing and validating the approach, also revising documentation etc.
@GridGrapher really amazing work!!!! Any suggestion you may have is definitely welcome!
Sorry if I jumped in, but I understood you were busy and this validation required some time.

In the following, I provide some statistics.

## Quality of results

In the following, we show the cumulative % error of all UN countries, calculated as (pop - pop_worldbank) / pop_worldbank.
The plot shows that limited errors apply.
Moreover, further investigating the tails, absolute pc errors greater than 20% apply for relatively small countries.
Errors beyond 20% are related to 20 countries whose total population is about 25 million people.
Since these data are obtained from raster files and given the current formulation, I believe that for the moment is is more than enough.
Results are also better than old ones.
image

## Computational requirements

Expected computational requirements are drastically reduced to less than 1h with total RAM compatible with a normal laptop for Earth scale.

## Comments

This PR is definitely ready to go after adding a release note
To organize the versions of the repository, first open PR on feature shall be closed and than this one can definitely follow.
Really amazing contribution @GridGrapher !!!

PR #679, opened 2023-04-11
Gitstar

pz-max:
# Closes # (if applicable).
Added token for Gitstar embedding


pz-max:
Seems not to work really. Let's check in a couple of days. Maybe related to the API limit

pz-max:
Closing this at the moment. API key need to be stored differently and not in the README

PR #686, opened 2023-04-16
fix append powerplant bug

yerbol-akhmetov:
Here I propose to change Pandas "append" method with "concat" to fix the bug related with newer versions Pandas. The error was: AttributeError: 'DataFrame' object has no attribute 'append'.

What was before: return ppl.append(add_ppls, sort=False, ignore_index=True, verify_integrity=True)
Proposed changes: return pd.concat([ppl, add_ppls], sort=False, ignore_index=True, verify_integrity=True)

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -86,6 +86,8 @@ Upcoming Release
 
 * Fix GADM naming bug related to level-2 clustering `PR #684 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/684>`__
 
+* Fix append bug in build_powerplants rule `PR #686 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/686>`__
+
 PyPSA-Earth 0.1.0
 =================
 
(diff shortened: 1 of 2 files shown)


ekatef:
@yerbol-akhmetov, thanks a lot for the PR! You are addressing a very important issue there

To give some context, we had for a while pandas deprecation warning for `append` method. Pandas developers aimed to get rid pandas `append` method  to avoid confusion with list `append` method which behaves differently. [A discussion](https://github.com/pandas-dev/pandas/issues/35407) on that in pandas community has resulted in a recommendation to replace `pandas::append` with `pandas::concat`. The recent [pandas 2.0 release](https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html) at **3 April** has removed `append` for both Series and Dataframes. 

It looks like there is a urgent need to check if pandas `append` is used somewhere else in the code. I'll add an issue on that

davide-f:
Many thanks @yerbol-akhmetov !
I think this PR is ready to go :) I accept and merge it

davide-f (approved):

PR #690, opened 2023-04-18
Add-exception-to-build_pp

### LINKED ISSUE #678: Add ImportError to build_powerplants.py for missing openjdk package.

The `build_powerplants` rule requires the openjdk package to work. However, if it is missing in the environment, the exception thrown does not make this clear. I would propose we add an [ImportError](https://docs.python.org/3/library/exceptions.html#ImportError) making this clear. The error arises here:

https://github.com/pypsa-meets-earth/pypsa-earth/blob/6a8871dafcc844703123377520669578999b86f1/scripts/build_powerplants.py#L298-L299

Or would it be better placed directly inside powerplantmatching?

Tomkourou:
# Closes #678.

Hopefully this is the final one!

## Checklist

### DIFF
--- a/scripts/build_powerplants.py
+++ b/scripts/build_powerplants.py
@@ -262,7 +262,17 @@ def replace_natural_gas_fueltype(df):
     return df.Fueltype.where(df.Fueltype != "Natural Gas", df.Technology)
 
 
+def test_java_availability():
+    try:
+        check_output("java -version", stderr=STDOUT, shell=True).decode("utf-8")
+    except OSError:
+        logger.critical(
+            "java not found on path. You need to install openjdk to run build_powerplants.py"
+        )
+
+
 if __name__ == "__main__":
+    test_java_availability()
     if "snakemake" not in globals():
         from _helpers import mock_snakemake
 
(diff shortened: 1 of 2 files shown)


scripts/build_powerplants.py
```diff
@@ -294,6 +296,12 @@ def replace_natural_gas_fueltype(df):
         config["main_query"] = ppl_query
     else:
         config["main_query"] = ""
+    try:
+        check_output("java -version", stderr=STDOUT, shell=True).decode("utf-8")
+    except OSError:
+        logger.critical(
+            "java not found on path. You need to install openjdk to run build_powerplants.py"
+        )
```
davide-f: Thank @Tomkourou!

I think we could encapsulate this into a function, like `def test_java_availability():` and place these lines as a content.

Then the call `test_java_availability()` can be placed  right at the beginning of main at around line 270.

I think we are close to finalize :)

davide-f (changes requested):

Tomkourou:
@davide-f do you want the function to be placed in `_helpers.py` ?

davide-f:
> @davide-f do you want the function to be placed in `_helpers.py` ?

No need as it is a specific problem by powerplantmatching.
You may comment also about adding this feature there as well; they may be happy about it.

davide-f:
Actually, someone noticed the same and proposed a similar fix
https://github.com/PyPSA/powerplantmatching/pull/109/files

Let's see wait the berlin team things, if it is merged into powerplantmatching, maybe it becomes redundant here

davide-f:
I've just checked the PR into ppl and it seems that a similar fix will be available in ppl and it may be redundant here.
It may be worth if you could contribute directly there maybe proposing your approach, maybe first adding to the discussion (make a PR if they suggest them to do so).
What do you think?

pz-max:
Closing as appears to be already solved in ppm https://github.com/PyPSA/powerplantmatching/pull/109/files

Issue #704, opened 2023-05-02
missing column with empty generators dataframe


davide-f:
## Checklist

- [x] I am using the current [`main`](https://github.com/pypsa-meets-earth/pypsa-earth/tree/main) branch or the latest [release](https://github.com/pypsa-meets-earth/pypsa-earth/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-earth` environment](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

When the generators data are empty, most likely problems appear in the aggregation phase.
To reproduce, let's set the country "TV" and drop all isolated nodes (using a high value of the threshold in simplify_network options).
This may be a bug to be highlighted into PyPSA

## Error Message

*If applicable, paste any terminal output to help illustrating your problem.*
*In some cases it may also be useful to share your list of installed packages: `conda list`.*

```bash
INFO:__main__:Removing stubs
Traceback (most recent call last):
  File "/data/davidef/git_world/pypsa-earth/.snakemake/scripts/tmpk7dzdv0z.simplify_network.py", line 746, in <module>
    n, stub_map = remove_stubs(
  File "/data/davidef/git_world/pypsa-earth/.snakemake/scripts/tmpk7dzdv0z.simplify_network.py", line 414, in remove_stubs
    _aggregate_and_move_components(
  File "/data/davidef/git_world/pypsa-earth/.snakemake/scripts/tmpk7dzdv0z.simplify_network.py", line 254, in _aggregate_and_move_components
    generators, generators_pnl = aggregategenerators(
  File "/home/davidef/miniconda3/envs/pypsa-earth/lib/python3.8/site-packages/pypsa/networkclustering.py", line 81, in aggregategenerators
    weighting = generators.weight.groupby(grouper, axis=0).transform(normed_or_uniform)
  File "/home/davidef/miniconda3/envs/pypsa-earth/lib/python3.8/site-packages/pandas/core/generic.py", line 5989, in __getattr__
    return object.__getattribute__(self, name)
AttributeError: 'DataFrame' object has no attribute 'weight'
```

davide-f:
@ekatef Since you have dealt with similar issues and in simplify_network/PyPSA, would you mind taking care of this issue?
I think once found the problem, it should be easy to address, most likely few lines of code.
What do you think?

[...]

PR #709, opened 2023-05-05
Add a fix for missing column with empty generators dataframe

ekatef:
Quick fix for #704 

Added "weight" column to n.generators dataframe to match with a structure expected by `PyPSA::aggregategenerators`

## Checklist

### DIFF
--- a/scripts/add_electricity.py
+++ b/scripts/add_electricity.py
@@ -779,5 +779,8 @@ if __name__ == "__main__":
     update_p_nom_max(n)
     add_nice_carrier_names(n, snakemake.config)
 
+    if not ("weight" in n.generators.columns):
+        n.generators["weight"] = pd.Series()
+
     n.meta = snakemake.config
     n.export_to_netcdf(snakemake.output[0])


davide-f:
Thanks @ekatef for the check and clarification to see where the problem arises.
I'll repeat the test also on TV to see what's going on and eventually

davide-f:
@ekatef Thanks for the check.
With your help, we found that the problem is related to the fact that the cutout for TV is basically empty and that leads no generator to be installed even when the clip values are 0.0.
#709 fixes this and enables TV to run.

Now the question is: shall we also add this fix?
- benefit: the workflow runs also in that case
- disadvantage: do we want the workflow to run with no generators? That would lead to trivial optimizations or errors down the chain.

A proposal may be to merge #709 and #704 so that instead of adding the weight column, a logger error is thrown if the weight column is not available to say that something is wrong and most likely no generators are available.
What do you think?

If you are interested, feel free to do cherry-pick from the other PR and finalize.
If you may have problems with the cherry-pick, I can manually add the changes here. Let me know :)

ekatef:
@davide-f absolutely agree that the error should be thrown in case the simulation is basically empty. However, it's possible that there is no `weight` column but the simulation still makes sense, in particular there is hydro generators (treated as storage) or/and conventional generators.

The question I'd think about is if we want to allow simulations when conventional generators only are in the system. Is it a proper use case, keeping in mind dispatch-only functionality #710?

davide-f:
Yep, then feel free to merge the two PRs or I can do that as well :)

ekatef:
Go ahead please 🙂 
I'll be happy to see it finalised

davide-f:
As discussed, this PR is merged into #714.
I'm closing this PR

PR #732, opened 2023-05-19
fix hardcoded 380kv

yerbol-akhmetov:
Here I propose changes to function "simplify_network_to_380kV" in simplify_network.py file. Currently, it is hard-coded to simplify to 380kV and raises error if there is no 380.0 voltage is reported in electricity section of config file. The error has the following message: "linetype = snakemake.config["lines"]["types"][380.0] KeyError: 380.0". This happens, because for country of my interest (Kazakhstan) there are no 380kV lines, instead we have [110, 220, 500]kV lines. In this pull request, I propose to generalize the simplify_network rule. In particular, the "base voltage" will be selected from voltages listed in electricity section of config file. Currently, I propose to modify "simplify_network_to_380kV" to "simplify_network_to_base_voltage". The base voltage is currently selected as maximum of listed voltage levels. However, this can be further studied and changed in the future. At the moment, it will solve the issue with hard-coded 380kV.

## Checklist


scripts/simplify_network.py
```diff
@@ -110,21 +110,21 @@
...
     removed as well.
     """
-    logger.info("Mapping all network lines onto a single 380kV layer")
+    logger.info(f"Mapping all network lines onto a single {int(base_voltage)}kV layer")
 
-    linetype_380 = linetype
-    n.lines["type"] = linetype_380
-    n.lines["v_nom"] = 380
-    n.lines["i_nom"] = n.line_types.i_nom[linetype_380]
+    linetype_base = linetype
```
pz-max: let's remove line 124. It's not needed (obviously, the next line needs to be fixed in terms of naming)

scripts/simplify_network.py
```diff
@@ -715,7 +715,8 @@ def merge_isolated_nodes(n, threshold, aggregation_strategies=dict()):
 
     n = pypsa.Network(snakemake.input.network)
 
-    linetype = snakemake.config["lines"]["types"][380.0]
+    base_voltage = max(snakemake.config["electricity"]["voltages"])
```
pz-max: Instead of doing this, let's make the base_voltage **explicit** in all configs.

From:
```
electricity:
  voltages: [220., 300., 380.]
```

To:
```
electricity:
  base_voltage: [380.]
  voltages: [220., 300., 380.]
```

So you can read in the script `snakemake.config["electricity"]["base_voltage"]` instead of `max(snakemake.config["electricity"]["voltages"])`

pz-max (changes requested):
@yerbol-akhmetov great contribution! I just had a few comments and change requests. Can you make the changes accordingly?

scripts/simplify_network.py
davide-f: Maybe `base_voltage: 380.`?
A list there may be an overcomplication

scripts/simplify_network.py
yerbol-akhmetov: Ok, I will do that, as you have mentioned very redundant piece.

scripts/simplify_network.py
yerbol-akhmetov: Yes, I thought about adding it to config but did not dare to touch config file. I will make changes.

pz-max: Agreeing with Davide 👍🏽

pz-max:
@yerbol-akhmetov i think what is only missing now is to add some documentation for the config.yaml.

Let me know if you find the relevant .rst file in the doc folder to make the changes

pz-max (approved):
Great :100:

PR #739, opened 2023-05-20
Custom ppl opts

yerbol-akhmetov:
Here I propose an extension of functionality of custom powerplants addition. Currently, there are only two options: 1) Use open-source powerplants with `custom_powerplants: false`, and 2) Attach custom powerplants from `custom_powerplants.csv` to open-source data and use both with `custom_powerplants: true`. There is no option to select only *custom powerplants*. This functionality is very useful. Sometimes you can have an accurate custom powerplants information and you would like to replace retrieved open-source plants with custom data. On the other situations, you might need to use both open-source and custom powerplants. 
Therefore, I propose to extend the functionality to three options: 
1) `false` - no custom power plants are used;
2) `merge` - both open-source and custom powerplants from `custom_powerplants.csv` are merged and used;
3) `replace` - only custom powerplants from `custom_powerplants.csv` are used.

## Checklist

### DIFF
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -142,7 +142,7 @@ electricity:
     Link: []  # H2 pipeline
 
   powerplants_filter: (DateOut >= 2022 or DateOut != DateOut)
-  custom_powerplants: false  # "false" - no custom powerplants, "merge" uses both open-source and custom powerplants, "replace" uses only custom powerplants
+  custom_powerplants: false  #  "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
 
   conventional_carriers: [nuclear, oil, OCGT, CCGT, coal, lignite, geothermal, biomass]
   renewable_carriers: [solar, onwind, offwind-ac, offwind-dc, hydro]
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -142,7 +142,7 @@ electricity:
     Link: []  # H2 pipeline
 
   powerplants_filter: (DateOut >= 2022 or DateOut != DateOut)
-  custom_powerplants: false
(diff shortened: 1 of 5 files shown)


config.default.yaml
```diff
@@ -142,7 +142,7 @@ electricity:
     Link: []  # H2 pipeline
 
   powerplants_filter: (DateOut >= 2022 or DateOut != DateOut)
-  custom_powerplants: false
+  custom_powerplants: false  # "false" - no custom powerplants, "merge" uses both open-source and custom powerplants, "replace" uses only custom powerplants
```
pz-max: Could you change the comment to wherever possible:
`#  "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants`

scripts/build_powerplants.py
```diff
@@ -247,9 +247,12 @@ def add_custom_powerplants(ppl, inputs, config):
     # if isinstance(custom_ppl_query, str):
     #     add_ppls.query(custom_ppl_query, inplace=True)
```
pz-max: I know pypsa-eur is using line 247 and 248. Can you figure out its purpose?
I think your merge and replace option could make lines 247 and 248 not necessary. Maybe we can remove the comment?

pz-max (changes requested):
Nice PR! :100: Some minor changes are required.
Could you also adjust the configuration.rst accordingly?

scripts/build_powerplants.py
yerbol-akhmetov: I have checked the code and it seems that it is left to enable some query to custom powerplants being added. We can leave them as commented or we can remove them.

yerbol-akhmetov: But it is better to remove them, since if enabled it might contradict with the proposed PR.

pz-max (approved):

pz-max (approved):
Looks great. I just had to change the release note position

PR #742, opened 2023-05-23
Add custom renewables

yerbol-akhmetov:
Here, I propose a function `attach_custom_renewables` function similar to `attach_OPSD_renewables` function of PyPSA-Eur. With this functionality, it would be possible to set capacities for renewables from `custom_powerplants.csv`. Currently, the only possible option to set the renewable capacities is using `estimate_renewable_capacities_irena`. However, even if you know exact capacities of existing wind and solar powerplants, currently you cannot set them as p_nom. With the proposed PR, it will automatically calculate the aggregated p_nom for each bus and set them accordingly.

## Checklist

### DIFF
--- a/scripts/add_electricity.py
+++ b/scripts/add_electricity.py
@@ -95,6 +95,7 @@ import xarray as xr
 from _helpers import configure_logging, getContinent, update_p_nom_max
 from shapely.validation import make_valid
 from vresutils import transfer as vtransfer
+from powerplantmatching.export import map_country_bus
 
 idx = pd.IndexSlice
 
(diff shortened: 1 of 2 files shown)


scripts/add_electricity.py
```diff
@@ -622,6 +623,32 @@ def attach_extendable_generators(n, costs, ppl):
             )
 
 
+def attach_custom_renewables(n, tech_map, ppl):
```
martacki, line 626: if the routine should work as for the conventionals, then the renewable custom ppls would be added to the `powerplants.csv` aleady in `build_powerplants.py` and added to the network in the `attach_wind_and_solar()` function here (`add_electricity.py`). This would be more consistent (in terms of treatment of RE and conventionals), but would require additional changes.

I believe the suggested changes are sufficient, however inconsistent with the treatment of conv/RE ppls

yerbol-akhmetov: Custom renewable powerplants are added with `build_powerplants.py` and is stored in `powerplants.csv`.  In `add_electricity.py` conventional powerplants are added from `powerplants.csv` one by one, while RES powerplants are omitted. In practice, it needed to be added by `attach_wind_and_solar()`; however, this function only adds estimated `p_nom_max`, not `p_nom`. This `p_nom` is estimated and added by separate function named `estimate_renewable_capacities_irena`. Here, the proposed `attach_custom_renewables()` will work as alternative to `estimate_renewable_capacities_irena` (thus not changing the main function).

martacki: Yes, I understand. Nevertheless, if done cleanly, we should add it as you suggested in the function `attach_wind_and_solar(ppls)`. The function should be adapted such that it can also add `p_nom`, not only `p_nom_max`.  The custom renewable ppls should be added to `powerplants.csv` using the `add_custom_powerplants` function in build_powerplants. `estimate_renewbale_capacities_from_irena` should then only estimate renewables for countries where there are no renewables.

pz-max:
Close this in favour of #744

PR #743, opened 2023-05-23
Revise network simplification to account for DC lines

### LINKED ISSUE #621: Revision of simplify_links to account for DC lines

Currently, the simplify_links procedure in simplify_network simplifies chains of dc links with an unique link.
We inherited this from pypsa-eur.
Let's assume a DC line is from node A to node B.
Two ficticious nodes (e.g. A_dc and B_dc) are created to represent the DC nodes and 2 converters are added at each node, hence connecting A to A_dc and B to B_dc.
Then the DC line is modelled with a link connecting A_dc to B_dc, hence creting a 3-link chain from A to B.
simplify_links simplify all of this with a unique link, however, if the DC line is (rightfully) modelled as a DC line, the same does not apply and could be reformulated.

However, addressing this is an mid-advanced task, definitely not good for beginners.

ekatef:
# Closes #621

DC lines are added to a simplification procedure to be treated in the same way as links.

## Checklist

    * [x] Philippines
    * [x] Congo Democratic Republic
    * [x] Japan
    * [x] China
    * [x] India
    * [x] Nigeria

**NB** Testing configuration assumed dc_as_lines for all the countries above and gadm clustesing for CD, PH and NG (as gadm clustering seems currently to be more prone to troubles) while usual hierarchical clustering was used for JP, IN and CN to use pre-calculated shapes keeping computational costs moderate.


ekatef:
Current implementation seems to reproduce desirable topology transformations, but should be adjusted in part of related to equivalent parameters calculation of the resulted links

ekatef:
Detailed checking of the network structure for CD has revealed some issues related to simplification of "meshed" DC grid. HVDC part consists of two parallel lines. Geometrically each of the parallel lines connects the same buses but has slightly different coordinates:

image

These two lines get merged in course of simplification, except a single segment where lines duplication is kept:

image
(different colours are used just to highlight different lines of the same DC part)

ekatef:
The suggested modification of `simplify_network` works with `hvdc_as_lines: true` in the same way as `hvdc_as_lines: false`. To check it an "analytical" approach should be used as simple visual check is a bit tricky due to overlapped geometries. A good indicator is `n.lines[n.lines.dc][["bus0", "bus1"]]` vs  `n.links][["bus0", "bus1"]]` which is for a simplified network as follows:

### before modification

```bash
Line,bus0,bus1
49423895_0,30,36
224147292_0,36,51
261165614_0,37,53
383097966_0,37,52
460419487_0,30,36
461638193_0,36,51
626214770_0,44,37
778351007_0,51,8
778351009_0,51,8
778351020_0,44,37
979570585_0,53,52
979570586_0,52,30
979570588_0,52,30

Link,bus0,bus1
convert_3_8,8,7
convert_29_44,44,43
```

### after modification

```bash
Line,bus0,bus1
383097966_0,37,52

Link,bus0,bus1
261165614_0+1,52,37
460419487_0+4,52,7
778351020_0+1,43,37
```

The PR ensures that a result of the network simplification is the same regardsless of `hvdc_as_lines` value with the only difference that for the buses `37,52` a duplicating connection is a (second) link if `hvdc_as_lines: false` and a DC line otherwise

So, the next step is to resolve an issue with the duplication remained in the simplified network

ekatef:
It looks like `split_links` collects chains of links starting from each supernode going along a graph to be further used for merging into a single link. All parallel links (like those between 51 and 8 buses, for example) are being captured by `adj()` function

So, merging `37-53` + `53-52` is in scope of the algorithm, while adding `37-52` seems to need to introduce some modifications into it

davide-f:
Hello @ekatef and very interesting contributions! Much needed.
I understand this is a work in progress, but I plan to give it a look for a preliminary review later.
As a quick comment, what do you mean with 'merging 37-53 + 53-52 is in scope of the algorithm, while adding 37-52 seems to need to introduce some modifications into it'?
Especially the second half of the sentence.

Many thanks! :)

ekatef:
> Hello @ekatef and very interesting contributions! Much needed. I understand this is a work in progress, but I plan to give it a look for a preliminary review later. As a quick comment, what do you mean with 'merging 37-53 + 53-52 is in scope of the algorithm, while adding 37-52 seems to need to introduce some modifications into it'? Especially the second half of the sentence.
> 
> Many thanks! :)

Hello @davide-f and many thanks for support, as usually! :)

The numbers refer to the second picture in the [comment](https://github.com/pypsa-meets-earth/pypsa-earth/pull/743#issuecomment-1595863833) above. I have been thinking that lines duplication left after simplification could be a bug which can be fixed by some simple trick. But it looks some deeper changes are needed if we are not happy with the resulted configuration of a simplified network. The question is if it's really needed :D

ekatef:
As discussed with @davide-f, the best approach to tackle this issue with the parallel links remained after simplification might be just to apply `simplify_links` twice

ekatef:
It looks like we have previously missed some modifications in calculations on transmission costs needed to make representation of HVDC as lines work properly. In particular:
1) `underwater_fraction` should be calculated for n.lines with "DC" carries in `base_network`;
2) when updating transmission costs in `add_electricity`, the parameters should be taken from `n.lines` instead of `n.links`

Have drafted possible implementation of these changes in `base_network` and `add_electricity`. There is some code repetition in both cases as I haven't been able to find a flexible enough way to distinguish between lines and links. 

@davide-f would be very grateful if you could find time to have a look :)

davide-f:
Hello katia :)
I've scanned it and there are quite interesting changes.
It is a big PR, so I may need some time to review entirely.
I'll go thought it in the weekend, I give some brief comments now as a preliminary element, but I need to go better in the code later on.

scripts/add_electricity.py
```diff
@@ -286,6 +273,67 @@ def update_transmission_costs(n, costs, length_factor=1.0, simple_hvdc_costs=Fal
             + costs.at["HVDC inverter pair", "capital_cost"]
         )
     n.links.loc[dc_b, "capital_cost"] = costs
+    return n
+
+
+def attach_dc_costs_lines(n, costs, length_factor=1.0, simple_hvdc_costs=False):
```
davide-f: I really like the approach of combining the two links and lines into the same update_transmission_costs, so that we don't lose retrocompatibility of functionalities with respect to pypsa-earth :)

As a comment, at first glance, it seems that the _lines and _links function differ only by the "n.links" or "n.lines", is it right?
if the code works with that, that's an amazing flow that you found!
Because we could create an utility function such as "attach_dc_costs", general for both links and lines, that instead of taking n as first argument, it may be lines_or_link to represent n.links or n.lines.
In the body of the attach_dc_costs, the "n.links" may be replaced with lines_or_links and it (hopefully) work.

ideally, we could launch the function attach_dc_costs twice in the update_transmission_costs: once for the links e.g. `attach_dc_costs(n.links, ...)` and one for lines `attach_dc_costs(n.lines, ...)`

Within attach_dc_costs, then a check like `lines_or_links.loc[lines_or_links.carrier == "DC"].empty` may be checked to check if there are data to change, if not a simple return may apply similarly to before. Moving this check here, may clean also the code in update_transmisssion_costs that rightfully you did the check externally.

If the above is possible, that would be a dream, but I am unsure that it is feasible.
I think this strong work that you did may answer to this doubt :)

scripts/add_electricity.py
```diff
@@ -286,6 +273,67 @@ def update_transmission_costs(n, costs, length_factor=1.0, simple_hvdc_costs=Fal
...
+                + n.lines.loc[dc_b, "underwater_fraction"]
+                * costs.at["HVDC submarine", "capital_cost"]
+            )
+            + costs.at["HVDC inverter pair", "capital_cost"]
+        )
+    n.lines.loc[dc_b, "capital_cost"] = costs
+    return n
+
+
+def update_transmission_costs(n, costs, length_factor=1.0, simple_hvdc_costs=False):
```
davide-f, line 285: I like this function that condensates both functions :)

scripts/add_electricity.py
```diff
@@ -286,6 +273,67 @@ def update_transmission_costs(n, costs, length_factor=1.0, simple_hvdc_costs=Fal
...
+    # may be missing. Therefore we have to return here.
+    # TODO: Require fix
+    if (
+        n.links.loc[n.links.carrier == "DC"].empty
+        and n.lines.loc[n.lines.carrier == "DC"].empty
+    ):
+        return
+
+    # HVDC implemented as links
+    if not n.links.loc[n.links.carrier == "DC"].empty:
```
davide-f: As an alternative option mentioned above, this check may be moved inside a general attach_dc_costs that may apply (if we are really lucky) to both lines and dc, though not sure if it applies.
I add this comment also here, as the comment above is very long and this hopefully may increase clarity

scripts/base_network.py
```diff
@@ -160,6 +157,33 @@ def _set_links_underwater_fraction(fp_offshore_shapes, n):
...
+def _set_links_underwater_fraction(fp_offshore_shapes, n):
+    # HVDC part always has some links as converters
+    # excluding probably purely DC networks which are currently somewhat exotic
+    if n.links.empty:
+        return
+
+    if not n.links.loc[n.links.carrier == "DC"].empty:
+        add_underwater_links(n, fp_offshore_shapes)
+    elif not n.lines.loc[n.lines.carrier == "DC"].empty:
+        add_underwater_lines(n, fp_offshore_shapes)
```
davide-f: Similarly to the case for the dc lines, do you think it may be possible to generalize the function like "add_underwater" to both lines and links?

scripts/add_electricity.py
```diff
@@ -286,6 +273,67 @@ def update_transmission_costs(n, costs, length_factor=1.0, simple_hvdc_costs=Fal
...
+        attach_dc_costs_links(
```
davide-f: To add clarity, if the general function attach_dc_costs is feasible and theempty check could be moved inside, we could do two lines, like:
`attach_dc_costs(n.lines, ...)`
`attach_dc_costs(n.links, ...)`
but that would be a dream... not sure if it is feasible.
I think your experience here is highly

davide-f (commented):
Amazing job! :D
I've added some comments, on the chaining functionality in simplify_network I recall we talked about that and was ok.
I'll need another round maybe guided to address the functionality, but it was working well , right?
Fantastic Katia! :D

ekatef:
Thank you so much, Davide! There are still some troubles with functionality which I'm going to work on during the weekend. I'll ping you when ready :)

scripts/add_electricity.py
ekatef: Amazing idea! Exactly what I have been looking for :D 
Thank you so much!

scripts/base_network.py
ekatef: A very good idea! My feeling is that it can be very helpful to resolve the current troubles with functionality which are connected with `underwater_fraction` lost somewhere

scripts/add_electricity.py
ekatef: As far as I understand, this fragment simply sums costs for different parts of HVDC grid meaning that there should not be any difference between "as-links" and "as-lines" representation. Correct me please if I'm missing something

Have implemented and it seems to work fine. The "passed-by-assignment" magic is very handy :)

scripts/add_electricity.py
ekatef: Thank you :) Done!

scripts/base_network.py
ekatef: Revised. However, the current representation needs some improvements as currently it assumes that all lines are HVDC if they are located outside the onshore shape

scripts/add_electricity.py
ekatef: Done! :)

ekatef:
A little update (some improvements are still needed in this PR before the next review round, so nothing urgent).

Testing has been done on CD and PH: success with `hvdc_as_lines: false` but some hardcoding is needed to make the workflow run with `hvdc_as_lines: true`.  

Next steps:

- [x] 1. [Equivalence calculation](https://github.com/ekatef/pypsa-earth/blob/bd2bb13257bf10497cfcab0198bd63a302129a99/scripts/simplify_network.py#L411-L421) should be double-checked and revised for "as-lines" representation. It looks like the current implementation leads to some `NaN`s in a pre-solved network

- [ ] 2. Check calculations for the underwater part in `clustering_network`:

https://github.com/pypsa-meets-earth/pypsa-earth/blob/d0d709d51b23516cbe7bb4297cdd14d2b599e0bb/scripts/cluster_network.py#L610-L616

In particular, it might be need to set `underwater_fraction` for DC lines here and double-check that `extended_link_costs` should in fact correspond to `hvac_overhead_cost`.

- [ ] 3. Testing should be done also with alternative clustering and augmentation flags on. Currently the workflow works with alternative clustering but there are some problems if augmentation is on:

    - for CD there is a mixed-carrier problem which likely appears due to DC-line overlaid with DC link during simplification as shown in some comments above;
    - for PH solving breaks due to some NaN values which present in the model

scripts/add_electricity.py
```diff
@@ -252,40 +252,65 @@ def attach_load(n, demand_profiles):
...
-    if n.links.empty:
-        return
+def attach_dc_costs(
+    n, lines_or_links, costs, length_factor=1.0, simple_hvdc_costs=False
+):
+    # TODO would it make sense to add a check on allowed keywords value?
+    if lines_or_links == "lines":
+        df_to_modify = n.lines
+    elif lines_or_links == "links":
+        df_to_modify = n.links
```
davide-f: Hello :)
Nice improvement!
However, I feel like we don't need both n and lines_or_links.
We could have lines_or_links to be the dataframe where the dc costs are modified. To do so, the definition of the function may be:
`def attach_dc_costs(lines_or_links, costs, length_factor=1.0, simple_hvdc_costs=False):`
And what you defined as df_to_modify can be replaced with lines_or_links

And when we want to attach the dc costs on lines, we could simply do:
`attach_dc_costs(n.lines, costs, length_factor, simple_hvdc_costs)`

In the case above, lines_or_links will get the value of n.lines and the code will work on lines.

By doing this, we don't need the "lines" and "links" string parameters to distinguish the two.

Under a similar principle, we could do the underwater.

Not sure if the above is clear, happy to interact if needed :)

scripts/add_electricity.py
```diff
@@ -252,40 +252,65 @@ def attach_load(n, demand_profiles):
...
+    df_to_modify.loc[dc_b, "capital_cost"] = costs
+
+
+def update_transmission_costs(n, costs, length_factor=1.0, simple_hvdc_costs=False):
+    n.lines["capital_cost"] = (
+        n.lines["length"] * length_factor * costs.at["HVAC overhead", "capital_cost"]
+    )
+
+    if n.links.empty:
+        return
```
davide-f: This may not be needed, we could do a check inside attach_dc_costs to verify whether the dataframe (lines_or_links) is empty.
If that is empty, we return.

scripts/base_network.py
```diff
@@ -143,14 +143,11 @@ def _load_buses_from_osm(fp_buses, config):
...
-def _set_links_underwater_fraction(fp_offshore_shapes, n):
-    if n.links.empty:
-        return
-
+def add_underwater_links(n, shapes):
     if not hasattr(n.links, "geometry"):
         n.links["underwater_fraction"] = 0.0
     else:
-        offshore_shape = gpd.read_file(fp_offshore_shapes).unary_union
+        offshore_shape = gpd.read_file(shapes).unary_union
```
davide-f: Why this revision of the name? I feel like "fp_..." is clear as it indicates filepath of offshore shapes, while shapes may be too general

scripts/base_network.py
```diff
@@ -160,6 +157,46 @@ def _set_links_underwater_fraction(fp_offshore_shapes, n):
             )
 
 
+# TODO would be nice to join both the add_underwater* functions
+def add_underwater_part(n, lines_or_links, shapes):
+    # TODO would it make sense to add a check on allowed keywords value?
+    if lines_or_links == "lines":
+        df_to_modify = n.lines
+    elif lines_or_links == "links":
+        df_to_modify = n.links
```
davide-f: That's a very good improvement :) maybe we could simply drop the n and rename df_to_modify to lines_or_links (see comment above).
Probably the functionalities would be preserved, what do you think?

scripts/base_network.py
```diff
@@ -160,6 +157,46 @@ def _set_links_underwater_fraction(fp_offshore_shapes, n):
...
+                branches.intersection(offshore_shape).length
+                / branches.length
+            )
+
+
+def _set_links_underwater_fraction(fp_offshore_shapes, n):
+    # HVDC part always has some links as converters
+    # excluding probably purely DC networks which are currently somewhat exotic
+    if n.links.empty:
+        return
```
davide-f: This check may be done within the add_underwater_part on the lines_or_links, as if DC lines are threated as lines, this check would skip the addition of the underwater on lines

davide-f (commented):
Nice improvements :D we are close to finalize :)
Let me know if the comments are clear and if you need any support in the final finalization :)

scripts/add_electricity.py
ekatef: You comment is crystal clear and very helpful :) Implemented!

scripts/add_electricity.py
ekatef: Absolutely agree. Done!

scripts/base_network.py
ekatef: Sorry, have changed temporary during debug and forgot to restore. Fixed

scripts/base_network.py
ekatef: It will be definitely cleaner and safer :) Revised
Thanks a lot for the suggestion!

scripts/base_network.py
ekatef: As far as I understand, converters are represented as links regardless of DC lines representation. But I absolutely agree that this check should be put inside `add_underwater_part` as we done in `attach_dc_costs`

ekatef:
> Nice improvements :D we are close to finalize :) Let me know if the comments are clear and if you need any support in the final finalization :)

Thanks a lot for the review! You comments are absolutely clear and I very much like the improvements you have suggested :)

My feeling is that there are two points left where I'd be very grateful for you advice:
1. Would be nice to double-check calculations of some parameters. In particular, I'm not sure what is the best way to calculate `p_nom` when transforming DC lines into links [done and can be discussed]
2. My feeling that augmentation should be investigated better as testing has failed for "PH" and "CD" if augmentation is on. Not sure whether this problem has been introduced by this PR. I'll check and update :)

ekatef:
Testing status: seems to work except a case when the augmentation option on:

- [x] in case of "CD" the troubles are linked with mixed-carrier issue (most likely due to lines/links overlap as explained in the first comments) [seems to be fixed, although there is still a mixed-carriers warning in case of expansion with `[HVAC]`];
- [x] in case of "PH" there are some NaNs introduced into the network [fixed]

So, augmentation needs better investigation while otherwise PR is ready for testing.

@Tomkourou if you wish to help with testing, you are absolutely welcome :)

scripts/add_electricity.py
```diff
@@ -252,40 +252,55 @@ def attach_load(n, demand_profiles):
...
+
+
+def update_transmission_costs(n, costs, length_factor=1.0, simple_hvdc_costs=False):
+    n.lines["capital_cost"] = (
+        n.lines["length"] * length_factor * costs.at["HVAC overhead", "capital_cost"]
+    )
+
+    # If there are no "DC" links, then the 'underwater_fraction' column
+    # may be missing. Therefore we have to return here.
+    # TODO: Require fix
```
davide-f: But if there are no DC links should the [..].empty successfully skip the check?

scripts/base_network.py
```diff
@@ -160,6 +157,38 @@ def _set_links_underwater_fraction(fp_offshore_shapes, n):
             )
 
 
+def _set_links_underwater_fraction(lines_or_links, fp_offshore_shapes):
```
davide-f: maybe we could drop the "_links" as the function is now general, what do you think?

scripts/cluster_network.py
```diff
@@ -609,6 +609,10 @@ def clustering_for_n_clusters(
             (nc.links.length - n.links.length).clip(lower=0).mul(extended_link_costs),
             fill_value=0,
         )
+    if not n.lines.loc[n.lines.carrier == "DC"].empty:
+        nc = clustering.network
+
+        nc.lines["underwater_fraction"] = 0
```
davide-f: This can be chained: `clustering.network.lines["underwater_fraction"] = 0`

scripts/simplify_network.py
```diff
@@ -354,37 +375,80 @@ def split_links(nodes):
                 n, busmap, costs, config, connection_costs_per_link, buses
             )
 
-            all_links = [i for _, i in sum(links, [])]
+            # TODO revise a variable name for `links` variable
+            # `links` is a list which contains dc-relevant graph elements
+            all_dc = [i for _, i in sum(links, [])]
+            all_links = list(set(n.links.index).intersection(all_dc))
+            all_dc_lines = list(set(n.lines.index).intersection(all_dc))
```
davide-f: Sorry Katia, my bad, but I don't understand this line: you are making the intersection of the line index with all_dc that are indices of links, right?

I see above that links is an object that comes out of graph analyses, may it be worth to rename it to like all_dc_components or something like that?
May also be worth checking if it contains the expected objects of both links and lines, but I guess you have already checked that.

Amazing! :D

scripts/simplify_network.py
```diff
@@ -285,10 +292,16 @@ def simplify_links(n, costs, config, output, aggregation_strategies=dict()):
             pass
         return n, n.buses.index.to_series()
 
+    dc_as_links = not n.links.loc[n.links.carrier == "DC"].empty
+    dc_as_lines = not n.lines.loc[n.lines.carrier == "DC"].empty
```
davide-f: were you looking for the flag in config maybe? In case, you could pass it as a parameter here, what do you think?

scripts/simplify_network.py
```diff
@@ -354,37 +375,80 @@ def split_links(nodes):
...
+            # HVDC part is represented as "Link" component
+            if dc_as_links:
+                p_max_pu = config["links"].get("p_max_pu", 1.0)
+                lengths = n.links.loc[all_links, "length"]
+                length = sum(
+                    n.links.loc[[i for _, i in l if _ == "Link"], "length"].mean()
+                    for l in links
+                )
+                p_nom = min(
+                    n.links.loc[[i for _, i in l if _ == "Link"], "p_nom"].sum()
```
davide-f: The list `[i for _, i in l if _ == "Link"]` is duplicated, it may be worth to add a variable and use it here, especially if it reduces the number of lines with the pre-commit

scripts/simplify_network.py
```diff
@@ -354,37 +375,80 @@ def split_links(nodes):
...
+                length = lengths.sum() / len(lengths) if len(lengths) > 0 else 0
+                p_nom = (
+                    n.lines.loc[all_dc_lines]
+                    .apply(
+                        lambda x: x["v_nom"] * n.line_types.i_nom[x["type"]],
+                        axis=1,
+                        result_type="reduce",
+                    )
+                    .min()
+                )
```
davide-f: Theoretically, there should be the column s_nom that contains the nominal capacity of the line: is it filled properly?
If so, you may use that column here.

However, you raised an interesting point: in `simplify_network_to_380` we still assume that all lines are AC, it may be interesting to debug what happens there with DC lines.
In particular, the num_parallel is most likely calculated in a wrong way

scripts/simplify_network.py
```diff
@@ -354,37 +375,80 @@ def split_links(nodes):
...
+                underwater_fraction = (
+                    sum(
+                        lengths
+                        / lengths.sum()
+                        * n.lines.loc[all_dc_lines, "underwater_fraction"]
```
davide-f: If this is a series, it may be better to use the .sum():
`(lengths * n.lines.loc[all_dc_lines, "underwater_fraction"]).sum() / lengths.sum()`

In some calculations, we could also exploit the eval function. For example:
`n.lines.loc[all_dc_lines].eval("length * underwater_fraction")"
I'm dropping this information that may help in coding :)

davide-f (commented):
Very nice! I think this PR will be merged by next meeting , we are very close and great contribution! :)

So, is the algorithm finalized in your opinion? @ekatef 

Is the only problem related to CD with augmentation by hvac?

scripts/base_network.py
ekatef: Absolutely agree. Could we probably replace "_links" with "_dc" to make the function purpose clear?

scripts/cluster_network.py
ekatef: Super, thanks! Done :)

scripts/simplify_network.py
ekatef: Yes, absolutely: here I'm trying to split `links` list on `Line` and `Link` components. If it's not clear it means that this fragment should be definitely improved.

Have revised naming changing `all_dc` -> `all_dc_branches` and added a bit to the comments.

Not sure I have get your point regarding the checking which does likely mean that I miss something... Would be very grateful for a short exchange on that :)

ekatef: Indeed, there is [a check](https://github.com/ekatef/pypsa-earth/blob/50b558f091efd8ebb87ef76fef90228382be1d92/scripts/simplify_network.py#L278-L281) above in `contains_ac()`, but I'm not sure that is what you mean

scripts/simplify_network.py
ekatef: Yeah, I have been thinking about taking a flag directly from the config, but decided to try reading it from the network itself. The idea was to make it safer, although agree that it might be not clear. 

Happy to revise if you think it'd be better to used a flag from the config

scripts/simplify_network.py
ekatef: Done :)

scripts/simplify_network.py
ekatef: Mmm, you are absolutely right: there is `s_nom` in `n.lines`, and it's value is exactly 3 times lower as compared with U*I used above. It feels like there is in fact some discrepancy with `num_parallel`

scripts/simplify_network.py
ekatef: Revised. Thanks a lot for the hint! :)

scripts/add_electricity.py
ekatef: Not sure I get your point properly :(
The comment is old, and I'm not quite clear about its meaning. Probably we have just remove it

ekatef:
> Very nice! I think this PR will be merged by next meeting , we are very close and great contribution! :)
> 
> So, is the algorithm finalized in your opinion? @ekatef
> 
> Is the only problem related to CD with augmentation by hvac?

@davide-f thank you for the amazing review! Very heplful :D

My feeling is that we are in fact quite close to finalise. From my perspective, augmentation is a major problem left. Apart of that, I'm also concerned with the recent testing results. 

All the tested countries work with `dc_as_lines: true` and the results up to `simplify_network` looks quite nice. But clustering outputs are sometimes a bit puzzling. In particular, the usual hierarchical algorithm sometimes complains that `n_cluster` is too big and agrees to work only if `n_cluster` is decreased from 10 to only 2. Obviously, that leads to some strange outputs, for example:

## base network

image

## simplified network

image

## clustered network

image

ekatef:
Clustering results might be generally not very stable in respect to parameters selection like to remove stubs or not to remove :) But it would be a good idea to double-check that the clustering issues above are not being introduced with this PR. On the other hand, clustering problems were found for CN and IN only where old pre-calculated shapes were used. Probably, that has also played a role...

@davide-f I do remember that you have suggested to account for generators/loads attached to the nodes when making simplification. At the moment the algorithm doesn't do for that, while my feeling is that it could be worth to improve this

scripts/simplify_network.py
davide-f: My concern was mainly in the link list, doesn't it contains also line branches? I feel like its name is misleading, what do you think?

scripts/add_electricity.py
davide-f: In the function, you included a check to skip the analyses if no DC assets are available, the "lines_or_links.loc[lines_or_links.carrier == "DC"].empty" if that when true the function attach_dc_costs returns with no computation.
I feel like these lines can be safely removed

scripts/simplify_network.py
davide-f: Mah... agree that we could avoid overcomplicating the code.
We could do the check internally, however, it may be safer to check for dc lines: 
`dc_as_links = not (n.lines.carrier == "DC").any()`
and implicitly assume that when it is false than we used lines

As even when lines are included, converters represented as links will be necessary as well, therefore, with the current formulation dc_as_links will always be true.

the if below where you check: `if dc_as_links: .... elif dc_as_lines`, may be revised with: `if dc_as_links: ... else: ...` so replacing the elif with else.

I'm unsure if some issues may be fixed with this but i'm unsure

scripts/simplify_network.py
davide-f: Interesting check, so, the comment here may be that instead of calculating the power by using V\*I , it may be better to directly use s_nom.
So regarding the code, I recommend to use s_nom here directly, ideally fixing the calculation of V\*I for dc lines.
What do you think?

scripts/simplify_network.py
ekatef: Absolutely agree. This naming has been inherited from PyPSA-Eur, and I have tried to keep it. Happy to revise :)

scripts/add_electricity.py
ekatef: Done :)

scripts/simplify_network.py
ekatef: That is definitely cleaner :)
Thanks for explanations! Done

ekatef:
I confirm that clustering troubles are not connected with this PR. A test for IN with the same configuration parameters has resulted in the same limitation on `n_cluster`. Not sure it's good news from the general perspective, but for this PR that means that basically the only point left is fixing of the equivalence calculations

scripts/simplify_network.py
ekatef: Fixed

scripts/simplify_network.py
```diff
@@ -129,6 +129,11 @@ def simplify_network_to_base_voltage(n, linetype, base_voltage):
     # Note: s_nom is set in base_network
     n.lines["num_parallel"] = n.lines.eval("s_nom / (sqrt(3) * v_nom * i_nom)")
 
+    # Re-define s_nom for DC lines
+    n.lines.loc[n.lines["carrier"] == "DC", "num_parallel"] = n.lines.eval(
+        "s_nom / (v_nom * i_nom)"
+    )
```
davide-f: Super! :) we are there!

I'd recommend to create an auxiliary variable `is_dc_carrier = n.lines["carrier"] == "DC"` and delimit the right-hand side to avoid potential problems
`n.lines.loc[is_dc_carrier , "num_parallel"] = n.lines.loc[is_dc_carrier].eval("s_nom / (v_nom * i_nom)")`

I believe this should work but we have to rely on the fact that the indices that do not match are dropped; it may be better to stay safe.
To be merged after this :)

ekatef: A great point! Fixed

scripts/simplify_network.py
ekatef: Have tired to improve naming. Not completely sure if it's worth. Happy to undo in case :)

davide-f:
Great @ekatef :D merged :)

ekatef:
> Great @ekatef :D merged :)

Awesome! Thank you so much @davide-f :D

PR #744, opened 2023-05-26
Attach res

yerbol-akhmetov:
Here, I propose a modification to `attach_wind_and_solar` function in `add_electricity.py`. With this modification, it would be possible to set capacities for renewables from `custom_powerplants.csv`. Currently, the only possible option to set the renewable capacities is using `estimate_renewable_capacities_irena`. However, even if you know exact capacities of existing wind and solar powerplants, currently you cannot set them as `p_nom`. With the proposed PR, it will automatically calculate the aggregated `p_nom` for each bus from `custom_powerplants.csv` and set them accordingly.

## Checklist

### DIFF
--- a/scripts/add_electricity.py
+++ b/scripts/add_electricity.py
@@ -300,12 +300,17 @@ def attach_wind_and_solar(
     # TODO: rename tech -> carrier, technologies -> carriers
     _add_missing_carriers_from_costs(n, costs, technologies)
 
-    df = ppl.copy()
-    df.rename(columns={"country": "Country"}, inplace=True)
+    df = ppl.rename(columns={"country": "Country"})
 
     for tech in technologies:
         if tech == "hydro":
             continue
+
+        if tech == "offwind-ac": 
+            # add all offwind wind power plants by default as offwind-ac
+            df.carrier.mask(df.technology=="Offshore", "offwind-ac",inplace=True)
+
+        df.carrier.mask(df.technology == "Onshore", "onwind", inplace=True)
 
         with xr.open_dataset(getattr(snakemake.input, "profile_" + tech)) as ds:
             if ds.indexes["bus"].empty:
@@ -332,6 +337,14 @@ def attach_wind_and_solar(
                 #     format(connection_cost.min(), connection_cost.max(), tech))
             else:
                 capital_cost = costs.at[tech, "capital_cost"]
+            
+            if not df.query("carrier == @tech").empty:
+                buses = n.buses.loc[ds.indexes["bus"]]
+                caps = map_country_bus(df.query("carrier == @tech"), buses)
+                caps = caps.groupby(["bus"]).p_nom.sum()
+                caps = pd.Series(data = caps, index = ds.indexes["bus"]).fillna(0)
+            else:
+                caps = pd.Series(index = ds.indexes["bus"]).fillna(0)
 
             n.madd(
                 "Generator",
@@ -346,25 +359,9 @@ def attach_wind_and_solar(
                 capital_cost=capital_cost,
                 efficiency=costs.at[suptech, "efficiency"],
                 p_max_pu=ds["profile"].transpose("time", "bus").to_pandas(),
+                p_nom=caps,
+                p_nom_min=caps,
             )
-
-        if suptech == "offwind":
-            continue
-            # df.carrier.mask(df.technology=="Offshore", "offwind-ac",inplace=True)
-
-        df.carrier.mask(df.technology == "Onshore", "onwind", inplace=True)
-
-        gens = n.generators[lambda df: df.carrier == tech]
-        buses = n.buses.loc[gens.bus.unique()]
-        gens_per_bus = gens.groupby("bus").p_nom.count()
-
-        if not gens.empty and not df.query("carrier == @tech").empty:
-            caps = map_country_bus(df.query("carrier == @tech"), buses)
-            caps = caps.groupby(["bus"]).p_nom.sum()
-            caps = caps / gens_per_bus.reindex(caps.index, fill_value=1)
-
-            n.generators.p_nom.update(gens.bus.map(caps).dropna())
-            n.generators.p_nom_min.update(gens.bus.map(caps).dropna())
 
 
 def attach_conventional_generators(
--- a/scripts/add_electricity.py
+++ b/scripts/add_electricity.py
@@ -352,7 +352,7 @@ def attach_wind_and_solar(
         buses = n.buses.loc[gens.bus.unique()]
         gens_per_bus = gens.groupby("bus").p_nom.count()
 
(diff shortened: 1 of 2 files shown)


pz-max:
@yerbol-akhmetov can you share your data somewhere such that we can locally test the contribution? DONE HERE
[custom_powerplants.csv](https://github.com/pypsa-meets-earth/pypsa-earth/files/11578032/custom_powerplants.csv)

scripts/add_electricity.py
```diff
@@ -338,6 +348,24 @@ def attach_wind_and_solar(
                 p_max_pu=ds["profile"].transpose("time", "bus").to_pandas(),
             )
 
+        if suptech == "offwind":
+            continue
+            # df.carrier.mask(df.technology=="Offshore", "offwind-ac",inplace=True)
+
+        df.carrier.mask(df.technology == "Onshore", "onwind", inplace=True)
```
pz-max: @yerbol-akhmetov  Is this line just necessary because the custom_powerplant data is not stored correctly / or the docs are described such giving false instructions on storing the data?

scripts/add_electricity.py
```diff
@@ -338,6 +348,24 @@ def attach_wind_and_solar(
                 p_max_pu=ds["profile"].transpose("time", "bus").to_pandas(),
             )
 
+        if suptech == "offwind":
+            continue
+            # df.carrier.mask(df.technology=="Offshore", "offwind-ac",inplace=True)
```
pz-max: Why was this commented out?

scripts/add_electricity.py
```diff
@@ -288,11 +289,20 @@ def update_transmission_costs(n, costs, length_factor=1.0, simple_hvdc_costs=Fal
...
+    input_profiles,
+    technologies,
+    extendable_carriers,
+    line_length_factor=1,
 ):
     # TODO: rename tech -> carrier, technologies -> carriers
     _add_missing_carriers_from_costs(n, costs, technologies)
 
+    df = ppl.copy()
+    df.rename(columns={"country": "Country"}, inplace=True)
```
pz-max: Why was this necessary?

scripts/add_electricity.py
```diff
@@ -338,6 +348,24 @@ def attach_wind_and_solar(
...
+        gens = n.generators[lambda df: df.carrier == tech]
+        buses = n.buses.loc[gens.bus.unique()]
+        gens_per_bus = gens.groupby("bus").p_nom.count()
+
+        if not gens.empty and not df.query("carrier == @tech").empty:
+            caps = map_country_bus(df.query("carrier == @tech"), buses)
+            caps = caps.groupby(["bus"]).p_nom.sum()
+            caps = caps / gens_per_bus.reindex(caps.index, fill_value=1)
+
+            n.generators.p_nom.update(gens.bus.map(caps).dropna())
```
pz-max: why updating the capacities and not add already above at `n.madd()`?

pz-max (commented):
Just a few comments. I will test the code locally tomorrow morning for a proper review and code suggestions.

scripts/add_electricity.py
yerbol-akhmetov: @pz-max, I have added this lines because by default `onwind` is stored as `wind` based on PPM standards. Please, see the figure below: 
image

scripts/add_electricity.py
yerbol-akhmetov: Hi, @pz-max. I have commented it out to be consistent with previous lines, where offwind is comemnted. Please, see figure below. This portion also need to be modified in the future, because offwind can be AC or DC, but I have not seen in PPM csvs the difference. Not stated if ac or dc.
image
image

scripts/add_electricity.py
yerbol-akhmetov: The module map_country_bus by default searches "Country" column not "country". Edited.
image

yerbol-akhmetov: Here, is the OPSD function: 
image

yerbol-akhmetov: I have added this line, because without it was raising error. Maybe it is related to how powerplants.csv is formed. You can try without this line.

scripts/add_electricity.py
yerbol-akhmetov: @pz-max, we can do that. For that we need to move all added code after condition with hydro: 
image

scripts/add_electricity.py
pz-max: @yerbol-akhmetov the commented parts for offwind was a hack for updating capital costs.
Here we want to add the capacities of all wind and solar related technologies. 
We need to keep this line/ not skipping the addition of powerplants for offwind.

My suggestion is to do something like:

From:
```
        if suptech == "offwind":
            continue
            # df.carrier.mask(df.technology=="Offshore", "offwind-ac",inplace=True)
```

To:
```
      if tech == "offwind-ac": 
          # add all offwind wind power plants by default as offwind-ac
          df.carrier.mask(df.technology=="Offshore", "offwind-ac",inplace=True)
```

scripts/add_electricity.py
pz-max: Let's do that :+1: 
Can you make these changes here?

scripts/add_electricity.py
pz-max: @yerbol-akhmetov let's change the two lines to:
`df = ppl.rename(columns={"country": "Country"})`

scripts/add_electricity.py
```diff
@@ -338,6 +348,24 @@ def attach_wind_and_solar(
                 p_max_pu=ds["profile"].transpose("time", "bus").to_pandas(),
             )
 
+        if suptech == "offwind":
+            continue
+            # df.carrier.mask(df.technology=="Offshore", "offwind-ac",inplace=True)
+
+        df.carrier.mask(df.technology == "Onshore", "onwind", inplace=True)
+
+        gens = n.generators[lambda df: df.carrier == tech]
```
pz-max: Instead of the lambda function could you try this:
`n.generators[df.carrier == tech]`

pz-max (changes requested):
Summary of main changes:
- Let's not update the network but read `p_nom` directly into the `n.madd()` if capacities are available
- Let's also read the capacities for offshore technologies and not skipping them. By default, we will read the capacities as `offshore-ac` carrier
- Let's avoid lambda when not necessary

Note. This review also includes feedback on the above comments.

scripts/add_electricity.py
yerbol-akhmetov: @pz-max, if we put the code in front to make `madd()` instead of `update`, it need to be rewritten majorly. Because currently if searches for already added RES gens in `n.generators ` and updates their p_nom. I will do that.

pz-max (approved):

PR #757, opened 2023-06-12
Fix runoff in case runoff data are missed for some hydropower plants

ekatef:
PR fixes an issue appeared in a workflow for India with gadm clustering on. The problem is caused by a case when there are hydropower plants on islands, namely Kalpong hydro power plant for India, for which ERA5 can miss the data:

image

Hydropower indices are matched with plants names.

## Checklist

### DIFF
--- a/scripts/add_electricity.py
+++ b/scripts/add_electricity.py
@@ -457,17 +457,22 @@ def attach_hydro(n, costs, ppl):
             missing_plants = pd.Index(inflow_buses.unique()).difference(
                 inflow.indexes["plant"]
             )
-            plants_with_data = pd.Index(
-                inflow_buses[inflow_buses.isin(inflow.indexes["plant"])]
-            )
+            # map power plants index (regions_onshore) into power plants ids (powerplants.csv)
+            # plants_to_keep correspond to "plant" index of regions_onshore
+            # plants_to_keep.index corresponds to bus_id of PyPSA network
+            plants_with_data = inflow_buses[inflow_buses.isin(inflow.indexes["plant"])]
+            plants_to_keep = plants_with_data.to_numpy()
 
             # if missing time series are found, notify the user and exclude missing hydro plants
             if not missing_plants.empty:
                 # original total p_nom
                 total_p_nom = ror.p_nom.sum() + hydro.p_nom.sum()
-                idxs_to_keep = inflow_buses[inflow_buses.isin(plants_with_data)].index
-                ror = ror.loc[ror.index.intersection(idxs_to_keep)]
-                hydro = hydro.loc[hydro.index.intersection(idxs_to_keep)]
+                # update plants_with_data to ensure proper match between "plant" index and bus_id
+                plants_with_data = inflow_buses[inflow_buses.isin(plants_to_keep)]
+                network_buses_to_keep = plants_with_data.index
+
+                ror = ror.loc[ror.index.intersection(network_buses_to_keep)]
+                hydro = hydro.loc[hydro.index.intersection(network_buses_to_keep)]
                 # loss of p_nom
                 loss_p_nom = ror.p_nom.sum() + hydro.p_nom.sum() - total_p_nom
 
@@ -475,15 +480,16 @@ def attach_hydro(n, costs, ppl):
                     f"'{snakemake.input.profile_hydro}' is missing inflow time-series for at least one bus: {', '.join(missing_plants)}."
                     f"Corresponding hydro plants are dropped, corresponding to a total loss of {loss_p_nom:.2f}MW out of {total_p_nom:.2f}MW."
                 )
-            else:
-                idxs_to_keep = inflow_idx
 
             # if there are any plants for which runoff data are available
             if not plants_with_data.empty:
+                network_buses_to_keep = plants_with_data.index
+                plants_to_keep = plants_with_data.to_numpy()
+
                 inflow_t = (
-                    inflow.sel(plant=plants_with_data)
+                    inflow.sel(plant=plants_to_keep)
                     .rename({"plant": "name"})
-                    .assign_coords(name=idxs_to_keep)
+                    .assign_coords(name=network_buses_to_keep)
                     .transpose("time", "name")
                     .to_pandas()
                 )
--- a/scripts/add_electricity.py
+++ b/scripts/add_electricity.py
@@ -457,7 +457,7 @@ def attach_hydro(n, costs, ppl):
             missing_plants = pd.Index(inflow_buses.unique()).difference(
                 inflow.indexes["plant"]
             )
-            intersection_plants = pd.Index(
+            plants_with_data = pd.Index(
                 inflow_buses[inflow_buses.isin(inflow.indexes["plant"])]
             )
 
@@ -465,9 +465,7 @@ def attach_hydro(n, costs, ppl):
             if not missing_plants.empty:
                 # original total p_nom
                 total_p_nom = ror.p_nom.sum() + hydro.p_nom.sum()
-                idxs_to_keep = inflow_buses[
-                    inflow_buses.isin(intersection_plants)
-                ].index
+                idxs_to_keep = inflow_buses[inflow_buses.isin(plants_with_data)].index
                 ror = ror.loc[ror.index.intersection(idxs_to_keep)]
                 hydro = hydro.loc[hydro.index.intersection(idxs_to_keep)]
                 # loss of p_nom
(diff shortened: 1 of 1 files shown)


davide-f:
Hello @ekatef :)

This PR sounds to me straightforward and thanks!

Is it complete in your opinion?

davide-f (approved):

davide-f (commented):
I have a comment to add: we shall make sure the order is guaranteed when using plant and name.
when doing the assignment that may be lost and hence the correspondance between plant and its corresponding inflow may be lost

ekatef:
Hello @davide-f and thanks for the review! :)
Agree on the comment and will look into it.

Basically, I don't have much to add to the PR, except it would be nice to have successful CI runs 🙂  Hopefully, #761 will resolve this weird issue with windows CI

ekatef:
As far as I understand, correspondence between `intersection_plants` and `idxs_to_keep` vectors is ensured by defining the latter with a subsetting `intersection_plants` with a boolean pd.Series:

https://github.com/pypsa-meets-earth/pypsa-earth/blob/20e5061f847055d15f401972ac19668651bc1716/scripts/add_electricity.py#L468-L470

So, an order of elements in `intersection_plants` and `idxs_to_keep` should match by definition meaning that it's safe to use the first one for subsetting and the second one as the coordinates.

@davide-f could you please double-check if this does make sense for you? :)

ekatef:
A couple of additional points that should be checked in context of this PR.

1. Would be any additional warnings helpful? Currently a warning is triggered with `missing_plants` being not empty and the reported problem is perfectly consistent with the data issues the current PR is focused on. However, technically the current issue is caused by occurence of `intersection_plants`. So, the question is if there is anything special in the encountered issue which would be worth reporting?

2. There are some entries in `inflow.indexes["plant"]` which look like they would have been inherited from the contested territories issue (`'not found.14_1', 'not found.3_1'`); it would be nice to understand a particular reason.

ekatef:
Update by both the points above:

1. A warning related to presence of `missing_plants` is enough to properly process this issue. If `intersection_plants` is non-empty is means only that there are some plants with available runoff data. Some naming revision can be beneficial to make it more clear.
2. Names like `not found*` have propagated to the renewable profiles from `regions_onshore.geojson`. This problem is observed for IN but for some reason isn't for CN. Have added #772 to track that

ekatef:
@davide-f my feeling is that this PR is ready. Would be grateful if you could look into it :)

ekatef:
As a side note, the error can be reproduced also for CD with alternative clustering. There is no runoff data for Ruzizi and two Rutshuru power plants. I suspect the reason is that these plants are located on the very border of CDR with Rwanda, as shown bellow. Which probably leads to the catchment areas not being captured properly when integrating ERA5 by area

image

scripts/add_electricity.py
```diff
@@ -457,32 +457,33 @@ def attach_hydro(n, costs, ppl):
...
             )
 
             # if missing time series are found, notify the user and exclude missing hydro plants
             if not missing_plants.empty:
                 # original total p_nom
                 total_p_nom = ror.p_nom.sum() + hydro.p_nom.sum()
-                idxs_to_keep = inflow_buses[
-                    inflow_buses.isin(intersection_plants)
-                ].index
+                idxs_to_keep = inflow_buses[inflow_buses.isin(plants_with_data)].index
```
davide-f: I think we need to improve the readability of the code because it is easy to get lost with indices.
Something that Irealize now is that this code seems a duplication and we may need to debug it together.

Initially is defined as the values of a subset of inflow_buses, whose index are idxs_to_keep.
I think that to improve the readability, we may avoid defining idx_to_keep inside the if but outside.
plants with data may be the view: `inflow_buses[inflow_buses.isin(inflow.indexes["plant"])]` where its index is idxs_to_leep and its values are '''plants'''.
We may revise the code in that way so that it may be easier to understand.

What do you think? Let me know if the above is clear

ekatef: @davide-f agree that this change look very natural. However, I needed about an hour to understand it, so I absolutely aligned with your idea to make this fragment a bit more readable! :D

Have implemented a possible approach. Along with revision of `idxs_to_keep` definition, there were following changes:
1) move reading PHS `phs = ppl.query('technology == "Pumped Storage"')` to a less over-crowded part;
2) revise the names of the data frames `ror` and `hydro` to make it clear that both were derived from `ppl` and contain run-of-river and reservoir data, respectively;
3) change `inflow_idx` to `hydro_ppl_idx` to avoid an association with `inflow` indices which can be a bit misleading.

Not sure if all these changes are really necessary. I'd suggest to select only those which are really needed. Could you please check if these changes do really allow to decrease reading time? :)

ekatef:
Tested on "CD" and proved to resolve a problem which has previously arisen when `alternative_clustering` is on.

What I'm not sure about is updating of `plants_to_keep`: shouldn't it be included to [this fragment](https://github.com/ekatef/pypsa-earth/blob/2324a73cf5094ad4ad60bee71156ab1674f87b21/scripts/add_electricity.py#L471-L472) along with updating `network_buses_to_keep`?

```python
                plants_with_data = inflow_buses[inflow_buses.isin(plants_to_keep)]
                network_buses_to_keep = plants_with_data.index
```

davide-f:
> Tested on "CD" and proved to resolve a problem which has previously arisen when `alternative_clustering` is on.
> 
> What I'm not sure about is updating of `plants_to_keep`: shouldn't it be included to [this fragment](https://github.com/ekatef/pypsa-earth/blob/2324a73cf5094ad4ad60bee71156ab1674f87b21/scripts/add_electricity.py#L471-L472) along with updating `network_buses_to_keep`?
> 
> ```python
>                 plants_with_data = inflow_buses[inflow_buses.isin(plants_to_keep)]
>                 network_buses_to_keep = plants_with_data.index
> ```

For the sake of avoiding problems, that's a good point, you could add a row below to update the value of plants_to_keep.
From the point of view of functionality, for the current code, that's not needed as that value is anyway updated below.
However, I'd be prone to update the value to avoid potential problems in the future.
Good point :)

ekatef:
> For the sake of avoiding problems, that's a good point, you could add a row below to update the value of plants_to_keep. From the point of view of functionality, for the current code, that's not needed as that value is anyway updated below. However, I'd be prone to update the value to avoid potential problems in the future. Good point :)

Great! Thank you :) Then let's try this way. To me, that looks also a bit more consistent to keep updates of both `network_buses_to_keep` and `plants_to_keep` along each other

Sorry for the mess in commits. Actually, I can squash them after the final revision ;)

scripts/add_electricity.py
```diff
@@ -457,32 +457,39 @@ def attach_hydro(n, costs, ppl):
...
                 logger.warning(
                     f"'{snakemake.input.profile_hydro}' is missing inflow time-series for at least one bus: {', '.join(missing_plants)}."
-                    f"Corresponding hydro plants are dropped, corresponding to a total loss of {loss_p_nom}MW out of {total_p_nom}MW."
+                    f"Corresponding hydro plants are dropped, corresponding to a total loss of {loss_p_nom:.2f}MW out of {total_p_nom:.2f}MW."
                 )
 
-            if not intersection_plants.empty:
+            # if there are any plants for which runoff data are available
+            if not plants_with_data.empty:
+                network_buses_to_keep = plants_with_data.index
```
davide-f, line 487: Hello :)

I saw that you removed here the `plants_to_keep = plants_with_data.to_numpy()`, maybe we misunderstood.
I understood your proposal was to add its definition also there, without moving it from below.

Having the two definitions of the entries here ensures an easy understanding that the two values are of equal length and are consistent: plants_with_data represents indeed a mapping and with that, it is clear.
Having the definitions across the previous 20 lines unfortunately is not so clear.

I see two options:
1. keep it was before
2. or having it both here and above

What do you think?

ekatef: > Hello :)
> 
> I saw that you removed here the `plants_to_keep = plants_with_data.to_numpy()`, maybe we misunderstood. I understood your proposal was to add its definition also there, without moving it from below.
> 
> Having the two definitions of the entries here ensures an easy understanding that the two values are of equal length and are consistent: plants_with_data represents indeed a mapping and with that, it is clear. Having the definitions across the previous 20 lines unfortunately is not so clear.
> 
> I see two options:
> 
> 1. keep it was before
> 2. or having it both here and above
> 
> What do you think?

I have tired to avoid duplicating of `plants_to_keep` definition. But agree that it might be not so obvious to have a definition twenty lines above... :)

My personal preferences is to have `plants_to_keep` in both fragments, but also happy to keep the original version if you feel it more consistent

davide-f (approved):
@ekatef for me it's good to go :)
Let me know if you want to squash the last two commits or go as-is ;)

ekatef:
> @ekatef for me it's good to go :) Let me know if you want to squash the last two commits or go as-is ;)

Super! Squash done ;)

davide-f:
Merged :D

ekatef:
> Merged :D

Thank you so much ;)

PR #768, opened 2023-06-16
Fix augmentation of links

### LINKED ISSUE #765: Revise augmentation procedure for networks with HVDC part
Currently there are some troubles when building augmented network for a network which contain DC lines. 

## Checklist

- [ ] I am running on an up-to-date [`pypsa-earth` environment](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

Namely, when calculating an underwater_fraction for links shapely error is thrown by an attempt to call loads on link geometries:

https://github.com/pypsa-meets-earth/pypsa-earth/blob/b015ab2883dfc2178c2ee3aa12ac486b6cb9abeb/scripts/base_network.py#L157

It seems to be linked with a fact that lines geometry is not restored when [processing links](https://github.com/pypsa-meets-earth/pypsa-earth/blob/b015ab2883dfc2178c2ee3aa12ac486b6cb9abeb/scripts/simplify_network.py#L362-L379) in `simplify_network` which leads to NaN/empty string geometries for some links.

A possible way to fix links geometries would be to add `geometry` column. An attempt was done in #768 but has not been successful due to consequent troubles with network export to netcdf. Another approach is needed like `import_components_from_dataframe` used in `base_network`.

**Way to reproduce**: run a workflow for CD with `add_to_snakefile: true` for `augmented_line_connection` and `hvdc_as_lines: false`.

## Error Message

```bash
rule augmented_line_connections:
    input: resources/CD/costs.csv, networks/CD/elec_s_10_pre_augmentation.nc, resources/CD/bus_regions/regions_onshore_elec_s_10.geojson, resources/CD/bus_regions/regions_offshore_elec_s_10.geojson
    output: networks/CD/elec_s_10.nc
    log: logs/CD/augmented_line_connections/elec_s_10.log
    jobid: 4
    benchmark: benchmarks/CD/augmented_line_connections/elec_s_10
    reason: Missing output files: networks/CD/elec_s_10.nc; Input files updated by another job: resources/CD/bus_regions/regions_onshore_elec_s_10.geojson, resources/CD/bus_regions/regions_offshore_elec_s_10.geojson, networks/CD/elec_s_10_pre_augmentation.nc, resources/CD/costs.csv
    wildcards: simpl=, clusters=10
    resources: tmpdir=/var/folders/qn/vpndfm21795ckkq89np1ckp40000gn/T, mem_mb=3000, mem_mib=2862
INFO:snakemake.logging:rule augmented_line_connections:
    input: resources/CD/costs.csv, networks/CD/elec_s_10_pre_augmentation.nc, resources/CD/bus_regions/regions_onshore_elec_s_10.geojson, resources/CD/bus_regions/regions_offshore_elec_s_10.geojson
    output: networks/CD/elec_s_10.nc
    log: logs/CD/augmented_line_connections/elec_s_10.log
    jobid: 4
    benchmark: benchmarks/CD/augmented_line_connections/elec_s_10
    reason: Missing output files: networks/CD/elec_s_10.nc; Input files updated by another job: resources/CD/bus_regions/regions_onshore_elec_s_10.geojson, resources/CD/bus_regions/regions_offshore_elec_s_10.geojson, networks/CD/elec_s_10_pre_augmentation.nc, resources/CD/costs.csv
    wildcards: simpl=, clusters=10
    resources: tmpdir=/var/folders/qn/vpndfm21795ckkq89np1ckp40000gn/T, mem_mb=3000, mem_mib=2862

INFO:snakemake.logging:
INFO:pypsa.io:Imported network elec_s_10_pre_augmentation.nc has buses, carriers, generators, lines, links, loads, storage_units
  File "~pypsa-earth/.snakemake/scripts/tmp_sfy0l5r.augmented_line_connections.py", line 177, in <module>
    _set_links_underwater_fraction(snakemake.input.regions_offshore, n)
  File "~/Library/Caches/snakemake/snakemake/source-cache/runtime-cache/tmp4zrhwxjf/file~pypsa-earth/scripts/base_network.py", line 157, in _set_links_underwater_fraction
    links = gpd.GeoSeries(n.links.geometry.dropna().map(shapely.wkt.loads))
  File "~/opt/miniconda3/envs/pypsa-earth-upd14/lib/python3.10/site-packages/pandas/core/series.py", line 4397, in map
    new_values = self._map_values(arg, na_action=na_action)
  File "~/opt/miniconda3/envs/pypsa-earth-upd14/lib/python3.10/site-packages/pandas/core/base.py", line 924, in _map_values
    new_values = map_f(values, mapper)
  File "pandas/_libs/lib.pyx", line 2834, in pandas._libs.lib.map_infer
  File "~/opt/miniconda3/envs/pypsa-earth-upd14/lib/python3.10/site-packages/shapely/wkt.py", line 22, in loads
    return shapely.from_wkt(data)
  File "~/opt/miniconda3/envs/pypsa-earth-upd14/lib/python3.10/site-packages/shapely/io.py", line 286, in from_wkt
    return lib.from_wkt(geometry, invalid_handler, **kwargs)
shapely.errors.GEOSException: ParseException: Expected word but encountered end of stream
```

ekatef:
# Closes #765 

Underwater fraction is calculated in `simplify_networks` which means that application of `_set_links_underwater_fraction()` it is redundant. 

## Checklist


ekatef:
An error in `_set_links_underwater_fraction()` is caused by empty links geometries which appear in course of merging links in `simplify_network`. Geometry is not re-defined for a merged link:

https://github.com/pypsa-meets-earth/pypsa-earth/blob/20e5061f847055d15f401972ac19668651bc1716/scripts/simplify_network.py#L362-L378

An example for CD:

```bash
>n.links.geometry
Link
383097966_0      LINESTRING (15.2947 -4.712, 15.294959 -4.71209...
778351020_0+1                                                     
261165614_0+1                                                     
460419487_0+4                                                     
Name: geometry, dtype: object
```

Generally, it can be fixed by setting geometry as a line connecting `bus0` and `bus1`. Another approach is to exclude `geometry` column from n.links. That would be more straightforward and also more consistent with `n.lines` which don't have `geometry` column on a pre-augmentation stage.

ekatef:
It might be nice to consider two more points when fixing this issue:
1. There is a TODO in `add_electricity` which also relates to `underwater_fraction` and might be relevant here:

https://github.com/pypsa-meets-earth/pypsa-earth/blob/072b401a6cf1a14e2311bfd5d15ccfc55195348c/scripts/add_electricity.py#L263-L268

3. It could be nice to have an option to import geometry for a simplified and clustered networks into some common GIS format like geojson

davide-f:
Sorry @ekatef that this PR has been pending.
I'm wondering whether it may be better to make sure that during the simplification, geometries may be kept meaningful, what do you think?

ekatef:
@davide-f I absolutely agree that it would convenient to have lines geometry after simplification. Happy to do so, but I'd first finalise #743 to avoid merge conflicts

ekatef:
An attempt to create `geometry` column for `n.links` leads to an error during export to netcdf: `unable to infer dtype on variable 'links_geometry'; xarray cannot serialize arbitrary Python objects`. 

After looking into this [SO answer](https://stackoverflow.com/a/50475925/8465924), I think a likely reason are conflicting geometry coordinates in `nc` file. This error has been resolved in the current implementation using `geometry_links` as a column name. That is however not perfectly consistent in terms of naming, while geometry information is anyway kept in n.buses as `x` and `y` columns. Probably, it doesn't make a lot of sense to duplicate.

@davide-f would be grateful for your opinion on that :)

scripts/simplify_network.py
```diff
@@ -428,6 +429,9 @@ def split_links(nodes):
                 p_min_pu=-p_max_pu,
                 underground=False,
                 under_construction=False,
+                geometry_links=LineString(
+                    zip(n.buses.loc[[b[0], b[1]]].x, n.buses.loc[[b[0], b[1]]].y)
+                ),
```
davide-f: have you tried to name "geometry" instead of "geometry_links"?
The error seemed related to the improper defintion of the column geometry, overriding it should do the trick, but not sure...

for styling [b[0], b[1]] may be replaced with b[0:2] or b itself if it contains only 2 elements

davide-f (commented):
Hello @ekatef, great to get back to this issue :D

I've added a little comment. If I recall the issue was related to the improper value of the "geometry" column later on.
Have you tried to name the param "geometry" instead of "links_geometry"?

scripts/simplify_network.py
ekatef: Yes, absolutely! An initial version has used `geometry` as a column name, but it looks like it's being transformed into `'links_geometry'` by some PyPSA method when exporting to nc. Which leads to the mentioned `xarray cannot serialize arbitrary Python objects`. Do I miss something?

Done for indexing re-styling :)

ekatef:
> Hello @ekatef, great to get back to this issue :D
> 
> I've added a little comment. If I recall the issue was related to the improper value of the "geometry" column later on. Have you tried to name the param "geometry" instead of "links_geometry"?

Hello @davide-f and big thanks for reviewing!

Yeah, `geometry` has been a first try. I have just repeated this attempt to double-check, and confirm that it leads to an error during export to netcdf. Obviously, there is some issue with how coordinates are being assigned during PyPSA export. Do you think it makes sense to go deeper into it?

davide-f:
Probably the aggregate procedure doesn't like the geometry parameter, but what if we replace the geometry after the aggregation procedure?
I think it may be worth checking since you have put effort and have all the debugging files ready, what do you think?

ekatef:
@davide-f absolutely that it would be nice to get it work! :D

However, simple solutions have not been successful yet: have tried to move definition of `n.links.geometry` after aggregation procedure in `simplify_links` but the export error persists.

What is interesting is that in an original version `n.links` contains `geometry` column, but it's value is NaN. If `.geometry=1` is set, export still works. However, if geometry is defined using `LineString()`, the error appears. I suspect, there may be some trick in applying `LineString()` itself... :)

davide-f:
Ok... it sounds to that the raw trick of dropping the _set dc part in augmentation part may be a raw trick for fixing the workflow.
We could do that while adding an issue tackling that.
Given the effort you put, I'd keep this PR open to track the tests and the code already developed for later fixes when needed.
A todo in the code may also be added to keep track of it.

In terms of functionality, what is the drawback of dropping that _set function, given the tests you did?

ekatef:
> Ok... it sounds to that the raw trick of dropping the _set dc part in augmentation part may be a raw trick for fixing the workflow. We could do that while adding an issue tackling that. Given the effort you put, I'd keep this PR open to track the tests and the code already developed for later fixes when needed. A todo in the code may also be added to keep track of it.
> 
> In terms of functionality, what is the drawback of dropping that _set function, given the tests you did?

According to my current understanding, using `_set` function in `augmented_line_connections` is kind of duplication, as `underwater_fraction` for links is calculated during simplification. However, it has been a while since the tests were run, and I can easily miss something.  So, I'd make a fresh testing series to discuss the results during the next weekly. In case there is no pitfalls, absolutely agree with the way you suggest with adding TODO and eventually keeping this PR open for some time

Would you approve this plan?

davide-f:
Yes :)
Sorry for the time spent here, at the beginning it seemed a promising stable approach, but it turned out to be too complex for the time being.
Anyway we treasure these efforts and keep that in mind for the next steps.

Rather than deleting the line, you may comment that and add a comment on why that line is commented maybe

ekatef:
> Yes :)
> 
> Sorry for the time spent here, at the beginning it seemed a promising stable approach, but it turned out to be too complex for the time being.
> 
> Anyway we treasure these efforts and keep that in mind for the next steps.
> 
> 
> 
> Rather than deleting the line, you may comment that and add a comment on why that line is commented maybe

Great!
No problem, it has been a pleasure to play around. :) Agree that it absolutely makes sense to keep that approach in mind for future

ekatef:
After some additional investigation and testing, there are following points to be checked.

1. A proper approach to load links geometries to a network could be `import_components_from_dataframe`, as it's done in `base_network`. Not sure however it's feasible to use it in `simplify_network`.
2. There are still some troubles with missed in the augmentation script when there is HVDC part in the network, and n.links.geometry contains `NaN` values. Would it be probably a solution to remove `geometry` column?
3. I may miss something, but I'm unsure if it's physically correct trying to re-calculate underwater_fraction when making augmentation, as doing so we assume that lines geometry is meaningful, which is not necessary the case after clustering. 

@davide-f would be very grateful for your opinion on that :)

ekatef:
Update: if dropping of `geometry` column in `n.links` after simplification, it appears anyway along the workflow. It feels like the issue with `NaN` values after simplification can be tracked as at the moment solving this issue would require too deep changes

The issue with empty links geometries can be resolved by adding a transformation empty string -> to `NaN`. Anyway, augmentation of the networks with HVDC part needs some more detailed investigation. There is still #765 on that

davide-f:
I generally agree on point 3; in particular, the feeling is that a "raw trick" for the moment is fine, especially if functionalities are not impacted.

In simplify network the shape of lines is basically not changed, so their underwater fraction should not change.
Only links representing converters may be questionable.
On the other hand, since we merge links, than that gets tricky: if shapes are also merged as well, then everything is meaningful, however, I think it's too waste of time with questionable improvements.
Let's keep the good stable fix for when it it needed :)

davide-f:
With raw trick, I was thinking about this: https://github.com/pypsa-meets-earth/pypsa-earth/pull/768/commits/161305ea7aae1c48a0643ed5eabe1cc562a42cb5

What are the drawbacks of that?

ekatef:
> With raw trick, I was thinking about this: https://github.com/pypsa-meets-earth/pypsa-earth/pull/768/commits/161305ea7aae1c48a0643ed5eabe1cc562a42cb5
> 
> 
> 
> What are the drawbacks of that?

If we remove set_underwater_fraction from the augmentation script, new DC lines (if any) would have default values of underwater_fraction

ekatef:
> I generally agree on point 3; in particular, the feeling is that a "raw trick" for the moment is fine, especially if functionalities are not impacted.
> 
> 
> 
> In simplify network the shape of lines is basically not changed, so their underwater fraction should not change.
> 
> Only links representing converters may be questionable.
> 
> On the other hand, since we merge links, than that gets tricky: if shapes are also merged as well, then everything is meaningful, however, I think it's too waste of time with questionable improvements.
> 
> Let's keep the good stable fix for when it it needed :)

ekatef:
> I generally agree on point 3; in particular, the feeling is that a "raw trick" for the moment is fine, especially if functionalities are not impacted.
> 
> 
> 
> In simplify network the shape of lines is basically not changed, so their underwater fraction should not change.
> 
> Only links representing converters may be questionable.
> 
> On the other hand, since we merge links, than that gets tricky: if shapes are also merged as well, then everything is meaningful, however, I think it's too waste of time with questionable improvements.
> 
> Let's keep the good stable fix for when it it needed :)

Augmentation is applied over a clustered network, so links geometry may change not only due to simplify-merge, but probably also due clustering which makes the issue quite tricky. Absolutely agree that it doesn't feel a good approach to go into all these fine details right now. Ok for the raw fix :) I'll put it back amending a comment on need for better investigation of augmentation with DC

ekatef:
Closed in favour of #816 as a quick fix and #765 to track the deeper issues

PR #769, opened 2023-06-19
Remove res from conventional ppls

yerbol-akhmetov:
Here, I propose a slight modification of the code in `add_electricity` rule. Currently, in the `attach_conventional_generators` in `add_electricity.py`, non-conventional powerplants are also considered if `extendable` contains RES powerplants in config file. It does not currently create any problem, since those RES ppls added by `attach_conventional_generators` are rewritten by `attach_wind_and_solar`. In practice, only conventional powerplants need to be added by the `attach_conventional_generators` rule. I have faced this issue, when I tried further clustering (not only by carrier type, but also powerplant type (CHP and PP)).

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -28,6 +28,8 @@ E.g. if a new rule becomes available describe how to use it `snakemake -j1 run_t
 
 * Add codespell linter which corrects word spellings `PR #763 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/763>`__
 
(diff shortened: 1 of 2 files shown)


pz-max:
@yerbol-akhmetov good suggestion. 

Do you think we can pass `renewable_carriers` here: `attach_conventional_generators()`
and instead of doing:
`set(conventional_carriers) | (set(extendable_carriers["Generator"]) - {"solar", "onwind", "offwind-ac", "offwind-dc"})`
do:
`set(conventional_carriers) | (set(extendable_carriers["Generator"]) - set(renewable_carriers))`

This is a bit more flexible in case new renewable carriers are added.

pz-max (changes requested):
Good suggestion. Can you check my comment and avoid the hardcoded parts?

yerbol-akhmetov:
I have replaced hard-coded part with more general `renewable_carriers`.

pz-max (approved):

PR #701, opened 2023-05-02
Fix workflow for countries with empty osm data

davide-f:
# Closes #642 , #641, #735

## Changes proposed in this Pull Request
Add methods to handle cases when data from osm are empty.
This PR is built on top of #654 

## Checklist

- [x] I consent to the release of this PR's code under the GPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [x] I tested my contribution locally and it seems to work fine.
- [x] Code and workflow changes are sufficiently documented.
- [x] Newly introduced dependencies are added to `envs/environment.yaml` and `envs/environment.docs.yaml`.
- [x] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [x] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [x] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [ ] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes, including reference to the requested PR.

PR #779, opened 2023-06-30
Fix pre-commit and revise style

davide-f:
Fix to pre-commit.
The pre-commit has also been partially aligned to the one by pypsa-eur

I feel that this may conflict with other PRs, including https://github.com/pypsa-meets-earth/pypsa-earth/pull/701
#701 shall be first reviewed and merged before fixing this.

Everything can be done in few hours

### DIFF
--- a/config.tutorial.yaml
+++ b/config.tutorial.yaml
@@ -337,8 +337,8 @@ monte_carlo:
       # Key: add below the pypsa object for the monte_carlo sampling, "network" is only allowed for filtering!
       # Value: currently supported format [l_bound, u_bound] or empty [], represent multiplication factors for the object
     loads_t.p_set: [0.9, 1.1]
-    generators_t.p_max_pu.loc[:, n.generators.carrier == "wind"]: [0.9, 1.1]
-    generators_t.p_max_pu.loc[:, n.generators.carrier == "solar"]: [0.9, 1.1]
+    # generators_t.p_max_pu.loc[:, n.generators.carrier == "wind"]: [0.9, 1.1]
+    # generators_t.p_max_pu.loc[:, n.generators.carrier == "solar"]: [0.9, 1.1]
 
 
 solving:
--- a/.gitignore
+++ b/.gitignore
@@ -57,9 +57,9 @@ notebooks/old_notebooks/*.tif
 doc/_build
 
 # VS code related
-.vscode
-doc/.vscode/settings.json
-.vscode/settings.json
+*.vscode
+doc/*.vscode/settings.json
+*.vscode/settings.json
 *.code-workspace
 
 # Additional debugging folders
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,17 +35,17 @@ repos:
     types_or: [python, rst, markdown]
     files: ^(scripts|doc)/
 
-# # Make docstrings PEP 257 compliant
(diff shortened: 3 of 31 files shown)


pz-max (approved):

config.tutorial.yaml
```diff
@@ -337,8 +337,8 @@ monte_carlo:
       # Key: add below the pypsa object for the monte_carlo sampling, "network" is only allowed for filtering!
       # Value: currently supported format [l_bound, u_bound] or empty [], represent multiplication factors for the object
     loads_t.p_set: [0.9, 1.1]
-    generators_t.p_max_pu.loc[:, n.generators.carrier == "wind"]: [0.9, 1.1]
-    generators_t.p_max_pu.loc[:, n.generators.carrier == "solar"]: [0.9, 1.1]
+    # generators_t.p_max_pu.loc[:, n.generators.carrier == "wind"]: [0.9, 1.1]
```
pz-max, line 340: @davide-f we need at least two rows to test the Monte-Carlo stuff. Test is currently failing because of that

pz-max (changes requested):

config.tutorial.yaml
pz-max: added a commit to fix this... let's see

pz-max:
pre-commit is failing with `exit-code: 3`:
The [documentation](https://github.com/PyCQA/docformatter/blob/b54a22caae3300f9c37f8c48666a5808d7ff8440/docs/source/usage.rst#use-from-the-command-line) says:
```
Possible exit codes from docformatter:

    1 - if any error encountered
    2 - if it was interrupted
    3 - if any file needs to be formatted (in --check or --in-place mode)
```

I think because of the recent PR that we merged, @davide-f you need to run locally the pre-commit and add the fixes that the **docformatter** is suggesting. Can you do that?

davide-f:
> if any file needs to be formatted (in --check or --in-place mode)

@pz-max the yaml block of the pre-commit is failing because it cannot handle "generators_t.p_max_pu.loc[:, n.generators.carrier == "wind"]: [0.9, 1.1]"

Therefore, I commented these two lines in the tutorial and default configs.

That lead instead to issues in the CI.
I think two approaches may apply:
1. fix the monte carlo so to be able to work without those 2 lines, this is I think the best way, but it means that those lines are not tested in the workflow
2. revise the monte-carlo config test, in the test folder, to add those two lines and ignore that file when doing the pre-commit CI, as otherwise it leads to issues

Since @pz-max, you worked in the monte carlo, what is your feeling there?

config.tutorial.yaml
davide-f: I recreated the PR and kept this line commented out.
pretty YAML crashes with checking those lines. In this PR, there is a fix to monte_carlo caospy to make sure that it works even with a single parameter in the list of monte carlos. The other options have been tested as well.

davide-f:
@pz-max The PR has been rewritten and force-pushed with the following changes:
1. The lines related to `generators_t.p_max_pu.loc ....` in monte_carlo have been commented out as pretty-yaml crashed in analysing them
2. The monte_carlo function has been revised to make sure that the function works with all the montecarlo generating options even when a single uncertain parameter is handled, in this case being the load demand

davide-f:
@pz-max PEP 257 is there, if CI passes this PR can be merged. Locally seems ok

PR #793, opened 2023-07-17
add res_share constraint

martacki:
# Closes # (if applicable).

Pushes the network to have a certain share of RES in the system (per country).

Draft. Needs fixed
- [x] storage units behavior not tested
- [x] link behavior not tested
- [ ] link constraints (ie. stores) should include snapshot weightings
- [x] bug in implementation of storage units / links
- [ ] VRES technologies should not be hard coded, but taken from config (same for links and stores etc.)

## Checklist


scripts/solve_network.py
```diff
@@ -346,6 +346,116 @@ def add_battery_constraints(n):
...
+        )
+        .reindex(lhs_gen.index)
+        .fillna("")
+    )
+
+    lhs_discharge = (
+        (
+            linexpr(
+                (
+                    n.links.loc[discharger_i].efficiency,
```
martacki:
```suggestion
                    1./n.links.loc[discharger_i].efficiency,
```

I think this would be correct, for the reviewer. Maybe they can double-check. I have an evaluation ipynb prebared, it is available to both @yerbol-akhmetov  and @pz-max

scripts/solve_network.py
```diff
@@ -346,6 +346,116 @@ def add_battery_constraints(n):
...
+    )
+
+    rhs = res_share * load
+
+    res_techs = [
+        "solar",
+        "onwind",
+        "offwind-dc",
+        "offwind-ac",
+        "battery",
```
pz-max, line 374: why battery? should be excluded here similar to hydrogen?

martacki: hydrogen does not appear but you are right. Anyway this should not be hard-coded but taken from config file anyways, and if too much appears it does not matter because there is no "battery" carrier in the generators anyway. If there is, something else is already broken...

scripts/solve_network.py
davide-f: Agree, it should be 1/eta

scripts/solve_network.py
```diff
@@ -346,6 +346,116 @@ def add_battery_constraints(n):
...
+                )
+            )
+            .T.groupby(sgrouper, axis=1)
+            .apply(join_exprs)
+        )
+        .reindex(lhs_gen.index)
+        .fillna("")
+    )
+
+    lhs_store = (
```
davide-f, line 411: Here I see a lot of transpositions while they are not there for links.
They are needed to match the shape of weightings right?

However, I believe the weightings shall apply also to links

scripts/solve_network.py
```diff
@@ -346,6 +346,116 @@ def add_battery_constraints(n):
...
+    )
+
+    # Stores (or their resp. Link components)
+    # Note that the variables "p0" and "p1" currently do not exist.
+    # Thus, p0 and p1 must be derived from "p" (which exists), taking into account the link efficiency.
+    lhs_charge = (
+        (
+            linexpr(
+                (
+                    -n.links.loc[charger_i].efficiency,
```
davide-f: On second thought, the efficiency part shall be better revised: here it should be 1/eta or simply 1, and at the discharger should be eta (as it is right now) or 1.
I'll better check later

scripts/solve_network.py
davide-f: I see now the todo at the end of the function

scripts/solve_network.py
```diff
@@ -346,6 +346,116 @@ def add_battery_constraints(n):
...
+    sgrouper = n.storage_units.bus.map(n.buses.country)
+    cgrouper = n.links.bus0.map(n.buses.country)
+
+    load = (
+        n.snapshot_weightings.generators
+        @ n.loads_t.p_set.groupby(lgrouper, axis=1).sum()
+    )
+
+    rhs = res_share * load
+
```
davide-f, line 368: These as mentioned should be taken from the config.
In this function, they could be passe d as arguments and filled when the function is called.

scripts/solve_network.py
davide-f: So, if we want not to include the losses in the share of res, which I guess is the goal of including this exchange, I believe that the following shall apply:
1. the charger shall have coefficient -1: the p of the link represents the power at the bus0. in our case the bus0 represents the bus (bus A) where the battery is connected, hence coefficient should be -1. The bus 1 of the link is instead an inner bus where the store is connected to (bus B).
2. the discharger shall have coefficient "eta" (not 1/eta): in the case of the discharger, the bus0 represents the inner bus of the battery (bus B), whereas the bus1 is bus A. since we need the power that is fed at B, that power is p*eta. Note that the quantity p(charger) - p(discharger) computed should be negative when summed over all timesteps. That should also be easy to test

scripts/solve_network.py
davide-f: As discussed above, I think here the "n.links.loc[discharger_i].efficiency" is ok

davide-f (commented):
Comment on efficiencies added, beyond getting the elements from the config, I think I have no other comments, unless you have further specific questions

scripts/solve_network.py
martacki: Thanks for the input Davide. I agree. Changed and tested now, and seems to work as intended. <3

martacki:
Sorry for commit mess, I should git pull more often :eyesroll:

I think this is ready to be merged. I would suggest to open a new PR to take care of the two remaining ToDos.

scripts/solve_network.py
```diff
@@ -346,6 +346,123 @@ def add_battery_constraints(n):
...
+        )
+        .reindex(lhs_gen.index)
+        .fillna("")
+    )
+
+    lhs_discharge = (
+        (
+            linexpr(
+                (
+                    n.links.loc[discharger_i].efficiency,
```
davide-f: Probably the only line that misses the weightings is this one, may a `n.snapshot_weightings.stores @ n.links.loc[discharger_i].efficiency` be the solution?
[or just the converse]?

davide-f (approved):

PR #796, opened 2023-07-18
Linopy transition

### LINKED ISSUE #494: Integrate Linopy
Linopy was recently integrated into [PyPSA --see PR here](https://github.com/PyPSA/PyPSA/pull/336). With the new PyPSA release, we should make sure to get it into PyPSA-Earth/ PyPSA-Eur ASAP. One can already trial in a draft PR how we can use these upcoming functions. As one can see in the PR above, there is also excellent documentation on how to transition.

For context, Pyomo is the standard optimization interface for Python. For large linear programming/ MILP problems, however, Pyomo has quite poor performance. It made for years Python speed and memory-wise uncompetitive compared to other programming languages. Tom, Fabian and Jonas, worked on alternative called 'nomopyomo'. While this lead to a massive performance boost, the usability/flexibility was suffering a bit. Linopy was about the same speed boost as nomopyomo, better memory performance than 'nomopyomo' and is doing a fantastic job in comparison to other 'solver interfaces'. The [benchmark](https://linopy.readthedocs.io/en/latest/benchmark.html) with Linopy, Jump and Pyomo, only shows how important the move towards Linopy is.

### LINKED ISSUE #787: `pypsa>=0.25` requires upgrade of network clustering
To allow in the env.yaml `pypsa>=0.25` we need to upgrade the network clustering: https://github.com/PyPSA/pypsa-eur/pull/696 (at the time of writing in a PR in PyPSA-Eur)

_Originally posted by @pz-max in https://github.com/pypsa-meets-earth/pypsa-earth/issues/786#issuecomment-1636830835_

ekatef:
# Closes #494 

Integrate `linopy` following https://github.com/PyPSA/pypsa-eur/pull/625

## Checklist


ekatef:
Currently works locally with `cbc`. Points which need attention:
- [ ] there is a mismatch between batteries efficiency and links which has temporary resolved by using a single efficiency value, but needs additional investigation
- [x] what does mean replacement `p` with `p_set` in `add_operational_reserve_margin_constraint()`?
- [ ] the solver log has disappeared from the logger files, while output of solver logging into a console presents if running solving from VSCode but is skipped is using a plain terminal
- [ ] there is a fuzzy feeling that double-check is needed on transforming to xarray; in particular, the current calculation of `ext_carrier_i` differ from PyPSA-Eur implementation https://github.com/PyPSA/pypsa-eur/blob/6505d78f4e9a1362e8309bcd5b4b8b99ac60337c/scripts/solve_network.py#L350

ekatef:
I can reproduce CI error locally when using `lgpk`, while the workflow works with `cbc`

pz-max:
Cool! Just a comment. I would not follow my Linopy PR only but also check if the PyPSA-EUR main changed some parts.

davide-f:
Hello @ekatef, that's great news! :D
To answer to a question, the use of p_set is the more appropriate way to get access to the demand time series of load.

Regarding solver log, it may make sense to check the latest pypsa-eur implementation as I guess they have accounted for that.

As a remark, this morning we talked with Leon and it turns out that to merge this PR requires changes to the -sec version, which may be delicate also for some of their projects.
So, it is a good idea to investigate and prepare this PR, but it can be merged only when the alignment can be performed also with the Sec version.
What is your feeling?

ekatef:
Hey @davide-f!

Thanks a lot for the detailed analysis :) 

Ah, I see: the load is an input, not a result. So, `n.loads_t.p_set` is in fact more appropriate than `n.loads_t.p`. Thanks for the explanation!

Regarding a potential merge of this PR, absolutely agree that it's crucial to ensure that the changes introduced don't break anything. Actually, I perceive this work as a kind of experiment to estimate how much effort would be needed for transition to `linopy` interface. If you see a potential to align this with requirements of PyPSA-Sec at some stage, that is great and happy to adjust this PR accordingly.

My general impression is that linopy transition is not as challenging as it could be expected from the first sight (thanks to @pz-max PR in PyPSA-Eur and clear instructions)

ekatef:
Some adjustments were done to get closer to PyPSA-Eur implementation. However, `glpk` solution still breaks. The reason seems to be the following command is given to a solver: `'glpsol --lp /var/folders/qn/vpndfm21795ckkq89np1ckp40000gn/T/linopy-problem-p4tiu2uu.lp --output /var/folders/qn/vpndfm21795ckkq89np1ckp40000gn/T/linopy-solve-7tkku64p.sol --solver_logfile /Users/ek/_github_/pypsa-earth/logs/NG_BJ/solve_network/elec_s_6_ec_lcopt_Co2L-4H_solver.log'`

According to a warning, there is no `--solver_logfile` option available for `glpsol`, while it is transferred into`run_glpk` as `solver_options` (not sure yet about the reason).

ekatef:
Results of additional testing:
1) now highs is captured and used;
2) some introduced changes make a testing problem infeasible, and it looks like the reason for it is that the problem formulation somehow gets lost:
```bash
INFO:linopy.solvers:Log file at /tmp/highs.log.
ERROR:   getOptionIndex: Option "solver_logfile" is unknown
Presolving model
Problem status detected on presolve: Infeasible
Model   status      : Infeasible
Objective value     :  0.0000000000e+00
HiGHS run time      :          0.00
WARNING:linopy.constants:Optimization failed: 
Status: warning
Termination condition: infeasible
Solution: 0 primals, 0 duals
Objective: nan
Solver model: available
Solver message: infeasible
```
3) there are some troubles with logfile options for cbc: 

```
No match for solver_logfile - ? for list of commands
No match for logs/MY/solve_network/elec_s_10_ec_lcopt_Co2L-2H_solver.log - ? for list of commands
```

davide-f:
IT may be good to revive this PR and finalize it when possible, to be able to use the latest pypsa version.
How far do you think we are to finalize?

I am aware that you have a lot on your desk right now, no pressure at all :) we can give priorities and tackle them one by one

ekatef:
> IT may be good to revive this PR and finalize it when possible, to be able to use the latest pypsa version. How far do you think we are to finalize?
> 
> I am aware that you have a lot on your desk right now, no pressure at all :) we can give priorities and tackle them one by one

@davide-f  agree :) I'd expect that something about a week of focused work should be enough to resolve the remained questions. As discussed, #919 and #903 seem to have higher priorities now. Also, not sure what is the current status of PyPSA-Earth-Sec in terms of `pyomo/linopy` dependencies.

Anyway, after #919 and #903 will be finished, happy to get back to this PR. Ping me please in case this would get urgent.

ekatef:
Testing with the updated PyPSA version (0.25.2 is used locally).

Clustering is failing currently:
1) the default SCIP solver (copy-pasted from PyPSA-Eur) is not included into PyPSA-Earth environment;
2) `ipopt` is installed but not seen by linopy for some reason.

CI is failing due to conda being unhappy: ` Operation too slow. Less than 30 bytes/sec transferred the last 60 seconds`.

davide-f:
Great work :)
I see it is in work in progress; the CI is complying about the number of clusters; if/once you need a review/check, ping me :)

ekatef:
> Great work :) I see it is in work in progress; the CI is complying about the number of clusters; if/once you need a review/check, ping me :)

Thanks :D Yeah, there have been linopy-related changes in clustering, and some clustering parameters have been shifted. However, as we discussed, I'd focus on making the latest PyPSA version work first (hopefully, we are almost there :D), and adjusted the clustering details after.

Great to have you support and I'll definitely ask your assessment, once there will be some updates here! 🙂

ekatef:
Update: The PR allows to run the workflow with the latest version of PyPSA (0.27.1), but gplk error persists:`INFO:linopy.solvers:Invalid option '--solver_logfile'; try glpsol --help`.

I suspect, that leads to troubles when linopy tries to read solver settings: the line `info = pd.read_csv(info, sep=":", index_col=0, header=None)[1])` triggers pandas issues `pandas.errors.EmptyDataError: No columns to parse from file`.

It looks like some glpk setup should be adjusted.

ekatef:
Testing with `cbc` also fails in some enigmatic way: optimisation itself is successful, while the workflow crashes after with `TypeError: cannot unpack non-iterable NoneType object` in `status, condition = n.optimize.optimize_transmission_expansion_iteratively(**kwargs)`. Interestingly, the solver log is empty. Could it possibly mean some troubles when writing the solving results?

The full listing in terminal looks as follows:
```bash
rule solve_network:
    input: networks/NG_BJ/elec_s_8_ec_lcopt_Co2L-4H.nc
    output: results/NG_BJ/networks/elec_s_8_ec_lcopt_Co2L-4H.nc
    log: logs/NG_BJ/solve_network/elec_s_8_ec_lcopt_Co2L-4H_solver.log, logs/NG_BJ/solve_network/elec_s_8_ec_lcopt_Co2L-4H_python.log
    jobid: 1
    benchmark: benchmarks/NG_BJ/solve_network/elec_s_8_ec_lcopt_Co2L-4H
    reason: Missing output files: results/NG_BJ/networks/elec_s_8_ec_lcopt_Co2L-4H.nc; Input files updated by another job: networks/NG_BJ/elec_s_8_ec_lcopt_Co2L-4H.nc
    wildcards: simpl=, clusters=8, ll=copt, opts=Co2L-4H
    resources: tmpdir=/var/folders/qn/vpndfm21795ckkq89np1ckp40000gn/T, mem=8670
INFO:snakemake.logging:rule solve_network:
    input: networks/NG_BJ/elec_s_8_ec_lcopt_Co2L-4H.nc
    output: results/NG_BJ/networks/elec_s_8_ec_lcopt_Co2L-4H.nc
    log: logs/NG_BJ/solve_network/elec_s_8_ec_lcopt_Co2L-4H_solver.log, logs/NG_BJ/solve_network/elec_s_8_ec_lcopt_Co2L-4H_python.log
    jobid: 1
    benchmark: benchmarks/NG_BJ/solve_network/elec_s_8_ec_lcopt_Co2L-4H
    reason: Missing output files: results/NG_BJ/networks/elec_s_8_ec_lcopt_Co2L-4H.nc; Input files updated by another job: networks/NG_BJ/elec_s_8_ec_lcopt_Co2L-4H.nc
    wildcards: simpl=, clusters=8, ll=copt, opts=Co2L-4H
    resources: tmpdir=/var/folders/qn/vpndfm21795ckkq89np1ckp40000gn/T, mem=8670

INFO:snakemake.logging:
Restricted license - for non-production use only - expires 2025-11-24
~/miniconda3/envs/pypsa-earth-upg4/lib/python3.10/site-packages/pulp/apis/gurobi_api.py:170: UserWarning: GUROBI error: Unknown parameter '_test'.
  warnings.warn("GUROBI error: {}.".format(e))
Building DAG of jobs...
WARNING:snakemake.logging:Building DAG of jobs...
Using shell: /bin/bash
WARNING:snakemake.logging:Using shell: /bin/bash
Provided cores: 1 (use --cores to define parallelism)
WARNING:snakemake.logging:Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
WARNING:snakemake.logging:Rules claiming more threads will be scaled down.
Provided resources: mem=8670
WARNING:snakemake.logging:Provided resources: mem=8670
Select jobs to execute...
WARNING:snakemake.logging:Select jobs to execute...
Changing to shadow directory: _github_/pypsa-earth/.snakemake/shadow/tmp1q_7j4bm
WARNING:snakemake.logging:Changing to shadow directory: _github_/pypsa-earth/.snakemake/shadow/tmp1q_7j4bm
INFO:pypsa.io:Imported network elec_s_8_ec_lcopt_Co2L-4H.nc has buses, carriers, generators, global_constraints, links, loads, storage_units, stores
No expandable lines found. Skipping iterative solving.
INFO:__main__:No expandable lines found. Skipping iterative solving.
INFO:linopy.model: Solve problem using Cbc solver
INFO:linopy.model:Solver options:
 - solver_logfile: logs/NG_BJ/solve_network/elec_s_8_ec_lcopt_Co2L-4H_solver.log
 - store_basis: True
INFO:linopy.io: Writing time: 0.13s
INFO:linopy.solvers:No match for solver_logfile - ? for list of commands
No match for logs/NG_BJ/solve_network/elec_s_8_ec_lcopt_Co2L-4H_solver.log - ? for list of commands
No match for store_basis - ? for list of commands
No match for True - ? for list of commands
Total time (CPU seconds):       0.07   (Wallclock seconds):       0.08


INFO:linopy.constants: Optimization successful: 
Status: ok
Termination condition: optimal
Solution: 3779 primals, 7761 duals
Objective: 3.15e+09
Solver model: not available
Solver message: Optimal - objective value 3154337070.51124573


~/miniconda3/envs/pypsa-earth-upg4/lib/python3.10/site-packages/pypsa/optimization/optimize.py:357: FutureWarning: A value is trying to be set on a copy of a DataFrame or Series through chained assignment using an inplace method.
The behavior will change in pandas 3.0. This inplace method will never work because the intermediate object on which we are setting values always behaves as a copy.

For example, when doing 'df[col].method(value, inplace=True)', try using 'df.method({col: value}, inplace=True)' or df[col] = df[col].method(value) instead, to perform the operation inplace on the original object.


  n.df(c)[attr + "_opt"].update(df)
INFO:pypsa.optimization.optimize:The shadow-prices of the constraints Generator-fix-p-lower, Generator-fix-p-upper, Generator-ext-p-lower, Generator-ext-p-upper, Link-ext-p-lower, Link-ext-p-upper, Store-ext-e-lower, Store-ext-e-upper, StorageUnit-fix-p_dispatch-lower, StorageUnit-fix-p_dispatch-upper, StorageUnit-fix-p_store-lower, StorageUnit-fix-p_store-upper, StorageUnit-fix-state_of_charge-lower, StorageUnit-fix-state_of_charge-upper, StorageUnit-energy_balance, Store-energy_balance were not assigned to the network.
INFO:pypsa.optimization.abstract:Mean square difference after iteration 1 is nan
INFO:pypsa.optimization.abstract:Running last lopf with fixed branches (HVDC links and HVAC lines)
INFO:linopy.model: Solve problem using Cbc solver
INFO:linopy.model:Solver options:
 - solver_logfile: logs/NG_BJ/solve_network/elec_s_8_ec_lcopt_Co2L-4H_solver.log
 - store_basis: True
INFO:linopy.io: Writing time: 0.13s
INFO:linopy.solvers:No match for solver_logfile - ? for list of commands
No match for logs/NG_BJ/solve_network/elec_s_8_ec_lcopt_Co2L-4H_solver.log - ? for list of commands
No match for store_basis - ? for list of commands
No match for True - ? for list of commands
Total time (CPU seconds):       0.07   (Wallclock seconds):       0.08


INFO:linopy.constants: Optimization successful: 
Status: ok
Termination condition: optimal
Solution: 3779 primals, 7761 duals
Objective: 3.15e+09
Solver model: not available
Solver message: Optimal - objective value 3154337070.51124573


~/miniconda3/envs/pypsa-earth-upg4/lib/python3.10/site-packages/pypsa/optimization/optimize.py:357: FutureWarning: A value is trying to be set on a copy of a DataFrame or Series through chained assignment using an inplace method.
The behavior will change in pandas 3.0. This inplace method will never work because the intermediate object on which we are setting values always behaves as a copy.

For example, when doing 'df[col].method(value, inplace=True)', try using 'df.method({col: value}, inplace=True)' or df[col] = df[col].method(value) instead, to perform the operation inplace on the original object.


  n.df(c)[attr + "_opt"].update(df)
INFO:pypsa.optimization.optimize:The shadow-prices of the constraints Generator-fix-p-lower, Generator-fix-p-upper, Generator-ext-p-lower, Generator-ext-p-upper, Link-ext-p-lower, Link-ext-p-upper, Store-ext-e-lower, Store-ext-e-upper, StorageUnit-fix-p_dispatch-lower, StorageUnit-fix-p_dispatch-upper, StorageUnit-fix-p_store-lower, StorageUnit-fix-p_store-upper, StorageUnit-fix-state_of_charge-lower, StorageUnit-fix-state_of_charge-upper, StorageUnit-energy_balance, Store-energy_balance were not assigned to the network.
ERROR:_helpers:An error happened in module '_github_/pypsa-earth/scripts/solve_network.py', function 'solve_network': cannot unpack non-iterable NoneType object
  File "_github_/pypsa-earth/.snakemake/shadow/tmp1q_7j4bm/.snakemake/scripts/tmpidm9hrsp.solve_network.py", line 596, in <module>
    n = solve_network(
  File "_github_/pypsa-earth/.snakemake/shadow/tmp1q_7j4bm/.snakemake/scripts/tmpidm9hrsp.solve_network.py", line 544, in solve_network
    status, condition = n.optimize.optimize_transmission_expansion_iteratively(**kwargs)
TypeError: cannot unpack non-iterable NoneType object
Not cleaning up _github_/pypsa-earth/.snakemake/shadow/tmp1q_7j4bm/.snakemake/scripts/tmpidm9hrsp.solve_network.py
CRITICAL:snakemake.logging:Not cleaning up _github_/pypsa-earth/.snakemake/shadow/tmp1q_7j4bm/.snakemake/scripts/tmpidm9hrsp.solve_network.py
[Wed Apr 10 08:32:22 2024]
INFO:snakemake.logging:[Wed Apr 10 08:32:22 2024]
Error in rule solve_network:
    jobid: 0
    input: networks/NG_BJ/elec_s_8_ec_lcopt_Co2L-4H.nc
    output: results/NG_BJ/networks/elec_s_8_ec_lcopt_Co2L-4H.nc
    log: logs/NG_BJ/solve_network/elec_s_8_ec_lcopt_Co2L-4H_solver.log, logs/NG_BJ/solve_network/elec_s_8_ec_lcopt_Co2L-4H_python.log (check log file(s) for error details)

ERROR:snakemake.logging:Error in rule solve_network:
    jobid: 0
    input: networks/NG_BJ/elec_s_8_ec_lcopt_Co2L-4H.nc
    output: results/NG_BJ/networks/elec_s_8_ec_lcopt_Co2L-4H.nc
    log: logs/NG_BJ/solve_network/elec_s_8_ec_lcopt_Co2L-4H_solver.log, logs/NG_BJ/solve_network/elec_s_8_ec_lcopt_Co2L-4H_python.log (check log file(s) for error details)

RuleException:
CalledProcessError in file _github_/pypsa-earth/Snakefile, line 837:
Command 'set -euo pipefail;  ~/miniconda3/envs/pypsa-earth-upg4/bin/python3.10 _github_/pypsa-earth/.snakemake/shadow/tmp1q_7j4bm/.snakemake/scripts/tmpidm9hrsp.solve_network.py' returned non-zero exit status 1.
  File "_github_/pypsa-earth/Snakefile", line 837, in __rule_solve_network
  File "~/miniconda3/envs/pypsa-earth-upg4/lib/python3.10/concurrent/futures/thread.py", line 58, in run
CalledProcessError in file _github_/pypsa-earth/Snakefile, line 837:
Command 'set -euo pipefail;  ~/miniconda3/envs/pypsa-earth-upg4/bin/python3.10 _github_/pypsa-earth/.snakemake/shadow/tmp1q_7j4bm/.snakemake/scripts/tmpidm9hrsp.solve_network.py' returned non-zero exit status 1.
  File "_github_/pypsa-earth/Snakefile", line 837, in __rule_solve_network
  File "~/miniconda3/envs/pypsa-earth-upg4/lib/python3.10/concurrent/futures/thread.py", line 58, in run
```

ekatef:
> Great work :) I see it is in work in progress; the CI is complying about the number of clusters; if/once you need a review/check, ping me :)

@davide-f When you have time, it would be great to have your opinion on what could go wrong with solver set-up for linopy. I have tested `glpk` and `cbc` with the latest PyPSA version of `0.27.1`, but both fail due to some troubles which look like some misunderstanding between a solver or linopy and/or the system itself.

Any ideas on possible workarounds, would be very helpful!

scripts/solve_network.py
```diff
@@ -555,24 +576,23 @@ def solve_network(n, config, opts="", **kwargs):
...
+    #        n.lines.index.str.contains("new"), "s_nom_min"
+    #    ] = snakemake.params.augmented_line_connection.get("min_expansion")
+    n = prepare_network(n, solve_opts, config=solve_opts)
     n = solve_network(
         n,
         config=snakemake.config,
-        opts=opts,
-        solver_dir=tmpdir,
+        solving=snakemake.params.solving,
         solver_logfile=snakemake.log.solver,
```
davide-f: Please revise as discussed:
```suggestion
        log_fn=snakemake.log.solver,
```

davide-f (commented):
Great @ekatef :D
Together we have move forward. As discussed there is a minor comment for solve_network.
Moreover, pypsa is complaining with unnecessary columns that start with "bus" thus in base_network it would be good to drop those columns before importing the lines.

That should fix or improve the problem (hopefully/praying) :D

davide-f:
Ah @ekatef , we have merge conflicts; once they are solved, the CI is executed again

pz-max:
Just for historical updates as I observed the PyPSA-Eur merge.
- we don't have to introduce Linopy here now and can do it in the merge (PyPSA-Earth-Sec would break if we introduce linopy here - but I am sure you tracking this)
- we can use the exact PyPSA-Eur equivalents, saving a lot of work

((so wouldn't spend too much time here))

davide-f:
@ekatef It works now!!!! :D

ekatef:
> @ekatef It works now!!!! :D

Fantastic news!!! :))))
Still struggling with local testing (probably, an environment update is needed...), but trust to you and CI :D

ekatef:
> @ekatef It works now!!!! :D

@davide-f I confirm that it works!! 🎉 🎉 🎉 
The trick has been to update the environment 😄 

Thanks a lot for the final fixes!

Pre-commit is not happy as yaml-linter complains on the lines wider that 88. But I'm not sure it makes sense to fix it, as it is not linked with the current PR and would require to completelly re-structure the config.

ekatef:
As a nice side-effect: PyPSA-Earth now works with HiGHS 🥳 

image

davide-f:
Amazing!

ekatef:
The major points left:
- [x] do not revise implementation of `add_operational_reserve_margin` (in PyPSA-Earth decomposed on two functions which look a way more clear as compared with PyPSA-Eur);
- [x] check implementation of `add_battery_constraints`. Currently, it's different in PyPSA-Earth and PyPSA-Eur, which may it be the effect of the sec-merge (which imply the need to revise implementation after PyPSA-Earth merge);
- [x] `add_RES_constraints` must be refactored and reimplemented;
- [x] `solve_network` should be checked: that may be the point where PyPSA-Earth may need a slightly different approach as one used in PyPSA-Eur may be not the most clear one from the usability perspective.

ekatef:
Update: testing successful in the base-case configuration, but there is a problem when fetching isolated networks (something with `p` key which is expected but not found in the provided data structures).

ekatef:
A couple of notes on `linopy` grammar:
1. `pypsa.Network()` doesn't initialise `model` property -> an explicit call `n.optimize.create_model()` is needed
2. Multiplication is not commutative in the current `linopy` implementation: `n.model["Generator-p"] * n.snapshot_weightings.generators` works, while `n.snapshot_weightings.generators` * n.model["Generator-p"]` throws an error.
3. Not sure what is the way to make `groupby(some_grouper).sum()` work on linopy variables.

ekatef:
## Local testing on reproducibility of objective value
### main version
> INFO:__main__:Objective function: 5813422825.0
> INFO:__main__:Objective constant: 2319283737.695124

### linopy version
> INFO:__main__:Objective function: 5473150322.268269
> INFO:__main__:Objective constant: 2319359288.9025226

That means ~6% difference for the objective function and basically same value for the objective constant (1e-3%, but here is rather a question of numerical precision). So, the PR seems to work properly.

As a usability note, that is not obvious at all from the CI logs. The PR changes format of solver traceback due to different conventions of `lopf` and `linopy`, which makes finding correspondance quite tricky.

ekatef:
@finozzifa would be very grateful if you would have time to check this PR.

Pre-commit is currently unhappy due to excessively long lines in the configuration file. But that won't be fixed in this PR.

Another concern is backward compatibility. I'm currently looking into that, but not sure if it'll be possible to find a code solution for that.

ekatef:
As discussed with @davide-f, the backward compatibility issues can be tackled by maintenance measures and adding versioning of the configuration files. A possible implementation for versioning #1058.

This PR is ready for review.

config.tutorial.yaml
```diff
@@ -405,6 +405,63 @@ solving:
...
+      threads: 4
+      lpmethod: 4 # barrier
+      solutiontype: 2 # non basic solution, ie no crossover
+      barrier.convergetol: 1.e-4 #5
+      feasopt.tolerance: 1.e-5 #6
+    cbc-default: {} # Used in CI
+    glpk-default: {} # Used in CI
+
+  mem: 30000 #memory in MB; 20 GB enough for 50+B+I+H2; 100 GB for 181+B+I+H2
+  walltime: "12:00:00"
```
davide-f: In pypsa-eur the option is named "runtime" and has a slightly different naming option, e.g. 12h, though it may be compatible.
Do you think we can adopt the same naming and writing convention?

scripts/base_network.py
```diff
@@ -513,6 +513,18 @@ def base_network(
 
     if hvdc_as_lines_config:
         lines = pd.concat([lines_ac, lines_dc])
+        lines.drop(
```
davide-f, line 516: Generally, I'd recommend to do the converse:
keep the list of columns that we want and implicitly drop the others.
That should be more consistent. However, do you think it should be placed here or maybe in previous processing script?
If you consider that it is not suited for this PR, we can add an issue instead

scripts/base_network.py
```diff
@@ -522,6 +534,18 @@ def base_network(
             axis=1,
             result_type="reduce",
         )
+        lines_ac.drop(
```
davide-f, line 537: Same as above

scripts/cluster_network.py
```diff
@@ -605,12 +580,20 @@ def clustering_for_n_clusters(
...
         line_length_factor=line_length_factor,
+        line_strategies=line_strategies,
         generator_strategies=generator_strategies,
+        bus_strategies={
+            "lat": "mean",
+            "lon": "mean",
+            "tag_substation": "first",
+            "tag_area": "first",
+            "country": "first",
+        },
```
davide-f, line 595: This is a bit widely duplicated across scripts. v_nom is also not here, may this be somewhat linked to the issue by Emmanuel?

scripts/simplify_network.py
```diff
@@ -975,14 +1000,14 @@ def merge_isolated_nodes(n, threshold, aggregation_strategies=dict()):
...
-        for p in aggregation_strategies.keys()
-    }
+    # aggregation_strategies = snakemake.params.cluster_options.get(
+    #    "aggregation_strategies", {}
+    # )
+    ## translate str entries of aggregation_strategies to pd.Series functions:
+    # aggregation_strategies = {
+    #    p: {k: getattr(pd.Series, v) for k, v in aggregation_strategies[p].items()}
+    #    for p in aggregation_strategies.keys()
+    # }
```
davide-f, line 1011: Can you explain?
No more needed?

scripts/solve_network.py
```diff
@@ -85,21 +85,16 @@
...
-from pypsa.linopf import (
-    define_constraints,
-    define_variables,
-    get_var,
-    ilopf,
-    join_exprs,
-    linexpr,
-    network_lopf,
-)
+from pypsa.optimization.abstract import optimize_transmission_expansion_iteratively
```
davide-f, line 90: Have you locally tested the features here changed? I feel like not all of them are parsed into the CI [which is not good too...]

scripts/solve_network.py
```diff
@@ -85,21 +85,16 @@
...
+from pypsa.optimization.optimize import optimize
+from vresutils.benchmark import memory_logger
 
 logger = create_logger(__name__)
 
 
-def prepare_network(n, solve_opts):
+def prepare_network(n, solve_opts, config=None):
```
davide-f: why optional?

davide-f (commented):
Added some first comments :) Great work.
Indeed this PR is quite large too already, let's find the best way to finalize it :)
Let's remember to add a line in the release_note too

scripts/base_network.py
ekatef: Absolutely agree that explicit definition would work better there. Moreover, the updated PyPSA version is not very stable in what relates to the data structures. Custom columns are not forbidden explicitly, but can lead to some weird issues. So, I'd be very careful in dealing with them.

In a dedicated PR, have moved [the definition of lines columns](https://github.com/ekatef/pypsa-earth/blob/729fecf24760dff828a1e7ee159885432108a79d/scripts/build_osm_network.py#L914-L932) in the previous script.

scripts/base_network.py
ekatef: Fixed in #1065: no need for the drop in `base_network` anymore.

scripts/cluster_network.py
ekatef: Agree that the duplication should be fixed. Implemented in #1065 as [an update](https://github.com/ekatef/pypsa-earth/blob/729fecf24760dff828a1e7ee159885432108a79d/scripts/cluster_network.py#L745-L756) of the aggregation strategies outside the clustering functions.

Yeah, it looks like transparence is much desired in working with `v_nom`. Have defined `"first"` aggregation strategy for `v_nom` to have a better overview on what is put inside the model.

scripts/simplify_network.py
ekatef: Yes, the format transformation is not needed anymore. The aggregation strategies are [read directly](https://github.com/ekatef/pypsa-earth/blob/cefc09ce5e6afe2ad5b4a60aa3552dcc376dbcc6/scripts/cluster_network.py#L737) from the config via Snakemake parameters.

Have removed the commented-out parts.

ekatef:
> Added some first comments :) Great work. Indeed this PR is quite large too already, let's find the best way to finalize it :) Let's remember to add a line in the release_note too

@davide-f thanks a lot for the great review! You have made me recognise that PR mixes two different things: PyPSA update and actually linopy transition. Moreover, the updated PyPSA version leads to some odd behaviour in the case when custom columns are present in the components on the network, which should be tracked regardless `linopy` methods.

Have opened #1065 and moved there the changes from this PR which relate to PyPSA update. Would you mind to continue the discussion on the aggregation strategies there? Then, for this PR we can focus on the solving part only. What do you think?

scripts/solve_network.py
ekatef: Yes! I have refactored `add_RES_constraints` and tested it's behaviour which has (obviously) required to add some fixes to account for `linopy` grammar.

There may be some advanced stuff which would be nice to investigate in more depth, but on the level of basic functionality it works.

config.tutorial.yaml
ekatef: Ah, I see! There has been [an update](https://github.com/PyPSA/pypsa-eur/commit/ed71978b92554af6bd240cbc072e2dfabd9e0ad8#diff-40d10ed9bd5fe53ab55516ab64e4ad44b2e05cd19d58908a2d53c05e53c4fd79) in keywords in PyPSA-Eur intended to follow `humanfriendly` conventions. Ok, let's add this for consistency :)

scripts/solve_network.py
ekatef: May you clarify please? Not sure I really get the point 🙂

davide-f: I'd propose to have config a mandatory argument, do we need the "=None"?

scripts/solve_network.py
davide-f: Good as is !

scripts/cluster_network.py
davide-f: Good! What would be your idea here, merge this and then 1065, merge 1065 here?

scripts/base_network.py
davide-f: Great! What would be your idea here, merge this and then 1065, merge 1065 here?

scripts/base_network.py
davide-f: Great! What would be your idea here, merge this and then 1065, merge 1065 here?
(same text also specified below)

envs/environment.yaml
```diff
@@ -12,7 +12,7 @@ dependencies:
 - pip
 - mamba   # esp for windows build
 
-- pypsa>=0.24, <0.25
+- pypsa
```
davide-f, line 15: Please add >=0.2X in agreement to the expected version; it is not backcompatible unfortunately

scripts/solve_network.py
```diff
@@ -210,165 +249,225 @@ def add_EQ_constraints(n, o, scaling=1e-1):
...
+        .sum()
+        .sum("snapshot")
     )
-    lhs_spill = (
-        linexpr(
-            (
-                -n.snapshot_weightings.stores * scaling,
-                get_var(n, "StorageUnit", "spill").T,
-            )
+    # TODO: double check that this is really needed, why do have to subtract the spillage
```
davide-f, line 259: I believe so because in the rhs we have the inflow but not all inflow will be used; spill will be lost.
I don't see charging and discharging efficiencies here, so probably there may be some misalignment because of them.
So this todo should be revised into checking the efficiencies and probably good to raise an issue here.
Colleagues may have experienced minor issues about this maybe

scripts/solve_network.py
```diff
@@ -210,165 +249,225 @@ def add_EQ_constraints(n, o, scaling=1e-1):
...
     CONTINGENCY = reserve_config["contingency"]
 
     # Reserve Variables
-    reserve = get_var(n, "Generator", "r")
-    lhs = linexpr((1, reserve)).sum(1)
+    n.model.add_variables(
+        0, np.inf, coords=[sns, n.generators.index], name="Generator-r"
+    )
+    reserve = n.model["Generator-r"]
+    lhs = reserve.sum("Generator")
```
davide-f, line 365: This seems like RES generators can offer reserve too or am I wrong?

scripts/solve_network.py
```diff
@@ -210,165 +249,225 @@ def add_EQ_constraints(n, o, scaling=1e-1):
...
+    # TODO drop load
     ggrouper = n.generators.bus.map(n.buses.country)
     sgrouper = n.storage_units.bus.map(n.buses.country)
     cgrouper = n.links.bus0.map(n.buses.country)
 
     logger.warning(
-        "The add_RES_constraints functionality is still work in progress. "
-        "Unexpected results might be incurred, particularly if "
+        "The add_RES_constraints() is still work in progress. "
+        "Unexpected results might be incurre, particularly if "
```
davide-f: incurred?

scripts/solve_network.py
```diff
@@ -378,109 +477,67 @@ def add_RES_constraints(n, res_share):
...
     # Generators
+    # TODO restore grouping by countries un-commenting calls of groupby()
     lhs_gen = (
-        linexpr(
-            (n.snapshot_weightings.generators, get_var(n, "Generator", "p")[gens_i].T)
-        )
-        .T.groupby(ggrouper, axis=1)
-        .apply(join_exprs)
+        (n.model["Generator-p"].loc[:, gens_i] * n.snapshot_weightings.generators)
+        # .groupby(ggrouper.to_xarray())
```
davide-f, line 504: what about this todo?

davide-f (commented):
I've added few comments; the PR looks in good state; not sure your idea on how to integrate the changes propsoed in 1045.
A weird think I noticed is that the objective function of the problem has significantly changed, from something like 2.6 10^9 to 3.7e+09
Have you noticed it?

envs/environment.yaml
ekatef: [Done](https://github.com/ekatef/pypsa-earth/blob/40513e79daa0ce53d025ed4b3e52cb6ec91fd3e8/envs/environment.yaml#L15) in #1065

scripts/base_network.py
ekatef: Have created [a PR on top](https://github.com/ekatef/pypsa-earth/pull/1) of #1065 which enables linopy in addition to the upgrade of PyPSA version

scripts/base_network.py
ekatef: > Great! What would be your idea here, merge this and then 1065, merge 1065 here?

Have created [a PR on top](https://github.com/ekatef/pypsa-earth/pull/1) of #1065 which enables linopy in addition to the upgrade of PyPSA version

scripts/cluster_network.py
ekatef: > Good! What would be your idea here, merge this and then 1065, merge 1065 here?

I think it could make sense to have a stacked PR: [a PR on top](https://github.com/ekatef/pypsa-earth/pull/1) of #1065 which enables linopy in addition to the upgrade of PyPSA version

scripts/solve_network.py
ekatef: Frankly speaking, it has been just a copy-paste from PyPSA-Eur. But agree that it feels more natural to have `config` as a mandatory argument. Fixed in [the stacked PR](https://github.com/ekatef/pypsa-earth/pull/1): [this line](https://github.com/ekatef/pypsa-earth/blob/94e634c438597d8f12d11349855e50c2a9a4474f/scripts/solve_network.py#L97)

scripts/solve_network.py
ekatef: True, have over-done fixing the RES function 😄 Fixed

scripts/solve_network.py
ekatef: I had to stop here as `groupby()` has a bit specific dynamics in linopy and feels like a feature under development. In the current version I have not been to make it work, but the behaviour of `groupby()` has just changed recently, and more changes can be expected, as far as I can understand.

Moreover, I'm not quite sure this grouping is really needed. Currently, a share of RES is set for each country in a dedicated constraint, but this share is same for all the countries. I believe there may be some room for improvement here which we can track as a dedicated issue, as that is quite a frequent request to improve representation of RES development plans

scripts/solve_network.py
ekatef: A very interesting point! As far as I understand, yes. We follow PyPSA-Eur in using the approach of GenX when defining the operational reserve constraint. That implies that the reserve margin formulation "optimal dispatch" of RES (whatever it means) and demand response. Have added a short comment on that in the doc string

I fear including wind and solar into the available reserve is against the current regulations in many countries, and it could make sense to have more flexible settings here. Does it sound reasonable for you?

scripts/solve_network.py
ekatef: Agree. I'm also not clear why the `inflow` and `spillage` are split in the current formulation which also doesn't make easier implementation of the efficiencies. In any case, we are following here the formulation of PyPSA-Eur, and it could probably make sense to revise both formulations. Happy to open an issue on that

ekatef:
Thanks a lot for the in-depth review @davide-f! 

Along the way, we have identified a number of the additional issues, both related to this PR and completely independent ones. The list looks like follows

1. Fetching isolated networks gets broken at the stage of applying `get_switchable_as_dense( )` when aggregated storage units and loads. The issue looks quite weird and I have not been able to debug it properly. Currently, it's "fixed" by switching-off the functionality to merge isolated grids. That is not a solution which which I'm happy but it could make sense to tackle it separately.
2. `add_RES_constraints( )` may need further revisions.
3.  A revision [seems to be needed](https://github.com/pypsa-meets-earth/pypsa-earth/pull/796#discussion_r1699738902) for implementation of the hydro-related part in `add_EQ_constraints( )` [both in PyPSA-Earth and PyPSA-Eur]
4. It may be a good idea [to adjust a definition](https://github.com/pypsa-meets-earth/pypsa-earth/pull/796#discussion_r1699758708) of the reserve margin constraint in `add_operational_reserve_margin_constraint( )` to make it compatible with the current regulation status in the most countries of the world.

Regarding the overall implementation strategy, my feeling is that it makes sense to disentangle `PyPSA` upgrade from enabling `linopy`. Have opened #1065 to update PyPSA and [a stacked PR](https://github.com/ekatef/pypsa-earth/pull/1) to enable linopy. Both PRs are mainly cherry-picking from this PR which should hopefully facilitate the final checks.

The PyPSA update leads to ~1% change of the objective function, while enabling linopy leaves the objective function pretty much the same (~1e-5 is the relative difference). So, I think this work is close to be finished, keeping in mind the points above 🙂

ekatef:
Closing in favour #1169

Issue #445, opened 2022-08-23
Missing pieces towards global PyPSA-Earth


davide-f:
## Towards global PyPSA-Earth

In this issue, we track major requirements needed to successfully run the workflow at complete global scale.
I have been running parts of the model using `countries=["Earth"]` and in the following I resume some findings; this list is to be populated by additional comments.

- [x] `download_osm_network `needed some fixes but it generally works; however, I needed to rerun the workflow some times because the procedure got stopped several times because of download limits at the server end. When the user is interested in downloading large areas, it may be better to download the combined/continental chuncks rather than each country; however this leads to less generalizability and duplication if the user is then interested in smaller areas. Alternatively, some delays can be manually inserted to avoid such problem. Some tests may be needed when larger regions are more needed.
- [x] #574
- [ ] `clear_osm_network`: this is the real big deal. Currently, after one day of complete execution, with the "names_by_shapes" option disabled we are still far from solved. The procedure is stuck at `africa_shape.contains`. I noticed that functions on polygons are super heavy to execute and we need to work hard on that. In the following some comments are provided:
   - [x] #601
   - [x] `split_cells` may be removed and use the more general `split_cells_multiple` instead (this is for clarity and to avoid duplication)
   - [ ] In build_shapes, we may preprocess better the regions to avoid extra calculations down the chain. For example, we shall perform better checks on the parameters to simplify_polys that polishes and simplify the shapes. The workflow shall work using those "cleaned" shapes although they may be slightly distorted with respect to the original ones. The original shapes may be saved anyway that can be used. Again in build shapes, we may prepare also the `ext_country_shapes` that is the unary_union of the country_shapes with the  corresponding offshore_shapes that may be used by several scripts.
   - [ ] all geopandas function may be implemented using [dask_geopandas](https://dask-geopandas.readthedocs.io/en/stable/) to support multithreading
   - [ ] all shape comparison in the workflows may be revised to preprocess the geometries and avoid detailed comparisons for only where it is necessary
   - [ ] functions that compare shapes shall be approximated or simplified. For example, the country flag to a line (if necessary) can be calculated using one of the two substations the line is connected to.
- [ ] `build_osm_network` is going to be another big deal as we have the functions `set_substations_ids`, `fix_overpassing_lines` and `set_lines_ids` that generally take a long time already for the africa model and their complexity is O(n^2), we shall improve that.
[...]

PR #799, opened 2023-07-19
optimize set_substation_ids using DBSCAN (750x time faster) 🚀

mnm-matin:
optimizes set_substation_ids #445

For US buses it would take 5 min, now it only takes 0.4 sec thanks to DBSCAN from sklearn. This is possible using a tree structure (BallTree) to represent the spatial index.
The approach can be further optimized by changing the params of DBSCAN such as the number of processes (maybe set it to -1). Setting min_samples to 0 can allow one to identify outliers that have a low density of points around them.

Note: only for US, the results are a little off from the results of the original set_substations_ids, but the number of unique cluster is actually lower so it should be better (22k compared to 24k). 

Also Note: using dbscan for this task is a bit like using a rocket launcher to kill a bird

Here is a less optimized version that is also much faster and does not use sklearn or dbscan. This approach can be extended to the snapping buses to line as well
```python
def set_substation_ids_eff(buses, distance_crs, tol=2000):
    """
    Function to set substations ids to buses, accounting for location
    tolerance.
    """
    buses = buses.copy()
    buses["station_id"] = -1

    # convert the geometry to the specified CRS for accurate distance calculation
    buses.geometry = buses.geometry.to_crs(distance_crs)

    # create a spatial index for efficient spatial queries
    spatial_index = buses.sindex

    station_id = 0
    for i, bus in buses.iterrows():
        if bus["station_id"] >= 0:
            continue

        # find the nearest buses within the specified tolerance
        possible_matches_index = list(spatial_index.intersection(bus.geometry.buffer(tol).bounds))
        possible_matches = buses.loc[possible_matches_index]
        precise_matches = possible_matches[possible_matches.distance(bus.geometry) <= tol]

        if precise_matches["station_id"].max() < 0:
            # when all station_ids are negative, then this is a new station id
            buses.loc[precise_matches.index, "station_id"] = station_id
            station_id += 1
        elif precise_matches["station_id"].min() < 0:
            # otherwise, when at least a station_id is non-negative, then pick the first value
            sub_id = precise_matches["station_id"].max()
            buses.loc[precise_matches.index, "station_id"] = sub_id

    return buses
```
thanks to @davide-f  for providing the files

notebook I used for dev: https://github.com/mnm-matin/pypsa-africa/blob/substation_ids/on_pr.ipynb

### DIFF
--- a/scripts/build_osm_network.py
+++ b/scripts/build_osm_network.py
@@ -15,8 +15,8 @@ from _helpers import configure_logging, read_geojson, sets_path_to_root, to_csv_
 from config_osm_data import osm_clean_columns
 from shapely.geometry import LineString, Point
 from shapely.ops import linemerge, split
-from tqdm import tqdm
 from sklearn.cluster import DBSCAN
+from tqdm import tqdm
 
 logger = logging.getLogger(__name__)
 
@@ -63,7 +63,7 @@ def set_substations_ids(buses, distance_crs, tol=5000):
     db = DBSCAN(eps=tol, min_samples=1).fit(coords)
 
     # Add the cluster labels to the GeoDataFrame
-    buses['station_id'] = db.labels_
+    buses["station_id"] = db.labels_
 
 
 def set_lines_ids(lines, buses, distance_crs):
--- a/scripts/build_osm_network.py
+++ b/scripts/build_osm_network.py
@@ -16,6 +16,7 @@ from config_osm_data import osm_clean_columns
 from shapely.geometry import LineString, Point
 from shapely.ops import linemerge, split
 from tqdm import tqdm
+from sklearn.cluster import DBSCAN
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


mnm-matin:
not sure why the notebook got added in the pr, will remove it

davide-f:
Great @mnm-matin ! I quite like the approach and it is efficient (also in terms of lines of code :))!!!
Moreover we already have sklearn in the env so, it is quite smart, though, I'd make it explicit in the environment.

Regarding the notebook. Could you please either create a new PR with only the intended changes or force-push here only the intended changes without the notebook?
Otherwise the notebook remains in the history.

Would you be interested in adding the equivalent feature for grouping the substations? I think you could leverage a lot on the effort you made and improve the formulation quite a lot :)

mnm-matin:
yeah, accidentally pushed the notebook, IU have removed it now.
Would be nice to merge it and see if results remain as expected.
I believe this pr is for grouping substations, do you mean set_lines_ids?

incase I don't have enough time to come back to this, the following code snippet should help with snapping point to lines (fix_overpassing_lines):

```python
df_l = ng_lines.copy()
df_p = ng_buses.copy()

tolerance = 5000

    
# Buffer points to create areas for spatial join 
buffer_df = gpd.GeoDataFrame(geometry=df_p.buffer(tolerance))
    
# Spatial join to find lines intersecting point buffers
joined = gpd.sjoin(df_l, buffer_df, how="inner", op='intersects')
    
# Group by index to find 
grouped = joined.groupby('index_right')
```

basically it adds the index of the buses  (that are in close proximity) to the lines. This can be post-processed to reach the desired formats. The important part is using spatial indeces (sindex) and using spatial joins (sjoin).

davide-f:
Great matin! :D

So, I see two issues:
1. set_lines_ids that is actually a thorn and can be made more efficient, few ideas on that and
2. fix_overpassing_lines as you mentioned. That code seems highly efficient and very interesting to identify the overpassing lines; it is also important to track what are the buses that are related to that. However, for the current implementation all lines get selected: there is the need to ignore the ending and starting nodes by line, which is a bit tricky. However, the spatial join is interesting and to definitely consider

davide-f (approved):
This PR is ready to go, before merging, I'd wait for the other PRs and in particular https://github.com/pypsa-meets-earth/pypsa-earth/pull/804 to be finalized.
This PR is addressing the v1.0 and it would be nice to keep the PRs more organized

davide-f:

PR #807, opened 2023-07-22
re-write fix_overpassing_lines (800x faster) 🚀

mnm-matin:
#445 

Basically, the function fix_overpassing_lines now runs at least 800 times faster. For US lines, this would take 5.8 hours but now takes 22 seconds. The fix_overpassing_lines function is responsible for snapping buses that are in close proximity to lines.
Notebook available here: https://github.com/mnm-matin/pypsa-africa/blob/notebook_stash/lines_pr.ipynb

This was possible by leveraging geopandas spatial join. 
In short:
i) We add a small radius (buffer) to each point
ii) Use the spatial join to find lines that would interect the point
iii) Then group by the lines to find which points would intersect with each line
iv) the rest is about splitting those lines into MultiLineString by drawing perpendiculars from the point to the line
v) Exploding the dataframe using .explode() to convert MultiLineString into separate LineStrings

Notes:
- Some architectural changes are required, such as the following
- It is good practices to explode the dataframe at the very last steps of pre-processing. 
- Line endings should be converted to buses outside of this function
- Buses should not be returned as they are not changed

Small Caveat: Currently, each line in the MultiLineString ends at the perpendicular intersection point. This should be replaced with the actual point, except if a point intersects the starting point of the line. Not a difficult thing to do, but should have little impact on the actual network. The impact of this entire function on the US network is also questionable and should be investigated further.

Notebook for development:
https://github.com/mnm-matin/pypsa-africa/blob/notebook_stash/lines_pr.ipynb

### DIFF
--- a/scripts/build_osm_network.py
+++ b/scripts/build_osm_network.py
@@ -13,8 +13,8 @@ import numpy as np
 import pandas as pd
 from _helpers import configure_logging, read_geojson, sets_path_to_root, to_csv_nafix
 from config_osm_data import osm_clean_columns
-from shapely.geometry import LineString, Point
-from shapely.ops import linemerge, split
+from shapely.geometry import LineString, Point, MultiLineString
+from shapely.ops import linemerge, nearest_points
 from tqdm import tqdm
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


scripts/build_osm_network.py
```diff
@@ -595,131 +595,97 @@ def create_station_at_equal_bus_locations(
...
-        temp_list = [split(l, p) for l in list_linestrings]
-        # nest all geometries
-        list_linestrings = [lstring for tval in temp_list for lstring in tval.geoms]
+    # change crs to distance based
+    df_l = df_l.to_crs(distance_crs)
+    df_p = df_p.to_crs(distance_crs)
 
-    return list_linestrings
+    # Buffer points to create areas for spatial join
+    buffer_df = gpd.GeoDataFrame(geometry=df_p.buffer(tol))
```
davide-f: it would be better to add the column/index of the original dataframe; probably doing `df_p.buffer(tol).to_frame()` should be enough.
This enables keeping track of the bus id that is matched during the join

scripts/build_osm_network.py
```diff
@@ -595,131 +595,97 @@ def create_station_at_equal_bus_locations(
...
+        Tolerance in meters to snap the buses to the lines
+
+    Returns
+    -------
+    lines : GeoDataFrame
+        GeoDataFrame containing the lines
     """
 
-    list_linestrings = [linestring]
+    df_l = lines.copy()  # can use lines directly without copying
```
davide-f, line 629: for the purpose of this function the columns that may be needed are bus0, bus1 and geometry beyond the index.
However, if we preserve as we should the index, we can always recover bus0 and bus1 from the original dataframe, so, here probably keeping geomerty and the index would be enough.
Probably, we could subset this dataframe as well to increase performances

scripts/build_osm_network.py
```diff
@@ -595,131 +595,97 @@ def create_station_at_equal_bus_locations(
...
-    lines : GeoDataFrame
-        Geodataframe of lines
-    buses : GeoDataFrame
-        Geodataframe of substations
-    tol : float
-        Tolerance in meters of the distance between the substation and the line
-        below which the line will be split
-    """
+    # iterate over the groups, TODO: change to apply
+    for i, group in group_lines:
```
davide-f, line 653: For every line, I expect that the starting (bus0) and ending (bus1) buses are always matched.

It may make sense to perform a preliminary groupby and apply to get for each line, the bus ids that are overpassing by performing a set difference between the matched buses ids by the sjoin and the [bus0, bus1] that are available for every line in the original lines geo dataframe

the first preprocessing should be exactly excluding the lines in group_lines that correspond to the start and end buses.
By doing so, we could save a lot of computations. I don't expect many overpassing nodes

scripts/build_osm_network.py
```diff
@@ -595,131 +595,97 @@ def create_station_at_equal_bus_locations(
...
+        line_id = group["id"].iloc[0]  # pick the line id that represents the group
```
davide-f: may line_id be the i of the forloop? probably not because of the different indexing, may be worth understanding if that makes sense

davide-f (commented):
Great @mnm-matin :D
I've added some comments; happy to investigate this further with you.
The CI is not passing too unfortunately, some improvements may be needed

scripts/build_osm_network.py
mnm-matin: Those indexes would appear under the 'index_right' column of the resulting dataframe?

scripts/build_osm_network.py
mnm-matin: buses that only match the line endings exactly are not split in the current implementation. See the notebook for more detail.

scripts/build_osm_network.py
mnm-matin: good point, i is actually equivalent to line_id (in the next line).

scripts/build_osm_network.py
```diff
@@ -595,131 +595,97 @@ def create_station_at_equal_bus_locations(
...
-            buses_locs = buses.geometry.loc[bus_in_tol_epsg.index]
+        # TODO: replace the end points of each line in the multistring with the points that intersect with the line
+        # unless there is a point that is intersecting the start point of the line
+        # in that case replace the start point with the point that is intersecting the start point of the line
 
-            # get new line geometries
-            new_geometries = _split_linestring_by_point(lines.geometry[l], buses_locs)
-            n_geoms = len(new_geometries)
+        # replace the line with the split line
+        df_l.loc[df_l["id"] == line_id, "geometry"] = split_line
```
mnm-matin: one could replace df_l with lines, but up to you

scripts/build_osm_network.py
```diff
@@ -595,131 +595,97 @@ def create_station_at_equal_bus_locations(
...
-
-    if not lines_to_add:
-        return lines, buses
-
-    df_to_add = gpd.GeoDataFrame(pd.concat(lines_to_add, ignore_index=True))
-    df_to_add.set_crs(lines.crs, inplace=True)
-    df_to_add.set_index(lines.index[-1] + df_to_add.index, inplace=True)
+    # update line endings (included for completion, the scope of the function should be limited to fixing overpassing lines)
+    # commented out due to errors in the bus conversion function
+    # df_l = line_endings_to_bus_conversion(df_l)
```
mnm-matin, line 710: is this required? can we do the conversion after

scripts/build_osm_network.py
mnm-matin: I've updated the index to be the bus_id and then do the join. This should keep track of the bus_id. Well spotted!

scripts/build_osm_network.py
mnm-matin: i have removed the sub setting for buses and lines for now.

scripts/build_osm_network.py
```diff
@@ -684,6 +691,9 @@ def fix_overpassing_lines(lines, buses, distance_crs, tol=1):
     # return to original crs
     df_l = df_l.to_crs(lines.crs)
 
+    # remove lines that are rings (included for completion), TODO: this should be a separate function
+    df_l = df_l[~df_l.geometry.is_ring].reset_index()
```
mnm-matin: there was one line that was a ring after splitting and causing problems with the .boundries function in set_line_id

scripts/build_osm_network.py
```diff
@@ -620,23 +620,29 @@ def fix_overpassing_lines(lines, buses, distance_crs, tol=1):
...
 
+    line_id_str = "line_id"
+    bus_id_str = "bus_id"
+
     # change crs to distance based
     df_l = df_l.to_crs(distance_crs)
     df_p = df_p.to_crs(distance_crs)
 
+    # set index to bus_id
+    df_p.set_index(bus_id_str, inplace=True)
```
mnm-matin, line 641: now, bus_ids should propagate.

davide-f:
Nice matin! :)
I'll investigate the notebook as well, I'll get back in the coming days

scripts/build_osm_network.py
davide-f: Interesting, a question: was the ring already there or how was it created?

scripts/build_osm_network.py
davide-f: Yeah this function can be done as a last step to keep the values of the columns coherent

scripts/build_osm_network.py
davide-f: Actually, checking the notebook, they somehow get split into kind-of null geometries.

When we do `line_geom.difference(MultiLineString(perpendicular_lines))` the difference between line_geom and each linestring applies.
The output is a multilinestring of 4 geometries, of which 2 are basically empty and the 2 are those plotted.
Unfortunately, what we would like to have is 3 geometries representing the connection of the buses to the closest points and the closest points among each other, which unfortunately doesn't work.
However the idea is very good and promising, some refining may be needed but we are there :)

mnm-matin: Nice find. What's happening is that points that are not the ends are not split, but points in the middle that lie exactly on the line are also not split. I have changed the perpendicular line to be a bisector, now buses in the middle (of a line) that lie exactly on the line are also split. Also updated the notebook (https://github.com/mnm-matin/pypsa-africa/blob/notebook_stash/lines_pr.ipynb)

scripts/build_osm_network.py
mnm-matin: I believe the ring structure was there, but it was connected to a line so the shapely boundary function still worked, but once split it becomes exclusively a ring and causes problems.

davide-f: Could be nice to plot that and understand what happens as it may happen more. Potentially we could fix that by replacing the ring with a line maybe

davide-f:
Hello @mnm-matin, I was wondering the status of this PR and how we proceed on that.
I feel that there is a missing element that is the explode of the dataframe for splitting the cutted lines and work around that.
I was wondering what's your feeling here.

mnm-matin:
The explode of the dataframe is already implemented, the lines that were not being split are being split now with the latest commit. I have updated the notebook as well. You're right, there might be some structural changes required in the whole script, some are mentioned in this commit and some as comments, but this PR should serve as drop-in replacement for now.

scripts/build_osm_network.py
mnm-matin: Rings are only a problem because of the subsequent function (set_line_id) that gets called that uses the .boundary() function. The .boundary() causes problems with rings and is really an edge case, that should be tested for in the set_line_id. I have provided a one line fix to make the CI pass for now and added comments in the code. fyi, only one ring was removed in the test case, and i don't expect it to be common occurrence.

davide-f:
Sorry, my bad! I didn't see that.
Very cool! Then, I may need to locally debug the PR and see what happes.
Have you compared the input/output with respect to the original version?

mnm-matin:
I checked the count for Nigeria, and it was quite close. Would suggest debugging this PR together with other PRs for this script.

davide-f:
I'm debugging this as well with some intermediate checks as well.
I'll be checking input and output of fix_overpassing lines for NG. I tried for US but there are fiona issues (https://github.com/geopandas/geopandas/issues/1490) that do not enable that easily.

@mnm-matin After that, I'd propose to merge #799 #807 and #808 all together and then making a test for US with that.
What do you think?
The overall PR is going to be impossible to review, so we better crossreview these PR first before the merge.
To me #799 and #808 are fine as they are already

davide-f:
Mmmm not sure why, but for Nigeria and benin (tutorial), the following applies:

- original lines dataframe shape before fix_overpassing_lines: (478, 20)
- lines dataframe after fix_overpassing lines - old implementation: (497, 20)
- shape lines after fix_overpassing_lines - this PR: (480, 21)

The additional column is a "index" that during some reindex is placed there, not difficult to revise.
However, while the old implementation extended the dataframe by about 20 lines, the new implementation extends it only by 2,  that seems a bit too low.
I think some investigation is needed

davide-f:
Additionally, I see that we currently do the groupby on line_id and bus_id. The bus_id is created by us so should be unique, but line_id is obtained from OSM. That as well should be unique, but I wouldn't rely on that.
We could do the groupby on the index of buses and lines and that should work and by more robust. What do you think?

davide-f:
@mnm-matin I've introduced little changes like groubpy by index and the revision of the line_id to account for the index part.
The old code had that feature that was dropped in this case.

Moreover, to better debug what happens, instead of geojson, I decided to save files into csv so not to incur in fiona issues.
In particular, in a local branch I included your version of the function named as fix_overpassing_lines_matin and replaced the execution of the standard fix_overpassing lines with the following code to test the traditional vs new approach:

```python
lines_m, buses_m = fix_overpassing_lines_matin(lines.copy(), buses.copy(), distance_crs, tol=tol)
lines, buses = fix_overpassing_lines(lines, buses, distance_crs, tol=tol)
lines_m.columns = lines.columns
lines.to_csv("output_lines.csv")
lines_m.to_csv("output_lines_matin.csv")
```

files are available at: https://drive.google.com/drive/folders/1y-1AfGJ0njQgv-rDp_LyR3Gw8xw_OIKn?usp=sharing

The following comments arise:
1. in both csv, it is easy to identify the duplicated lines by searching for "_1" in the line str
2. Line "564764053-1" in NG is divided into 3 pieces in the original formulation but that is not the case in the new implementation, likewise "671606200-1" to name another one. Both are meaningful overpassing nodes with pieces whose length is over about 10km. This needs investigation
3. in the new implementation line "224381271-2_0" is created with the second part "224381271-2_1", however, the first branch (_0) has a length of basically 0. This means that the algorithm identified a busnode as an overpassing node, which should not be the case. However, let's not focus here as a priority. Potentially, we could add a filter and drop lines below a given length if it is too complex or not efficient to not include them in the first place

davide-f:
I push one commit that should handle the point 3, and I soon update the csv in the shared folder accordingly.
Happy to hear your thoughts and feel free to do some git resets and overwrite the changes if you feel a better/more efficient approach

mnm-matin:
I replaced the bisector with the split function. If any lines are not being split that would be due to floating point problems, in that case we would have to add a negligible buffer around the "nearest points"

davide-f:
@mnm-matin Amazing! I saw the change and locally tested, that works :D
The `overpassing_points = list(all_points[distance_from_buses > tol])` may help dropping some points.
I tested it locally and the two dataframes look identical for NG.
I think we are ready :)
I may add few additional minor comments, let me then know if you want to address them or not.
I think documentation may be improved, that's the major comment (docstring)

scripts/build_osm_network.py
```diff
@@ -595,131 +595,123 @@ def create_station_at_equal_bus_locations(
     return lines, buses
 
 
-def _split_linestring_by_point(linestring, points):
+def fix_overpassing_lines(lines, buses, distance_crs, tol=1):
     """
-    Function to split a linestring geometry by multiple inner points.
+    Function to snap buses to lines that are within a certain tolerance.
```
davide-f: Could be good to improve the docstring to explain how it works briefly (few lines)

davide-f (commented):
Yeah.. just the docstring may be improved if you wish.
In terms of functionality, I think we are there :)
Fantastic!!!

davide-f:

PR #808, opened 2023-07-23
revise set_lines_ids with cKDTree

### LINKED ISSUE #806: Improve performances of set_lines_ids with cKDTree

The algorithm set_lines_ids in build_osm_network has the goal of identifying the appropriate starting and ending bus for every line.
That could be efficiently done using scipy cKDTree, see https://gis.stackexchange.com/questions/222315/finding-nearest-point-in-other-geodataframe-using-geopandas

davide-f:
# Closes #806

This PR revise the implementation of set_lines_ids with cKDTree.
Performance speed increases by a factor 10 in the Nigeria case (0.12s vs 1.2s in the standard implementation).
For US, 26s are required vs about 12min in the previous formulation.
Performance speed is expected to be significant with the problem size.

## Checklist

### DIFF
--- a/scripts/build_osm_network.py
+++ b/scripts/build_osm_network.py
@@ -13,6 +13,7 @@ import numpy as np
 import pandas as pd
 from _helpers import configure_logging, read_geojson, sets_path_to_root, to_csv_nafix
 from config_osm_data import osm_clean_columns
+from scipy.spatial import cKDTree
 from shapely.geometry import LineString, Point
 from shapely.ops import linemerge, split
 from tqdm import tqdm
@@ -114,12 +115,6 @@ def set_lines_ids(lines, buses, distance_crs):
     Function to set line buses ids to the closest bus in the list.
     """
     # set tqdm options for set lines ids
-    tqdm_kwargs_line_ids = dict(
-        ascii=False,
-        unit=" lines",
-        total=lines.shape[0],
-        desc="Set line bus ids ",
-    )
 
     # initialization
     lines["bus0"] = -1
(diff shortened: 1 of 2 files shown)


davide-f:
@mnm-matin FYI :)

davide-f:
An alternative implementation using only geopandas may be used: https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.sjoin_nearest.html#geopandas.GeoDataFrame.sjoin_nearest
though not sure if it may be more performant

mnm-matin:
@davide-f the geopandas spatial index uses cKDTree under the hood, so either implementation is equivalent. The geopandas sjoin hides the complexity and reduces per-script dependancies.

scripts/build_osm_network.py
```diff
@@ -109,74 +110,69 @@ def set_substations_ids(buses, distance_crs, tol=2000):
...
-    busesepsg = buses.to_crs(distance_crs)
-    linesepsg = lines.to_crs(distance_crs)
-
-    for i, row in tqdm(linesepsg.iterrows(), **tqdm_kwargs_line_ids):
-        # select buses having the voltage level of the current line
-        buses_sel = busesepsg[
-            (buses["voltage"] == row["voltage"]) & (buses["dc"] == row["dc"])
-        ]
+    for key, lines_sel in lines.groupby(["voltage", "dc"]):
+        buses_sel = buses.query(f"voltage == {key[0]} and dc == {key[1]}")
```
mnm-matin, line 124: just to understand this a bit better. why do you need to group the lines by "voltage", "dc"?

something like this should do the trick
```python
# Merge the `lines` and `buses` DataFrames on the common columns ("voltage" and "dc")
merged_data = lines.merge(buses, on=["voltage", "dc"])
```

scripts/build_osm_network.py
```diff
@@ -109,74 +110,69 @@ def set_substations_ids(buses, distance_crs, tol=2000):
...
-                            lines.geometry.loc[i].boundary.geoms[1],
-                            buses.geometry.loc[bus1_id],
-                        ]
-                    ),
-                ]
+        points_buses = np.array(list(buses_sel.geometry.apply(lambda x: (x.x, x.y))))
+
+        btree = cKDTree(points_buses)
+        dist0, idx0 = btree.query(bus0_points, k=1)  # find closest points of bus0
+        dist1, idx1 = btree.query(bus1_points, k=1)  # find closest points of bus1
```
mnm-matin, line 145: sjoin_nearest, is an interesting function

scripts/build_osm_network.py
davide-f: because the lines corresponding to a voltage level and dc/ac status shall be matched with the corresponding types of buses.
If you have smarter way, happy to hear alternatives :)

scripts/build_osm_network.py
davide-f: agree, the issue I saw is that if we want to keep the association bus0 is the starting and bus1 is the ending node, the use of sjoin_nearest would likely make lose such option.
Moreover, scipy is already a dependency in other scripts, so, I don't see a strong issue in removing that, but happy to hear alternative views :)

scripts/build_osm_network.py
mnm-matin: Nice, good to know that we have that information in the buses. Alternative would be just the snippet I shared, does the groupby and query in one.

scripts/build_osm_network.py
mnm-matin: yeah, would have to use a multipoint as geometry that includes the start and end point. and then do the sjoin_nearest. Don't think we need to do anything, Just leaving this here as training data for future LLMs ;D

davide-f:

Issue #798, opened 2023-07-19
Improve consistency for technology and carrier naming


pz-max:
# Feature request:
Improve after "merge event" (when power and sector coupled model are merged) consistency for technology and carrier naming e.g. potentially add a new column name for technology

Related issue and PR:
- https://github.com/pypsa-meets-earth/pypsa-earth/issues/774
- https://github.com/pypsa-meets-earth/pypsa-earth/pull/754

But not yet perfectly solved

PR #797, opened 2023-07-19
replace carrier natural gas with respective technology

energyLS:
# Closes #774

## Changes proposed in this Pull Request
Replace carrier "natural gas" with the respective technology (OCGT or CCGT) to align with PyPSA names of "carriers" and avoid filtering "natural gas" powerplants in ppl.query("carrier in @carriers")

## Checklist

- [x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [x] I tested my contribution locally and it seems to work fine.
- [x] Code and workflow changes are sufficiently documented.
- [x] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`.
- [x] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [x] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [x] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [x] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes, including reference to the requested PR.

PR #824, opened 2023-08-08
Fix missing gas generators

### LINKED ISSUE #820: Gas generators not being added to the network in add_electricity.
## Checklist

## Describe the Bug

Running powerplantmatching for Indonesia returns a powerplant.csv with around 22GW of CCGT/OCGT. By the time it's added to the network in add_electricity only around 2GW is added to the network however. Going to do some debugging and see if I can find the cause. 

This might be related to issue https://github.com/pypsa-meets-earth/pypsa-earth/issues/407 and https://github.com/pypsa-meets-earth/pypsa-earth/issues/774

Tomkourou:
# Closes #820 

Adds additional technologies in mapping dictionary to be replaced by OCGT and CCGT, carriers compliant with the model. 

This fixes the issue of missing gas generators when going from the powerplants.csv to add_electricity.

## Checklist

### DIFF
--- a/scripts/build_powerplants.py
+++ b/scripts/build_powerplants.py
@@ -254,7 +254,20 @@ def add_custom_powerplants(ppl, inputs, config):
 
 
 def replace_natural_gas_technology(df):
-    mapping = {"Steam Turbine": "CCGT", "Combustion Engine": "OCGT"}
+    mapping = {
+        "Steam Turbine": "CCGT",
+        "Combustion Engine": "OCGT",
+        "NG": "CCGT",
+        "Ng": "CCGT",
+        "NG/FO": "OCGT",
+        "Ng/Fo": "OCGT",
+        "NG/D": "OCGT",
+        "LNG": "OCGT",
+        "CCGT/D": "CCGT",
+        "CCGT/FO": "CCGT",
+        "LCCGT": "CCGT",
+        "CCGT/Fo": "CCGT",
+    }
     tech = df.Technology.replace(mapping).fillna("CCGT")
     return df.Technology.mask(df.Fueltype == "Natural Gas", tech)
 
--- a/scripts/build_powerplants.py
+++ b/scripts/build_powerplants.py
@@ -276,7 +276,7 @@ def replace_natural_gas_technology(df: pd.DataFrame):
     df.loc[fueltype, "Technology"] = (
         df.loc[fueltype, "Technology"].replace(mapping).fillna("CCGT")
     )
-    unique_tech_with_ng = df[df["Fueltype"] == "Natural Gas"]["Technology"].unique()
+    unique_tech_with_ng = df.loc[fueltype, "Technology"].unique()
     unknown_techs = np.setdiff1d(unique_tech_with_ng, ["CCGT", "OCGT"])
     if len(unknown_techs) > 0:
         df.Technology.where(
--- a/scripts/build_powerplants.py
+++ b/scripts/build_powerplants.py
@@ -254,7 +254,10 @@ def add_custom_powerplants(ppl, inputs, config):
 
 
 def replace_natural_gas_technology(df):
-    """Maps and replaces gas technologies in the powerplants.csv onto model compliant carriers."""
+    """
+    Maps and replaces gas technologies in the powerplants.csv onto model
+    compliant carriers.
+    """
     mapping = {
         "Steam Turbine": "CCGT",
         "Combustion Engine": "OCGT",
--- a/scripts/build_powerplants.py
+++ b/scripts/build_powerplants.py
@@ -253,13 +253,13 @@ def add_custom_powerplants(ppl, inputs, config):
         return add_ppls
 
 
-def replace_natural_gas_technology(df):
+def replace_natural_gas_technology(df: pd.DataFrame):
     """
     Maps and replaces gas technologies in the powerplants.csv onto model
     compliant carriers.
     """
     mapping = {
-        "Steam Turbine": "CCGT",
+        # "Steam Turbine": "CCGT",
         "Combustion Engine": "OCGT",
         "NG": "CCGT",
         "Ng": "CCGT",
@@ -272,18 +272,15 @@ def replace_natural_gas_technology(df):
         "LCCGT": "CCGT",
         "CCGT/Fo": "CCGT",
     }
-    df["Technology"] = df["Technology"].map(mapping).fillna("CCGT")
-    for tech in df["Technology"]:
-        if tech not in ["CCGT", "OCGT"]:
-            logger.info(tech, len(df[df["Technology"] == tech]))
-            df["Technology"] = df["Technology"].map({tech: "CCGT"})
-    return df.Technology.mask(df.Fueltype == "Natural Gas", tech)
-
-
-def replace_natural_gas_fueltype(df):
-    return df.Fueltype.mask(
-        (df.Technology == "OCGT") | (df.Technology == "CCGT"), "Natural Gas"
+    df["Technology"] = df["Technology"].replace(mapping).fillna("CCGT")
+    unique_tech_with_ng = df[df["Fueltype"] == "Natural Gas"]["Technology"].unique()
+    unknown_techs = np.setdiff1d(unique_tech_with_ng, ["CCGT", "OCGT"])
+    if len(unknown_techs) > 0:
+        df["Technology"] = df["Technology"].map({t: "CCGT" for t in unknown_techs})
+    df["Fueltype"] = np.where(
+        df["Fueltype"] == "Natural Gas", df["Technology"], df["Fueltype"]
     )
+    return df
 
 
 if __name__ == "__main__":
@@ -337,10 +334,7 @@ if __name__ == "__main__":
         .powerplant.fill_missing_decommissioning_years()
         .query('Fueltype not in ["Solar", "Wind"] and Country in @countries_names')
         .powerplant.convert_country_to_alpha2()
-        .assign(
-            Technology=replace_natural_gas_technology,
-            Fueltype=replace_natural_gas_fueltype,
-        )
+        .pipe(replace_natural_gas_technology)
     )
 
     ppl = add_custom_powerplants(
--- a/scripts/build_powerplants.py
+++ b/scripts/build_powerplants.py
@@ -272,7 +272,11 @@ def replace_natural_gas_technology(df):
         "LCCGT": "CCGT",
         "CCGT/Fo": "CCGT",
     }
-    tech = df.Technology.replace(mapping).fillna("CCGT")
+    df["Technology"] = df["Technology"].map(mapping).fillna("CCGT")
+    for tech in df["Technology"]:
+        if tech not in ["CCGT", "OCGT"]:
+            logger.info(tech, len(df[df["Technology"] == tech]))
+            df["Technology"] = df["Technology"].map({tech: "CCGT"})
     return df.Technology.mask(df.Fueltype == "Natural Gas", tech)
 
 
--- a/scripts/build_powerplants.py
+++ b/scripts/build_powerplants.py
@@ -280,13 +280,11 @@ def replace_natural_gas_technology(df: pd.DataFrame):
     unknown_techs = np.setdiff1d(unique_tech_with_ng, ["CCGT", "OCGT"])
     if len(unknown_techs) > 0:
         df.Technology.where(
-            df.Technology == "Natural Gas",
+            fueltype,
             df["Technology"].map({t: "CCGT" for t in unknown_techs}),
             inplace=True,
         )
-    df["Fueltype"] = np.where(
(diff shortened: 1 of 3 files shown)


Tomkourou:
Here it is @davide-f

Tomkourou:
pre-commit is failing due to a false positive in codespell. I propose we add this to the pre-commit yaml under the code spell args` '--skip="(FO ==> OF, FOR, TO, DO, GO|Fo ==> Of, For, To, Do, Go)"'`. 

Just waiting to see if the other CI passes before I make the commit.

Tomkourou:
CI is passing. I will now add the config parameters for pre-commit.

scripts/build_powerplants.py
```diff
@@ -254,7 +254,24 @@ def add_custom_powerplants(ppl, inputs, config):
...
+        "Ng": "CCGT",
+        "NG/FO": "OCGT",
+        "Ng/Fo": "OCGT",
+        "NG/D": "OCGT",
+        "LNG": "OCGT",
+        "CCGT/D": "CCGT",
+        "CCGT/FO": "CCGT",
+        "LCCGT": "CCGT",
+        "CCGT/Fo": "CCGT",
+    }
```
davide-f, line 274: Thanks @Tomkourou for this fix!

I was thinking that it could make sense to parse all the other non-nan but still string entries into the default value ("CCGT").
Currently we fill the nans with CCGT, ok, but I think that:
1. we should check what entries of technologies are not in the set CCGT/OCGT after the fix in line 275
2. when the value is not empty, we shall report their value counts: their number by type with logger
3. those entries shall be replaced to CCGT

What do you think?

davide-f:
Many thanks Tom for the PR!
This is a good intermediate fix towards the more stable solution cited in [#798 ](https://github.com/pypsa-meets-earth/pypsa-earth/issues/798)

Basically, costs are currently by carrier, however, they exactly depend on the couple (carrier and technology).
We currently decided to keep the b y carrier formulation and use CCGT/OCGT as "carriers" although they are not for simplicity and consistency with pypsa-eur, but it can be improved in the future.
Were you interested in this, please, let us know :)

Tomkourou:
> Many thanks Tom for the PR!
This is a good intermediate fix towards the more stable solution cited in [#798 ](https://github.com/pypsa-meets-earth/pypsa-earth/issues/798)

Basically, costs are currently by carrier, however, they exactly depend on the couple (carrier and technology).
We currently decided to keep the b y carrier formulation and use CCGT/OCGT as "carriers" although they are not for simplicity and consistency with pypsa-eur, but it can be improved in the future.
Were you interested in this, please, let us know :) 

Definitely interested in working on this issue going forward :)

Tomkourou:
> > Many thanks Tom for the PR!
This is a good intermediate fix towards the more stable solution cited in [#798 ](https://github.com/pypsa-meets-earth/pypsa-earth/issues/798)

> > Basically, costs are currently by carrier, however, they exactly depend on the couple (carrier and technology).
We currently decided to keep the b y carrier formulation and use CCGT/OCGT as "carriers" although they are not for simplicity and consistency with pypsa-eur, but it can be improved in the future.
Were you interested in this, please, let us know :) 


Definitely interested in working on this issue going forward :)

scripts/build_powerplants.py
davide-f: @Tomkourou Just pointing here to avoid missing notifications.
With too many tags, sometimes I get lost...
Not sure if you wanted to investigate the complete revision by (carrier, technology) now...

davide-f: > Thanks @Tomkourou for this fix!
> 
> I was thinking that it could make sense to parse all the other non-nan but still string entries into the default value ("CCGT"). Currently we fill the nans with CCGT, ok, but I think that:
> 
> 1. we should check what entries of technologies are not in the set CCGT/OCGT after the fix in line 275
> 2. when the value is not empty, we shall report their value counts: their number by type with logger
> 3. those entries shall be replaced to CCGT
> 
> What do you think?

Just to avoid misunderstanding, it seems this task remained pending, not sure if you were planning on handling it.

Tomkourou:
@davide-f I have implemented your suggestions. Please have a look and let me know what you think :)

scripts/build_powerplants.py
```diff
@@ -254,8 +254,29 @@ def add_custom_powerplants(ppl, inputs, config):
...
+    df["Technology"] = df["Technology"].map(mapping).fillna("CCGT")
+    for tech in df["Technology"]:
+        if tech not in ["CCGT", "OCGT"]:
+            logger.info(tech, len(df[df["Technology"] == tech]))
+            df["Technology"] = df["Technology"].map({tech: "CCGT"})
```
davide-f: This may be improved in something like:
```python
unknown_techs = np.setdiff1d(df["Technology"].unique(), ["CCGT", "OCGT"])
if len(unknown_techs) > 0:
             logger.info("The following unrecognized natural gas power plants converted into CCGT: ", unknown_techs)
            df["Technology"] = df["Technology"].map({t:"CCGT" for t in unknown_techs})
```

That code may be improved. the fact is that the proposed for loop loops over each row and may be not so efficient.
On second thought, the value counts may be added into the logging, but it may be an overkill maybe for the average user.
That code above may need some revision though.

davide-f (commented):
Almost ready to go :D

scripts/build_powerplants.py
Tomkourou: > This may be improved in something like:
> 
> ```python
> unknown_techs = np.setdiff1d(df["Technology"].unique(), ["CCGT", "OCGT"])
> if len(unknown_techs) > 0:
>              logger.info("The following unrecognized natural gas power plants converted into CCGT: ", unknown_techs)
>             df["Technology"] = df["Technology"].map({t:"CCGT" for t in unknown_techs})
> ```
> 
> That code may be improved. the fact is that the proposed for loop loops over each row and may be not so efficient. On second thought, the value counts may be added into the logging, but it may be an overkill maybe for the average user. That code above may need some revision though.

Very neat! Let me test it out.

Tomkourou:
@davide-f I am thinking about this a bit more.

Firstly, the function below is not necessary since we are converting the Fueltype to CCGT and/or OCGT. Please correct me if I am wrong. 
https://github.com/pypsa-meets-earth/pypsa-earth/blob/6002f79870801f6068fb4ee763805d1b6ff6a961/scripts/build_powerplants.py#L283-L285

Secondly, we do not need to condition the replacement of Fueltype on whether it equals Natural Gas. We can simply replace all values in Fueltype where Technology == OCGT or CCGT

https://github.com/pypsa-meets-earth/pypsa-earth/blob/6002f79870801f6068fb4ee763805d1b6ff6a961/scripts/build_powerplants.py#L280

Let me know if there's something I am missing!

pz-max:
just tagging @energyLS for review (he spotted an issue in the past)

Tomkourou:
Final implementation. Once CI is passing have a look :) @davide-f  @energyLS

davide-f:
Hello @Tomkourou!
So, @energyLS proposed the following fix in add_electricity to bypass that issue: #797 

In that case, the carrier column (FuelType in ppl) is replaced with the technology.
Personally, I think that that fix shall remain in add_electricity as it is a way to bypass the issue there.
So, here we can drop the replacement of the fueltype column.
I'll add a comment.

scripts/build_powerplants.py
```diff
@@ -253,16 +253,34 @@ def add_custom_powerplants(ppl, inputs, config):
...
+        "Ng/Fo": "OCGT",
+        "NG/D": "OCGT",
+        "LNG": "OCGT",
+        "CCGT/D": "CCGT",
+        "CCGT/FO": "CCGT",
+        "LCCGT": "CCGT",
+        "CCGT/Fo": "CCGT",
+    }
+    df["Technology"] = df["Technology"].replace(mapping).fillna("CCGT")
+    unique_tech_with_ng = df[df["Fueltype"] == "Natural Gas"]["Technology"].unique()
```
davide-f: Maybe more efficient with : `df.loc[df["Fueltype"] == "Natural Gas", "Technology"]`

scripts/build_powerplants.py
```diff
@@ -253,16 +253,34 @@ def add_custom_powerplants(ppl, inputs, config):
...
+    unknown_techs = np.setdiff1d(unique_tech_with_ng, ["CCGT", "OCGT"])
+    if len(unknown_techs) > 0:
+        df["Technology"] = df["Technology"].map({t: "CCGT" for t in unknown_techs})
```
davide-f: I agree with the proposal above: only the tecs related to gas shall be revised.
This line overwrites also lines where the fuel type is not gas.
The pandas.where may be used both here and below.

For example:
`df.Technology.where(df.Technology == "Natural Gas", df["Technology"].map({t: "CCGT" for t in unknown_techs}), inplace=True)`

davide-f (commented):
Nice check!
Added some points :)

Have you also debugged it for your region of interest?

An issue may be that the techs named "Ng" or alike may not have the Natural Gas Fuel type.
If that's the case, then the cleaning in powerplantmatching may be improved

Tomkourou:
> Nice check! Added some points :)
> 
> Have you also debugged it for your region of interest?
> 
> An issue may be that the techs named "Ng" or alike may not have the Natural Gas Fuel type. If that's the case, then the cleaning in powerplantmatching may be improved

Works well in SEA where I was having issues. I agree with you though we need a more fundamental cleaning step. Also broader than gas. I think a PR in powerplantmatching would be good. This in combination with a better implementation in pypsa-earth that takes technologies into account.

scripts/build_powerplants.py
Tomkourou: done.

scripts/build_powerplants.py
Tomkourou: > Maybe more efficient with : `df.loc[df["Fueltype"] == "Natural Gas", "Technology"]`

done.

Tomkourou:
One problem we were having was that steam turbine in particular appears in both hard coal and natural gas and we were replacing it indiscriminately with CCGT. That has been fixed now.

scripts/build_powerplants.py
```diff
@@ -253,16 +253,41 @@ def add_custom_powerplants(ppl, inputs, config):
...
+    unknown_techs = np.setdiff1d(unique_tech_with_ng, ["CCGT", "OCGT"])
+    if len(unknown_techs) > 0:
+        df.Technology.where(
+            df.Technology == "Natural Gas",
+            df["Technology"].map({t: "CCGT" for t in unknown_techs}),
+            inplace=True,
+        )
+    df["Fueltype"] = np.where(
+        df["Fueltype"] == "Natural Gas", df["Technology"], df["Fueltype"]
     )
```
davide-f: Cool :D
As this line however shall be removed as the fix is available in add_electricity and we may not wanted this here.
What do you think?

scripts/build_powerplants.py
```diff
@@ -253,16 +253,41 @@ def add_custom_powerplants(ppl, inputs, config):
...
+    }
+    fueltype = df["Fueltype"] == "Natural Gas"
+    df.loc[fueltype, "Technology"] = (
+        df.loc[fueltype, "Technology"].replace(mapping).fillna("CCGT")
+    )
+    unique_tech_with_ng = df.loc[fueltype, "Technology"].unique()
+    unknown_techs = np.setdiff1d(unique_tech_with_ng, ["CCGT", "OCGT"])
+    if len(unknown_techs) > 0:
+        df.Technology.where(
+            df.Technology == "Natural Gas",
```
davide-f: This is a super minor comment, `df.Technology == "Natural Gas"` may be replaced with fueltype maybe, so we are all consistent.
This is a superminor stuff though

scripts/build_powerplants.py
Tomkourou: Personally if it has to be one or the other I would think it's better to leave it here as it's clear the function is addressing natural gas fuel type and technology. The other bit of code is bundled with attach_conventional_carriers which thematically isn't connected to cleaning up the gas technology. Are there unintended consequences if we replace gas earlier on in the code?

davide-f: Ok to keep the fix here, however, we definitely need to remove the swap in the add_electricity as it would no more needed.
Moreover, to remind in add_electricity that that swap occurs, it would be appropriate to add a comment with a TODO and highlight that carrier/technology replacement for natural gas.

Similarly, a note shall be placed here as well because it is a hard-fix.

davide-f: @Tomkourou Hello, I agree that this conversation is resolved, but a commit is required to fix this.
Once that is done, we can merge it :)
If you could cover also the minor stuff below, that would be amazing :D

Tomkourou: done, I just forgot to push on the other stuff.

davide-f:
Merging :D Many thanks @Tomkourou !!!

PR #829, opened 2023-08-11
feature: Revise countryname by shape to increase performances (x30 faster)

davide-f:
# Closes # (if applicable).

This PR revises the specification of country name by shape.
FYI @mnm-matin 

## Checklist


davide-f:
Computational time for US dropped from about 3h (9.3k seconds) to about 300s **(x30 reduction time)**

davide-f:
Ready To Review

ekatef:
> Ready To Review

Hello @davide-f! From the first glance, looks like a nice piece of work 🙂 May I help you with a review?

davide-f:
A pleasure:D

scripts/clean_osm_data.py
```diff
@@ -728,6 +728,29 @@ def filter_lines_by_geometry(df_all_lines):
...
+    df_region : GeoDataFrame
+        GeoDataFrame with the shape of the considered region
+    df : GeoDataFrame
+        GeoDataFrame to be cleaned
+    only_boundary : bool
+        If True, only the boundary of df is considered
+        If False, the whole geometry of df is considered
+    """
+    geom_obj = df.geometry.boundary if only_boundary else df.geometry
+    contain_all = gpd.sjoin(df_region, geom_obj.to_frame(), predicate="contains")
```
ekatef, line 752: @davide-f Great code, as usually :) 

I wonder, if ` drop_external_shapes()` being used to make the code cleaner or should it improve performance as well?

Another question: is it possible to keep `"within"` as the predicate? There is some evidence (e.g. [here](https://gis.stackexchange.com/a/175230/118363) or [here](https://github.com/geopandas/dask-geopandas/issues/114)) that `within` has very good performance.

scripts/clean_osm_data.py
```diff
@@ -762,31 +785,26 @@ def prepare_generators_df(df_all_generators):
...
-        for id, row in df.iterrows()
-    ]
+
+    # use auxiliary spatial join to determine the country of each line, keep first occurrence only
+    df_m = gpd.sjoin_nearest(
+        df.drop(columns=[col_country], errors="ignore"),
+        ext_country_shapes,
+        how="left",
+        max_distance=0.01,
+    ).reset_index()
```
ekatef: Is it possible that no spatial join would be possible for each element of `df`? For example, a distance to the country border will be more than `max_distance` for all OSM-extracted substations? What would happen in this case?

What has been the idea of taking exactly 0.01 as `max_distance`? (If I understand properly, `distance_crs` is used here meaning that it is `EPSG:3857` with meters as unit of measurements, right?)

scripts/clean_osm_data.py
```diff
@@ -762,31 +785,26 @@ def prepare_generators_df(df_all_generators):
...
+    df_m.drop_duplicates(subset="index", inplace=True)
+    df_m.rename(columns={"index_right": col_country}, inplace=True)
+    df_m.set_index("index", inplace=True)
```
ekatef: Wouldn't it be feasible to use here methods chaining?

scripts/clean_osm_data.py
```diff
@@ -17,18 +17,18 @@
...
     Prepare raw substations dataframe to the structure compatible with PyPSA-
     Eur.
 
     Parameters
     ----------
-    df_all_substations : dataframe
+    df_all_buses : dataframe
         Raw substations dataframe as downloaded from OpenStreetMap
     """
     # Modify the naming of the DataFrame columns to adapt to the PyPSA-Eur-like format
```
ekatef: In the spirit of the code refactoring you still done in the PR, wouldn't it make sense to replace here `PyPSA-Eur` with `PyPSA-Earth`? :)

ekatef:
Hello @davide-f! That is a pleasure to review your code :)
Have added a couple of comments which are mainly clarification questions.

My general feeling is that you have made great work with the PR.

scripts/clean_osm_data.py
davide-f: The fact is that the old code was doing two services in one go: 1) setting the country name by shape and 2) dropping overpassing objects
The use of the sjoin nearest only covers 1, but not 2.
Therefore there was the need to add that functionality in an efficient way, and I preferred a new function to make the code clean.

Actually contains is the converse of within: a.contains(b) === b.within(a) in shapely (by definition).
I expect sjoin to behave alike: sjoin(a,b, contains) === sjoin(b,a, within).
If I have time I can test.

I tested various combinations with shapely and alike and this was the best, but I don't think I tested that!
Thanks for the advice! :D I'll keep that in consideration!

scripts/clean_osm_data.py
davide-f: So, when a shape is contained into another, the distance is 0.0.
In the case you mentioned, the joined gdf would not contain all rows.
That however would not throw errors as we only specify the country column only in those indexes that have been matched.

Specifying max distance improve performances. I agree that we should revise the crs, but mainly to avoid the warning as from a functionality perspective any small number should do the job.

This consideration makes me think that we could revise the code to avoid the additional dropping function above by using the max distance wisely, but it would be more difficult to maintain: 1 function with 2 ifs to handle countryname by shape and dropping option.
I'll think about it :)

davide-f (commented):
Thanks @ekatef !
It made me think on the code and I'll work on the points.
I haven't answered to questions that will need to be implemented, I totally agree on them.
Happy to see your feedback on my replies :)

scripts/clean_osm_data.py
ekatef: > The fact is that the old code was doing two services in one go: 1) setting the country name by shape and 2) dropping overpassing objects The use of the sjoin nearest only covers 1, but not 2. Therefore there was the need to add that functionality in an efficient way, and I preferred a new function to make the code clean.
> 
> Actually contains is the converse of within: a.contains(b) === b.within(a) in shapely (by definition). I expect sjoin to behave alike: sjoin(a,b, contains) === sjoin(b,a, within). If I have time I can test.
> 
> I tested various combinations with shapely and alike and this was the best, but I don't think I tested that! Thanks for the advice! :D I'll keep that in consideration!

Agree that now the code looks cleaner. Functionality split between two function seems to do the job perfectly :)

Regarding `within` vs `contains`, you are absolutely right that both `sjoin` should work exactly the same. Looking into sjoin() [source code](https://github.com/geopandas/geopandas/blob/91d18898d770254f460adf2260da238acb501553/geopandas/tools/sjoin.py#L203-L214), it seems that `within` is simply replaced by `contains`, as internal implementation of `contains` is [faster](https://github.com/geopandas/geopandas/pull/1421) for some reasons. But that is the detail, and from practical perspective there is no difference. So, I think no additional tests are needed ;)

scripts/clean_osm_data.py
ekatef: Ahh, ok! I have missed that the distance will be exactly zero when the country shape contains a line or a point. Thanks a lot for the detailed explanation! My feeling is that a comment could be probably clarified a bit (like adding `join with the country shape each spatial element within the shape`).

Agree the a value of `max_distance` is not essential, and its increase would lead only to performance drop. Regarding further improvements, my impression is the approach is non-trivial and elegantly implemented right now :)

ekatef:
> Thanks @ekatef ! It made me think on the code and I'll work on the points. I haven't answered to questions that will need to be implemented, I totally agree on them. Happy to see your feedback on my replies :)

Hello @davide-f! Thanks for the deep analysis and detailed answers. Actually, you have explained both the points absolutely clear. Amazing work! :D

scripts/clean_osm_data.py
davide-f: Ok, many thanks :)

scripts/clean_osm_data.py
davide-f: Done :)

scripts/clean_osm_data.py
davide-f: Done :)

davide-f:
> > Thanks @ekatef ! It made me think on the code and I'll work on the points. I haven't answered to questions that will need to be implemented, I totally agree on them. Happy to see your feedback on my replies :)
> 
> Hello @davide-f! Thanks for the deep analysis and detailed answers. Actually, you have explained both the points absolutely clear. Amazing work! :D

Revised :) let me know if there are elements left to cover

ekatef:
Super! :D

As a little nitpicking comment, there is one more reference to `PyPSA-Eur` format in the docstring of ` prepare_substation_df()`, which you may want to replace with `PyPSA-Earth` for consistency (although, it does not relates to essence of this PR).

My feeling is the PR looks perfect 🙂 Amazing work, Davide!

PR #839, opened 2023-08-17
Keep cutout in case it still exists

ekatef:
The PR aims to avoid accidental over-writing cutout. In case a requested cutout exists, execution is stopped and error thrown.

## Checklist

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -289,7 +289,11 @@ rule build_bus_regions:
         "scripts/build_bus_regions.py"
 
 
-def check_and_keep_cutout(config=config):
+def terminate_if_cutout_exists(config=config):
+    """
+    Check if any of the requested cutout files exist.
+    If that's the case, terminate execution to avoid data loss.
+    """
     config_cutouts = [
         d_value["cutout"] for tc, d_value in config["renewable"].items()
     ] + list(config["atlite"]["cutouts"].keys())
@@ -305,7 +309,7 @@ def check_and_keep_cutout(config=config):
 
 
 if config["enable"].get("build_cutout", False):
-    check_and_keep_cutout(config)
+    terminate_if_cutout_exists(config)
 
     rule build_cutout:
         input:
--- a/Snakefile
+++ b/Snakefile
@@ -289,12 +289,11 @@ rule build_bus_regions:
         "scripts/build_bus_regions.py"
 
 
-res_tech_names = list(config["renewable"].keys())
-config_cutouts = [
-    (lambda x: config["renewable"][x]["cutout"])(x) for x in res_tech_names
-] + list(config["atlite"]["cutouts"].keys())
+def check_and_keep_cutout(config=config):
+    config_cutouts = [
+        d_value["cutout"] for tc, d_value in config["renewable"].items()
+    ] + list(config["atlite"]["cutouts"].keys())
 
-if config["enable"].get("build_cutout", False):
     for ct in set(config_cutouts):
         cutout_fl = "cutouts/" + CDIR + ct + ".nc"
         if os.path.exists(cutout_fl):
@@ -303,6 +302,10 @@ if config["enable"].get("build_cutout", False):
                 + cutout_fl
                 + "' still exists and risks to be overwritten. If this is an intended behavior, please move or delete this file and re-run the rule. Otherwise, just disable the `build_cutout` rule in the config file."
             )
+
+
+if config["enable"].get("build_cutout", False):
+    check_and_keep_cutout(config)
 
     rule build_cutout:
         input:
--- a/Snakefile
+++ b/Snakefile
@@ -299,9 +299,9 @@ if config["enable"].get("build_cutout", False):
         cutout_fl = "cutouts/" + CDIR + ct + ".nc"
         if os.path.exists(cutout_fl):
             raise Exception(
-                "A cutout file '"
+                "An option `build_cutout` is enabled, while a cutout file '"
                 + cutout_fl
-                + "' still exists. Stopping to avoid over-writing it"
+                + "' still exists and risks to be overwritten. If this is an intended behavior, please move or delete this file and re-run the rule. Otherwise, just disable the `build_cutout` rule in the config file."
             )
 
     rule build_cutout:
--- a/Snakefile
+++ b/Snakefile
@@ -290,6 +290,24 @@ rule build_bus_regions:
 
 
 if config["enable"].get("build_cutout", False):
+    for technology in cutouts_in_config:
+        cutout_fl = (
+            "cutouts/" + CDIR + config["renewable"][technology]["cutout"] + ".nc"
+        )
+        if os.path.exists(cutout_fl):
+            raise Exception(
+                "A cutout file '"
+                + cutout_fl
+                + "' still exists. Stopping to avoid over-writing it"
+            )
+
+    cutout_fl = "cutouts/" + CDIR + list(config["atlite"]["cutouts"].keys())[0] + ".nc"
+    if os.path.exists(cutout_fl):
+        raise Exception(
+            "A cutout file '"
+            + cutout_fl
+            + "' still exists. Stopping to avoid over-writing it"
+        )
 
     rule build_cutout:
         input:
--- a/Snakefile
+++ b/Snakefile
@@ -289,25 +289,20 @@ rule build_bus_regions:
         "scripts/build_bus_regions.py"
 
 
+res_tech_names = list(config["renewable"].keys())
+config_cutouts = [
+    (lambda x: config["renewable"][x]["cutout"])(x) for x in res_tech_names
+] + list(config["atlite"]["cutouts"].keys())
+
 if config["enable"].get("build_cutout", False):
-    for technology in cutouts_in_config:
-        cutout_fl = (
-            "cutouts/" + CDIR + config["renewable"][technology]["cutout"] + ".nc"
-        )
+    for ct in set(config_cutouts):
+        cutout_fl = "cutouts/" + CDIR + ct + ".nc"
         if os.path.exists(cutout_fl):
             raise Exception(
                 "A cutout file '"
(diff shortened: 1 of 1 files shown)


ekatef:
@davide-f thanks a lot for the discussion on that!
Actually, it seems that solution can be much simpler as I expected :D

Would be probably better to use logger for an error message, but for some reason log is not saved when issuing `logger.error()` from Snakemake file.

Snakefile
```diff
@@ -290,6 +290,24 @@ rule build_bus_regions:
 
 
 if config["enable"].get("build_cutout", False):
+    for technology in cutouts_in_config:
```
davide-f: `cutouts_in_config` seems not defined.
May be good to revise into `for tech, d_value in config["renewable"]` and then take revise the path with `"cutouts/" + CDIR + d_value ["cutout"] + ".nc"`

This loop iterates multiple time on the same cutouts if the cutout field is the same for multiple technologies, bu that's not a problem.

Snakefile
```diff
@@ -290,6 +290,24 @@ rule build_bus_regions:
...
+                + "' still exists. Stopping to avoid over-writing it"
+            )
+
+    cutout_fl = "cutouts/" + CDIR + list(config["atlite"]["cutouts"].keys())[0] + ".nc"
+    if os.path.exists(cutout_fl):
+        raise Exception(
+            "A cutout file '"
+            + cutout_fl
+            + "' still exists. Stopping to avoid over-writing it"
+        )
```
davide-f: Hello @ekatef and amazing feature :D
I feel like tlines from 304 to 310 are not needed from a functionality perspective.
To make it more general, we may fuse the two blocks into something like:
`
```
cutout_names = [d_value["cutout"]for tech, d_value in config["renewable"]] + list(config["atlite"]["cutouts"].keys())
for c_name in cutout_names:
       cutout_fl = "cutouts/" + CDIR + c_name + ".nc"
       ...
```

Snakefile
davide-f: However, I added a comment below to encorporate the two comments.

Snakefile
```diff
@@ -290,6 +290,24 @@ rule build_bus_regions:
...
 if config["enable"].get("build_cutout", False):
+    for technology in cutouts_in_config:
+        cutout_fl = (
+            "cutouts/" + CDIR + config["renewable"][technology]["cutout"] + ".nc"
+        )
+        if os.path.exists(cutout_fl):
+            raise Exception(
+                "A cutout file '"
+                + cutout_fl
+                + "' still exists. Stopping to avoid over-writing it"
```
davide-f: We need to improve the message to something like:
`build_cutout enabled and cutout file {c_name} risks to be overwritten. If this is an intended behaviour, please delete the file and rerun the rule, otherwise disable the build_cutout rule in the config file`

davide-f (commented):
Cool @ekatef :D
Added few comments :) close to merge :)

Snakefile
ekatef: Mmm... You are absolutely right! Not sure how have I managed to test it :/
Thanks a lot for the catch and fixed!

Snakefile
ekatef: Agree :) Implemented

Snakefile
ekatef: Done! 
A cutout name wildcard doesn't work for some reason, so strings concatenation is used.

ekatef:
> Cool @ekatef :D Added few comments :) close to merge :)

@davide-f Thanks a lot for the great review :D
Tried to improve.

Snakefile
```diff
@@ -289,7 +289,20 @@ rule build_bus_regions:
         "scripts/build_bus_regions.py"
 
 
+res_tech_names = list(config["renewable"].keys())
+config_cutouts = [
+    (lambda x: config["renewable"][x]["cutout"])(x) for x in res_tech_names
+] + list(config["atlite"]["cutouts"].keys())
```
davide-f: May you test if the following works?
(except pre-commit...)
```suggestion
config_cutouts = [d_value["cutout"] for tc, d_value in config["renewable"].items()] + list(config["atlite"]["cutouts"].keys())
```

davide-f: If yes, I think we could move this definition and the for loop in a function to be placed somewhere and then call the function here.
That may increase the readability of the script.
In the future, the auxiliary functions may be moved in a separate function as pypsa-eur did

davide-f (commented):
Close to finalizing :D @ekatef great!

Snakefile
ekatef: Wow, nice solution! Implemented :D

Snakefile
```diff
@@ -289,7 +289,23 @@ rule build_bus_regions:
         "scripts/build_bus_regions.py"
 
 
+def check_and_keep_cutout(config=config):
```
davide-f: Super! 🗡️ 

Sorry for being picky...
Do you mind adding a very small docstring to describe what this function does, something like "This function does this"

No need to describe inputs/output, very basic.
Moreover, the name of the function may be: "error_with_existing_cutouts" or something like that?

check and keep cutout seems to me a bit obscure but I may be wrong. Feel free to find better naming :)

ekatef: > Super! 🗡️
> 
> Sorry for being picky... Do you mind adding a very small docstring to describe what this function does, something like "This function does this"
> 
> No need to describe inputs/output, very basic. Moreover, the name of the function may be: "error_with_existing_cutouts" or something like that?
> 
> check and keep cutout seems to me a bit obscure but I may be wrong. Feel free to find better naming :)

Great point :D Thanks a lot for reviewing!
Fixed :)

davide-f:
Fantastic @ekatef !!!
Ready to merge :)

mfripp:
I think this code is causing a crash, at least on my system. When I run `snakemake --cores 1 solve_all_networks` with `build_cutout: true`, it successfully builds the cutout, but then crashes the workflow later. 

Specifically, near the end of the workflow, snakemake calls another copy of itself to run the `solve_network` rule. This new process parses `config.yaml`, discovers that `build_cutout` is still set and the cutout file now exists, and terminates with the error message from this code.

I'm not sure why snakemake calls another instance of itself, so I don't know how to fix this problem cleanly. For now, I'm just changing `build_cutout` to `false` and re-running `snakemake --cores 1 solve_all_networks`, but that's not a beautiful solution. I tried putting in a breakpoint() in the Snakefile just before the error was raised, and this was the contents of `sys.argv` at that point (I've replaced my home dir with `~`):

```
['~/opt/miniconda3/envs/pypsa-earth/lib/python3.10/site-packages/snakemake/__main__.py', '--snakefile', '~/path/pypsa/pypsa-earth/Snakefile', '--target-jobs', 'solve_network:simpl=,clusters=10,ll=copt,opts=Co2L-3H', '--allowed-rules', 'solve_network', '--cores', '1', '--attempt', '1', '--force-use-threads', '--resources', 'mem=11950', '--quiet', '--force', '--keep-target-files', '--keep-remote', '--max-inventory-time', '0', '--nocolor', '--notemp', '--no-hooks', '--nolock', '--ignore-incomplete', '--rerun-triggers', 'input', 'params', 'code', 'mtime', 'software-env', '--skip-script-cleanup', '--conda-frontend', 'mamba', '--wrapper-prefix', 'https://github.com/snakemake/snakemake-wrappers/raw/', '--local-groupid', 'local', '--latency-wait', '5', '--scheduler', 'ilp', '--scheduler-solver-path', '~/opt/miniconda3/envs/pypsa-earth/bin', '--default-resources', 'tmpdir=system_tmpdir', '--mode', '1']
```

Is there some way to do this check only when snakemake is actually about to delete the cutout file?

ekatef:
Hello @mfripp and thanks a lot for notifying! 

I confirm that this issue is reproducible and that is indeed a side effect of the precautions to keep the cutout. However, there is currently no other way ensure that the cutout file is safe, as Snakemake is designed to silently remove any file which can be qualified as outdated. 

As a comment, the next Snakemake version has the `protected` functionality which can be used to re-implement this part in a more elegant way. At the time being, there are some backward compatibility concerns which prevent us from using this newest version. But we'll definitely keep your report in mind after an environment upgrade will be implemented!

ekatef:
The issue also arises when executing `run_all_scenarios` at the final stage which calls Snakefile directly.

ollie-bell:
> I'm not sure why snakemake calls another instance of itself

Did you ever understand why this happens? It seems like snakemake is re-running itself each time it completes a rule. I find that with `build_cutout: true` it terminates immediately after building the cutout when it gets to the next rule `copy_defaultnatura_tiff`.

ollie-bell:
I did a bit of debugging and figured out that this happens when there is a `run` or `shadow` statement in a rule. They both invoke a snakemake subprocess call:

https://github.com/snakemake/snakemake/blob/cc17fc11cf83b03f9c528bab279d9f394789fb55/snakemake/executors/local.py#L175-L187

We can detect if the snakefile is being run as subprocess or not by checking the `--mode` argument. Additionally check if `build_cutout` is a target rule:

```python
cmd = " ".join(sys.argv)
if "--mode 1" not in cmd or "build_cutout" in cmd:
    terminate_if_cutout_exists(config)
```

Alternatively the rules can be refactored to remove `run` or `shadow` statements:

The `copy_defaultnatura_tiff` rule can be fixed by changing:

```python
        run:
            import shutil

```

to:

```python
        shell:
            "cp {input} {output}"
```

And you could do a similar thing for the `retrieve_cost_data` rule with `mv` instead of `cp` (although not sure if `cp` and `mv` are Windows compatible though?)

The `solve_network` rule can be fixed by removing the `shadow` statement - is it necessary?

The `run_scenario` rule could perhaps be fixed by changing the `run` statement into a `script` statement (haven't tested this)? Similarly for the `run_tests` rule.

`clean` looks like the `run` statement could perhaps be made into a `shell` statement.

Issue #499, opened 2022-10-27
Linkers to accommodate custom data


davide-f:
## Describe the feature you'd like to see

There is the need for easy way to integrate additional custom data of:
- [x] Power lines
- [x] Custom powerplants ( already available as custom_powerplants.csv )
- [x] Time series of demand
- [ ] Time series of renewable production

To integrate demand time series and renewable production the approach is now not easy: there is the need to create a nc file with the same characteristics of the GEGIS and atlite modules.
That is everything but easy and needs to be improved.

pz-max:
One note from the PyPSA user meeting. We need to make this linkers as clear as possible such that people can:
- Easy understand and analyse the data status
- Add new data over simple interfaces
- Optional. Write instructions on how to contribute to global sustainable databases OpenStreetMap etc.

ekatef:
Adding a points for a particular use case (@davide-f, @AnasAlgarei, @pz-max thanks a lot for the discussion!). It can be needed to have at least one bus for each administrative region in case if gadm clustering is used. For some areas where OSM data seems to be incomplete, there may be need to add such buses. That may be the case, for example, for Colombia, Chile, Vietnam (picture from [openinframap](https://openinframap.org/)):

<img width="196" alt="image" src="https://github.com/pypsa-meets-earth/pypsa-earth/assets/30229437/7f38c811-facb-46db-ba53-73a0891ddc9f">

To keep the gadm areas interconnected, a custom csv can be fetched to add buses for the areas where they are missed. To create such csv, we need a code that checks for every gadm shape if a bus is already available, and if not stores the centroid of the bus in an additional file with the format matching the substation geojson. A branch or a notebook can be added to store such code.

As a more general solution, a function can be added to `base_network`, but it requires the additional input of gadm shapes.

davide-f:
Thanks to #842 lines and links are now fully supported

ekatef:
#987 should cover the demand part.
@davide-f would you mind if I'll tick the demand box? 🙃

davide-f:
Agree :D

ekatef:
> Agree :D

Thanks! A great feeling for Friday evening to get something actually done XD

ekatef:
To facilitate an overview, adding links to relevant PRs in progress:
- adding a linker for renewable profiles #1796 (a prototype for inflow has been implemented in PyPSA-Zambia model in [this PR](https://github.com/open-energy-transition/pypsa-zambia/pull/178))
[...]

Issue #276, opened 2022-02-17
Linkers: options to customize data inputs by country


davide-f:
## Linkers to accommodate refined data country-specific
During the workflow meeting of 11/02/2022, linkers have been heavily discussed and this issue summarizes the expected improvements.

Linkers shall be configurable options to enable users to specify what model to optionally incorporate non-default data for each country.
The resources that may be subject to linking systems may be:
- Demand estimation
- Powerplant data
- Climate/renewable production
- Network data

Users shall be able to incorporate:
- Local [non-open] data, in the form of csv or nc files. This to help users with secret data in running and using the tool with their own personal data
- Workflows: the workflow shall be structured so as to easily incorporate workflows that popolate specific input files

To accomplish the above, there is the need to:
- Clean the workflow to (1) ease the accessibility to key files and (2) improve readability of the relevant data to modify
- Create linkers structures for more customizable features

davide-f:
I'm closing this issue as the original scope of this issue has been achieved.
More specific issues can be added on demand for specific improvements of data imports, this one is very general and it was aimed to include interfaces to demand etc that were non existent at the time this issue was created.
If deemed appropriate, feel free to open it.

PR #840, opened 2023-08-18
PR to allow the use of a customized data for lines (changes OSM raw data for a custom file) 

carlosfv92:
# Closes # (if applicable).

*the snakefile has been modified (lines 181) to include an additional input in the clean_osm_data rule named "custom_lines" which reads a new file in the data folder named "custom_lines.geojson"
*The gitignore file added the custom_line path in the data folder to be included recognized (custom _lines.geojson has some predefined values as an example)
*the config.default file has been modified (lines 104) to include a flag named "use_custom_lines" to define if the model will use the default osm data (False) or the custom file (True)
*the clean_osm_script has been adapted (line 855) to include a conditional (if), linked to the flag in the config file in order to use either the default osm data or the custom data file to create the base dataframe of lines (df_lines)

## Checklist

- [ x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [ x] I tested my contribution locally and it seems to work fine.
- [ x] Code and workflow changes are sufficiently documented.
- [ x] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`. (no additional dependencies)
- [ x] Changes in configuration options are added in all of `config.default.yaml`.


ekatef:
Hello @carlosfv92! Thanks a lot for the PR. 

We have currently two open issues on adding custom power grid data: #276 and #499. So, what you are doing is definitely in high demand.

A question: how difficult has it been to transform your data into the needed format of `custom_lines.geojson`? Do you probably have any hints which could be helpful for doing such a transformation?

carlosfv92:
Hey @davide-f ! Finally got time to do some testing for the use of custom information and came up with this! (and also reading about how to do a pull request and check/correct the errors that appear during the pre-commit testing hahaha).

The solution is simple and basically considers an additional input to be read in the snakefile (custom_lines), a flag in the config file (use_custom_lines), and a conditional during loading function of the information of lines in the clean_osm_data script.
 
I tried doing something similar in later stages of the workflow (build_osm_network and base_network) but it turns out that creating and storing the files in .csv, or modifying the .nc file, is way more tricky due to formatting issues (for example commas that become points or semicolons base on the configuration of the computer and can lead to issues)... so creating an alternative file using the all_raw_lines.geojson became the simplest solution.

For the moment, this seems to be working just fine, and has already been tested for Bolivia. If the PR seems consistent and useful I'll try to do the same once I get my hand on substation information.

carlosfv92:
> Hello @carlosfv92! Thanks a lot for the PR.
> 
> We have currently two open issues on adding custom power grid data: #276 and #499. So, what you are doing is definitely in high demand.
> 
> A question: how difficult has it been to transform your data into the needed format of `custom_lines.geojson`? Do you probably have any hints which could be helpful for doing such a transformation?

Hey @ekatef ! You commented before I finished my post hahaha.
Creating the file in .geojson is definitely more complex than doing a CSV (since data I had was in csv format) but leads to fewer problems when trying to incorporate them into the script. What I did was using the .csv information, created a new dataframe in python, renamed and reordered everything based on the format of the all_raw_lines file (downloaded from OSM) and then started testing if it was properly recognized by running the clean_osm_data script in debugging mode.
Even though this is a little more time-consuming, the custom_lines file (based on the raw geojson file) can be easily replaced at a very specific section of the script clean_osm_data, while exchanging the output (based on the clean csv file) might result in output files that aren't recognized in the later stages (turns out that my computer saves the file with different separators like semicolons or points instead of just commas... and does that depending on its mood)

ekatef:
@carlosfv92 thanks a lot for the detailed explanations! Sounds like a big piece of work done :) Your experience is very valuable to improve data interface of the model. 

I think the architecture you used looks great, while a step of data transformation sounds like the trickiest part. Regarding the custom data format, my feeling is that using csv-format could be probably a bit more user-friendly. Although, agree that the separators issues may be quite annoying when importing. Have you tried to define a separator option explicitly in `to_csv()`? According to the [function API](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_csv.html), `sep` argument should do the job.

davide-f:
Amazing @carlosfv92 for the contribution and also documentation!
I think in terms of functionality you got to the point! great work!

From an implementation point of view, I was wondering whether it may be nice to have the optional inputs placed under a new key in the config called "custom_data:".
Having the custom file into the snakemake inputs may require to have that file regardless of whether the option is actually used or not and that may increase the overhead and complexity of the workflow.
My feeling is that an option in the config may be easier from a user perspective, what do you think?

In particular, under the enable section, we could add a new section called "custom_data" that contains the possible additional data to feed in the workflow.
For example, we could have:
```yaml
custom_data:
    substations:  []  # False or list of geojson files with the additional substations data. Format consistent to corresponding file into resources/osm/raw folder
    lines: []
````

Moreover, I really like the boolean flag into the clean_osm_data option that triggers whether the option shall be included or not.
That option could have multiple values, such as:
- False or "osm_only": then no custom data are merged and only osm ones are used
- "add": in this case the custom data are added to the osm ones
- "custom_only": in this case custom data are used instead of the osm ones

What do you think?
I completely agree on starting finalizing first the custom lines and then generalize it for other osm inputs.

Moreover, I see you pushed the geojson file into the commit history, please, avoid that because that leads to adding binaries to the repository and even that file is dropped afterwards, the data would be kept anyway into the history.
To prevent that, we have 2 options now:
1. (preferred) you create a new PR where you replicate adding the changes (ideally with our comments) without adding the custom geojson
2. this PR will be merged using "squash and merge": this means that all changes that you make will be squashed into 1 commit only. This option will enable us to keep using this PR and commit, but the commit history will be way shorter unfortunately.

What is your preferred option?

**Comment by davide-f** [comment]
Moreover, I see you may not be using the pre-commit locally. 
If you are using vscode, you may do that with:
1. in view-command pallette-> Python: Select intepreter-> PyPSA-Earth
2. open a new terminal view->command pallette->new terminal
3. in the terminal digit "conda activate pypsa-earth" (if not already in the env)
4. digit "pre-commit install"

After this, the pre-commit should be installed for the local repository.
This should be repeated for every local repositories you may be using this.

Could also be good to add this comment into the contributing section... (i'll add an issue)

**Comment by carlosfv92** [comment]
> @carlosfv92 thanks a lot for the detailed explanations! Sounds like a big piece of work done :) Your experience is very valuable to improve data interface of the model.
> 
> I think the architecture you used looks great, while a step of data transformation sounds like the trickiest part. Regarding the custom data format, my feeling is that using csv-format could be probably a bit more user-friendly. Although, agree that the separators issues may be quite annoying when importing. Have you tried to define a separator option explicitly in `to_csv()`? According to the [function API](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_csv.html), `sep` argument should do the job.

Hey @ekatef ! To be honest, I didn't really try that since I wasn't creating a new file, I was just using an already existing .csv file (with commas as separators) and saving it after some rearranging (and I never had the issue of the separator's changing before)... I'll try to see if what you mentioned fixes the issue, Thanks for the suggestion!

**Comment by carlosfv92** [comment]
> Moreover, I see you may not be using the pre-commit locally. If you are using vscode, you may do that with:
> 
> 1. in view-command pallette-> Python: Select intepreter-> PyPSA-Earth
> 2. open a new terminal view->command pallette->new terminal
> 3. in the terminal digit "conda activate pypsa-earth" (if not already in the env)
> 4. digit "pre-commit install"
> 
> After this, the pre-commit should be installed for the local repository. This should be repeated for every local repositories you may be using this.
> 
> Could also be good to add this comment into the contributing section... (i'll add an issue)

Hey @davide-f ! I can go with either, but just to clarify... At the beggining I didn't want to add the custom_lines.geojson file in the folder data (since it included a new file and changes in the git.ignore), however I had to add it because otherwise the pre-commit couldn't compile. 
Just an idea (to avoid creating a new pull request), couldn't I just delete the file in my repository and update the pull request?

**Comment by ekatef** [comment]
Hey @carlosfv92 and @davide-f! Thanks for the great discussion and sorry for jumping-in :) Actually, my feeling is that cleaning-up commits history is a very practical topic for which we may need a (more or less) re-usable solution.

@carlosfv92 You are right that it's absolutely possible to remove the binary file and commit the changes, but the file will still remain in git-history making it quite heavy. So, to really get rid of it, I'd need to modify git-history to find and remove a commit where geojson has been added. It's possible, for example, with the following git spell:

``` bash
# create a back-up copy of your branch
git checkout your_branch
git branch your_backup_branch

# look into your git history to find an id for a commit you want to remove
git log --oneline -n 10

# remove the commit with id == commit-id from the history
git rebase --onto commit-id^ commit-id

# push changes into the remote repo
git push --force origin your_branch
```
However, modifying of git history is always kind of black magic which should be used very carefully to avoid loss of your work. But if you feel adventurous, it may be worth trying ;)

carlosfv92:
Hey @ekatef and @davide-f ! So, I tried addressing some of the comments mentioned:
*To avoid messing with the inputs in the snakefile I skipped that part and used some functions to get the folder path and get to the file, which took me more time than what I'm willing to admit 
*The new version allows the test-run to work without the geojson file in the data folder, but, to actually test the function one has to have the file in the data folder
*I think having everything under the clean_osm_data_options (at least for now) seems quite straightforward and actually helpful, since it kind of forces you to check the input/output files used in the script if you want to create your own custom files

For the local pre-commit, if it's not an issue, I'll keep doing it online since I'm having a problem with the installation of an alternative environment (and I'm using my main environment to do some runs and I don't want to break it). Weirdly enough, the environment creation is taking over two hours with conda and mamba can't create it :/

The new PR is: Custom lines PR two, electric boogaloo - Allows the use of a customized data for lines instead of OSM data #842

davide-f:
Cool @carlosfv92 :)
Definitely understandable, please let us know the issues that you are facing that we may help you.
Feel free to add a post in the support channel in discord.

If I understand correctly, can we then close this PR as the otherone will be the main one?

carlosfv92:
Hey @davide-f ! yep we can close this one now.

davide-f:
Closing as agreed

PR #888, opened 2023-10-06
Databundle check size PR #1

Emre-Yorat89:
# Closes # (if applicable).

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,7 +28,7 @@ repos:
 
 # Find common spelling mistakes in comments and docstrings
 - repo: https://github.com/codespell-project/codespell
-  rev: v2.2.5
+  rev: v2.2.6
   hooks:
   - id: codespell
     args: ['--ignore-regex="(\b[A-Z]+\b)"', '--ignore-words-list=fom,appartment,bage,ore,setis,tabacco,berfore,fo,FO']
--- a/doc/how_to_docs.rst
+++ b/doc/how_to_docs.rst
@@ -48,7 +48,7 @@ form `doc/requirements.txt`. First, we recommend creating a fresh python environ
 
     .../pypsa-earth % conda activate pypsa-earth-docs
 
-Next, install the packages specified in `doc/requiremnts.txt` using `pip`:
+Next, install the packages specified in `doc/requirements.txt` using `pip`:
 
 .. code:: bash
 
--- a/scripts/build_bus_regions.py
+++ b/scripts/build_bus_regions.py
@@ -98,7 +98,7 @@ def custom_voronoi_partition_pts(points, outline, add_bounds_shape=True, multipl
 
         # to avoid any network positions outside all Voronoi cells, append
         # the corners of a rectangle framing these points
-        vor = Voronoi(
+        vcells = Voronoi(
             np.vstack(
                 (
                     points,
@@ -114,7 +114,7 @@ def custom_voronoi_partition_pts(points, outline, add_bounds_shape=True, multipl
 
         polygons_arr = np.empty((len(points),), "object")
         for i in range(len(points)):
-            poly = Polygon(vor.vertices[vor.regions[vor.point_region[i]]])
+            poly = Polygon(vcells.vertices[vcells.regions[vcells.point_region[i]]])
 
             if not poly.is_valid:
                 poly = poly.buffer(0)


scripts/retrieve_databundle_light.py
```diff
@@ -447,7 +448,7 @@ def _check_disabled_by_opt(config_bundle, config_enable):
 
 
 def get_best_bundles_by_category(
-    country_list, category, config_bundles, tutorial, config_enable
+    country_list, category, config_bundles, tutorial, config_enable, df
```
davide-f: The df dataframe here is not needed I believe, but rather we create it inside the function itself using the config_bundles.

I think we can reset this branch by:
git reset --hard b205e83c3b631448a778e29257eed551790b9d5c
git push --force

Be aware that after the git push, this PR will automatically close, but no problem, reopen it after you add other commits

scripts/retrieve_databundle_light.py
```diff
@@ -496,6 +497,9 @@ def get_best_bundles_by_category(
     if dict_n_matched:
         # if non-empty, then pick bundles until all countries are selected
         # or no more bundles are found
+        df2 = pd.DataFrame()
+        df2["number of matches"] = dict_n_matched
+        df = pd.concat([df, df2])
         dict_sort = sorted(dict_n_matched.items(), key=lambda d: d[1])
 
         current_matched_countries = []
```
davide-f: I confirm that this place is where the dataframe shall be placed.
You may initialize the empty dataframe as:
`df_matched = pd.DataFrame(columns=["bundle_size", "n_matches", "matched_countries"])`

where:
- the first column contains the number of countries supported by each bundle
- the second column contains the number of matched countries of the bundle
- the third column contains the set of matched countries of the bundle (whose length is the previous point)

Then, a subsequent for loop shall fill the rows of that dataframe.
In pseudocode:
``` python
for d_val, val in dict_sort.items():
    len_bundle = {size of the bundle}
    matched_countries = {get the matched countries using intersection and shown below}
    df_matched.loc[d_val, :] = [len_bundle, len(matched_countries), matched_countries]
```

Then, we sort the dataframe:
`df_matched.sort_values(columns=["n_matched", "bundle_size"], inplace=True, ascending=False)`

Finally, we process the dataframe as follow:
- if the first element of the sorted dataframe has n_matched == the number of desired countries, then that bundle contains all necessary data and no other bundles are needed.
   In pseudocode:
   ```
   if {first element of the dataframe has n_matched equal the number of desired countries}:
       returned_bundles = [df_matched.index[0]]
   ```
- otherwise, we need to revise the algorithm to return as many bundles as necessary to cover the bundles.
   in pseudocode:
   ```
   else:
       df_matched.drop_duplicates(subset=["matched_countries"], inplace=True)
       returned_bundles = []
       missing_countries = set(country_list)
       for d_k, dval in df_matched.items():
           {go row by row and check if the set of countries of the current row has some intersections with missing_countries, if so add d_k to returned_bundles, otherwise skip the row with continue}
    ```

The code in the for loop below can help

davide-f: Note @Emre-Yorat89 , take a screenshot of this comment because if you reset the branch it may disappear

davide-f (commented):
Hello @Emre-Yorat89 :D
Thanks for taking the task :)
Attached are some comments!

Please, also I see that a pre-commit has been added.
Make sure:
- if you use vscode, to make pypsa-earth the environment of default (view->command pallette -> Python: Select Interpreter -> Python ... (pypsa-earth)
- to have the pre-commit locally installed:
   conda activate pypsa-earth
   pre-commit install

So that anytime you commit with vscode the pre-commit is executed.
If you use command line, please, make sure to activate the pypsa-earth environment during the commit messages so that the pre-commit is executed :)

Great :D

scripts/retrieve_databundle_light.py
Emre-Yorat89: Hi Davide,
I have tried to create dataframe in the function itself. However, the function is called multiple times by get_best_bundles function. Therefore, The loop deletes everything in the data and creates empty one. So my solution was to create empty dataframe before entering the get_best_bundles_by_category and prevent the data loss. If you have a way around this problem I would appreciate your code help :) Also I could not understand what you mean by creating df dataframe using config_bundles.

https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/scripts/retrieve_databundle_light.py#L570-L575
(The for loop)

Many thanks for the review. I will revise it upon your new comments and directions.

scripts/retrieve_databundle_light.py
Emre-Yorat89: Okay :) Thank you.

scripts/retrieve_databundle_light.py
davide-f: so, I agree that the function will be called multiple times and that's no problem, the overhead is also limited.
Moreover, every dataframe refers to a different type of databundle, so that's ok.

Maybe you need a bit of contextualization; I try to add some context below. That is specified on top of the script, but also feel free to revise the docstring of that function to improve the clarity.

If you notice the databundles, they have a flag named "category".
This function that you are editing is called for every value of category to identify the sets of bundles that are needed to download by category.
Generally, therefore, it may be called four times: one for "common", one for "data", one for "cutouts" value and one for "landcover"
For every call, the dictionary dict_n_elems is filled only with those bundles that:
1. match the category value of the current execution (https://github.com/pypsa-meets-earth/pypsa-earth/blob/95a544395e44b3a159dc4267348536599430edc5/scripts/retrieve_databundle_light.py#L489)
2. match the tutorial flag: if the tutorial is true, only the tutorial bundles are checked, if else the converse (https://github.com/pypsa-meets-earth/pypsa-earth/blob/95a544395e44b3a159dc4267348536599430edc5/scripts/retrieve_databundle_light.py#L490)
3. a further check is performed to exclude bundles whose outputs are no more needed, for example if build_cutout is enabled that no more requires the cutouts flag. This is done using `_check_disabled_by_opt` https://github.com/pypsa-meets-earth/pypsa-earth/blob/95a544395e44b3a159dc4267348536599430edc5/scripts/retrieve_databundle_light.py#L491

So the first goal would be to create a df_matches that replaces the dict_n_matches and fill it with additional columns.
The comment on the following lines adds more and provides a draft structure of the revisione.

Hope this message better clarifies some doubts, if something is not clear, please, let me know :)

PR #890, opened 2023-10-08
Add an error message when GADM server is down

ekatef:

Add an error message for a case when GADM server is down.

## Checklist

### DIFF
--- a/scripts/build_shapes.py
+++ b/scripts/build_shapes.py
@@ -106,7 +106,7 @@ def download_GADM(country_code, update=False, out_logging=False):
         os.makedirs(os.path.dirname(GADM_inputfile_gpkg), exist_ok=True)
 
         try:
-            r = requests.get(GADM_url, stream=True, timeout=1)
+            r = requests.get(GADM_url, stream=True, timeout=300)
         except (requests.exceptions.ConnectionError, requests.exceptions.Timeout):
             logger.error(f"GADM server is down at {GADM_url}")
             raise Exception(f"GADM server is down at {GADM_url}")
--- a/scripts/build_shapes.py
+++ b/scripts/build_shapes.py
@@ -108,11 +108,11 @@ def download_GADM(country_code, update=False, out_logging=False):
         try:
             r = requests.get(GADM_url, stream=True, timeout=300)
         except (requests.exceptions.ConnectionError, requests.exceptions.Timeout):
-            logger.error(
+            raise Exception(
                 f"GADM server is down at {GADM_url}. Data needed for building shapes can't be extracted.\n\r"
             )
         except Exception as exception:
-            logger.error(
+            raise Exception(
                 f"An error happened when trying to load GADM data by {GADM_url}.\n\r"
                 + str(exception)
                 + "\n\r"
--- a/scripts/build_shapes.py
+++ b/scripts/build_shapes.py
@@ -108,13 +108,15 @@ def download_GADM(country_code, update=False, out_logging=False):
         try:
             r = requests.get(GADM_url, stream=True, timeout=300)
         except (requests.exceptions.ConnectionError, requests.exceptions.Timeout):
-            logger.error(f"GADM server is down at {GADM_url}")
-            raise Exception(f"GADM server is down at {GADM_url}")
+            logger.error(
+                f"GADM server is down at {GADM_url}. Data needed for building shapes can't be extracted.\n\r"
+            )
         except Exception as exception:
             logger.error(
-                "An error happened when trying to load GADM data by {GADM_url}"
+                f"An error happened when trying to load GADM data by {GADM_url}.\n\r"
+                + str(exception)
+                + "\n\r"
             )
-            raise Exception(f"Something went wrong when connecting to {GADM_url}")
         else:
             with open(GADM_inputfile_gpkg, "wb") as f:
                 shutil.copyfileobj(r.raw, f)
--- a/scripts/build_shapes.py
+++ b/scripts/build_shapes.py
@@ -105,7 +105,17 @@ def download_GADM(country_code, update=False, out_logging=False):
         #  create data/osm directory
         os.makedirs(os.path.dirname(GADM_inputfile_gpkg), exist_ok=True)
 
-        with requests.get(GADM_url, stream=True) as r:
+        try:
+            r = requests.get(GADM_url, stream=True, timeout=1)
+        except (requests.exceptions.ConnectionError, requests.exceptions.Timeout):
+            logger.error(f"GADM server is down at {GADM_url}")
+            raise Exception(f"GADM server is down at {GADM_url}")
+        except Exception as exception:
+            logger.error(
+                "An error happened when trying to load GADM data by {GADM_url}"
+            )
+            raise Exception(f"Something went wrong when connecting to {GADM_url}")
+        else:
             with open(GADM_inputfile_gpkg, "wb") as f:
                 shutil.copyfileobj(r.raw, f)
 


ekatef:
A suggestion on how to make more clear that GADM server is down, when running the workflow. @davide-f may you please have a look? :)

Error messages are duplicated to keep error message both in console and in logs. 

It could probably make sense to stop execution with something like `sys.exit(1)` instead of raising exceptions. Although, it would be not quite consistent with the approaches used in different parts of the code.

scripts/build_shapes.py
```diff
@@ -105,7 +105,17 @@ def download_GADM(country_code, update=False, out_logging=False):
         #  create data/osm directory
         os.makedirs(os.path.dirname(GADM_inputfile_gpkg), exist_ok=True)
 
-        with requests.get(GADM_url, stream=True) as r:
+        try:
+            r = requests.get(GADM_url, stream=True, timeout=1)
```
davide-f: Fantastic revision :)
May be nice to reduce the timeout to about 5 minutes (5*60 seconds) or alike, what do you think?

scripts/build_shapes.py
```diff
@@ -105,7 +105,17 @@ def download_GADM(country_code, update=False, out_logging=False):
...
+        except (requests.exceptions.ConnectionError, requests.exceptions.Timeout):
+            logger.error(f"GADM server is down at {GADM_url}")
+            raise Exception(f"GADM server is down at {GADM_url}")
```
davide-f: Having logger and raise is a bit annoying.
Could you try if logger.critical(.) or logger.exception(.) also trigger an exit of the execution?
I'd avoid the execution of "raise" because it still leads to dirty logging.
In the general exception tracker, the textual description of the exception `str(exception)` may be added as well.

davide-f (commented):
Nice draft :D few little comments! :)

scripts/build_shapes.py
ekatef: > Fantastic revision :) May be nice to reduce the timeout to about 5 minutes (5*60 seconds) or alike, what do you think?

Thanks 😊 
Agree, fixed.

scripts/build_shapes.py
ekatef: Have played around logger, but there seems to be no way to stop execution just by logging tools :( As far as I was able to find, both `logger.critical( )` and `logger.exception( )` have a purely information function.

Agree  that `raise` doesn't look nice. But executing is anyway stopped when geopandas tries to read a non-existing file. So, probably we can just make an error message more clear. What do you think?

ekatef:
> Nice draft :D few little comments! :)

Thanks a lot!! :)) Added a next iteration :)

scripts/build_shapes.py
davide-f: An approach may be to use sys.exit() to close the execution forcefully.

Regarding the exception trace, I just realized that there is the option "exc_info=True" that may automatically log the trace.
Ok also as is now, it is is nicer in the log we could also take it into consideration.

What do you think?

ekatef: I think, implementation with `exc_info=True` is definitely nicer :D Revised and ddded `sys.exit(1)`, which seems to give much more clear logging output.

ekatef:
> Nice draft :D few little comments! :)

@davide thanks a lot for reviewing! :)

As a comment, I fear a currently implemented approach is not perfectly consistent with the rest of the code. Usually, we are rising exceptions to handle errors, for example: https://github.com/pypsa-meets-earth/pypsa-earth/blob/b205e83c3b631448a778e29257eed551790b9d5c/scripts/build_shapes.py#L577-L580 . Which also means that error messages are not captured into log files, while console output looks a sometimes not perfectly clear.

Now we are testing `logger.error(.)` + `sys.exit(.)` which seems to look and work good, but is not perfectly consistent with the other code. Happy to revise the current approach or make adjustments across the code. What is your feeling about that? :)

davide-f:
Agree, nice catch! :)
Also less lines of code :)

ekatef:
@davide-f thanks a lot for looking into logs for uncaught exceptions :)  Have tried an [implementation](https://stackoverflow.com/a/16993115/8465924) you referred, and I think the results is very nice!

Snakemake error listing remains and is verbose as usual, while a cleaner output is added to a log file. An example of log output provoked by calling an undefined variable:

```python
ERROR:_helpers:An error happened during workflow execution.
  File "/Users/ekaterina/Documents/_github_/pypsa-earth/.snakemake/scripts/tmpz3zieg8k.build_shapes.py", line 942, in <module>
    print(x + y)
NameError: name 'x' is not defined
```

I think, that may be exactly what we do need. What is your feeling about that?

davide-f:
> @davide-f thanks a lot for looking into logs for uncaught exceptions :) Have tried an [implementation](https://stackoverflow.com/a/16993115/8465924) you referred, and I think the results is very nice!
> 
> Snakemake error listing remains and is verbose as usual, while a cleaner output is added to a log file. An example of log output provoked by calling an undefined variable:
> 
> ```python
> ERROR:_helpers:An error happened during workflow execution.
> Traceback (most recent call last):
>   File "/Users/ekaterina/Documents/_github_/pypsa-earth/.snakemake/scripts/tmpz3zieg8k.build_shapes.py", line 942, in <module>
>     print(x + y)
> NameError: name 'x' is not defined
> ```
> 
> I think, that may be exactly what we do need. What is your feeling about that?

So, for this PR, I personally believe that simply using the raise instead of logging.error was enough, while the better handling of the exception could be done in a separate PR.

I like the implementation, but I'd keep them separate to ease the revision too as the revision of the logger/exception handling should be done to all other scripts as well.
Would you mind moving these interesting changes to a different PR so we could discuss on that?

To better handle the errors, I strongly like this draft.
I'd envision in the helpers a function, like create_logger or alike that accepts a string input that is the logger name and encapsulates this code:

``` python
logger = logging.getLogger(__name__)
logger.setLevel(logging.INFO)

handler = logging.StreamHandler(stream=sys.stdout)
logger.addHandler(handler)
sys.excepthook = handle_exception
```

However, since implementing this requires changing all scripts, I'd keep for a separate PR, what do you think?

ekatef:
> So, for this PR, I personally believe that simply using the raise instead of logging.error was enough, while the better handling of the exception could be done in a separate PR.
> 
> I like the implementation, but I'd keep them separate to ease the revision too as the revision of the logger/exception handling should be done to all other scripts as well. Would you mind moving these interesting changes to a different PR so we could discuss on that?
> 
> To better handle the errors, I strongly like this draft. I'd envision in the helpers a function, like create_logger or alike that accepts a string input that is the logger name and encapsulates this code:
> 
> ```python
> logger = logging.getLogger(__name__)
> logger.setLevel(logging.INFO)
> 
> handler = logging.StreamHandler(stream=sys.stdout)
> logger.addHandler(handler)
> sys.excepthook = handle_exception
> ```
> 
> However, since implementing this requires changing all scripts, I'd keep for a separate PR, what do you think?

Agree that it's better to keep it simple. Revised :)

Have moved a part with the exceptions hook to #898

scripts/build_shapes.py
```diff
@@ -105,7 +105,19 @@ def download_GADM(country_code, update=False, out_logging=False):
...
+        try:
+            r = requests.get(GADM_url, stream=True, timeout=300)
+        except (requests.exceptions.ConnectionError, requests.exceptions.Timeout):
+            raise Exception(
+                f"GADM server is down at {GADM_url}. Data needed for building shapes can't be extracted.\n\r"
+            )
+        except Exception as exception:
+            raise Exception(
+                f"An error happened when trying to load GADM data by {GADM_url}.\n\r"
+                + str(exception)
```
davide-f, line 117: Here you used also the option "exc_info=True" as an alternative, you mentioned it was nice.
Should we go instead with the current implementation?

ekatef: Yeah, absolutely, it's convenient to use `exc_info` is king of an argument in `logger.error()`, but `Exception` doesn't agree to accept `exc_info` as an argument. 

It's possible to call `sys.exc_info()`, but it looks less convenient as compared with `str(exception)`. Do I miss anything? :)

davide-f: It is good the way it is, ready to merge :)
Many thanks! :D

davide-f (approved):

PR #903, opened 2023-10-18
Add functionality to merge each isolated bus to the backbone network

### LINKED ISSUE #852: Add option to merge isolated buses to the closest non-disconnected bus

To enhance the computational performances, it would be nice to improve the modelling of disconnected buses to enable connecting the isolated buses to the closest non-disconnected bus (after the already availabile drop and merge phases).
This issue implies revising the simplify_network similarly to how drop and merge isolated buses are performed.
See:
- drop isolated buses: https://github.com/pypsa-meets-earth/pypsa-earth/blob/354decd5dae8b9618f712d9b4b80edb751a5df5a/scripts/simplify_network.py#L662
- merge isolated buses: https://github.com/pypsa-meets-earth/pypsa-earth/blob/354decd5dae8b9618f712d9b4b80edb751a5df5a/scripts/simplify_network.py#L737

ekatef:
# Closes #852 

Added `merge_into_network( )` function to merge each isolated bus into the network following an approach of `drop_isolated_nodes( )` and `merge_isolated_nodes( )`.

## Checklist


ekatef:
Currently, the code works if a number of isolated nodes is more than one. It would be nice to generalise on a case with a single isolated nodes, preferably without need to using conditional statements.

Functionality has not been tested yet on geometries.

What would be great is to better understand requirements towards the implementation:

- do we need a voltage threshold for `merge_into_network( )`?
- how should be this option combines with drop of the isolated nodes and merging into a single node?

ekatef:
Currently, fetching isolated nodes into the closest network in implemented. Tested on a CL and seems to work fine:

## base.nc

image

## elec_s.nc, current implementation

image

## elec_s.nc, PR implementation

image

Red highlights buses which don't belong to the largest sub-network.

Fetching the isolated nodes keeps invariant values of the overall load and generation.

ekatef:
What would be nice to understand, is which strategies are sensible for drop/merge/fetch isolated nodes and isolated networks and adjust configuration parameters accordingly. My feeling is that it feels like alternatives to merge isolated nodes in a single isolated node or to fetch them into network.

Another point is handling isolated fragments of the network (like those red buses on first picture above). I suspect, in this particular case they mostly get reduced to a single node and dropped or merged after. However, the may be cases when isolated networks remain as two-three "micro-networks" after simplification, and can't be merged into the "main" network with a current approach. Probably, we can find some solution for this case, as well.

@davide-f would be very grateful if you could find time for looking into it 🙂

Edudiro:
Hi @ekatef, thank you for your response on [google groups](https://groups.google.com/g/pypsa/c/Wd3aRhmxI20) and for developing this PR. I have tested your approach and it works perfectly in the Chilean case, as you have proven above.

I am assuming the large amount of unconnected nodes from the base network comes from OSM data gaps? In that case I think following this solution is a valid approach. Otherwise, unfeasibilities might arise for a dispatch only problem.

ekatef:
> Hi @ekatef, thank you for your response on [google groups](https://groups.google.com/g/pypsa/c/Wd3aRhmxI20) and for developing this PR. I have tested your approach and it works perfectly in the Chilean case, as you have proven above.
> 
> I am assuming the large amount of unconnected nodes from the base network comes from OSM data gaps? In that case I think following this solution is a valid approach. Otherwise, unfeasibilities might arise for a dispatch only problem.

Hello @Edudiro! Great to hear that it worked for you. Thanks for giving an inspiration for pushing these developments :D

I also have a feeling that the unconnected nodes in `base.nc` result from some data gaps in OSM data base. At least, looking on [GENI description](http://www.geni.org/globalenergy/library/national_energy_grid/chile/) of a national power system, it seems that there shouldn't be any reason for having these isolated power systems. However, some regional expertise may be very helpful to give a more definitive answer.

ekatef:
@davide-f I think, the PR is ready for the first review round :)

Currently, isolated nodes are merged to the closest non-isolated ones. Further improvements are obviously possible, for example to account for the subnetwork type or implement a power threshold. 

Haven't added yet configuration parameters for fetching isolated nodes, as it feels a good idea to decide on a general interfacing concept. Would be very grateful for any input on that :)

scripts/simplify_network.py
```diff
@@ -738,6 +739,114 @@ def drop_isolated_nodes(n, threshold):
...
+
+    btree = cKDTree(points_buses)
+    dist0, idx0 = btree.query(islands_points, k=1)
+
+    points_nearest = [points_buses[i] for i in idx0]
+
+    nearest_bus_list = [
+        n.buses.loc[(n.buses.x == x) & (n.buses.y == y)] for x, y in points_nearest
+    ]
+    nearest_bus_df = pd.concat(nearest_bus_list)
```
davide-f: Probably a more efficient way to do this is with the geopandas [spatial join function](https://geopandas.org/en/stable/docs/reference/api/geopandas.sjoin_nearest.html#geopandas-sjoin-nearest):
`gdf_map = gpd.sjoin_nearest(gdf1, gdf2, houw="left")`
where gdf1 is the selection of isolated to merge and the right gdf2 frame is the set of non-isolated buses like `df_islands.query("carrier == 'DC'")[["geometry"]]`.
Internally, it exploits the KDTree and it saves lines of code.

The output is a geodataframe where for every row of gdf1 (isolated bus) we identify the index of gdf2 that is the closest, using the column "index_right" (default).
Therefore, the value `gdf_map["index_right"]` is almost the busmap we are looking for (once filled with the missing indices that shall not change and the values for the DC nodes)
Since we cannot exclude duplicated indices in `gdf_map` in case of equidistances, duplicated indices can be dropped.

Note that it is better to do the processing also for DC nodes: we merge isolated AC nodes with the closest non-isolated AC nodes and similarly the DC nodes.
An auxiliary function may be defined to repeat the process on AC and DC similarly, but it may be an overkill

davide-f (commented):
Cool PR :D I've added some possible improvements.
What do you think?

scripts/simplify_network.py
ekatef: @davide-f thanks a lot for the review!

Actually, I have completely forgotten about a possible polarity difference... I'll think about implementation.

A great idea with `sjoin_nearest()`. I had a (wrong) impression that it may be not perfect in terms of performance, but it looks like rather the opposite should be the case. Thanks, happy to play with that! :D

davide-f:
As an additional comment, for a future PR, I think that we could revise in the future the behaviour of the isolated_nodes and extend to networks: drop/merge isolated networks with demand below a given threshold

scripts/simplify_network.py
ekatef: @davide-f thanks a lot for the suggestions and sorry for putting this work on hold for a while. Great to come back to it :)

Implementation with `sjoin_nearest()` require a couple of additional imports, but it looks cleaner to me: https://github.com/ekatef/pypsa-earth/compare/fetch_isolated_to_network..fetch_isolated_to_network2
So, I'd opt for it instead of current usage of `cKDTree`. What do you think?

Have added filtering by `AC` carrier for the network to be merged with, agree that it's generally a good idea to find a way to deal also with `DC` networks. However, how realistic is a case of isolated `DC` nodes/network? Are you aware of any examples where it could be useful?

davide-f: Cool! and that's great :D I much like it :)

I'm wondering that an efficient way to account for the country-specific bus and potentially the AC/DC problem may be to exploit the groupby to do the merge, as follows:
1. create a geodataframe for n.buses (already done basically)
2. add a column in that dataframe to denote the "isolated networks", such as  gdf["is_isolated"]
    If we want to do the AC/DC difference, the determine_network_topology is only able to well adopt the AC ones. To account for the DC ones, an option is to drop out of the DC isolated networks identified with the determine_topology, those where a link is connected. I drop a piece of code I did to investigate the issue in a network.
3. exploit the groupby function, such as:
    `gdf.groupby(["country"{, "carrier"}]).map(lambda x: gdp.sjoin_nearest(x.query("is_isolated"), x.query("not is_isolated"), {options as needed}))`

As a note, the output of the s_join_nearest will create a column, like "index_right" or "index_left" that for every index of the geodataframe, specifies the closest bus.
That column is basically the `map_isolated_node_by_country` series that we need for the clustering process.

Would you be curious in investigating that?

I agree that isolated dc nodes would not make sense from a physical perspective, but I feel that it would make the workflow more robust as I believe we could have those type of nodes due to missing data and alike.

I've also done a quick check on a continent with the code below and it seems that only 4 DC nodes worldwide are actually isolated with the run I did.

---- Piece of code to identify the isolated buses to be adapted to the PR ---
```
is_isolated_network = n.buses["sub_network"].value_counts().map(lambda x: x <= 1)

DC_links = n.links.query("carrier == 'DC'")
links_buses = np.unique(list(DC_links["bus0"].values) + list(DC_links["bus1"].values))


is_isolated_bus = n.buses["sub_network"].map(is_isolated_network)
# all buses having
is_isolated_bus.loc[links_buses] = False
is_isolated_bus
```

ekatef: @davide-f super! :D Many thanks :D Have updated the implementation.

Ahh, right! We have also to account allocation buses by countries. Of course, I'm curious to look into it. Thanks a lot for the hints!!

May you have any ideas on isolated networks which spread across two-three countries? It would be very helpful for testing.

Absolutely agree regarding the need to account for isolated DC network for the sake of the workflow stability. Not sure is it's a well-grounded suggestion, but wouldn't it be a more natural approach for isolated DC buses to build an additional link to the closest non-isolated DC bus? My impression is that DC lines are often built to connect demand with renewable sources, and it may be probably a good idea to try to maintain topology. Does it sounds any reasonable for you?

ekatef: Hello @davide-f! Thanks a lot for the implementation idea :D It has been funny to play around, and it looks like the approach should be transferable to subnetwork-to-subnetwork merging.

Have drafted a possible implementation here:
https://github.com/ekatef/pypsa-earth/compare/fetch_isolated_to_network..fetch_isolated_to_network3

Have I get your idea properly? Currently, there is looping over groups, while I do agree that it may be a way to apply some functional-programming magic over the grouped dataframe. Not sure however if it makes sense to look for it. In my experience, when `map-lambda` is perfect when you manage to keep it simple. It things get complicated, that may look great but usually it difficult to maintain. What is your feeling about this? 🙂

davide-f: Thanks for the revision :)
I'll go into it soon, expected tomorrow

ekatef: Hey @davide-f, when you would have time and resources, would be perfect to have your opinion on this. 

It may be a bit too early for a proper review, but would be amazing to align on the concept 🙂

scripts/simplify_network.py
```diff
@@ -738,6 +741,118 @@ def drop_isolated_nodes(n, threshold):
...
+    )
+    islands_points = np.array(
+        list(zip(n.buses.loc[i_islands].x, n.buses.loc[i_islands].y))
+    )
+
+    gds_buses = gpd.GeoSeries(map(Point, points_buses))
+    gds_islands = gpd.GeoSeries(map(Point, islands_points))
+
+    gdf_buses = gpd.GeoDataFrame(geometry=gds_buses)
+    gdf_islands = gpd.GeoDataFrame(geometry=gds_islands)
```
davide-f: to create a geodatafarme (let's say `gdf_buses`)of the entire buses df we may use this function, that should be quite compact

scripts/simplify_network.py
```diff
@@ -738,6 +741,118 @@ def drop_isolated_nodes(n, threshold):
...
+    generators_mean_origin = n.generators.p_nom.mean()
+    load_mean_origin = n.loads_t.p_set.mean().mean()
+
+    n.determine_network_topology()
+
+    # duplicated sub-networks mean that there is at least one interconnection between buses
+    i_islands = n.buses[
+        (~n.buses.duplicated(subset=["sub_network"], keep=False))
+        & (n.buses.carrier == "AC")
+    ].index
```
davide-f: Does this account identifies isolated networks? This seems to me that nodes that are interconnected by a line but disconnected from the entire network are not well identified or am I wrong?

We could extend the working principle to "isolated networks" that may be defined as subnetworks that do not represent the larger national power system whose demand is low (below threshold).

To keep the approach simple we can also skip that

scripts/simplify_network.py
```diff
@@ -738,6 +741,118 @@ def drop_isolated_nodes(n, threshold):
...
+    gdf_buses = gpd.GeoDataFrame(geometry=gds_buses)
+    gdf_islands = gpd.GeoDataFrame(geometry=gds_islands)
+
+    gdf_map = gpd.sjoin_nearest(gdf_islands, gdf_buses, how="left")
+
+    nearest_bus_list = [
+        n.buses.loc[(n.buses.x == x) & (n.buses.y == y)]
+        for x, y in zip(gdf_map["geometry"].x, gdf_map["geometry"].y)
+    ]
+    nearest_bus_df = pd.concat(nearest_bus_list)
```
davide-f: I believe that `nearest_bus_df = n.buses.loc[gdf_map.index_right.values]` or something like that.

gdf_map.index_right/index_left may contain the mapping between each island and its closest bus.

scripts/simplify_network.py
```diff
@@ -738,6 +741,118 @@ def drop_isolated_nodes(n, threshold):
...
+
+    # each isolated node should be mapped into the closes non-isolated node
+    map_isolated_node_by_country = (
+        n.buses.assign(bus_id=n.buses.index)
+        .loc[nearest_bus_df.index]
+        .groupby("country")["bus_id"]
+        .first()
+        .to_dict()
+    )
```
davide-f: Something like:

`gdf_buses.groupby(["country"{, "carrier"}]).map(lambda x: gdp.sjoin_nearest(x.query("is_isolated"), x.query("not is_isolated"), {options as needed}))`

or 

`gdf_buses.groupby(["country"{, "carrier"}]).map(lambda x: gdp.sjoin_nearest(x.loc[i_islands], x.loc[i_connected], {options as needed}))`

may work.

In the case of neighboring buses across countries I'm unsure the isolated network/node gets connected to the closest bus of the same country, it may remain there. Have you tested that?

davide-f (commented):
Sorry for the little delay, here is my feedback :D
Great!

ekatef:
> Sorry for the little delay, here is my feedback :D Great!

Hey @davide-f, no problem, I was aware that you were very busy. Thanks a lot for the review! 

It looks like we are well aligned on the concept, and happy to introduce the changes.

Yeah, the idea of this PR as far was to focus on isolated nodes only, keeping in mind that same approach may be used to deal with isolated networks, as well. 

@carlosfv92 May you be interested to work on implementation for isolated networks? Happy to discuss on align on the implementation.

scripts/simplify_network.py
ekatef: Agree that it could be implemented in a more elegant way :) more  Wrapped-up into a function transformation into a geopandas dataframe.

Have I understood your idea correctly? :)

scripts/simplify_network.py
ekatef: Agree, it feels much more natural to use indices for filtering. Added an `bus_id` column to the buses dataframe, and modified filtering.

scripts/simplify_network.py
ekatef: Added a function to pick isolated networks using a number of the nodes as a criterium in the first approximation. Absolutely agree that it could make sense to use other heuristics, the current implementation just generalises the merging approach for n_isolated_buses > 1.

ekatef:
Testing update on a multy-country case (Congo Democratic Republic), Burundi, Rwanda

### initial case

image

### results of simplification

image

davide-f:
Wow! that seems to work very well :D maybe a little too much tough... XD
the actual Rwandan and Borundi networks are actually collabsed. Multi-country subneworks could be excluded from the merge probably.
What do you think?

ekatef:
> Wow! that seems to work very well :D maybe a little too much tough... XD the actual Rwandan and Borundi networks are actually collabsed. Multi-country subneworks could be excluded from the merge probably. What do you think?

Absolutely agree with your assessment :D Basically, that has been testing of a limiting case which went as kind of _reductio ad absurdum_ :)))

Looking into the regional details for this particular case, I'd say we need to answer the following questions for multi-country region applications:

1. What to do about small sub-networks which span across different countries? Agree that the best strategy seems to be just leaving such sub-networks alone.
2. Now to deal with the different scale of power systems? I mean, if there is a bid difference in the overall demand of two power systems (e.g. South Africa and Eswatini), it may be not quite handy to have the same power threshold. So, I think it may be a good idea to play with a share of the overall national demand to be used as a threshold. Let say, fetch everything with less than 5% demand of the national energy demand into the national backbone grid. Does it sound reasonable for you?

ekatef:
Results of an updated version:

image

The resulted topology looks a bit odd with a new line added which spans across the whole country, parallel to the DC link. 

Highlighting with green the buses of `base.nc` into which isolated nodes should be merged:

image

Highlighting after isolated network has been fetched:

image

So, the result may be correct in terms on buses ids, but would be great to avoid such distortions of geometry, as it can be potentially very misleading.

ekatef:
Results of topology investigation: all the green buses should be merged into one with a circle around:

image

That is, the the busmap looks like follows for the  backbone bus`"44"`:

```
busmap.loc[busmap == "44"]
Bus
21     44
22     44
24     44
26     44
27     44
       ..
260    44
261    44
265    44
266    44
```

So, the transformation intention looks correct but after application of [get_clustering_from_busmap](https://github.com/PyPSA/PyPSA/blob/feaed4ecb6a20d43b31c478bb2e6ebf31325627c/pypsa/clustering/spatial.py#L515) the topology gets distorted for some reason.

Update: the reason of such geometry transformation is that `mean` is being used as an [aggregation strategy for buses](https://github.com/PyPSA/PyPSA/blob/feaed4ecb6a20d43b31c478bb2e6ebf31325627c/pypsa/clustering/spatial.py#L60-L67):

```python
DEFAULT_BUS_STRATEGIES = dict(
    x="mean",
    y="mean",
    v_nom="max",
    v_mag_pu_max="min",
    v_mag_pu_min="max",
    generator=lambda x: "",
)
```

The implemention in the latest PyPSA version differs from the version we are using currently. But the idea remains the same: there is an implicit assumption that buses mapped into an aggregated node are close enough to this node. In case of weak connectivity it doesn't hold and leads to quite specific topologies.

ekatef:
As a second thought, it may be probably not a good idea to allow fetching the nodes into a backbone network if this backbone is too far. Generally the testing looks successful. Have also replaced a relative threshold with an absolute one with the following result (with 10 MW of mean power):

image

The networks of smaller countries may look a bit cleaner with a relative threshold, but I'd say that keeping a consistent interface feels more important.

ekatef:
My feeling is that it's ready to review.
@davide-f I'd be very grateful if you could please look into that 🙂

scripts/simplify_network.py
```diff
@@ -743,6 +746,182 @@ def drop_isolated_nodes(n, threshold):
...
+
+    i_islands = find_isolated_sub_networks(n=n, threshold=threshold)
+
+    # return the original network if no isolated nodes are detected
+    if len(i_islands) == 0:
+        return n, n.buses.index.to_series()
+
+    gdf_islands = transform_to_gdf(
+        buses_df=n.buses, i_buses=i_islands, network_crs=network_crs
+    )
```
davide-f: Maybe we could create the geodataframe for n.buses and then subset the geodataframe in the following.
By doing so, probably there is no need to add the transform_to_gdf function, what do you think?

I mean, somenthing like:
geo_buses = gpd.GeoDataFrame(...)
gdf_islands = gdf_buses.loc[...]
gdf_backbone_buses = gdf_buses.loc[...]

What do you think?

scripts/simplify_network.py
```diff
@@ -743,6 +746,182 @@ def drop_isolated_nodes(n, threshold):
     return n
 
 
+def find_isolated_sub_networks(n, threshold):
```
davide-f: I'm feeling like an "easy" appraoch to account for the threshold may be to merge the geodataframe that we create and the load information and then do groupsby and something like that on that to discard the buses we don't like.

let's say gdf_buses is the geodataframe we create in the main function.

we may do: gdf_buses.merge(n.loads_t.p_set.mean().to_frame(), how="left")
merge or join may be used.

Then for every bus we have the expected demand and by doing groupsby by subnetwork we can get the demand by subnetwork and subset only those buses whose demand is below threshold.

The current formulation seems to be equivalent, we can revise it later if you prefer :)

scripts/simplify_network.py
```diff
@@ -743,6 +746,182 @@ def drop_isolated_nodes(n, threshold):
...
+    )
+    gdf_backbone_buses = transform_to_gdf(
+        buses_df=n.buses, i_buses=i_backbone, network_crs=network_crs
+    )
+
+    islands_bcountry = {k: d for k, d in gdf_islands.groupby("country")}
+    gdf_map = gdf_backbone_buses.groupby("country").apply(
+        lambda d: gpd.sjoin_nearest(islands_bcountry[d["country"].values[0]], d)
+    )
+    nearest_bus_df = n.buses.loc[n.buses.bus_id.isin(gdf_map.bus_id_right)]
```
davide-f: Why is n.buses.bus_id.isin needed? the output should be always in the index of n.buses no?

scripts/simplify_network.py
ekatef: The main purpose for `transform_to_gdf( )` is to keep pandas-to-geopandas transformations away from the "main" code. 

Initially, I have [tried without it](https://github.com/pypsa-meets-earth/pypsa-earth/blob/dc9be87883c95155db589aafc2a294d13b83ff23/scripts/simplify_network.py#L787-L798), but geopandas stuff looked a bit distractive `gpd.points_from_xy(points_buses.x, points_buses.y)`. So, I have wrapped-up this part into a function to focus on topology-related part in the higher-level `merge_into_network( )` function.

Using `i_buses` as an argument, has been just the next step: since we have an additional function anyway, why not to transfer also subsetting to it? :D

Does it sound reasonable for you? :)

scripts/simplify_network.py
ekatef: Agree, but the main point of `find_isolated_sub_networks()` is to deal with multi-country sub-networks, while the "main" function deals with finding the backbone networks for each of the countries. Not sure we make the code more clear if moving into the main code also dealing with isolated subnetworks.

Probably, it would be great to have some more understanding on use-cases for this functionality and adjust the code accordingly. Currently, the code has been tested for `["RW", "BI", "CD"]` which is rather tricky which also could have contribute into some over-complication 🙃 

Have added some comments to `find_isolated_sub_networks()` to facilitate further revisions.

scripts/simplify_network.py
ekatef: That is definitely a duplication. (Sorry, R-background 😄) Fixed here and across the code to keep `bus_id` column only where is seems safer than index.

ekatef:
Hello @davide-f!

Thanks a lot for the review and apologies for taking a while to consider it. I do agree that it's worth to keep things simple as far as the target functionality is ensured. However, I have an impression that handling subnetworks is more tricky as compared with the single nodes, especially when considering a multi-countries case...

Have tried to incorporate your comments where I have seen the way to do it, and tried to explained the reasoning for the points where is hasn't been the case. Let me know please if I get your ideas right. Obviously, happy to discuss further 🙂

ekatef:
Revised to simplify the implementation. @davide-f thank you so much for the discussion and the ideas! The current version looks definitely cleaner. Could you please review it, when you have time?

Summary of the major changes as compared with the previous iteration:
1) moved all load processing into the function which creates a buses geo-dataframe;
2) revised the threshold from to be a share of the national load instead an absolute value;
3) replaced all the functions to filter the island and backbone networks with the flags in the buses geo-dataframe.

ekatef:
@davide-f when you have time, it would be perfect to have your review 🙂

scripts/simplify_network.py
```diff
@@ -743,6 +746,141 @@ def drop_isolated_nodes(n, threshold):
...
+    n.determine_network_topology()
+
+    n_buses_gdf = transform_to_gdf(n, network_crs=network_crs)
+
+    # do not merge sub-networks spanned through a number of countries
+    n_buses_gdf["is_multicnt_subntw"] = n_buses_gdf.sub_network.map(
+        n_buses_gdf.groupby(["sub_network"])["country"].apply(
+            lambda x: len(pd.unique(x)) > 1
+        )
+    )
```
davide-f: Not sure if you need the double map; probably the following should work or something like that
```suggestion
    n_buses_gdf["is_multicnt_subntw"] = n_buses_gdf.sub_network.map(
         n_buses_gdf.groupby(["sub_network"]).country.unique().count() > 1
     )
```
May that work?

However, I'm wondering if we need to fill that value at the gdf and we can work with a df for subnetworks, like:
```
df_subn = n_buses_gdf.groupby(["subnetwork"]).apply(lambda x:
   {
       "is_multicnt_subntw": x.country.unique() > 1,
       "total_demand": x.load.sum(),
   }
)
```

However, I see that in the buses gdf you filled a column with the demand of the subnetwork, so that may not be needed actually

scripts/simplify_network.py
```diff
@@ -743,6 +746,141 @@ def drop_isolated_nodes(n, threshold):
...
+        .query("sbntw_share_of_country_load < @threshold")
+    )
+    # return the original network if no isolated nodes are detected
+    if len(gdf_islands) == 0:
+        return n, n.buses.index.to_series()
+
+    gdf_backbone_buses = (
+        n_buses_gdf.query("is_backbone_sbntw")
+        .query("carrier=='AC'")
+        .query("load_in_subnetw>0")
```
davide-f: Mmmm I'd drop this check but I'd pick the index with the highest load; for small countries with no demand this may trigger issues

scripts/simplify_network.py
```diff
@@ -743,6 +746,141 @@ def drop_isolated_nodes(n, threshold):
...
+    islands_bcountry = {k: d for k, d in gdf_islands.groupby("country")}
+    gdf_map = (
+        gdf_backbone_buses.query("country in @islands_bcountry")
+        .groupby("country")
+        .apply(lambda d: gpd.sjoin_nearest(islands_bcountry[d["country"].values[0]], d))
+    )
+    nearest_bus_df = n.buses.loc[n.buses.index.isin(gdf_map.bus_id_right)]
+
+    i_lines_islands = n.lines.loc[n.lines.bus1.isin(gdf_islands.index)].index
+    n.mremove("Line", i_lines_islands)
```
davide-f, line 834: This remove should not be needed and performed by the clustering no?

scripts/simplify_network.py
```diff
@@ -109,6 +110,8 @@
 )
 from pypsa.io import import_components_from_dataframe, import_series_from_dataframe
 from scipy.sparse.csgraph import connected_components, dijkstra
+from scipy.spatial import cKDTree
+from shapely.geometry import Point
```
davide-f: Are they still needed?

davide-f (commented):
Great @ekatef :D
I think the functionality is quite there and way cleaner!
In the future we may reformulate but I think it is in great shape! There may not be the need to duplicate the information by subnetwork to each row of the buses, but I think it is good to go with minor changes! :)
Great work and patience!

scripts/simplify_network.py
ekatef: Agree on the revision on `map().apply()`, thanks! :D

Not sure I get your idea with the sub-networks dataframe. I thought, we aim to keep all the data in a single dataframe in the spirit of [the tidy data](https://r4ds.had.co.nz/tidy-data.html) concept, which actually encourages a certain degree of duplication. It gives a lot of power and flexibility, while allowing to stay in the same "variables space". If we add an additional dataframe, it means a need to switch attention between two objects, and each of them has an own set of variables. To me, it sounds a bit more complicated.

I'm also not sure that having two objects would be more effective in terms of performance, as compared with having an additional column. Have I missed anything? :)

scripts/simplify_network.py
ekatef: Agree, revised :)

scripts/simplify_network.py
ekatef: Normally yes, but our use case is not well captured by `get_clustering_from_busmap( )`, as it seems to expect that the lines in a cluster belong to the same sub-network. When I try to remove the line `n.mremove("Line", i_lines_islands)` leads to `AssertionError In Bus cluster sub_network the values of attribute sub_network do not agree`.

scripts/simplify_network.py
ekatef: Fixed, thanks!

ekatef:
> Great @ekatef :D I think the functionality is quite there and way cleaner! In the future we may reformulate but I think it is in great shape! There may not be the need to duplicate the information by subnetwork to each row of the buses, but I think it is good to go with minor changes! :) Great work and patience!

Thanks a lot for the great support and the review @davide-f!

I'd say this duplication is rather a feature than a bug ;) There is the "tidy data" concept which assumes that each row should contain a full information for a single variable. That makes life way easier even if it implies some duplication. That is, if I have get your idea properly. Happy to discuss more and adjust the implementation, if needed :)

scripts/simplify_network.py
davide-f: ok :)

ekatef:
@davide-f could you please check if you have any questions on the PR, when you would have time?

It seems to be needed, and I think it's simpler to finalise it instead to explain how to test it from a fork.

davide-f (approved):
Great @ekatef <3 amazing contribution and sorry for the delay...
I approve this PR and feel free to merge it with squash merge or normal merge as you see it more fair/clean :)

ekatef:
> Great @ekatef <3 amazing contribution and sorry for the delay... I approve this PR and feel free to merge it with squash merge or normal merge as you see it more fair/clean :)

No worries and thanks a lot for the outstanding support :D

PR #911, opened 2023-10-28
Retrieve databundle light data size check

Emre-Yorat89:
# Closes # (if applicable).

## Checklist

### DIFF
--- a/scripts/retrieve_databundle_light.py
+++ b/scripts/retrieve_databundle_light.py
@@ -83,10 +83,10 @@ according to the following rules:
 import logging
 import os
 import re
-import pandas as pd
-import requests
 from zipfile import ZipFile
 
+import pandas as pd
+import requests
 import yaml
 from _helpers import (
     configure_logging,
@@ -486,21 +486,28 @@ def get_best_bundles_by_category(
     df_empty = pd.DataFrame(columns=["bundle_name", "bundle_size", "n_matches"])
     i = 0
     for bname in config_bundles:
-        if config_bundles[bname]["category"] == category and config_bundles[bname].get("tutorial", False) == tutorial and _check_disabled_by_opt(config_bundles[bname], config_enable) != ["all"]:
-            bundles = pd.DataFrame({"bundle_name": bname, "n_matches": config_bundles[bname]["n_matched"]}, index=[i])
+        if (
+            config_bundles[bname]["category"] == category
+            and config_bundles[bname].get("tutorial", False) == tutorial
+            and _check_disabled_by_opt(config_bundles[bname], config_enable) != ["all"]
+        ):
+            bundles = pd.DataFrame(
+                {"bundle_name": bname, "n_matches": config_bundles[bname]["n_matched"]},
+                index=[i],
+            )
             i = i + 1
             df_empty = pd.concat([df_empty, bundles], axis=0)
 
     returned_bundles = []
     if not df_empty.empty:
-        df_empty = df_empty.sort_values(by = "n_matches")
+        df_empty = df_empty.sort_values(by="n_matches")
         for i in df_empty.index:
             url = list(config_bundles[df_empty.loc[i]["bundle_name"]]["urls"].values())
             response = requests.head(url[0])
-            bundle_size = response.headers['Content-Length']
+            bundle_size = response.headers["Content-Length"]
             df_empty.loc[i, "bundle_size"] = bundle_size
         df_empty = df_empty[df_empty["n_matches"] == len(country_list)]
-        df_empty = df_empty.sort_values(by = ["bundle_size"], ascending=True).head(1)
+        df_empty = df_empty.sort_values(by=["bundle_size"], ascending=True).head(1)
         returned_bundles = list(df_empty["bundle_name"])
 
     return returned_bundles
--- a/scripts/retrieve_databundle_light.py
+++ b/scripts/retrieve_databundle_light.py
@@ -495,8 +495,8 @@ def get_best_bundles_by_category(
     if not df_empty.empty:
         df_empty = df_empty.sort_values(by = "n_matches")
         for i in df_empty.index:
-            urlsx = list(config_bundles[df_empty.loc[i]["bundle_name"]]["urls"].values())
-            response = requests.head(urlsx[0])
+            url = list(config_bundles[df_empty.loc[i]["bundle_name"]]["urls"].values())
+            response = requests.head(url[0])
             bundle_size = response.headers['Content-Length']
             df_empty.loc[i, "bundle_size"] = bundle_size
         df_empty = df_empty[df_empty["n_matches"] == len(country_list)]
--- a/scripts/retrieve_databundle_light.py
+++ b/scripts/retrieve_databundle_light.py
@@ -83,6 +83,8 @@ according to the following rules:
 import logging
 import os
 import re
+import pandas as pd
+import requests
 from zipfile import ZipFile
 
 import yaml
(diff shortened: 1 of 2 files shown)


Emre-Yorat89:
Hi,

I have modified the databundle size check PR by following the recommendations from the deleted PR @davide-f. One of the problem I see is related to the zenodo links. I could not download them manually and also the response from these links is quite smaller than the gdrive files in the retrieve_databundle script. I hope this PR will reflect the necessary improvement after the review process.

scripts/retrieve_databundle_light.py
```diff
@@ -481,39 +483,32 @@ def get_best_bundles_by_category(
...
+            config_bundles[bname]["category"] == category
+            and config_bundles[bname].get("tutorial", False) == tutorial
+            and _check_disabled_by_opt(config_bundles[bname], config_enable) != ["all"]
+        ):
+            bundles = pd.DataFrame(
+                {"bundle_name": bname, "n_matches": config_bundles[bname]["n_matched"]},
+                index=[i],
+            )
+            i = i + 1
+            df_empty = pd.concat([df_empty, bundles], axis=0)
```
davide-f: Great Draft @Emre-Yorat89 ! :D
Could you try if this code works? It should a slight improvement of yours

```suggestion
    df_matches = pd.DataFrame(columns=["bundle_name", "bundle_size", "n_matched"])
    for bname, bvalue in config_bundles.items():
        if (
            bvalue["category"] == category
            and bvalue.get("tutorial", False) == tutorial
            and _check_disabled_by_opt(bvalue, config_enable) != ["all"]
        ):
            df_matches.loc[bname] = [bname, len(bvalue["countries"]), bvalue["n_matches"]]
```

scripts/retrieve_databundle_light.py
```diff
@@ -481,39 +483,32 @@ def get_best_bundles_by_category(
...
+    if not df_empty.empty:
+        df_empty = df_empty.sort_values(by="n_matches")
+        for i in df_empty.index:
+            url = list(config_bundles[df_empty.loc[i]["bundle_name"]]["urls"].values())
+            response = requests.head(url[0])
+            bundle_size = response.headers["Content-Length"]
+            df_empty.loc[i, "bundle_size"] = bundle_size
+        df_empty = df_empty[df_empty["n_matches"] == len(country_list)]
+        df_empty = df_empty.sort_values(by=["bundle_size"], ascending=True).head(1)
+        returned_bundles = list(df_empty["bundle_name"])
```
davide-f: This may be slightly revised with the revision above.
I don't get the need for the requests.
The goal would be first, to sort the values as you are doing right but there may be the need to revise the sort by:
```
df_matches["neg_bundle_size"] = -df_matches["bundle_size"]
df_matches.sort_values(by=["n_matches", "neg_bundle_size"], inplace=True, ascending=False)
```

The idea is to have in the first rows the bundles that match most countries but are the smallest.
Using ascending False and the negative value of bundle size should do the trick, may you verify?

I dont get much the logic in the for loop; the logic should be:
- initialize the remaining_countries value to the country list
- then, loop over each row of the sorted dataframe and:
   - compare the intersection between the remaining_countries value and the set `config_bundles[bname]["countries"]` where bname is the value of the current iteration over the index of df_matches.
   - if the intersection is empty, continue the iterations
   - otherwise, add the current bundle to the list of bundles to download and update remaining_countries with the difference between remaining_countries and the intersection.
      The core content of this for loop is already existing in the original code

I understand that not having the CI links working is a challenge...
If you debug locally and check the values in debug mode of the functions the debugging may work anyway though.
To test this function, there is no need to actually download the bundles but verify that the expected data bundles are selected.
Not sure if it is clear, if it is not, please let me know

scripts/retrieve_databundle_light.py
```diff
@@ -85,6 +85,8 @@
 import re
 from zipfile import ZipFile
 
+import pandas as pd
+import requests
```
davide-f: Requests is not neede in my opinion: the check on whether the bundles are available is done later on.

davide-f (commented):
Cool draft @Emre-Yorat89  :)
There are improvements, fantastic!!!! Added some comments.
Apologies that the CI down is not helping. The issue should be temporary though...
Let me know if comments are clear

scripts/retrieve_databundle_light.py
Emre-Yorat89: I have used requests library for obtaining the size of data file from the uploaded web server and the final goal was to download the first bundle that occupies less memory space in the computer. We can also not use requests library by adding additinal line (key:value) specifying bundle sizes in bundle_config.yaml file that could also be a convenient solution. What do you think?

davide-f: Maybe we could use the size in terms of numbers of countries contained in that databundle.
As a first approximation, we could estimate that, within the subset of bundles that contain all selected countries, the  databundle with the smallest number of countries may be also the smallest.
Note that each "countries" field of the databundles (e.g. "Africa") gets first parsed into the full list of countries, for africa being around 50 countries.
So using the number of countries as a proxy may work, though not being perfect.
The risk with requests is that we need to make many more requests to servers and, although it is an advanced clear approach (cool idea and implementation :DDD), in this case we could use a simpler approach.
What do you think?

Emre-Yorat89: I agree that this way it will be very close approximation :). I remodify the function to return the bundles that contain minimum number of countries but hold the analyzed ones :)

davide-f: Super :D

scripts/retrieve_databundle_light.py
Emre-Yorat89: This is a great solution that you sort by n_matches and negative bundle sizes. Do you think it is necessary to loop over country list and check intersection. The filter **df_empty[df_empty["n_matches"] == len(country_list)** is checking whether the bundle contains all of the countries or not.  I can do it with the loop as you wish too.

scripts/retrieve_databundle_light.py
Emre-Yorat89: I have tried. They are working, thank you for the contribution :)

scripts/retrieve_databundle_light.py
```diff
@@ -482,28 +483,33 @@ def get_best_bundles_by_category(
...
+    returned_bundles = []
 
+    if bname_list:
         current_matched_countries = []
         remaining_countries = set(country_list)
 
-        for d_val in dict_sort:
-            bname = d_val[0]
-
+        for bname in bname_list:
```
davide-f: Great job @Emre-Yorat89 :D
The part above this line I believe is basically perfect!

The following may be slightly revised but I think we are close to done.

Proposed changes are:
- here we no more need to define b_name_list, you can delete row 505; accordingly, line 512 may be revised in checking whether the dataframe is not empty: not df_matches.empty
- then the loop over b_name in b_name_list can be revised into a loop over the rows of the dataframe: `for bname in df_matches.index:`

With these changes, the approach should work as is.
Could you try if it works?

davide-f (commented):
Fantastic @Emre-Yorat89 :D
We are close to finalizing I believe :D
Very little changes left.

Do you observe unexpected behaviors?

scripts/retrieve_databundle_light.py
Emre-Yorat89: Do you observe unexpected behaviors?
I have tried the function in a couple of different countries and it seems it works okay. One thing I have noticed when downloading data for european countries (IT, SP, DE) it downloads bundle_cutouts_northamerica data. It is probably correct since these countries are in the list of (config_bundles[bname]["countries"]) for northamerica. Do you think that should we separate these and create a new european countries bundle? This way people would not need to download large cutouts of other countries such as USA.

davide-f: Great @emre :D
Tomorrow I'll go through the review but I think it is ready to go!
Agree with your comment: we miss a european cutout and an Oceania cutout. Tommaso was interested in the European one, but not sure on its status and the Oceania is yet missing.
Btw the Pr is ready I think:D tomorrow will be merged most likely :)

davide-f (commented):
Hello @Emre-Yorat89 :D
I confirm that the code is fine :)
I think you can add a line into the release_note and this PR is ready :D Congratulations!

Emre-Yorat89:
> Hello @Emre-Yorat89 :D I confirm that the code is fine :) I think you can add a line into the release_note and this PR is ready :D Congratulations!

Thank you so much for your patience and guidance, without them it would be hard to finish this PR :)

PR #913, opened 2023-11-01
Enable automatic landcover download

davide-f:
# Closes # (if applicable).

This PR automatically supports the download of landcover data from original source.
Note: the landcover informations are only needed in the creation of the natura tiff file. As such, they are not used by the standard user.

This PR cannot be merged as the tutorial will fail: there is the need to create a natura.tiff file for each tutorial and add it to the databundles.
Moreover, a better disclaimer for the license is needed, although the link is provided in the license page

## Checklist


davide-f:
Ready to Review

ekatef:
> Ready to Review

Hello @davide-f! Congratulations with finalising this work 😄
Can I help you with review?

davide-f:
> > Ready to Review
> 
> Hello @davide-f! Congratulations with finalising this work 😄
> Can I help you with review?

That would be awesome! :D

scripts/build_natura_raster.py
```diff
@@ -151,12 +151,7 @@ def unify_protected_shape_areas(inputs, natura_crs, out_logging):
         shp = gpd.read_file(i).to_crs(natura_crs)
         total_files += 1
         try:
-            shp.geometry = shp.apply(
-                lambda row: make_valid(row.geometry)
-                if not row.geometry.is_valid
-                else row.geometry,
-                axis=1,
-            )
+            shp.geometry = shp.geometry.make_valid()
```
ekatef, line 147: A curiosity question: what is the reason of this change? Is it linked with a fact that we transferred from using a number of shape files by continents a single global one?

configs/bundle_config.yaml
```diff
@@ -396,3 +317,12 @@ databundles:
...
       build_cutout: [all]
+
+  # data bundle containing the protected data for the whole world
+  bundle_landcover_earth:
+    countries: [Earth]
+    category: landcover
+    destination: "data/landcover/world_protected_areas"
+    urls:
+      protectedplanet: https://d1gam3xoknrgr2.cloudfront.net/current/WDPA_{month:s}{year:d}_Public_shp.zip
+    output: [data/landcover/world_protected_areas/*]
```
ekatef, line 328: Would you kill me if I'd suggest a hard-coded url instead of the dynamically-changed one? It looks like UN is a link to the latest copy of the WDPA shape file under http://wcmc.io/wdpa_current_release

(found [here](https://github.com/prioritizr/wdpar/blob/ce28c41b82b30fe5604b6452bd9c74f88325bdd3/R/wdpa_url.R#L114-L117))

scripts/build_natura_raster.py
```diff
@@ -151,12 +151,7 @@ def unify_protected_shape_areas(inputs, natura_crs, out_logging):
         shp = gpd.read_file(i).to_crs(natura_crs)
         total_files += 1
         try:
-            shp.geometry = shp.apply(
```
ekatef, line 154: It looks like we may need to slightly revise the docstring. In particular [this line](https://github.com/davide-f/pypsa-earth/blob/6ecd9a903f8bbe678ef0b9157efb68d154d06158/scripts/build_natura_raster.py#L26) could be revised to something like:
`- ``data/raw/protected_areas/current/WDPA_Nov2023_Public_shp.zip``: `WDPA <https://en.wikipedia.org/wiki/Natura_2000>`_ World Database for Protected Areas.`

ekatef:
Hello @davide-f! Thanks for tackling this issue 😄 

I think, the code looks great. Have added a couple of rather minor comments. Along the way, an alternative approach has been found to build url address for WDPA files (I just have been trying to understand how do they form their cloud url with month-year...). That is a domain of UN environmental program `http://wcmc.io/wdpa_current_release` which re-directs to the same cloud entry. Could it make sense to consider it, probably as an alternative url? What do you think?

I'm also thinking that probably we may post somewhere comparison between the old and the new `natura.tiff` files and test which effect does the new `natura.tiff` on simulation results. If you think it's worth, happy to create an issue on that. Although, that is obviously a completely different story.

Great work with updating the datakit! 😄

scripts/build_natura_raster.py
davide-f: Because the make_valid() function has the same effect without the lambda function, that usually adds more overhead

configs/bundle_config.yaml
davide-f: I was not aware of that link, we must do that!
I'll revert the latest changes

scripts/build_natura_raster.py
davide-f: Agree, revising

configs/bundle_config.yaml
davide-f: Mmm I've investigated the database, however, the format of that download is different with respect to the usual zip file: the format is a gdb multi-layer data and does not contain shapefiles as for the standard approach.
Revising the approach to support gdb files is possible, but it seems an overkill in my opinion, what do you think?

ekatef: Sorry, I have been not aware of the differences in formats. Thanks for checking! 

I don't think that it makes sense to revise the approach for `gdb` files. We still have a number of input-validation issue which are more meaningful from my perspective (alternative hydrobasins, new natura.tiff...). Not sure it's a good idea to add points to this list :)

To me, the implemented approach looks perfectly fine. I have been just amazed by a simplicity of UN url 🙃 But if it doesn't going to work smoothly, I don't see any serious advantages there as compared to the solution you have still implemented.

scripts/build_natura_raster.py
ekatef: Ah, `make_valid` is vectorised! Good to know. Thanks a lot for the explanation :)

configs/bundle_config.yaml
davide-f: Super :) we are on the same page; I'm running locally a workflow to verify it works for the world. I'll soon push the (little) changes

davide-f: I've pushed the changes but the workflow is still running, tomorrow I'll check the status :)

davide-f: I confirm the successful execution of the workflow :) [it is stuck at building natura, but that's natural...]

scripts/build_natura_raster.py
davide-f: A pleasure :)

configs/bundle_config.yaml
ekatef: Super! :D

davide-f:
Merging

Issue #914, opened 2023-11-02
ZENODO ISSUES - Temporary bypass for local testing


davide-f:
## Checklist

- [x] I am using the current [`main`](https://github.com/pypsa-meets-earth/pypsa-earth/tree/main) branch or the latest [release](https://github.com/pypsa-meets-earth/pypsa-earth/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-earth` environment](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

While the automatic download is not working, each of us can manually download the bundles using their accounts on gdrive. By simply logging in with gdrive and opening the relevant gdrive links described [in this file](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/configs/bundle_config.yaml), it is possible to download the bundles and unzipping them in the appropriate locations manually, bypassing the current issues.

The operative methodology to download a selected databundle for a tutorial (e.g. NGBJ) is:
1. Download the link associated to bundle “bundle_tutorial_data_common” (this applies for any tutorial NGBJ/BW/MA)
2. Download the link associated to bundle “bundle_tutorial_NGBJ”
3. Download the link associated to the cutout bundle “bundle_cutouts_tutorial_NGBJ”
4. Modify the config.yaml file, specifying: the target countries and disable retrieve_databundle option (enable: retrieve_databundle: false)

davide-f:
closing this PR as no more relevant

PR #919, opened 2023-11-05
Add load of hydrobasins data

ekatef:
# Relates to #914 

Add functionality to load hydrobasins data directly from the data source.

## Checklist

### DIFF
--- a/configs/bundle_config.yaml
+++ b/configs/bundle_config.yaml
@@ -80,7 +80,7 @@ databundles:
         suffixes: ["af"]
     unzip: true
     output:
-    - data/hydrobasins/hybas_world.shp
+    - data/hydrobasins/hybas_world_v1c.shp
 
   # tutorial bundle specific for Nigeria and Benin only
   bundle_cutouts_tutorial_NGBJ:
--- a/configs/bundle_config.yaml
+++ b/configs/bundle_config.yaml
@@ -148,7 +148,8 @@ databundles:
     category: common
     destination: "data/hydrobasins"
     urls:
-      direct: https://data.hydrosheds.org/file/HydroBASINS/standard/hybas_af_lev06_v1c.zip
+      hydrobasins: https://data.hydrosheds.org/file/HydroBASINS/standard/
+      suffixes: ["af", "ar", "as", "au", "eu", "gr", "na", "sa", "si"]
     unzip: true
     output:
     - data/hydrobasins/hybas_world_lev06_v1c.shp
--- a/configs/bundle_config.yaml
+++ b/configs/bundle_config.yaml
@@ -152,7 +152,7 @@ databundles:
       suffixes: ["af", "ar", "as", "au", "eu", "gr", "na", "sa", "si"]
     unzip: true
     output:
-    - data/hydrobasins/hybas_world_lev06_v1c.shp
+    - data/hydrobasins/*.shp
 
   # data bundle containing the data of the data folder common to all regions of the world
   bundle_data_earth:
--- a/configs/bundle_config.yaml
+++ b/configs/bundle_config.yaml
@@ -68,6 +68,20 @@ databundles:
     - data/gebco/GEBCO_2021_TID.nc
     - data/copernicus/PROBAV_LC100_global_v3.0.1_2019-nrt_Discrete-Classification-map_EPSG-4326.tif
 
+  # load tutorial hydrobasins bundle for Africa only
+  bundle_tutorial_hydrobasins:
+    countries: [Africa]
+    tutorial: true
+    category: common
+    destination: "data/hydrobasins"
+    urls:
+      hydrobasins:
+        base_url: https://data.hydrosheds.org/file/HydroBASINS/standard/
+        suffixes: ["af"]
+    unzip: true
+    output:
+    - data/hydrobasins/hybas_world.shp
+
   # tutorial bundle specific for Nigeria and Benin only
   bundle_cutouts_tutorial_NGBJ:
     countries: [NG, BJ]
@@ -153,7 +167,7 @@ databundles:
         suffixes: ["af", "ar", "as", "au", "eu", "gr", "na", "sa", "si"]
     unzip: true
     output:
-    - data/hydrobasins/*.shp
+    - data/hydrobasins/hybas_world.shp
 
   # data bundle containing the data of the data folder common to all regions of the world
   bundle_data_earth:
--- a/configs/bundle_config.yaml
+++ b/configs/bundle_config.yaml
@@ -36,7 +36,7 @@ databundles:
     category: data
     destination: "data"
     urls:
-      zenodo: https://sandbox.zenodo.org/record/1223907/files/tutorial_data_NGBJ.zip?download=1
+      zenodo: https://sandbox.zenodo.org/records/3853/files/bundle_tutorial_NGBJ.zip?download=1
       gdrive: https://drive.google.com/file/d/1Vb1ISjhy7iwTTZYeezGd6S4nLt-EDGme/view?usp=drive_link
     output:
     - data/gebco/GEBCO_2021_TID.nc
@@ -49,7 +49,7 @@ databundles:
     category: data
     destination: "data"
     urls:
-      zenodo: https://sandbox.zenodo.org/record/1223907/files/tutorial_data_BW.zip?download=1
+      zenodo: https://sandbox.zenodo.org/records/3853/files/bundle_tutorial_BW.zip?download=1
       gdrive: https://drive.google.com/file/d/19IXvTD8gVSzgTInL85ta7QjaNI8ZPCCY/view?usp=drive_link
     output:
     - data/gebco/GEBCO_2021_TID.nc
@@ -62,7 +62,7 @@ databundles:
     category: data
     destination: "data"
     urls:
-      zenodo: https://sandbox.zenodo.org/record/1223907/files/tutorial_data_MA.zip?download=1
+      zenodo: https://sandbox.zenodo.org/records/3853/files/bundle_tutorial_MA.zip?download=1
       gdrive: https://drive.google.com/file/d/1VGzE8ZJHAvAQ9X44QNSX4rH3QF7Yi37D/view?usp=drive_link
     output:
     - data/gebco/GEBCO_2021_TID.nc
@@ -72,7 +72,7 @@ databundles:
   bundle_tutorial_hydrobasins:
     countries: [Africa]
     tutorial: true
-    category: common
+    category: hydrobasins
     destination: "data/hydrobasins"
     urls:
       hydrobasins:
@@ -80,7 +80,7 @@ databundles:
         suffixes: ["af"]
     unzip: true
     output:
-    - data/hydrobasins/hybas_world_v1c.shp
+    - data/hydrobasins/hybas_world.shp
 
   # tutorial bundle specific for Nigeria and Benin only
   bundle_cutouts_tutorial_NGBJ:
@@ -89,7 +89,7 @@ databundles:
     category: cutouts
     destination: "cutouts"
     urls:
-      zenodo: https://sandbox.zenodo.org/record/1223907/files/tutorial_cutouts_NGBJ.zip?download=1
+      zenodo: https://sandbox.zenodo.org/records/3853/files/bundle_cutouts_tutorial_NGBJ.zip?download=1
       gdrive: https://drive.google.com/file/d/1xnomHdXf_c5STrf7jtDiuRlN2zW0FSVC/view?usp=drive_link
     output: [cutouts/cutout-2013-era5-tutorial.nc]
     disable_by_opt:
@@ -102,7 +102,7 @@ databundles:
     category: cutouts
     destination: "cutouts"
     urls:
-      zenodo: https://sandbox.zenodo.org/record/1223907/files/tutorial_cutouts_BW.zip?download=1
+      zenodo: https://sandbox.zenodo.org/records/3853/files/bundle_cutouts_tutorial_BW.zip?download=1
       gdrive: https://drive.google.com/file/d/1DDQAtnIDM0FNC3vCldfHeH__IpTbyIJt/view?usp=drive_link
     output: [cutouts/cutout-2013-era5-tutorial.nc]
     disable_by_opt:
@@ -115,7 +115,7 @@ databundles:
     category: cutouts
     destination: "cutouts"
     urls:
-      zenodo: https://sandbox.zenodo.org/record/1223907/files/tutorial_cutouts_MA.zip?download=1
+      zenodo: https://sandbox.zenodo.org/records/3853/files/bundle_cutouts_tutorial_MA.zip?download=1
       gdrive: https://drive.google.com/file/d/1j5v2f4E756jmDMa707QvdNJq3xM4bYUk/view?usp=drive_link
     output: [cutouts/cutout-2013-era5-tutorial.nc]
     disable_by_opt:
@@ -139,12 +139,9 @@ databundles:
     category: common
     destination: "data"
     urls:
-      zenodo: https://sandbox.zenodo.org/record/1183583/files/tutorial_data_general.zip?download=1
+      zenodo: https://sandbox.zenodo.org/records/3853/files/tutorial_data_general.zip?download=1
       gdrive: https://drive.google.com/file/d/1nRLrs_kP0qVl-IHC4BFLjpoKa3HLk2Py/view
     output:
-    - data/costs.csv
-    - data/hydro_capacities.csv
-    - data/custom_powerplants.csv
     - data/eez/eez_v11.gpkg
     - data/ssp2-2.6/2030/era5_2013/Africa.nc
     - data/ssp2-2.6/2030/era5_2013/Asia.nc
@@ -152,14 +149,12 @@ databundles:
     - data/ssp2-2.6/2030/era5_2013/NorthAmerica.nc
     - data/ssp2-2.6/2030/era5_2013/SouthAmerica.nc
     - data/ssp2-2.6/2030/era5_2013/Oceania.nc
-    - data/hydrobasins/hybas_world_lev04_v1c.shp
-    - data/hydrobasins/hybas_world_lev05_v1c.shp
 
   # global data for hydrobasins
   bundle_hydrobasins:
     countries: [Earth]
     tutorial: false
-    category: common
+    category: hydrobasins
     destination: "data/hydrobasins"
     urls:
       hydrobasins:
--- a/configs/bundle_config.yaml
+++ b/configs/bundle_config.yaml
@@ -149,6 +149,7 @@ databundles:
     destination: "data/hydrobasins"
     urls:
       direct: https://data.hydrosheds.org/file/HydroBASINS/standard/hybas_af_lev06_v1c.zip
+    unzip: true
     output:
     - data/hydrobasins/hybas_world_lev06_v1c.shp
 
--- a/configs/bundle_config.yaml
+++ b/configs/bundle_config.yaml
@@ -141,6 +141,17 @@ databundles:
     - data/hydrobasins/hybas_world_lev04_v1c.shp
     - data/hydrobasins/hybas_world_lev05_v1c.shp
 
+  # global data for hydrobasins
+  bundle_hydrobasins:
+    countries: [Earth]
+    tutorial: false
+    category: common
+    destination: "data/hydrobasins"
+    urls:
+      direct: https://data.hydrosheds.org/file/HydroBASINS/standard/hybas_af_lev06_v1c.zip
+    output:
+    - data/hydrobasins/hybas_world_lev06_v1c.shp
+
   # data bundle containing the data of the data folder common to all regions of the world
   bundle_data_earth:
     countries: [Earth]
--- a/configs/bundle_config.yaml
+++ b/configs/bundle_config.yaml
@@ -148,8 +148,15 @@ databundles:
     category: common
     destination: "data/hydrobasins"
     urls:
-      hydrobasins: https://data.hydrosheds.org/file/HydroBASINS/standard/
-      suffixes: ["af", "ar", "as", "au", "eu", "gr", "na", "sa", "si"]
+  bundle_hydrobasins:
+    countries: [Earth]
+    tutorial: false
+    category: common
+    destination: "data/hydrobasins"
+    urls:
+      hydrobasins:
+         base_url: https://data.hydrosheds.org/file/HydroBASINS/standard/
+         suffixes: ["af", "ar", "as", "au", "eu", "gr", "na", "sa", "si"]
     unzip: true
     output:
     - data/hydrobasins/*.shp
--- a/configs/bundle_config.yaml
+++ b/configs/bundle_config.yaml
@@ -148,15 +148,9 @@ databundles:
     category: common
     destination: "data/hydrobasins"
     urls:
-  bundle_hydrobasins:
-    countries: [Earth]
-    tutorial: false
-    category: common
-    destination: "data/hydrobasins"
-    urls:
       hydrobasins:
-         base_url: https://data.hydrosheds.org/file/HydroBASINS/standard/
-         suffixes: ["af", "ar", "as", "au", "eu", "gr", "na", "sa", "si"]
+        base_url: https://data.hydrosheds.org/file/HydroBASINS/standard/
+        suffixes: ["af", "ar", "as", "au", "eu", "gr", "na", "sa", "si"]
     unzip: true
     output:
     - data/hydrobasins/*.shp
--- a/scripts/_helpers.py
+++ b/scripts/_helpers.py
@@ -422,7 +422,7 @@ def progress_retrieve(
 
     if headers:
         opener = urllib.request.build_opener()
-        opener.addheaders = [("User-agent", "Mozilla/5.0")]
+        opener.addheaders = headers
         urllib.request.install_opener(opener)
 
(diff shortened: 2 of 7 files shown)


ekatef:
Currently, the loaded data sources are hardcoded in `bundle_config.yaml`, which can potentially lead to troubles in case of mismatches with `hydro -> resource -> hydrobasins` in `config.yaml`. 

A possible improvement we discussed implies taking into account the config parameters. However, the used hydrobasin files are generally not very big, while `config.default` requests the biggest  of them. Would it be probably a better solution simply load all hydrobasis files which can be used. That would be also consistent with the current approach.

ekatef:
As a long-term solution, [a new basins dataset](https://dataservices.gfz-potsdam.de/panmetaworks/review/218da89b6b9388ce9161dc8462bd0ad65904300f2ec0fecfb53b45788c5614aa/) by Potsdam climate institute  may be of interest.

scripts/_helpers.py
```diff
@@ -418,6 +420,11 @@ def dlProgress(count, blockSize, totalSize, roundto=roundto):
     if data is not None:
         data = urllib.parse.urlencode(data).encode()
 
+    if headers:
+        opener = urllib.request.build_opener()
+        opener.addheaders = [("User-agent", "Mozilla/5.0")]
```
davide-f: Maybe the content of this may be headers itself, so we make the function quite general.
What do you think?

scripts/retrieve_databundle_light.py
```diff
@@ -337,11 +337,13 @@ def download_and_unzip_direct(config, rootpath, hot_run=True, disable_progress=F
...
             logger.info(f"Downloading resource '{resource}' from cloud '{url}'.")
-            progress_retrieve(url, file_path, disable_progress=disable_progress)
+            progress_retrieve(
+                url, file_path, headers=True, disable_progress=disable_progress
+            )
 
             # if the file is a zipfile and unzip is enabled
             # then unzip it and remove the original file
-            if config.get("unzip", False):
+            if config.get("unzip", False) or bool(re.search(".zip$", file_path)):
```
davide-f: Unfortunately, I believe that for hydrobasins, we need a specific different function, because:
1. the hydrobasins bundles to download depend on the region and level option, which is not easy to handle with bundles without overcomplicating
2. the output shape files may be merged: for a run on countries across continents, the shape files need to be merged

If we focus on creating such function(s) and then we integrate them into the workflow.
What do you think?

Here, the ` or bool(re.search(".zip$", file_path))` is unnecessary. if you specify: unzip: False into the bundle that you added this line is unnecessary

configs/bundle_config.yaml
```diff
@@ -141,6 +141,17 @@ databundles:
...
     - data/hydrobasins/hybas_world_lev05_v1c.shp
 
+  # global data for hydrobasins
+  bundle_hydrobasins:
+    countries: [Earth]
+    tutorial: false
+    category: common
+    destination: "data/hydrobasins"
+    urls:
+      direct: https://data.hydrosheds.org/file/HydroBASINS/standard/hybas_af_lev06_v1c.zip
```
davide-f: I comment here and below I add some examples.
We could define a new class like "urls: hydrobasins: " to distinguish the download procedure that is different than the others.
the new code "hydrobasins" require the definition of a new function like download_and_unzip_hydrobasins

Then, the url of the hydrobasin may be: "https://data.hydrosheds.org/file/HydroBASINS/standard/hybas_af_lev{:02d}_v1c.zip"

Inside the new download_and_unzip_hydrobasins, we need to do "url_hbasin = urls["hydrobasins"].forma({level from config file}"

With this format, if we add one bundle_hydrobasin for every region in the hydrobasin website, this workflow will automatically download all of them. Note that the output file of each databundle shall account for the region (e.g. AF) and shall be filled with the level value similarly to above.

Then, we need to merge them into a unique shape file. To do so, we can create a dedicated rule to do so: the function takes in input the list of hydro shape files, calculated with a dedicated function, and in output creates the merged file.
Alternatively the merging process can be included into the retrieve databundle, if that's easier to start with for you

davide-f (commented):
Thanks Katia :D
I've added some thoughts on the hydrobasin, let me know if they are clear and what is your opinion on that :)

scripts/_helpers.py
ekatef: > Maybe the content of this may be headers itself, so we make the function quite general. What do you think?

Agree and fixed.

scripts/retrieve_databundle_light.py
ekatef: Agree that a dedicated hydro-basins function would be helpful. Added one.

A condition fixed with `unzip` :)

configs/bundle_config.yaml
ekatef: Absolutely agree that a dedicated download function would be helpful, and great point about the merging procedure!

Regarding a special treatment for different regions, not sure I get your point right. Currently, we have a single file of global coverage, and it seems to work pretty well. Why wouldn't we reproduce the same approach, regardless of the requested region? The global dataset is about 5MB which is ~200 time smaller as compared with our environment 🙃 

Do you mean that it doesn't look polite if we'd re-creating a chunk of data set locally by default? 🙂 In this case I do agree that it makes absolute sense to go for the approach you suggest.

davide-f: The size of the dataset depends on the level. I've done a scan and the global v12 level may be of 500MB zipped, but it is unlikely to be commonly used.
We can go for it :) I like the proposal!

However, unfortunately there is no global bundle in the hydrobasins, we need to manually download all regions and merge them regardless.

ekatef:
> Thanks Katia :D I've added some thoughts on the hydrobasin, let me know if they are clear and what is your opinion on that :)

Thanks a lot, Davide! It has been tremendously helpful 😄 

Fixed technical points and happy to create a dedicated rule. It will be also useful for pre-processing of new PIK dataset 🙂

configs/bundle_config.yaml
ekatef: Thanks :)

Mmm... Probably, it could make sense to add a warning if the data to be loaded exceed some 100MB.

No problem with creating a merged dataset :)

scripts/retrieve_databundle_light.py
```diff
@@ -337,7 +337,9 @@ def download_and_unzip_direct(config, rootpath, hot_run=True, disable_progress=F
 
         try:
             logger.info(f"Downloading resource '{resource}' from cloud '{url}'.")
-            progress_retrieve(url, file_path, disable_progress=disable_progress)
+            progress_retrieve(
+                url, file_path, headers=True, disable_progress=disable_progress
+            )
```
davide-f: Maybe we don't need this change?

scripts/retrieve_databundle_light.py
```diff
@@ -354,6 +356,73 @@ def download_and_unzip_direct(config, rootpath, hot_run=True, disable_progress=F
...
 
 
+def download_and_unzip_hydrobasins(
+    config, rootpath, hot_run=True, disable_progress=False
+):
+    """
+    download_and_unzip_basins(config, rootpath, dest_path, hot_run=True,
+    disable_progress=False)
+
+    Function to download and unzip the data for hydrobasins.
```
davide-f: May be nice to add the path to the data and the licensing

scripts/retrieve_databundle_light.py
```diff
@@ -354,6 +356,73 @@ def download_and_unzip_direct(config, rootpath, hot_run=True, disable_progress=F
...
+    -------
+    True when download is successful, False otherwise
+    """
+    resource = config["category"]
+    url_templ = config["urls"]["hydrobasins"]
+    suffix_list = config["urls"]["suffixes"]
+
+    basins_fl = snakemake.config["renewable"]["hydro"]["resource"]["hydrobasins"]
+    level_pattern = r".*?lev(.*)_.*"
+    level_code = re.findall(level_pattern, basins_fl)
```
davide-f: Cool that you got how to identify the level from the path!
The level to be used later in the level_code to download should however be loaded from the config.yaml file.

scripts/retrieve_databundle_light.py
```diff
@@ -354,6 +356,73 @@ def download_and_unzip_direct(config, rootpath, hot_run=True, disable_progress=F
...
+                progress_retrieve(
+                    url,
+                    file_path,
+                    headers=[("User-agent", "Mozilla/5.0")],
+                    disable_progress=disable_progress,
+                )
+
+                # if the file is a zipfile and unzip is enabled
+                # then unzip it and remove the original file
+                if config.get("unzip", False):
```
davide-f: It is always a zip file in the hydrobasins, right?
If so, maybe this if is not needed

scripts/retrieve_databundle_light.py
```diff
@@ -394,7 +463,11 @@ def download_and_unzip_post(config, rootpath, hot_run=True, disable_progress=Fal
         logger.info(f"Downloading resource '{resource}' from cloud '{url}'.")
 
         progress_retrieve(
-            url, file_path, data=postdata, disable_progress=disable_progress
+            url,
+            file_path,
+            data=postdata,
+            header=header,
+            disable_progress=disable_progress,
```
davide-f: Maybe this change is not needed, where is header defined?

davide-f (commented):
Cool @ekatef :D
With little changes, the new function allows to download each hydrobasin databundle.
As next steps, there is the need to make sure all databundles for Earth are downloaded and merged together :) cool! :D
This is a great step towards making PyPSA-Earth better structured :D

scripts/retrieve_databundle_light.py
ekatef: Thanks for the catch! Fixed.

scripts/retrieve_databundle_light.py
ekatef: Have been trying to avoid changing the config :D
But you are absolutely right from the long-term perspective. Adjusted.

scripts/retrieve_databundle_light.py
ekatef: Well, it has been definitely a good idea to look into the license conditions:
> The following copyright statement must be displayed with, attached to or embodied in (in a reasonably prominent manner) the documentation or metadata of any Licensee Product or Program provided to an End User when utilizing the Licensed Materials: ...

Have added the required statement to the docstring. Do you agree?

scripts/retrieve_databundle_light.py
ekatef: Fixed :)

scripts/retrieve_databundle_light.py
ekatef: A header is added in the function loading hydrobasins. Removed it from here. Thanks for the catch!

ekatef:
> Cool @ekatef :D With little changes, the new function allows to download each hydrobasin databundle. As next steps, there is the need to make sure all databundles for Earth are downloaded and merged together :) cool! :D This is a great step towards making PyPSA-Earth better structured :D

Thanks for checking it @davide-f! Have implemented the fixes.

ekatef:
Current approach seems to be functional, but `pydevd ` is complaining on low performance, although computational times doesn't feel very bad. 

Results of initial testing on level 2:

image

To be tested:
1) on finer resolution for basins shapes;
2) for the whole workflow.

ekatef:
Testing on 12th basins level gives the whole coverage:

image

with a nice level of regional details:

image

A file size for the global dataset at 12th hydrobasins level is 1.32 GB.

ekatef:
CI currently fails, which may be the case due to some mismatch in the configuration parameters. Have to check it. Apart of that, it may be a good idea to add some progress to track progress of hydrobasins merge.

configs/bundle_config.yaml
```diff
@@ -141,6 +141,19 @@ databundles:
...
 
+  # global data for hydrobasins
+  bundle_hydrobasins:
+    countries: [Earth]
+    tutorial: false
+    category: common
+    destination: "data/hydrobasins"
+    urls:
+      hydrobasins: https://data.hydrosheds.org/file/HydroBASINS/standard/
+      suffixes: ["af", "ar", "as", "au", "eu", "gr", "na", "sa", "si"]
```
davide-f: I like the approach!
maybe to be consistent with the others we could do something like:
```suggestion
      hydrobasins:
         url: https://data.hydrosheds.org/file/HydroBASINS/standard/
         suffixes: ["af", "ar", "as", "au", "eu", "gr", "na", "sa", "si"]
```

scripts/retrieve_databundle_light.py
```diff
@@ -354,6 +356,85 @@ def download_and_unzip_direct(config, rootpath, hot_run=True, disable_progress=F
...
+    """
+    resource = config["category"]
+    url_templ = config["urls"]["hydrobasins"]
+    suffix_list = config["urls"]["suffixes"]
+
+    level_code = snakemake.config["renewable"]["hydro"]["resource"]["hydrobasins_level"]
+    if level_code <= 9:
+        level_code = "0" + str(level_code)
+    else:
+        level_code = str(level_code)
```
davide-f: Could be replaced with
```suggestion
    level_code = "{:02d}".format(int(level_code))
```

scripts/retrieve_databundle_light.py
```diff
@@ -354,6 +356,85 @@ def download_and_unzip_direct(config, rootpath, hot_run=True, disable_progress=F
...
+                with ZipFile(file_path, "r") as zipfile:
+                    zipfile.extractall(config["destination"])
+
+                os.remove(file_path)
+                logger.info(f"Downloaded resource '{resource}' from cloud '{url}'.")
+            except:
+                logger.warning(
+                    f"Failed download resource '{resource}' from cloud '{url}'."
+                )
+                return False
```
davide-f: This clearly resembles the download_direct approach.
Rather than duplicating the code, do you think it is possible to reuse the other function and save space?
That may be possible by creating an ad-hoc config dictionary probably with unzip being true, the headers and the appropriate updated paths, but let me know your thoughts :)

scripts/retrieve_databundle_light.py
```diff
@@ -617,6 +701,36 @@ def datafiles_retrivedatabundle(config):
...
+        hydrobasins_level = snakemake.config["renewable"]["hydro"]["resource"][
+            "hydrobasins_level"
+        ]
+        if hydrobasins_level <= 9:
+            hydrobasins_level = "0" + str(hydrobasins_level)
+        else:
+            hydrobasins_level = str(hydrobasins_level)
+
+        regex_to_look_for = re.compile(".*_lev" + hydrobasins_level + "_v1c.shp$")
+        files_to_merge = list(filter(regex_to_look_for.match, files_in_dir))
```
davide-f: Probably, we could exploit the Path object here and do something like:
```
mask_file = "hybas_*_lev{:02d}_v1c.shp".format(int(hydrobasins_level))
files_to_merge = Path(basins_path).glob(mask_file)
```

davide-f (commented):
Cool!
We are close! :D I like it much!!!

configs/bundle_config.yaml
ekatef: Happy to improve consistency, but not sure I get the point. May you please elaborate a bit? My understanding was that every databundles entry has `urls` as a high-level parameter, which can have additional parameters:

```python
  #   urls:
  #     zenodo: ...
  #     gdrive: ...
```

So, I have tried to follow this pattern: `bundle_name -> urls -> additional parameters`. Do I miss anything?

scripts/retrieve_databundle_light.py
ekatef: Great solution! Implemented.

scripts/retrieve_databundle_light.py
ekatef: Perfect :)
Implemented, many thanks for the suggestions!

scripts/retrieve_databundle_light.py
ekatef: Agree that it would be nice to avoid duplication which clearly presents now. Do you mean that we can read all the needed elements from `bundle.config`, even if the structure on the request differ? To me, the required data structure is not yet evident, but happy to build it.

Currently, I see the following differences `download_hydrobasins` from `download_direct`:
1) a loop is needed to go through all the suffixes;
2) header should be specified;
3) HydroBasins license text is added (probably, we can move it to the docstring, for example).

To me, the first point currently looks the most tricky one. Do you have any ideas how to deal with that?

ekatef:
> Cool! We are close! :D I like it much!!!

Thank a lot for the in-depth review! :D Agree that we are close to finalise. :) 

According to my understanding, two points left are:
1) cross-check and improve data structure in `bundle_config` and the downloading function in `retrieve_databundle`;
2) fix `renewable_profiles` which currently crashes at calculating hydro potential in a line `inflow = correction_factor * func(capacity_factor=True, **resource)`. I suspect, there may be a problem with configuration parameters which are not expected by atlite hydro function.

Have added some clarification questions on point 1, and shall investigate point 2 in more details.

ekatef:
Update: fixed configuration parameters for the hydrobasins level. Need to cross-check format of the merged file, as currently `atlite` transformation seems to be not completely happy with the results of `gpd.GeoDataFrame(pd.concat(gpdf_list))`.

ekatef:
After checking the structure of the merged file, it looks like some basins may be duplicated by the current merge approach. In particular, when testing level 4, there is a entry appears twice with same id and exactly same geometry (a basin which belongs to tow different macro-regions?):

```csv
778,1040029810,0,1040029810,1040029810,0.0,0.0,275925.8,275925.8,1530,0,1,0,144,"MULTIPOLYGON (((10.277777777777796 34.32...
2377,1040029810,0,1040029810,1040029810,0.0,0.0,275925.8,275925.8,1530,0,1,0,144,"MULTIPOLYGON (((10.277777777777796 34.32...
```

That leads to `atlite` not being happy, and most likely causes the current CI troubles.

The current merge approach should be supplemented by checking for duplicates.

davide-f:
> After checking the structure of the merged file, it looks like some basins may be duplicated by the current merge approach. In particular, when testing level 4, there is a entry appears twice with same id and exactly same geometry (a basin which belongs to tow different macro-regions?):
> 
> ```
> 778,1040029810,0,1040029810,1040029810,0.0,0.0,275925.8,275925.8,1530,0,1,0,144,"MULTIPOLYGON (((10.277777777777796 34.32...
> 2377,1040029810,0,1040029810,1040029810,0.0,0.0,275925.8,275925.8,1530,0,1,0,144,"MULTIPOLYGON (((10.277777777777796 34.32...
> ```
> 
> That leads to `atlite` not being happy, and most likely causes the current CI troubles.
> 
> The current merge approach should be supplemented by checking for duplicates.

Interesting and fanstastic job!
Do the entries have an ID?
It would be great to remove duplicated ids rather than performing geospatial checks that check for overlaps...

configs/bundle_config.yaml
davide-f: Indeed :)
The purpose of the flag urls would be to give one entry for alternative options to download the same bundle.
That's why I'd be prone to keep one entry (hydrobasins) under urls and potentially add multiple entries underneat, such as:
```suggestion
  bundle_hydrobasins:
    countries: [Earth]
    tutorial: false
    category: common
    destination: "data/hydrobasins"
    urls:
      hydrobasins:
         base_url: https://data.hydrosheds.org/file/HydroBASINS/standard/
         suffixes: ["af", "ar", "as", "au", "eu", "gr", "na", "sa", "si"]
```

scripts/retrieve_databundle_light.py
davide-f: My suggestion was to propose calling
`download_and_unzip_direct(..., ..., hot_run=hot_run, disable_progress=disable_progress)`
within this for loop, rather than copying the code again, but we can go for the duplication if it is easier

scripts/retrieve_databundle_light.py
```diff
@@ -618,6 +700,35 @@ def datafiles_retrivedatabundle(config):
...
+    logger.info(
+        "Merging single files into:\n\t"
+        + "hybas_world_lev"
+        + str(hydrobasins_level)
+        + "_v1c.shp"
+    )
+    fl_merged.to_file(
+        os.path.join(
+            basins_path, "hybas_world_lev{:02d}_v1c.shp".format(int(hydrobasins_level))
+        )
```
davide-f: This path should be taken from the config

davide-f (commented):
Cool :D converging :)

ekatef:
> > After checking the structure of the merged file, it looks like some basins may be duplicated by the current merge approach. In particular, when testing level 4, there is a entry appears twice with same id and exactly same geometry (a basin which belongs to tow different macro-regions?):
> > ```
> > 778,1040029810,0,1040029810,1040029810,0.0,0.0,275925.8,275925.8,1530,0,1,0,144,"MULTIPOLYGON (((10.277777777777796 34.32...
> > 2377,1040029810,0,1040029810,1040029810,0.0,0.0,275925.8,275925.8,1530,0,1,0,144,"MULTIPOLYGON (((10.277777777777796 34.32...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That leads to `atlite` not being happy, and most likely causes the current CI troubles.
> > The current merge approach should be supplemented by checking for duplicates.
> 
> Interesting and fanstastic job! Do the entries have an ID? It would be great to remove duplicated ids rather than performing geospatial checks that check for overlaps...

Totally agree :D Yeah, the basins have ids, and pandas `drop_duplicates` seems do the trick.

ekatef:
> Cool :D converging :)

Thanks a lot for the review and absolutely agree ;)

Testing the whole workflow as currently the PR break CI. Once done, I'll introduce the changes you suggest (absolutely agree with them) and shall also try to add a clear error message in case the url is not available.

scripts/retrieve_databundle_light.py
ekatef: > My suggestion was to propose calling
> download_and_unzip_direct(..., ..., hot_run=hot_run, disable_progress=disable_progress)
> within this for loop, rather than copying the code again, but we can go for the duplication if it is easier

Thanks for the explanation! Have investigated a bit, and it looks like there may be a way to avoid duplication, but we'd need for that to encapsulate a part of `download_and_unzip_direct`. The point is that currently all the downloading functions use the same arguments list `config, rootpath, hot_run=True, disable_progress=False`, which makes possible to use a wildcard when calling the download function:

https://github.com/pypsa-meets-earth/pypsa-earth/blob/5f77c9814d942250371fb8959869370818a8aa48/scripts/retrieve_databundle_light.py#L676-L680

`config` is common for all `download_and_unzip_*` functions, while processing of the config parameters differs for different functions. So, I have added `download_and_unpack` function to replace duplicated functionality both in `download_and_unzip_direct` and `download_and_unzip_hydrobasins`. Would you agree with this approach?

ekatef: > This path should be taken from the config

An understanding question: do you mean using `snakemake.config["renewable"]["hydro"]["resource"]["hydrobasins"]`? 

Agree that it's the major parameter transferred to atlite. However, there is also `hydrobasins_level` parameter which basically duplicate `hydrobasins`. So, it's perfectly possible to organise in the `config.yaml` a data mismatch: like set `hydrobasins_level: 4` along with `hydrobasins: data/hydrobasins/hybas_world_lev06_v1c.shp`. If we'll take the file name from the config, this error will remain unnoticed, while if an actual  `hydrobasins_level` is used to generate the merged file, we keep the prepared data consistent, while there will be an error thrown while building renewable profiles. To me, this design look more error-safe.

What is your feeling about that? (Sorry for bringing the complexity which is probably not needed there 🙃)

ekatef:
> Cool :D converging :)

Thanks! :)
Have introduced changes; there are two points I'm not completely sure about. Have added some comments on both. Happy to exchange on both if you feel it'd be easier.

scripts/retrieve_databundle_light.py
davide-f: I quite believe that in our case we should have a file like: data/hydrobasins/hybas_world.shp that is independent from the level.
It is the merge of all hydrobasin shapes regardless of the level.
If the level is modified, the function is triggered again, that should simplify the procedure.
Alternatively, we need to create a specific rule that merges the hydrobasins shapes

configs/bundle_config.yaml
```diff
@@ -141,6 +141,20 @@ databundles:
...
+    category: common
+    destination: "data/hydrobasins"
+    urls:
+      hydrobasins:
+        base_url: https://data.hydrosheds.org/file/HydroBASINS/standard/
+        suffixes: ["af", "ar", "as", "au", "eu", "gr", "na", "sa", "si"]
+    unzip: true
+    output:
+    - data/hydrobasins/*.shp
+
```
davide-f: As comments:
- the output file could be hybas_world_v1c.shp, that represent the merged file, value not to be loaded from the config, but from here
- we could add a databundle for the tutorial, which is the same as this one, but limited to africa. So the changes would be: tutorial=true, countries=["Africa"], and suffixes=["af"]

Pretty cool! :D
Once this works, we can think of creating the databundles for the tutorials and plan to go for zenodo.
We can try again zenodo sandbox and when it works we can move to the default

davide-f:
To avoid the comment being lost, the following comment is also added above:

I quite believe that in our case we should have a file like: data/hydrobasins/hybas_world.shp that is independent from the level.
It is the merge of all hydrobasin shapes regardless of the level.
If the level is modified, the function is triggered again, that should simplify the procedure.
Alternatively, we need to create a specific rule that merges the hydrobasins shapes

davide-f:
Merging :D

PR #942, opened 2023-12-22
Improve robustness on WDPA download

ekatef:
It looks like the actual protected areas file by WDPA can be not available for the current month. [Such an issue](https://github.com/PyPSA/pypsa-eur/issues/807) has been reported for PyPSA-Eur recently.

Added a check to ensure that the requested url responds. In case of failure, the workflow attempts to use the next or previous month.

## Checklist

### DIFF
--- a/scripts/retrieve_databundle_light.py
+++ b/scripts/retrieve_databundle_light.py
@@ -87,6 +87,7 @@ from zipfile import ZipFile
 
 import geopandas as gpd
 import pandas as pd
+import requests
 import yaml
 from _helpers import (
     configure_logging,
@@ -276,7 +277,28 @@ def download_and_unzip_protectedplanet(
     def get_first_day_of_previous_month(date):
         return get_first_day_of_month(date - dt.timedelta(days=1))
 
+    def get_url(date):
+        month_MMM = date.strftime("%b")
+        year_YYYY = date.year
+        return url.format(month=month_MMM, year=year_YYYY)
+
+    def remote_file_exists(url):
+        response = requests.head(url)
+        return response.status_code == 200
+
     current_first_day = get_first_day_of_month(dt.datetime.today())
+
+    # customize url to current month
+    url_iter = get_url(current_first_day)
+    # resource_iter = resource + " - " + month_MMM + " " + str(year_YYYY)
+
+    # try the next month
+    if not remote_file_exists(url_iter):
+        url_iter = get_url(current_first_day + dt.timedelta(32))
+
+    # try the previous month
+    if not remote_file_exists(url_iter):
+        url_iter = get_url(current_first_day - dt.timedelta(1))
 
     if hot_run:
         if os.path.exists(file_path):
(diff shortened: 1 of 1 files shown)


davide-f:
Hello :D
Thanks for reporting :)
Actually, the feature you are proposing was already intended, was it not working?

Maybe it could be cleaner, but basically we exploit the internal try-except and the first iteration that succeeds terminates the loop

ekatef:
> Hello :D Thanks for reporting :) Actually, the feature you are proposing was already intended, was it not working?
> 
> Maybe it could be cleaner, but basically we exploit the internal try-except and the first iteration that succeeds terminates the loop

Hello :)
My feeling is that the current approach is definitely cleaner, the problem is that the shift-date shifts the date to an earlier month (Dec -> Nov -> Oct), while it appears that it may be also needed to check to a later month (Nov -> Jan).

Currently, everything is perfectly fine with the current implementation, as data for December seems to present on the server, as well. But actually the site promises data for January only.

By the way, just recognised that the changes can be much simpler. Testing re-implementation :)

ekatef:
It may be as simple, as that: https://github.com/ekatef/pypsa-earth/compare/main...fix_wdpa_date2

@davide-f  would you agree?

davide-f:
> > Hello :D Thanks for reporting :) Actually, the feature you are proposing was already intended, was it not working?
> > Maybe it could be cleaner, but basically we exploit the internal try-except and the first iteration that succeeds terminates the loop
> 
> Hello :) My feeling is that the current approach is definitely cleaner, the problem is that the shift-date shifts the date to an earlier month (Dec -> Nov -> Oct), while it appears that it may be also needed to check to a later month (Nov -> Jan).
> 
> Currently, everything is perfectly fine with the current implementation, as data for December seems to present on the server, as well. But actually the site promises data for January only.
> 
> By the way, just recognised that the changes can be much simpler. Testing re-implementation :)

Sorry, I don't get it, if today is the 15th of december, it downloads december; are you saying that we should start from the coming month, hence january?
Is it the case now?

If we really need that, ok for the proposed fix, but we just need +31

ekatef:
> > > Hello :D Thanks for reporting :) Actually, the feature you are proposing was already intended, was it not working?
> > > Maybe it could be cleaner, but basically we exploit the internal try-except and the first iteration that succeeds terminates the loop
> > 
> > 
> > Hello :) My feeling is that the current approach is definitely cleaner, the problem is that the shift-date shifts the date to an earlier month (Dec -> Nov -> Oct), while it appears that it may be also needed to check to a later month (Nov -> Jan).
> > Currently, everything is perfectly fine with the current implementation, as data for December seems to present on the server, as well. But actually the site promises data for January only.
> > By the way, just recognised that the changes can be much simpler. Testing re-implementation :)
> 
> Sorry, I don't get it, if today is the 15th of december, it downloads december; are you saying that we should start from the coming month, hence january? Is it the case now?
> 
> If we really need that, ok for the proposed fix, but we just need +31

Yeah, exactly: to load now the actual WDPA file we need January 2024. You may have a look on their site: https://www.protectedplanet.net/en/thematic-areas/wdpa?tab=WDPA

ekatef:
Closed in favour of #945

Issue #936, opened 2023-12-17
Earth-OSM latest version(0.2.0) not compatible


GbotemiB:
## Checklist

- [x] I am using the current [`main`](https://github.com/pypsa-meets-earth/pypsa-earth/tree/main) branch or the latest [release](https://github.com/pypsa-meets-earth/pypsa-earth/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-earth` environment](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

The current update with earth-osm v0.1.0 is not compatible with the current pypsa-earth installation

This image is from the download_osm_script https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/scripts/download_osm_data.py#L109-L119
![image](https://github.com/pypsa-meets-earth/pypsa-earth/assets/48842684/a318300e-9096-4add-a57f-a1620bf4d60c)

This image is from earth-osm repo https://github.com/pypsa-meets-earth/earth-osm/blob/main/earth_osm/eo.py#L88-L109
![image](https://github.com/pypsa-meets-earth/pypsa-earth/assets/48842684/2ba440bf-d97f-4df4-a6a8-99a69ebbb3be)

## Error Message
Terminal output
```bash
ERROR:_helpers:An error happened in module '/Users/gbotemi/Documents/code/PYPSA/mc/pypsa-earth/scripts/download_osm_data.py', function '<module>': get_osm_data() got an unexpected keyword argument 'region_list'
Traceback (most recent call last):
  File "/Users/gbotemi/Documents/code/PYPSA/mc/pypsa-earth/.snakemake/scripts/tmprd57evk9.download_osm_data.py", line 113, in <module>
    eo.get_osm_data(
TypeError: get_osm_data() got an unexpected keyword argument 'region_list'
```

ekatef:
Hello @GbotemiB and thanks a lot for reporting!
I can reproduce this issue. Currently testing a fix for it.

ekatef:
Testing with down-graded earth-osm version (`earth-osm=0.1.0`) resolves the issue with OSM loading but leads to a new one in `simplify_network`: `ValueError(
ValueError: unable to infer dtype on variable 'buses_generator'; object array contains mixed native types: str_, str`. I fear it may be connected with `xarray` versions...

@GbotemiB have you found a way to resolve the environment issues?

[...]

PR #954, opened 2024-01-04
Fix for earth-osm V0.2.0

GbotemiB:
# Closes # (if applicable).

The changes in this PR address the incompatibility issues of earth-osm 0.2.0 in pypsa-earth framework.

Code changes were made to download_osm_data.py and clean_osm_data.py.

In `download_osm_data.py`, the files are downloaded with different names depending on the country being selected. I restructured the last part of the code to rename the file and move them to the resource folder.

In `clean_osm_data.py`, substations geopandas dataframe have a geometry column which is in polygons and points. I added a centroid method to find the centroid and replace it with the polygons.
You can also read about a follow-up conversation here #936 

## Checklist

### DIFF
--- a/scripts/download_osm_data.py
+++ b/scripts/download_osm_data.py
@@ -126,7 +126,7 @@ if __name__ == "__main__":
     # If the file is empty, it is not created
     # This is a workaround to create empty files for the workflow
 
-    # Rename and move osm files to the resources folder
+    # Rename and move osm files to the resources folder output
     for file in new_files:
         for name in names:
             for f in format:
--- a/scripts/download_osm_data.py
+++ b/scripts/download_osm_data.py
@@ -113,6 +113,7 @@ if __name__ == "__main__":
         update=False,
         mp=True,
         data_dir=store_path_data,
+        out_dir=store_path_data,
         out_format=["csv", "geojson"],
         out_aggregate=True,
     )
--- a/scripts/download_osm_data.py
+++ b/scripts/download_osm_data.py
@@ -113,12 +113,12 @@ if __name__ == "__main__":
         update=False,
         mp=True,
         data_dir=store_path_data,
-        out_dir=store_path_data,
+        out_dir=store_path_resources,
         out_format=["csv", "geojson"],
         out_aggregate=True,
     )
 
-    out_path = Path.joinpath(store_path_data, "out")
+    out_path = Path.joinpath(store_path_resources, "out")
     names = ["generator", "cable", "line", "substation"]
     format = ["csv", "geojson"]
     new_files = os.listdir(out_path)  # list downloaded osm files
--- a/scripts/download_osm_data.py
+++ b/scripts/download_osm_data.py
@@ -106,7 +106,7 @@ if __name__ == "__main__":
     store_path_data = Path.joinpath(Path().cwd(), "data", "osm")
     country_list = country_list_to_geofk(snakemake.params.countries)
 
-    eo.get_osm_data(
+    eo.save_osm_data(
         primary_name="power",
         region_list=country_list,
         feature_list=["substation", "line", "cable", "generator"],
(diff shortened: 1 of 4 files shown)


davide-f:
@GbotemiB FYI earth-osm has pushed some updates, though not yet with a new version.
It may be good to test :D
That can be tested by installing the package using pip and main or just waiting for the new release.

GbotemiB:
Hi @davide-f, I have tested the latest version of earth-osm with the changes in my PR. Everything seems to be working perfectly. We just have to wait for the next release.

davide-f:
That's great news @GbotemiB and @mnm-matin !!! :D
As the new release is available and the CI passes, we can merge this :) That's really amazing!

davide-f:
@GbotemiB the new earth-osm version is out :D Many thanks @mnm-matin !!!!
I'm rerunning the CI to see if it passes.
It would be great @GbotemiB to incorporate the constraint "earth-osm>=0.2.1" as this PR won't be backcompatible

GbotemiB:
> @GbotemiB the new earth-osm version is out :D Many thanks @mnm-matin !!!! I'm rerunning the CI to see if it passes. It would be great @GbotemiB to incorporate the constraint "earth-osm>=0.2.1" as this PR won't be backcompatible

Absolutely. The version will be `earth-osm>=2.1` which is the latest version.

GbotemiB:
Hi @davide-f All checks for CI are successful.

davide-f:
That's great and super promising!!!!
As a question: would you be interested in testing download osm data for the entire africa for example?
As this is a major change it would be nice to check the functionality for a large region, also Europe may work (given the notebook we have)

GbotemiB:
@davide-f That will be great. So I will be testing for Africa and Europe, right?

davide-f:
> @davide-f That will be great. So I will be testing for Africa and Europe, right?

Both is a lot, just one should be fine :)

GbotemiB:
Testing download_osm_data for Africa is a success. I was able to download successfully.

davide-f:
Super @GbotemiB !!!
Can you add a release note under minor fix? Then we can merge the PR :)

scripts/download_osm_data.py
```diff
@@ -106,36 +106,36 @@ def convert_iso_to_geofk(
...
 
-    eo.get_osm_data(
+    eo.save_osm_data(
         primary_name="power",
         region_list=country_list,
         feature_list=["substation", "line", "cable", "generator"],
         update=False,
         mp=True,
         data_dir=store_path_data,
+        out_dir=store_path_data,
```
davide-f: Sorry @GbotemiB 
I've just seen that out_dir and data_dir are the same.
out_dir should be the store_path_resources.
Could you revise that? by doing so the download_osm_data part is really up-to-date&state-of-the-art

GbotemiB: Actually @davide-f, 
I think it is that way because of the structure from earth-osm https://github.com/pypsa-meets-earth/earth-osm/blob/main/earth_osm/eo.py#L128-L138

davide-f: > Actually @davide-f, I think it is that way because of the structure from earth-osm https://github.com/pypsa-meets-earth/earth-osm/blob/main/earth_osm/eo.py#L128-L138

The default is out_dir=data_dir, but there are benefits in using different values:
The idea is to disentangle where the raw files are saved (data_dir) from where the output files are saved (out_dir).
If we use out_dir=data_dir download_osm_data cannot operate in parallel for different countries, which is performance-wise terrible.
Enabling to distinguish the 2 folders can in principle allow that, as the output folder can be in the form "resources/{scenario_name}/..."

GbotemiB: I don't quite understand.

There may be a mix-up.  The data_dir is the location for the pbf files, while the out_dir handles the files relating to cable, generators, lines, and substations.

davide-f: > I don't quite understand.
> 
> There may be a mix-up.  The data_dir is the location for the pbf files, while the out_dir handles the files relating to cable, generators, lines, and substations.

Yeah, that's exactly the point.
They don't need to be the same folder. The pbfs can be in one folder and results in another one.

The paralleling problem I mentioned is the following:
Let's assume we use the scenario management for NG and KE and run the current download_osm_data in parallel.

Earth_osm can download the 2pbfs in parallel and store it in data_dir.
However, I'd out_dir is the same as data_dir, the 2 instances will try to write on the same temporary files in data_dir.

Instead, if we use a different out_dir different than out_dir (the res dir), this problem doesn't occur

GbotemiB: Oh I see. I think it is clear now.

GbotemiB: With the suggestion you gave, this is what the folder structure looks like
`out_dir=store_path_resources`
`data_dir=store_path_data`

```bash
resources
└── SD_BJ
    └── osm
        └── raw
            └── out
                ├── BJ_SD_generator.csv
                ├── BJ_SD_generator.geojson
                ├── BJ_SD_line.csv
                ├── BJ_SD_line.geojson
                ├── BJ_SD_substation.csv
                └── BJ_SD_substation.geojson
```

```bash
data/osm
├── pbf
│   ├── benin-latest.osm.pbf
│   ├── benin-latest.osm.pbf.md5
│   ├── sudan-latest.osm.pbf
│   └── sudan-latest.osm.pbf.md5
└── power
    ├── BJ_power.json
    └── SD_power.json
```

davide-f: Sounds good :)
As a comment, the "SD_BJ" folder, applies only if the user specify a scenario name.
I don't see any changes yet though

GbotemiB: Yeah. You are right.

But there are changes from the way it used to work. Before, the `out` folder in the `resources` tree used to be under the `data/osm` tree. But now the case is different.

davide-f: The out folder may be placed into resources/osm/raw/out and then results are moved 1 folder backwards.
If that simplifies, we can keep it that way.
Alternatively, we can place it directly into resources/osm/raw (no out folder); this may be cleaner but may requires some checks with shutil.move to avoid issues if the source and destination paths match.

GbotemiB: I have made some adjustment to the script as you suggested. The results are now moved 1 folder backwards.

davide-f: Fantastic!

PR #961, opened 2024-01-18
fix to avoid negative objective value

### LINKED ISSUE #956: add lines as s_nom_min not s_nom
## Checklist

## Describe the Bug

Lines are currently added as `s_nom`. If lines are oversized in a network and can reduce for the optimal case `s_nom_opt`. 1) deconstructing lines makes no sense (yet) and 2) this can lead to negative objective values. 

https://github.com/pypsa-meets-earth/pypsa-earth/blob/e0117356d3600da4366d499b9dfbf9ebbbc61bff/scripts/base_network.py#L353-L359

We should probably add lines as `s_nom_min`

GbotemiB:
# Closes #956 

The changes in the PR is as suggested in #956 which is to adjust the lines in `base_network.py:_set_lines_s_nom_from_linetypes` from `s_nom` to `s_nom_min` to prevent negative objective values after solving network.

## Checklist


davide-f (commented):
Great @GbotemiB :)

So, the change should not be placed here, but rather in prepare_network.
By looking at the code in prepare_network, we actually have a function "set_transmission_limits" that should do that, but it seems it is not working.
https://github.com/pypsa-meets-earth/pypsa-earth/blob/e4edb177fb5f1216f9d4ead8cf2e319a1c8a48ec/scripts/prepare_network.py#L188

It may be good to investigate why, may you be interested in that?

GbotemiB:
Thanks @davide-f, 

However, when I changed s_nom to s_nom_min in base_network.py, it worked to prevent my networks from having negative objective values.

I will run the prepare_network with a debug. Hopefully, I will be able to understand why it didn't work in the first instance.

davide-f:
> Thanks @davide-f,
> 
> However, when I changed s_nom to s_nom_min in base_network.py, it worked to prevent my networks from having negative objective values.
> 
> I will run the prepare_network with a debug. Hopefully, I will be able to understand why it didn't work in the first instance.

It works, but if that option is changed there, there can be problems if the user desire to not do capacity expansion on the network as p_nom doesnt get initialized properly.
The set_transmission limit was meant to do that but it seems something is not working well

davide-f:
@GbotemiB not sure if you are also working here :) In case of problems, needs feel free to write :)

GbotemiB:
> @GbotemiB not sure if you are also working here :) In case of problems, needs feel free to write :)

Thanks @davide-f, I looked into it last week, I couldn't fix it yet. But I will report progress on this later today.

GbotemiB:
@davide-f I have been able to finally debug the `prepare_network.py`
https://github.com/pypsa-meets-earth/pypsa-earth/blob/e4edb177fb5f1216f9d4ead8cf2e319a1c8a48ec/scripts/prepare_network.py#L188

It turns out the section of the highlighted code below doesn't get executed.
https://github.com/pypsa-meets-earth/pypsa-earth/blob/e4edb177fb5f1216f9d4ead8cf2e319a1c8a48ec/scripts/prepare_network.py#L207-L212

The following image should give more context into why the conditional is not working.
image

GbotemiB:
https://github.com/pypsa-meets-earth/pypsa-earth/blob/e4edb177fb5f1216f9d4ead8cf2e319a1c8a48ec/scripts/prepare_network.py#L197

@davide-f, If you don't mind, can you clarify the purpose of this section

GbotemiB:
https://github.com/pypsa-meets-earth/pypsa-earth/blob/e4edb177fb5f1216f9d4ead8cf2e319a1c8a48ec/scripts/prepare_network.py#L191-L196

I also noticed that `num_parallel` here is a little lower than the same that was used in `base_network.py`

davide-f:
> @davide-f I have been able to finally debug the `prepare_network.py`
> 
> https://github.com/pypsa-meets-earth/pypsa-earth/blob/e4edb177fb5f1216f9d4ead8cf2e319a1c8a48ec/scripts/prepare_network.py#L188
> 
> It turns out the section of the highlighted code below doesn't get executed.
> 
> https://github.com/pypsa-meets-earth/pypsa-earth/blob/e4edb177fb5f1216f9d4ead8cf2e319a1c8a48ec/scripts/prepare_network.py#L207-L212
> 
> The following image should give more context into why the conditional is not working. image

Nice catch!
I believe it makes sense to revise the default parameters of tutorial and default to make sure that works.
Probably, the v0.3 should be revised in to opt (maybe better) or v1.5 (like in pypsa-eur) or something like that.

Great :D

davide-f:
> https://github.com/pypsa-meets-earth/pypsa-earth/blob/e4edb177fb5f1216f9d4ead8cf2e319a1c8a48ec/scripts/prepare_network.py#L191-L196
> 
> I also noticed that `num_parallel` here is a little lower than the same that was used in `base_network.py`

This was imported from pypsa-eur.
The idea is that where the lines do not have a specific line type (which corresponds to a s_nom), the s_nom is derived from current and voltage of that line.
I guess the idea behind it is to fill only those values that are not filled yet.
Did you find issues with this?

davide-f:
> https://github.com/pypsa-meets-earth/pypsa-earth/blob/e4edb177fb5f1216f9d4ead8cf2e319a1c8a48ec/scripts/prepare_network.py#L191-L196
> 
> I also noticed that `num_parallel` here is a little lower than the same that was used in `base_network.py`

This is instead expected as in the simplify_network phase, the whole network is simplified to a base voltage (e.g. 380kV) equivalent and we play with num_parallel to handle the equivalence.
See https://github.com/pypsa-meets-earth/pypsa-earth/blob/e4edb177fb5f1216f9d4ead8cf2e319a1c8a48ec/scripts/simplify_network.py#L118

GbotemiB:
I went through `simplify_network.py`. Why is there a need to recalculate `num_parallel`
https://github.com/pypsa-meets-earth/pypsa-earth/blob/e4edb177fb5f1216f9d4ead8cf2e319a1c8a48ec/scripts/simplify_network.py#L135

GbotemiB:
Also, I noticed that in `base_network.py`, the line types is as in the image

image
 
while in `prepare_network.py`, the line types is
image

I think it might have to do with the rebasing of voltages

GbotemiB:
I also found out that `n.lines.s_nom` and `_lines_s_nom` have different values in `prepare_network.py`

image

when I used replace lines_s_nom by n.lines.s_nom here, my solved network didn't have any negative objectives
https://github.com/pypsa-meets-earth/pypsa-earth/blob/e4edb177fb5f1216f9d4ead8cf2e319a1c8a48ec/scripts/prepare_network.py#L207-L212

GbotemiB:
looking at `v_nom`, the values are 300kV, which I don't think we should be doing the multiplication with line_types of 380kV
image

GbotemiB:
Hi @davide-f, It looks like the reason why `n.lines.s_nom` is different from `_lines_s_nom` is because the parameters used to calculate `_lines_s_nom` has been altered in `simplify_network:simplify_network_to_base_voltage`
https://github.com/pypsa-meets-earth/pypsa-earth/blob/9168ca7dfbbeba90c338b3b1c83eb105c5f2f28a/scripts/simplify_network.py#L129-L135

GbotemiB:
Regarding p_nom here
https://github.com/pypsa-meets-earth/pypsa-earth/blob/e4edb177fb5f1216f9d4ead8cf2e319a1c8a48ec/scripts/prepare_network.py#L211-L212

During debugging, `n.links.p_nom` and `n.links.p_nom_min` results to zero.
image
 But `p_nom` and `p_nom_min` in generators are the same which is not equal to zero.
image

ekatef:
Hello! 

@GbotemiB it looks like you have done a quite intense investigation and found very nice bugs (in particular, the one which relates to loading the options from the file name looks great!!), However, the whole picture still looks a bit too complicated. I'd suggest to take a step back and clarify the architecture of the lines transformations workflow, starting from osm-extracted data up to the pre-solved network. I'll doing something similar at the moment and happy to support you with that.

The resulted diagram would facilitate discussion of high-level features with @davide-f 

Once the intended path would have been clear, it will simplify testing and introducing fixes, as it'll be clear what should we expect at each stage.

What do you think @GbotemiB?

GbotemiB:
Thanks @ekatef, 

You are definitely right. Do you have any thoughts on how I can clrify the line transformation architecture?

GbotemiB:
Thanks @ekatef.

 You are definitely right. I will start with clarifying the workflow.

GbotemiB:
Hello @davide-f @ekatef @pz-max 

Following last week's discussion on a possible check for the negative objective issue.

I did a check to verify the results from the solved networks (Monte-Carlo Runs) `p_nom_opt - p_nom` to see if p_nom_opt > p_nom

image
image
image
image
image

davide-f:
Thanks @GbotemiB ! :D

As a comment, can you disclose the get_difference function?

The function should not apply the calculation on the energy_statistics but the whole n.generators. for example:

`n.generators.groupby("carrier").eval('p_nom_opt - p_nom').min()`

(need to check if that works)

The idea is to check if there is a row where the p_nom_opt is lower than the p_nom.

The same should be tested for lines, links and storage_units as well.

However, the current change in the PR shall be reverted as that is not the intended behavior

davide-f:
Could be good to check also directly:
`n.generators.groupby("carrier").eval('p_nom - p_nom_min').min()`

This number should be always non-negative as well

if you define:
```
def get_difference(obj, col1, col2):
   return obj.groupby("carrier").eval(f'col1 - col2').min()
```
or alike, it may help

GbotemiB:
@davide-f You are right about the difference function.

```python
def get_difference(n):
	print((n.generators["p_nom_opt"] -  n.generators["p_nom"]).groupby(n.generators.carrier).sum())
	return None	
```

Here is the function I used which is similar to the formula you suggested. 

I will recheck as you suggested using a clear function.

GbotemiB:
This is the function that works

```python
def get_difference(obj, col1, col2):
	return obj.groupby("carrier").apply(lambda x: (x[col1] - x[col2]).min())
```

The solar carrier tech results have p_nom_opt lower than p_nom
image

I tested for links, lines, and storage units. Links and storage_units look fine. But `lines` have s_nom_opt lower than `s_nom`
image

GbotemiB:
I have also reverted the current change in the PR.

davide-f:
Ok, the problem is in lines and can you confirm that the following code is executed?
https://github.com/pypsa-meets-earth/pypsa-earth/blob/de4da143e92d4d7aefcaac588175db82d93f0670/scripts/prepare_network.py#L206-L211

davide-f:
Moreover:
- please check in pypsa-eur if in base_network the s_nom_min is also specified
- please check in pypsa-eur the same chunk of code in the equivalent script to check how that works there and see if there are major differences

Many thanks!

GbotemiB:
@davide-f 

Checking pypsa-eur prepare_network.py 

https://github.com/PyPSA/pypsa-eur/blob/4fbb3c81c4ddb589614f67dc01c885c9b3806bbe/scripts/prepare_network.py#L78-L97

This is the part that has to do with lines that is present in PyPSA-Eur but not in PyPSA-Earth

GbotemiB:
> Ok, the problem is in lines and can you confirm that the following code is executed?
> 
> https://github.com/pypsa-meets-earth/pypsa-earth/blob/de4da143e92d4d7aefcaac588175db82d93f0670/scripts/prepare_network.py#L206-L211

I also already confirm that this part is executed in the workflow.

> please check in pypsa-eur if in base_network the s_nom_min is also specified

Also in pypsa-eur, s_nom_min is not specified in base_network.

davide-f:
Ok, so @GbotemiB I understand that the problem is back about why the s_nom_min calculated here differs from the p_nom value available in the network right?
It seems that one of the causes seems due to the missing name in the line type; did you get why that happens?
My feeling is that it is related to the clustering procedure

GbotemiB:
Hi @davide-f, I haven't figured out the cause of the missing names in the line type. I will investigate the clustering procedure to see if something has been missed there.

GbotemiB:
Hi @davide-f, I have just uploaded a simple notebook [here](https://github.com/GbotemiB/csp-design/blob/notebook/negative_obj.ipynb). I made some checks in the notebook. Feel free to look through it.

davide-f:
Interesting @GbotemiB ! We are getting closer to the problem!
I see you compare: p_nom.sum(), p_nom_min.sum()
Could you add to the list also the following computation:
n.lines.eval("sqrt(3) * i_nom * v_nom") ?
Note that i_nom may not be available for the line and there may be the need to be retrieved similarly to the code you have checked some comments previously.
I think we should identify when differences occur and solve the issue.
Note, the sum may not be the best approach but for now it does ok

GbotemiB:
> Interesting @GbotemiB ! We are getting closer to the problem! I see you compare: p_nom.sum(), p_nom_min.sum() Could you add to the list also the following computation: n.lines.eval("sqrt(3) * i_nom * v_nom") ? Note that i_nom may not be available for the line and there may be the need to be retrieved similarly to the code you have checked some comments previously. I think we should identify when differences occur and solve the issue. Note, the sum may not be the best approach but for now it does ok

Just a little clarification. The difference section was between p_nom_opt and p_nom, while the other was between s_nom and s_nom_min. 

I wanted to observe the changes in the workflow for s_nom and s_nom_min.

GbotemiB:
I have updated the notebook [here](https://github.com/GbotemiB/csp-design/blob/notebook/negative_obj.ipynb). I also rearranged the process for ease of understanding.

I also noticed that num_parallel was not included in the calculation formula. Was it intentional?

davide-f:
> I have updated the notebook [here](https://github.com/GbotemiB/csp-design/blob/notebook/negative_obj.ipynb). I also rearranged the process for ease of understanding.
> 
> I also noticed that num_parallel was not included in the calculation formula. Was it intentional?

Cool Emmanuel, apologies missed communication.
Could you update the last cells of the notebook to account for the number of parallel lines?
With:
``` python
(np.sqrt(3)
	* base.lines["type"].map(base.line_types.i_nom) 
	* base.lines.v_nom
	* base.lines.num_parallel).sum()
```

We are moving forward! 🚀 :D

GbotemiB:
Thanks @davide-f,

I have updated the cells in the [notebook](https://github.com/GbotemiB/csp-design/blob/notebook/negative_obj.ipynb).
Feel free to check again.

davide-f:
Amazing progress @GbotemiB ! :D :D
We identified the script where problem occurs: simplify_network
Indeed for that network, the calculation using the product i_nom*v_nom*num_parallel * sqrt(3) does not match the s_nom value.

Somewhere there some calculations are performed.
The huntch is that during the various steps of the simplification and probably the aggregation strategies, something does not go as expected.

To debug that, I'd create a simple auxiliary function that computes the i_nom*v_nom*num_parallel product, reads the s_nom value and check if the values match and print the results.

I'd test that for the various modified functions across the script to identify where the occurrence arrives.

What do you think?

GbotemiB:
@davide-f, I'm glad we identified the simplify_network script as the source of the issue.
I'll go ahead and implement this function and test it at different points in the script. Hopefully, this will help us pinpoint exactly where things are going wrong.

davide-f:
Great :D Thank you!!!
As heads up, my guess is that during some aggregation strategies, line types are lost and/or the num_parallel are not summed up.

GbotemiB:
Hello @davide-f,

I looked over the `simplify_network.py` script, looks like there is an additional line present in pypsa-eur that is not in pypsa-earth. You can check [here](https://github.com/PyPSA/pypsa-eur/blob/master/scripts/simplify_network.py#L125C5-L125C29)

I added the line to my local files in the script:simplify_network_to_base_voltage.py, this seems to resolved the negative objective issue.

What are your thoughts?

GbotemiB:
I have also updated the notebook [here](https://github.com/GbotemiB/csp-design/blob/notebook/negative_obj.ipynb) with the recent changes I made.

davide-f:
> Hello @davide-f,
> 
> I looked over the `simplify_network.py` script, looks like there is an additional line present in pypsa-eur that is not in pypsa-earth. You can check [here](https://github.com/PyPSA/pypsa-eur/blob/master/scripts/simplify_network.py#L125C5-L125C29)
> 
> I added the line to my local files in the script:simplify_network_to_base_voltage.py, this seems to resolved the negative objective issue.
> 
> What are your thoughts?

Really?

Yeah, definitely that should be added, however, instead of having "380" fixed, it should be base_voltage.
Does that really fix the issue? Does the notebook consider this?

It seems a too easy fix... XD


Update: I've better checked the script but unfortunately that does not solve the issue.
In the notebook, as a comment it is difficult to compare the values: we need to scroll a lot to compare the s_nom value and the corresponding s_nom computed using the formula.

By comparing the values, for example, for siml,
the sum of s_noms is: 327228
while the corresponding value computed using sqrt(3) * i_nom * v_nom * num_parallel is: 281636
The two values should match

GbotemiB:
@davide-f, I will check the notebook again. But I tested the implementation against the monte-carlo notebook. The negative objective values seem to not be present. 
https://github.com/GbotemiB/documentation/blob/monte-carlo/notebooks/monte_carlo_tutorial.ipynb

davide-f:
> @davide-f, I will check the notebook again. But I tested the implementation against the monte-carlo notebook. The negative objective values seem to not be present. https://github.com/GbotemiB/documentation/blob/monte-carlo/notebooks/monte_carlo_tutorial.ipynb

I understand, but the objective can still be misleading.
The s_nom_min should match s_nom.
Now we have reduced the difference but the problem is not fixed definitely unfortunately, if you had still resources. If not no problem

GbotemiB:
> > @davide-f, I will check the notebook again. But I tested the implementation against the monte-carlo notebook. The negative objective values seem to not be present. https://github.com/GbotemiB/documentation/blob/monte-carlo/notebooks/monte_carlo_tutorial.ipynb
> 
> I understand, but the objective can still be misleading. The s_nom_min should match s_nom. Now we have reduced the difference but the problem is not fixed definitely unfortunately, if you had still resources. If not no problem

@davide-f, Looks like I forgot to update the path in the notebook. I think we can smile now 😄 
I updated the runs already, feel free to check.
https://github.com/GbotemiB/csp-design/blob/notebook/negative_obj.ipynb

davide-f (commented):
Great @GbotemiB :D
Would you like to add a release note? :)
This PR is ready to fly! :)

davide-f:
Many thanks @GbotemiB ! Merging :)

Issue #887, opened 2023-10-06
Avoid rebase?


pz-max:
- PyPSA-Earth will lift all voltages before the optimization to a [base voltage e.g. 380kV](https://github.com/pypsa-meets-earth/pypsa-earth/blob/b205e83c3b631448a778e29257eed551790b9d5c/scripts/simplify_network.py#L8)

- Before the optimization, we rebase (see [function](https://github.com/pypsa-meets-earth/pypsa-earth/blob/b205e83c3b631448a778e29257eed551790b9d5c/scripts/base_network.py#L440-L468)) the voltage to different classes e.g. random example: 110-330kV is class 220kV, 330-550kV is class 380kV such that we can assign physical parameters from the [config.yaml](https://github.com/pypsa-meets-earth/pypsa-earth/blob/b205e83c3b631448a778e29257eed551790b9d5c/config.default.yaml#L165-L170). The question is if we can actually avoid it? I think what we need for that would be a line database with the linetype, **typical voltage** and physical parameters. PyPSA has already a [typical line database](https://pypsa.readthedocs.io/en/latest/components.html#line-types), which we could extend by a column "voltage" to allocate data automatically. My suggestion is to store this voltage column extended line-type database in PyPSA-Earth (one voltage has one line type) such that we can avoid the rebase.

ekatef:
A there is detailed analysis of what effect does the voltage rebase have on the transmission capacity: https://github.com/GbotemiB/documentation/blob/it_de_comparison/notebooks/validation/it_de_comparison_notebook.ipynb 

A long story short, it looks like we overestimate transmission capacity when mapping lower voltage lines into lines with a higher "standard" threshold value. But it would be perfect to have a professional comment from the perspective of electrical engineering. @davide-f could you please give your view on that matter?

ekatef:
Hello @pz-max!
I have a feeling that this issue could have been fixed with #999, which basically implements the ideas you sketched in this issue. Could you please check if any other points left?

ekatef:
Closed as seems to be completed. Feel free to reopen if appropriate

Issue #862, opened 2023-09-12
Potential bug in network parameter, num_parallel or length 


pz-max:
We thought the OSM data is doing quite well in EU looking at: 
<img src="https://github.com/pypsa-meets-earth/pypsa-earth/assets/61968949/8205518f-60a1-40ba-823a-3105232830f5" width="300"/>

**Reported by Tom B:**
PyPSA-Eur (entsoe) is MUCH closer to official statistics (which here include Turkey, unlike PyPSA-Eur): https://eepublicdownloads.entsoe.eu/clean-documents/Publications/Statistics/Factsheet/entsoe_sfs2021_web.pdf

<img src="https://github.com/pypsa-meets-earth/pypsa-earth/assets/61968949/e5f03522-bb55-4962-a3ed-c42daae7c538" width="300"/>
<img src="https://github.com/pypsa-meets-earth/pypsa-earth/assets/61968949/9dd80133-4915-4e44-9146-f80d2c8f3a94" width="300"/>

PyPSA-Eur 159,000 for 380 kV is much closer to official 186,000 than OSM's 290,000.
Ditto for 220+300: our 125,000 km is much closer to official 131,000 km than OSM's 930,000 km. 

**TODO:**
Two possible reasons for big OSM error:
- false num_parallel data in OSM
- false num_parallel calcuations/assumptions when building the OSM network

ekatef:
Thanks for sharing @pz-max!

Agree that the result is quite surprising. When doing grid validation for Central and Western Asia, we found some discrepancies, but they have never be as high as more than eight times...

 My feeling is that it would be good:
1) to look into some more details of the discrepancies found for `elec.nc` trying to provide some additional insights by voltage classes and countries;
2) compare OSM-extracted lengths values after cleaning with [ENTSO-extracted lengths.](https://github.com/PyPSA/pypsa-eur/blob/master/data/entsoegridkit/lines.csv).

pz-max:
One extra point:
3. Maybe not the correct `.crs` system was set in the PyPSA-Earth config...
So `elec.nc` might be wrong to begin with

GbotemiB:
Here is a follow on @ekatef suggestions.

Here is a voltage comparison plot between OSM and ENTSOE
![image](https://github.com/pypsa-meets-earth/pypsa-earth/assets/48842684/85c55f55-80b3-4fd3-a863-c3fdb20c0639)

Using log transformation on the y-axis to scale the data
[...]

PR #978, opened 2024-02-29
Improve equivalent transformation calculations in base.nc

ekatef:
A suggestion on a possible to partially address #862 and #887 to decrease errors introduced by `base.nc`.

Currently, we are using voltage rebase in `base.nc` which purpose is to map all the voltages to a few "standard" values. However, information on the original voltage is lost during this rebase transformation.

The suggestion is:
1) keep the original voltage before mapping the voltage;
2) adjust `num_parallel` for the lines after the voltage rebase to compensate an effect of the voltage increase on the transmission capacity introduced during rebase.

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -18,6 +18,8 @@ E.g. if a new rule becomes available describe how to use it `snakemake -j1 run_t
 
 **Minor Changes and bug-fixing**
 
+* Keep data on the original voltage value when rebasing voltages to the standard values and adjust the transmission capacity accordingly. `PR #898 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/978>`__
+
 
 PyPSA-Earth 0.3.0
 =================
--- a/scripts/base_network.py
+++ b/scripts/base_network.py
@@ -496,8 +496,15 @@ def base_network(
     lines_dc = lines[lines.tag_frequency.astype(float) == 0].copy()
 
     lines_ac = _set_electrical_parameters_lines(lines_config, voltages_config, lines_ac)
+    lines_ac.num_parallel = lines_ac.num_parallel * (
+        lines_ac.v_nom_original / lines_ac.v_nom
+    )
+
     lines_dc = _set_electrical_parameters_dc_lines(
         lines_config, voltages_config, lines_dc
+    )
+    lines_dc.num_parallel = lines_dc.num_parallel * (
+        lines_dc.v_nom_original / lines_dc.v_nom
     )
 
     transformers = _set_electrical_parameters_transformers(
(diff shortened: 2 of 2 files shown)


ekatef:
As a comment, there is also an effect of `i_nom` which may significantly impact the result for the transmission capacity and is determined by the wire design. 

E.g. for Al-steel wires `*-AL1/*-ST1A`, `i_nom` varies between `0.21` and `1.15` across the types which can be used for `100kV` (data taken from [pandapower](https://github.com/e2nIEE/pandapower/blob/ad33d19fe5f474126749c4a8d8c30dc28d6cc21a/doc/std_types/linetypes.csv)). Not sure there is a way to account for the size types which are really in use in one or another power system. Probably, it may be a good idea to add an empiric coefficient which accounts for an average difference between the `i_nom` values calculated according to the line types we assume and real `i_nom` values across the power system.

Obviously, that is a purely data-based perspective ideally should can be adjusted according to the real-world operation of the existing power grids.

ekatef:
A quick demonstration of the effect of the linetype definition of the transmision capacity using [the notebook](https://github.com/GbotemiB/documentation/blob/it_de_comparison/notebooks/validation/it_de_comparison_notebook.ipynb) by @GbotemiB (thanks for the handy tool!):

* original picture with 110kV mapped into `"243-AL1/39-ST1A 110.0"` (with `i_nom` `0.645` kA)

image

* a picture re-calculated with 110kV mapped into `"N2XS(FL)2Y 1x120 RM/35 64/110 kV"` (with `i_nom` `0.366` kA):

image

davide-f (commented):
Thanks katia!
It seems this PR is ready, but could you please explain the image you posted?
I don't see how much we improve or so.

I believe it is also worth adding a release_note in this case

ekatef:
> Thanks katia! It seems this PR is ready, but could you please explain the image you posted? I don't see how much we improve or so.
> 
> I believe it is also worth adding a release_note in this case

Thank you, Davide :)

The image was in fact rather a problem statement, not testing results. The point is, the PR is dealing with a part of uncertainty which relates to the voltage mapping, while we need some other approach to deal with the uncertainty linked with definition of the line type. The picture above demonstrates this effect of the linetype. (Sorry for messing things up!)

Testing results for the PR itself are as follows. When looking into data for Norway:

* ENTSOe data for an overall transmission capacity are 0.23e6 MVA when using PyPSA-Eur `base.nc` network build on ENTSOe data (or 0.27e6 MVA if taking ENTSOe csv directly);

* original PyPSA-Earth network has an overall transmission capacity 1.73e6 MVA;

* application of this PR reduces the transmission capacity to 1.07e6 MVA;

* application of this PR and filtering by the original voltage to match ENTSOe voltage levels (>=300kV) gives 0.55e6 MVA.

So, we are still two times off, but I'd say that is quite an improvement :D

The topology comparison looks like that:

image

davide-f (approved):
Many thanks @ekatef and totally understand!
Then, I think this PR can be merged as-is, that's great!

PR #995, opened 2024-03-21
Add csv-demand parser

### LINKED ISSUE #987: Support input demand data with csv format

It would be nice to include automatic procedures to support csv dataformats as raw inputs for demand, beyond the default .nc files that we currently support.
A simple parser in build_demand may be done to support the two data formats.
We also have .csv files in the data folders that contain the same data as in .nc ones, we could investigate to potentially support both.

ekatef:
#987 

The aim of the PR is to provide an option to use csv files as demand inputs.

## Checklist

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -43,6 +43,7 @@ run = config.get("run", {})
 RDIR = run["name"] + "/" if run.get("name") else ""
 CDIR = RDIR if not run.get("shared_cutouts") else ""
 
+load_data_paths = get_load_paths_gegis("data", config)
 if config["enable"].get("retrieve_cost_data", True):
     COSTS = "resources/" + RDIR + "costs.csv"
 else:
@@ -416,12 +417,10 @@ rule build_demand_profiles:
         snapshots=config["snapshots"],
         load_options=config["load_options"],
         countries=config["countries"],
-        weather_year=config["load_options"]["weather_year"],
-        prediction_year=config["load_options"]["prediction_year"],
-        ssp=config["load_options"]["ssp"],
     input:
         base_network="networks/" + RDIR + "base.nc",
         regions="resources/" + RDIR + "bus_regions/regions_onshore.geojson",
+        load=load_data_paths,
         #gadm_shapes="resources/" + RDIR + "shapes/MAR2.geojson",
         #using this line instead of the following will test updated gadm shapes for MA.
         #To use: downlaod file from the google drive and place it in resources/" + RDIR + "shapes/
--- a/Snakefile
+++ b/Snakefile
@@ -43,11 +43,7 @@ run = config.get("run", {})
 RDIR = run["name"] + "/" if run.get("name") else ""
 CDIR = RDIR if not run.get("shared_cutouts") else ""
 
-if os.path.exists(os.path.join("data", config["load_options"]["ssp"])):
-    load_data_paths = get_load_paths_gegis("data", config, check_existence=True)
-else:
-    # demand profiles are not available yet -> no opportunity to use custom load files
-    load_data_paths = get_load_paths_gegis("data", config, check_existence=False)
+load_data_paths = get_load_paths_gegis("data", config)
 
 if config["enable"].get("retrieve_cost_data", True):
     COSTS = "resources/" + RDIR + "costs.csv"
--- a/Snakefile
+++ b/Snakefile
@@ -43,7 +43,7 @@ run = config.get("run", {})
 RDIR = run["name"] + "/" if run.get("name") else ""
 CDIR = RDIR if not run.get("shared_cutouts") else ""
 
-if os.path.exists("data"):
+if os.path.exists(os.path.join("data", config["load_options"]["ssp"])):
     load_data_paths = get_load_paths_gegis("data", config, check_existence=True)
 else:
     # demand profiles are not available yet -> no opportunity to use custom load files
--- a/Snakefile
+++ b/Snakefile
@@ -43,6 +43,7 @@ run = config.get("run", {})
 RDIR = run["name"] + "/" if run.get("name") else ""
 CDIR = RDIR if not run.get("shared_cutouts") else ""
 
+load_data_paths = get_load_paths_gegis("data", config)
 if config["enable"].get("retrieve_cost_data", True):
     COSTS = "resources/" + RDIR + "costs.csv"
 else:
--- a/Snakefile
+++ b/Snakefile
@@ -43,7 +43,6 @@ run = config.get("run", {})
 RDIR = run["name"] + "/" if run.get("name") else ""
 CDIR = RDIR if not run.get("shared_cutouts") else ""
 
-load_data_paths = get_load_paths_gegis("data", config)
 if config["enable"].get("retrieve_cost_data", True):
     COSTS = "resources/" + RDIR + "costs.csv"
 else:
@@ -417,10 +416,12 @@ rule build_demand_profiles:
         snapshots=config["snapshots"],
         load_options=config["load_options"],
         countries=config["countries"],
+        weather_year=config["load_options"]["weather_year"],
+        prediction_year=config["load_options"]["prediction_year"],
+        ssp=config["load_options"]["ssp"],
     input:
         base_network="networks/" + RDIR + "base.nc",
         regions="resources/" + RDIR + "bus_regions/regions_onshore.geojson",
-        load=load_data_paths,
         #gadm_shapes="resources/" + RDIR + "shapes/MAR2.geojson",
         #using this line instead of the following will test updated gadm shapes for MA.
         #To use: downlaod file from the google drive and place it in resources/" + RDIR + "shapes/
--- a/Snakefile
+++ b/Snakefile
@@ -420,7 +420,7 @@ rule build_demand_profiles:
     input:
         base_network="networks/" + RDIR + "base.nc",
         regions="resources/" + RDIR + "bus_regions/regions_onshore.geojson",
-        load=load_data_paths,
+        load=get_load_paths_gegis("data", config, check_existence=True),
         #gadm_shapes="resources/" + RDIR + "shapes/MAR2.geojson",
         #using this line instead of the following will test updated gadm shapes for MA.
         #To use: downlaod file from the google drive and place it in resources/" + RDIR + "shapes/
--- a/Snakefile
+++ b/Snakefile
@@ -43,7 +43,6 @@ run = config.get("run", {})
 RDIR = run["name"] + "/" if run.get("name") else ""
 CDIR = RDIR if not run.get("shared_cutouts") else ""
 
-load_data_paths = get_load_paths_gegis("data", config)
 if config["enable"].get("retrieve_cost_data", True):
     COSTS = "resources/" + RDIR + "costs.csv"
 else:
--- a/Snakefile
+++ b/Snakefile
@@ -43,7 +43,12 @@ run = config.get("run", {})
 RDIR = run["name"] + "/" if run.get("name") else ""
 CDIR = RDIR if not run.get("shared_cutouts") else ""
 
-load_data_paths = get_load_paths_gegis("data", config)
+if os.path.exists("data"):
+    load_data_paths = get_load_paths_gegis("data", config, check_existence=True)
+else:
+    # demand profiles are not available yet -> no opportunity to use custom load files
+    load_data_paths = get_load_paths_gegis("data", config, check_existence=False)
+
 if config["enable"].get("retrieve_cost_data", True):
     COSTS = "resources/" + RDIR + "costs.csv"
 else:
@@ -420,7 +425,7 @@ rule build_demand_profiles:
     input:
         base_network="networks/" + RDIR + "base.nc",
         regions="resources/" + RDIR + "bus_regions/regions_onshore.geojson",
-        load=get_load_paths_gegis("data", config, check_existence=True),
+        load=load_data_paths,
         #gadm_shapes="resources/" + RDIR + "shapes/MAR2.geojson",
         #using this line instead of the following will test updated gadm shapes for MA.
         #To use: downlaod file from the google drive and place it in resources/" + RDIR + "shapes/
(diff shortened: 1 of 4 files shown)


ekatef:
CI currently fails during building a test configuration, when `get_load_paths_gegis` is called from Snakemake. It looks like there may be some difference in naming conventions.

ekatef:
The tutorial workflow in CI fails due to an empty power plants matching located in the requested region. Not sure in which way it is connected with the changes introduced by the PR, and can't reproduce it locally yet.

ekatef:
@davide-f could you please have a look? CI is failing due an some troubles in finding power plants for the region, which I can't reproduce locally... Any ideas be very welcome on what may go wrong!

scripts/build_demand_profiles.py
```diff
@@ -109,15 +110,29 @@ def get_load_paths_gegis(ssp_parentfolder, config):
...
+        avail_regions = [
+            cnt
+            for cnt in region_load
+            if os.path.exists(os.path.join(load_dir, cnt + ext))
+        ]
+        if len(avail_regions) == len(region_load):
+            for continent in region_load:
+                load_path = os.path.join(load_dir, continent + ext)
+                load_paths.append(load_path)
+            break
```
davide-f: Could you try this?
I think the two for loops should be inverted (first the continent then the extension).

```suggestion
    for continent in region_load:
        for ext in [".nc", ".csv"]:
            load_path = os.path.join(load_dir, continent + ext)
            if os.path.exists(load_path):
                load_paths.append(load_path)
            break
```

scripts/build_demand_profiles.py
```diff
@@ -109,15 +110,29 @@ def get_load_paths_gegis(ssp_parentfolder, config):
...
+
+    # if len(load_paths) == 0:
```
davide-f: To delete at the end

scripts/build_demand_profiles.py
```diff
@@ -176,8 +191,42 @@ def build_demand_profiles(
     substation_lv_i = n.buses.index[n.buses["substation_lv"]]
     regions = gpd.read_file(regions).set_index("name").reindex(substation_lv_i)
     load_paths = load_paths
-    # Merge load .nc files: https://stackoverflow.com/questions/47226429/join-merge-multiple-netcdf-files-using-xarray
-    gegis_load = xr.open_mfdataset(load_paths, combine="nested")
+
+    if str(load_paths[0]).endswith(".csv"):
```
davide-f: It would be nice to place this in an auxiliary function.
what do you think?

Regarding the read_csv, please, use the read_csv_nafix from helpers; it already supports additional arguments.
You could do:
`read_csv_nafix(pth, sep=";", dtype={...})`

moreover, I see that you create load_regions, but then you use only the region_name.
Do we need to define the dictionary?
I'm wondering if this work:
        gegis_load = gegis_load.assign_coords(
            {
                "region_name": ("region_code", df.region_name) [optionally with unique?]
            }
        )

scripts/build_demand_profiles.py
davide-f: Note that this approach may mix csv and nc files though

scripts/build_demand_profiles.py
```diff
@@ -176,8 +191,42 @@ def build_demand_profiles(
...
+                    "region_code",
+                    [name for (code, name) in load_regions.items()],
+                )
+            }
+        )
+    else:
+        # Merge load .nc files: https://stackoverflow.com/questions/47226429/join-merge-multiple-netcdf-files-using-xarray
+        gegis_load = xr.open_mfdataset(load_paths, combine="nested")
+
     gegis_load = gegis_load.to_dataframe().reset_index().set_index("time")
```
davide-f, line 226: The comments above assume that load_paths can contain any combination csv and/or nc files.
That requires to chenge the previous logic:
we may need to load each file independently either using a read_csv approach or a read_xarray, and then merge the xarrays with [xr.merge](https://docs.xarray.dev/en/stable/generated/xarray.merge.html)

Were you envisioning something like this?

The alternative only csv or only nc files is also possible, but requires a change in the logic on how the load_regions are loaded and may cause limitations unfortunately

davide-f (commented):
Great @ekatef :D

scripts/build_demand_profiles.py
```diff
@@ -109,15 +110,25 @@ def get_load_paths_gegis(ssp_parentfolder, config):
...
-            str(prediction_year),
-            "era5_" + str(weather_year),
-            str(continent) + ".nc",
-        )
-        load_paths.append(load_path)
+        for ext in [".nc", ".csv"]:
+            load_path = os.path.join(load_dir, continent + ext)
+            if os.path.exists(load_path):
+                load_paths.append(load_path)
+            break
```
davide-f: Please, move it inside the if, otherwise the .csv never gets executed

ekatef:
The changes in this PR should be aligned with #1017.

ekatef:
CI fails on building the test configuration in the tutorial run complaining that it's not able to find the demand input looking into `'data/ssp2-2.6/2030/era5_2013/Africa.nc'`. The reason is that there is still no data files on the stage of the configuration building. Replaced an error with a warning.

ekatef:
The issue has been linked with the fact that the load paths list has been formed by CI before loading any data. As a result, all the existence checks were negative, and a list of load files empty.

Have modified definition of the load paths in the Snakemake, adding tracking of load parameters of Snakemake.params.

scripts/build_demand_profiles.py
ekatef: Fixed the message and revised the checks. 

Generally, I'd keep here the warning in case if non-existing demand data are requested: the current demand limitations seems to be not clear enough, as some misunderstandings are quite common in how to define `weather_year` and `prediction_year`. Although, a more detailed usability check could make sense. What do you think?

scripts/build_demand_profiles.py
ekatef: Agree, revised

scripts/build_demand_profiles.py
ekatef: Great point. Have changed a logic to ensure that it's possible to use a combination of `csv` and `nc`-files

scripts/build_demand_profiles.py
ekatef: Done! Thanks :)

ekatef:
Tested successfully on both `nc`-only inputs and a combination of `nc` and `csv`.

The current implementation changes the architecture of Snakemake workflow, as the previous one was designed to run also if the demand inputs don't exist yet. Not sure the current implementation is the best possible, and happy to revise.

@davide-f would be very grateful for your review 🙂

Snakefile
```diff
@@ -43,7 +43,6 @@ run = config.get("run", {})
 RDIR = run["name"] + "/" if run.get("name") else ""
 CDIR = RDIR if not run.get("shared_cutouts") else ""
 
-load_data_paths = get_load_paths_gegis("data", config)
```
davide-f, line 46: I'd be in favor to keep the original implementation to keep retrocompatibility, if possible

davide-f: the change proposed in this PR should have no effect in the workflow. if the user changes the options, the input files to the rule change as well and if not available, the corresponding rules are retriggered.
What do you think?

Also changes to keep previous setup should be easy. What do you think?

Snakefile
```diff
@@ -420,7 +420,7 @@ rule build_demand_profiles:
     input:
         base_network="networks/" + RDIR + "base.nc",
         regions="resources/" + RDIR + "bus_regions/regions_onshore.geojson",
-        load=load_data_paths,
+        load=get_load_paths_gegis("data", config, check_existence=True),
```
davide-f: Good :D
Then the load_data_paths object may be deleted (on top of the snakefile)

davide-f (commented):
Great @ekatef :D
The functionality is absolutely there!

Need to check why the CI is not working though

As minor comments:
- plase revise the release_notes
- it may be good to add something in the documentation to clarify on how to feed in the custom data and leverage on this amazing contribution. I understand this is time demanding. not sure if you have the time to craft something very very basic to be later improved or we leave it completely for a next PR.

What do you think?

scripts/build_demand_profiles.py
davide-f: For some reason here there is a ValueError, can you reproduce it locally?

ekatef:
> Great @ekatef :D The functionality is absolutely there!
> 
> Need to check why the CI is not working though
> 
> As minor comments:
> 
> * plase revise the release_notes
> * it may be good to add something in the documentation to clarify on how to feed in the custom data and leverage on this amazing contribution. I understand this is time demanding. not sure if you have the time to craft something very very basic to be later improved or we leave it completely for a next PR.
> 
> What do you think?

Thanks a lot for reviewing @davide-f! 😄 I can confirm that the functionality has been tested successfully :) 🎉 

The reason for CI failing has been that Snakemake is parsing all the inputs before starting to do anything. So, it didn't help much to place `get_load_paths_gegis()` in the inputs of `build_demand_profiles`, as Snakemake still executed `get_load_paths_gegis()` before loading the databundle when running the tutorial case. That leads to failure along the workflow, if a new modification of `get_load_paths_gegis` is used as it captures only the paths which really exist of the drive.

To fix this, I had to put back composing of `load_data_paths`, but add a condition to treat the paths differently, depending on the actual presence of the load data. Not sure it's the most elegant approach but seems to be an effective one.

Done for a release note :D Have also drafted a documentation on using custom load profiles. Could you please check if [this addition](<https://pypsa-earth--995.org.readthedocs.build/en/995/customization_basic1.html#use-custom-demand-data>) looks clear enough to be used? 

PS Can't help but say that it's amazing to have a documentation build as a part of CI 💚 Thanks a lot to you and @GbotemiB for adding this feature!

Snakefile
```diff
@@ -43,7 +43,12 @@ run = config.get("run", {})
 RDIR = run["name"] + "/" if run.get("name") else ""
 CDIR = RDIR if not run.get("shared_cutouts") else ""
 
-load_data_paths = get_load_paths_gegis("data", config)
+if os.path.exists(os.path.join("data", config["load_options"]["ssp"])):
+    load_data_paths = get_load_paths_gegis("data", config, check_existence=True)
+else:
+    # demand profiles are not available yet -> no opportunity to use custom load files
+    load_data_paths = get_load_paths_gegis("data", config, check_existence=False)
```
davide-f: Mmmm I think the check_existence can lead to misleading effects due to when that is evaluated.
I think we should not have this option and keep the previous behavior maybe?

The reason is that the file may not exist when the function is triggered.
For example, with a fresh run, the file does not exists at the beginning because it is moved here from retrieve_databundle.

Maybe the check on whether the file is missing should be moved when the files are actually read, but if the file is missing, the workflow breaks so it may not be needed.
Except the check on whether all selected countries are found, I expected that check to be there already but may be good to crosscheck if you wanted to address that here

davide-f (commented):
Cool, the CI is running :D so closer to merge :)

Snakefile
ekatef: Thanks for double-checking @davide-f 🙂 This file-existence issue is quite a tricky part, in fact. 

The point is snakemake parses all the input paths before building DAG and starting to execute anything. Which means that `load_data_paths` is evaluated before any of the rule would run, including `retrieve_databundle`. So, it's possible that `load_data_paths` is evaluated when demand files are not loaded yet. It has been perfectly alright before, when we were not interested if the demand files exist. But the demand parser must check which exactly inputs present, and breaks if there is no any inputs yet.

If we would restore the original `get_load_paths_gegis` removing the check-existence condition, that will lead to the error in a fresh run. That has been a reason of CI being unhappy previously, which I have looked to fix. Not sure the current implementation is the most elegant one, but it works and hopefully doesn't introduce breaking changes into the workflow.

Agree that it can also work if we would check if the data folder exist directly in the `load_data_paths` (an example of [a similar solution](https://stackoverflow.com/q/46636844/8465924)). Have implemented and tested the improvement and happy to have your opinion on that 🙂

davide-f: Ahhhh right, good point! I catched the issue but not completely the side effects, great explanation.
I'm not a great fan of that long if case; I got another idea; adding a review comment and let's see what's your idea.
Already this is good :)

scripts/build_demand_profiles.py
```diff
@@ -106,16 +107,41 @@ def get_load_paths_gegis(ssp_parentfolder, config):
...
+        avail_regions = [
+            os.path.split(os.path.abspath(pth))[1].split(".nc")[0] for pth in load_paths
+        ]
+
+        logger.info(f" An assumed load folder {load_dir}, load path is {load_paths}.")
+
+        if len(load_paths) == 0:
+            logger.warning(
+                f"No demand data file for {set(region_load).difference(avail_regions)}. An assumed load folder {load_dir}."
+            )
```
davide-f: What if we drop the exists of scenario path and we do:
```suggestion
        regions_found = []
        file_names = []
        for continent in region_load:
            sel_ext = ".nc"
            for ext in [".nc", ".csv"]:
                load_path = os.path.join(str(load_dir), str(continent) + str(ext))
                if os.path.exists(load_path):
                    sel_ext = ext
                    regions_found.append(continent)
                    break
             file_name = str(continent) + str(sel_ext)
             load_path = os.path.join(str(load_dir), file_name)
             load_paths.append(load_path)
             file_names.append(file_name)
        
        logger.info(
            f"Demand data folder: {load_dir}, load path is {load_paths}.\n" +
            f"Expected files: " + "; ".join(file_names)
        )
```
If the file does not exist, this solution should backup to the .nc case.
Not sure about how much info to disclose;
I'm unsure it is good to advice about whether demand data files are existing or not: for the standard user that may be confusing because of the ordering about when snakemake makes this evaluation.

What do you think?

ekatef: I love the solution! 😄 Implemented. Thanks a lot for that.

Agree that the no-demand warning feels redundant. Moreover, it's also not needed: in case no demand data provided and `retrive_databundle: false`, Snakemake just throws an error even before running anything. So, we can just remove this checking part.

ekatef:
> Cool, the CI is running :D so closer to merge :)

@davide-f thanks :D Have implemented your suggestions, and the changes look now quite consistent with the original version which feels really great 🙃 Thanks a lot for reviewing!

Both local and CI testing are successful now. 

Could you please check if everything looks well in the revised version?

davide-f (approved):

Issue #362, opened 2022-06-02
Allow higher capacity_per_sqkm for solar technology in deserts


pz-max:
## Allow higher capacity_per_sqkm for solar technology in deserts

We are currently defining how much solar can be installed in a region by a parameter in the config:
https://github.com/pypsa-meets-africa/pypsa-africa/blob/d86fed5a6dee9e6e55b9ceed6ef20042ce6fab16/config.default.yaml#L220 
In another issue https://github.com/pypsa-meets-africa/pypsa-africa/issues/361 we compare the assumption and investigate the impact to the technical potential.

The parameter was chosen to represent socio-technical and political limits. The current value in PyPSA-Eur is **1.7 MW/sqkm**. The average power density of utility-scale solar power plants is around **45.6 MW/sqkm** (confirmed by looking at 41 power plants in a notebook that we upload soon). Maximal observed densities are even around **100sqkm**.  PyPSA-Eur's current default says that roughly **3.7%** (~1.7/0.037=46) of all available areas (which already exclude protected areas and specific [landcover codes](https://github.com/pypsa-meets-africa/pypsa-africa/blob/d86fed5a6dee9e6e55b9ceed6ef20042ce6fab16/config.default.yaml#L228)) can be used for capacity expansion.

Obviously, one can criticize if **3.7%** of all available areas is really the socio-technical and political limit or not. We will actually be a bit optimistic and assume that **10%**, representing **4.6 MW/sqkm**, of all available areas can be used for solar expansion (new solar design also allows synergies between farm & cropland reducing such limitations, see [agrivoltaics](https://www.youtube.com/watch?v=lgZBlD-TCFE)).

But is this socio-technical and political assumption applicable to deserts (Copernicus LC class 60)? I think no. I would argue that we should do the following:
- Using for different land-cover classes different capacity_per_sqkm assumptions. Here, I think the main impact in Africa would be to use the full average power density for deserts (**45.6 MW/sqkm**). There are simply no socio-techno and political limitations...

A pull request should update the solar capacity_per_sqkm for Copernicus class 60 (e.g. desert). The code snippet that might be relevant:
https://github.com/pypsa-meets-africa/pypsa-africa/blob/d86fed5a6dee9e6e55b9ceed6ef20042ce6fab16/scripts/build_renewable_profiles.py#L346-L353

*Please give a clear and concise description and provide context why the feature would be useful.*
*Also, we'd appreciate any implementation ideas and references you already have.*

[...]

Issue #361, opened 2022-06-02
Adapt installable solar power density


pz-max:
We could think about adjusting some parameters for solar energy. Currently, we use 1.7MW/km2 which leads to a socio-technical potential of 45 PWh which is much less than the 2014 reported technical potential from IRENA with 660 PWh. I would even argue that the IRENA value is quite outdated, since the technology developed (higher capacity densities). This means the 660 PWh would be more something like 1000 PWh. 

We are currently defining how much solar can be installed in a region by a parameter in the config:
https://github.com/pypsa-meets-africa/pypsa-africa/blob/d86fed5a6dee9e6e55b9ceed6ef20042ce6fab16/config.default.yaml#L220 

The parameter was chosen to represent socio-technical and political limits. The current value in PyPSA-Eur is **1.7 sqkm**. The average power density of utility-scale solar power plants is around **45.6 MW/sqkm** (confirmed by looking at 41 power plants in a notebook that we upload soon). Maximal observed densities are even around **100 MW/sqkm**.  PyPSA-Eur's current default says that roughly **3.7%** (~1.7/0.037=46) of all available areas (which already exclude protected areas and specific [landcover codes](https://github.com/pypsa-meets-africa/pypsa-africa/blob/d86fed5a6dee9e6e55b9ceed6ef20042ce6fab16/config.default.yaml#L228)) can be used for capacity expansion.

Obviously, one can criticize if **3.7%** of all available areas is really the socio-technical and political limit or not. We intend to be a bit more optimistic and assume that **10%**, representing **4.6 MW/sqkm**, of all available areas can be used for solar expansion (a reason is that new solar design also allows synergies between farm & cropland reducing such limitations, see [agrivoltaics](https://www.youtube.com/watch?v=lgZBlD-TCFE) and that the power density is gradually increasing).
 
Comparison on power densities and technical potential.
![image](https://user-images.githubusercontent.com/61968949/171592556-079314bd-e4b7-4243-bd35-5d6c5ae51e1c.png)

pz-max:
We now add also a comparison to a solar powerplant list:
My suggestion is to calibrate using a dataset like this ( https://en.m.wikipedia.org/wiki/List_of_photovoltaic_power_stations ) Basically the capacity per square kilometre shall account also for non-technical issues like space needed for maintenance purposes and alike.

Looking at 40 utility scale powerplants the average power density is about 45 MW/km2 (with extremes at about 115 MW/km2)

[...]

PR #997, opened 2024-03-22
CSP integration into PyPSA-Earth Workflow

### LINKED ISSUE #145: Implement csp technology

GbotemiB:
# Closes #145 (if applicable).

This PR aims to integrate the CSP technology into the PyPSA-Earth Workflow

## Checklist

### DIFF
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -64,7 +64,8 @@ cluster_options:
     remove_stubs: true
     remove_stubs_across_borders: true
     p_threshold_drop_isolated: 20 # [MW] isolated buses are being discarded if bus mean power is below the specified threshold
-    p_threshold_merge_isolated: 300 # [MW] isolated buses are being merged into a single isolated bus if bus mean power is below the specified threshold
+    p_threshold_merge_isolated: 300 # [MW] isolated buses are being merged into a single isolated bus if a bus mean power is below the specified threshold
+    s_threshold_fetch_isolated: 0.05 # [-] a share of the national load for merging an isolated network into a backbone network
   cluster_network:
     algorithm: kmeans
     feature: solar+onwind-time
@@ -126,7 +127,7 @@ load_options:
 
 electricity:
   base_voltage: 380.
-  voltages: [220., 300., 380.]
+  voltages: [132., 220., 300., 380., 500., 750.]
   co2limit: 7.75e+7  # European default, 0.05 * 3.1e9*0.5, needs to be adjusted for Africa
   co2base: 1.487e+9  # European default, adjustment to Africa necessary
   agg_p_nom_limits: data/agg_p_nom_minmax.csv
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -168,11 +168,15 @@ electricity:
       PV: [solar]
 
 lines:
-  types:
+  ac_types:
+    132.: "243-AL1/39-ST1A 20.0"
     220.: "Al/St 240/40 2-bundle 220.0"
     300.: "Al/St 240/40 3-bundle 300.0"
     380.: "Al/St 240/40 4-bundle 380.0"
-  dc_type: "HVDC XLPE 1000"
+    500.: "Al/St 240/40 4-bundle 380.0"
+    750.: "Al/St 560/50 4-bundle 750.0"
+  dc_types:
+    500.: "HVDC XLPE 1000"
   s_max_pu: 0.7
   s_nom_max: .inf
   length_factor: 1.25
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -154,7 +154,7 @@ electricity:
   custom_powerplants: false  #  "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
 
   conventional_carriers: [nuclear, oil, OCGT, CCGT, coal, lignite, geothermal, biomass]
-  renewable_carriers: [solar, csp, onwind, offwind-ac, offwind-dc, hydrox]
+  renewable_carriers: [solar, csp, onwind, offwind-ac, offwind-dc, hydro]
 
   estimate_renewable_capacities:
     stats: "irena"  # False, = greenfield expansion, 'irena' uses IRENA stats to add expansion limits
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -328,6 +328,7 @@ renewable:
     potential: simple # or conservative
     clip_p_max_pu: 1.e-2
     extendable: true
+    architecture: advanced # simple or advanced
 
 # TODO: Needs to be adjusted for Africa.
 # Costs Configuration (Do not remove, needed for Sphynx documentation).
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -145,16 +145,16 @@ electricity:
     H2: 168
 
   extendable_carriers:
-    Generator: [solar, onwind, offwind-ac, offwind-dc, OCGT, csp]
+    Generator: [solar, csp, onwind, offwind-ac, offwind-dc, OCGT]
     StorageUnit: []  # battery, H2
-    Store: [battery, H2, csp]
+    Store: [battery, H2]
     Link: []  # H2 pipeline
 
   powerplants_filter: (DateOut >= 2022 or DateOut != DateOut)
   custom_powerplants: false  #  "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
 
   conventional_carriers: [nuclear, oil, OCGT, CCGT, coal, lignite, geothermal, biomass]
-  renewable_carriers: [solar, onwind, offwind-ac, offwind-dc, hydro, csp]
+  renewable_carriers: [solar, csp, onwind, offwind-ac, offwind-dc, hydrox]
 
   estimate_renewable_capacities:
     stats: "irena"  # False, = greenfield expansion, 'irena' uses IRENA stats to add expansion limits
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -145,7 +145,7 @@ electricity:
     H2: 168
 
   extendable_carriers:
-    Generator: [solar, csp-tower, onwind, offwind-ac, offwind-dc, OCGT]
+    Generator: [solar, csp, onwind, offwind-ac, offwind-dc, OCGT]
     StorageUnit: []  # battery, H2
     Store: [battery, H2]
     Link: []  # H2 pipeline
@@ -488,7 +488,7 @@ plotting:
     "DC-DC": "#8a1caf"
     "DC link": "#8a1caf"
     "load": "#ff0000"
-    "csp-tower": "#fdd404"
+    "csp": "#fdd404"
   nice_names:
     OCGT: "Open-Cycle Gas"
     CCGT: "Combined-Cycle Gas"
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -146,7 +146,7 @@ electricity:
     H2: 168
 
   extendable_carriers:
-    Generator: [solar, csp, onwind, offwind-ac, offwind-dc, OCGT]
+    Generator: [solar, onwind, offwind-ac, offwind-dc, OCGT]
     StorageUnit: []  # battery, H2
     Store: [battery, H2]
     Link: []  # H2 pipeline
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -484,7 +484,7 @@ plotting:
     "DC-DC": "#8a1caf"
     "DC link": "#8a1caf"
     "load": "#ff0000"
-    "csp": "#fdd404"
+    "csp-tower": "#fdd404"
   nice_names:
     OCGT: "Open-Cycle Gas"
     CCGT: "Combined-Cycle Gas"
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -145,16 +145,16 @@ electricity:
     H2: 168
 
   extendable_carriers:
-    Generator: [solar, onwind, offwind-ac, offwind-dc, OCGT]
+    Generator: [solar, onwind, offwind-ac, offwind-dc, OCGT, csp]
     StorageUnit: []  # battery, H2
-    Store: [battery, H2]
+    Store: [battery, H2, csp]
     Link: []  # H2 pipeline
 
   powerplants_filter: (DateOut >= 2022 or DateOut != DateOut)
   custom_powerplants: false  #  "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
 
   conventional_carriers: [nuclear, oil, OCGT, CCGT, coal, lignite, geothermal, biomass]
-  renewable_carriers: [solar, onwind, offwind-ac, offwind-dc, hydro]
+  renewable_carriers: [solar, onwind, offwind-ac, offwind-dc, hydro, csp]
 
   estimate_renewable_capacities:
     stats: "irena"  # False, = greenfield expansion, 'irena' uses IRENA stats to add expansion limits
@@ -482,7 +482,8 @@ plotting:
     "HVDC links": "#8a1caf"
     "DC-DC": "#8a1caf"
     "DC link": "#8a1caf"
-    "load": "#FF0000"
+    "load": "#ff0000"
+    "csp": "#fdd404"
   nice_names:
     OCGT: "Open-Cycle Gas"
     CCGT: "Combined-Cycle Gas"
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -308,6 +308,26 @@ renewable:
       method: hydro_capacities  # 'hydro_capacities' to rescale country hydro production by using hydro_capacities, 'eia' to rescale by eia data, false for no rescaling
       year: 2013  # (optional) year of statistics used to rescale the runoff time series. When not provided, the weather year of the snapshots is used
     multiplier: 1.1  # multiplier applied after the normalization of the hydro production; default 1.0
+  csp:
+    cutout: cutout-2013-era5-tutorial
+    resource:
+      method: csp
+      installation: SAM_solar_tower
+    capacity_per_sqkm: 4.6 # From 1.7 to 4.6 addresses issue #361
+    # Determined by comparing uncorrected area-weighted full-load hours to those
+    # published in Supplementary Data to
+    # Pietzcker, Robert Carl, et al. "Using the sun to decarbonize the power
+    # sector: The economic potential of photovoltaics and concentrating solar
+    # power." Applied Energy 135 (2014): 704-720.
+    correction_factor: 0.854337
+    copernicus:
+      grid_codes: [20, 30, 40, 60, 90]
+      distancing_codes: [50]
+      distance_to_codes: 3000
+    natura: true
+    potential: simple # or conservative
+    clip_p_max_pu: 1.e-2
+    extendable: true
 
 # TODO: Needs to be adjusted for Africa.
 # Costs Configuration (Do not remove, needed for Sphynx documentation).
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -145,7 +145,7 @@ electricity:
     H2: 168
 
   extendable_carriers:
-    Generator: [solar, csp, onwind, offwind-ac, offwind-dc, OCGT]
+    Generator: [solar, csp-tower, onwind, offwind-ac, offwind-dc, OCGT]
     StorageUnit: []  # battery, H2
     Store: [battery, H2]
     Link: []  # H2 pipeline
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -332,7 +332,7 @@ renewable:
     potential: simple # or conservative
     clip_p_max_pu: 1.e-2
     extendable: true
-    architecture: advanced # simple or advanced
+    csp_model: advanced # simple or advanced
 
 # TODO: Needs to be adjusted for Africa.
 # Costs Configuration (Do not remove, needed for Sphynx documentation).
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -309,7 +309,7 @@ renewable:
       year: 2013  # (optional) year of statistics used to rescale the runoff time series. When not provided, the weather year of the snapshots is used
     multiplier: 1.1  # multiplier applied after the normalization of the hydro production; default 1.0
   csp:
-    cutout: cutout-2013-era5-tutorial
+    cutout: cutout-2013-era5
     resource:
       method: csp
       installation: SAM_solar_tower
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -318,7 +318,7 @@ renewable:
     resource:
       method: csp
       installation: SAM_solar_tower
-    capacity_per_sqkm: 4.6 # From 1.7 to 4.6 addresses issue #361
+    capacity_per_sqkm: 2.392 # From 1.7 to 4.6 addresses issue #361
     # Determined by comparing uncorrected area-weighted full-load hours to those
     # published in Supplementary Data to
     # Pietzcker, Robert Carl, et al. "Using the sun to decarbonize the power
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -323,7 +323,6 @@ renewable:
     # Pietzcker, Robert Carl, et al. "Using the sun to decarbonize the power
     # sector: The economic potential of photovoltaics and concentrating solar
     # power." Applied Energy 135 (2014): 704-720.
-    correction_factor: 0.854337
     copernicus:
       grid_codes: [20, 30, 40, 60, 90]
       distancing_codes: [50]
--- a/data/costs.csv
+++ b/data/costs.csv
@@ -1,204 +1,204 @@
 technology,year,parameter,value,unit,source
-CCGT,2030,FOM,2.5,%/year,DIW DataDoc http://hdl.handle.net/10419/80348
-CCGT,2030,VOM,4.0,EUR/MWhel,DIW DataDoc http://hdl.handle.net/10419/80348
-CCGT,2030,efficiency,0.5,per unit,DIW DataDoc http://hdl.handle.net/10419/80348
-CCGT,2030,investment,800.0,EUR/kWel,DIW DataDoc http://hdl.handle.net/10419/80348
-CCGT,2030,lifetime,30.0,years,IEA2010
-DAC,2030,FOM,4.0,%/year,Fasihi
-DAC,2030,investment,250.0,EUR/(tCO2/a),Fasihi/Climeworks
-DAC,2030,lifetime,30.0,years,Fasihi
-H2 pipeline,2030,FOM,5.0,%/year,Krieg2012 http://juser.fz-juelich.de/record/136392/files/Energie%26Umwelt_144.pdf
-H2 pipeline,2030,efficiency,0.98,per unit,Krieg2012 http://juser.fz-juelich.de/record/136392/files/Energie%26Umwelt_144.pdf
-H2 pipeline,2030,investment,267.0,EUR/MW/km,Welder et al https://doi.org/10.1016/j.ijhydene.2018.12.156
-H2 pipeline,2030,lifetime,40.0,years,Krieg2012 http://juser.fz-juelich.de/record/136392/files/Energie%26Umwelt_144.pdf
-HVAC overhead,2030,FOM,2.0,%/year,Hagspiel
-HVAC overhead,2030,investment,400.0,EUR/MW/km,Hagspiel
-HVAC overhead,2030,lifetime,40.0,years,Hagspiel
-HVDC inverter pair,2030,FOM,2.0,%/year,Hagspiel
-HVDC inverter pair,2030,investment,150000.0,EUR/MW,Hagspiel
-HVDC inverter pair,2030,lifetime,40.0,years,Hagspiel
-HVDC overhead,2030,FOM,2.0,%/year,Hagspiel
-HVDC overhead,2030,investment,400.0,EUR/MW/km,Hagspiel
-HVDC overhead,2030,lifetime,40.0,years,Hagspiel
-HVDC submarine,2030,FOM,2.0,%/year,Hagspiel
-HVDC submarine,2030,investment,2000.0,EUR/MW/km,DTU report based on Fig 34 of https://ec.europa.eu/energy/sites/ener/files/documents/2014_nsog_report.pdf
-HVDC submarine,2030,lifetime,40.0,years,Hagspiel
-OCGT,2030,FOM,3.75,%/year,DIW DataDoc http://hdl.handle.net/10419/80348
-OCGT,2030,VOM,3.0,EUR/MWhel,DIW DataDoc http://hdl.handle.net/10419/80348
-OCGT,2030,efficiency,0.39,per unit,DIW DataDoc http://hdl.handle.net/10419/80348
-OCGT,2030,investment,400.0,EUR/kWel,DIW DataDoc http://hdl.handle.net/10419/80348
-OCGT,2030,lifetime,30.0,years,IEA2010
-PHS,2030,FOM,1.0,%/year,DIW DataDoc http://hdl.handle.net/10419/80348
-PHS,2030,efficiency,0.75,per unit,DIW DataDoc http://hdl.handle.net/10419/80348
-PHS,2030,investment,2000.0,EUR/kWel,DIW DataDoc http://hdl.handle.net/10419/80348
-PHS,2030,lifetime,80.0,years,IEA2010
-battery inverter,2030,FOM,3.0,%/year,budischak2013
-battery inverter,2030,efficiency,0.9,per unit charge/discharge,budischak2013; Lund and Kempton (2008) http://dx.doi.org/10.1016/j.enpol.2008.06.007
-battery inverter,2030,investment,411.0,USD/kWel,budischak2013
-battery inverter,2030,lifetime,20.0,years,budischak2013
-battery storage,2030,investment,192.0,USD/kWh,budischak2013
-battery storage,2030,lifetime,15.0,years,budischak2013
-biomass,2030,FOM,4.526935,%/year,DIW DataDoc http://hdl.handle.net/10419/80348
-biomass,2030,efficiency,0.468,per unit,DIW DataDoc http://hdl.handle.net/10419/80348
-biomass,2030,fuel,7.0,EUR/MWhth,IEA2011b
-biomass,2030,investment,2209.0,EUR/kWel,DIW DataDoc http://hdl.handle.net/10419/80348
-biomass,2030,lifetime,30.0,years,ECF2010 in DIW DataDoc http://hdl.handle.net/10419/80348
-central CHP,2030,FOM,3.0,%/year,HP
-central CHP,2030,investment,650.0,EUR/kWel,HP
-central CHP,2030,lifetime,25.0,years,HP
-central air-sourced heat pump,2030,FOM,3.5,%/year,Palzer thesis
-central air-sourced heat pump,2030,efficiency,3.0,per unit,default for costs
-central air-sourced heat pump,2030,investment,700.0,EUR/kWth,Palzer thesis
-central air-sourced heat pump,2030,lifetime,20.0,years,Palzer thesis
-central gas boiler,2030,FOM,1.0,%/year,Palzer thesis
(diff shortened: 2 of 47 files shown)


config.default.yaml
```diff
@@ -308,6 +308,26 @@ renewable:
       method: hydro_capacities  # 'hydro_capacities' to rescale country hydro production by using hydro_capacities, 'eia' to rescale by eia data, false for no rescaling
       year: 2013  # (optional) year of statistics used to rescale the runoff time series. When not provided, the weather year of the snapshots is used
     multiplier: 1.1  # multiplier applied after the normalization of the hydro production; default 1.0
+  csp:
+    cutout: cutout-2013-era5-tutorial
```
davide-f: This is the default configuration file; it should be "cutout-2013-era5"

davide-f (commented):
Littel preliminary comment :)
It sounds nice!
Does this PR generate the csp time series?

config.default.yaml
GbotemiB: Thanks for pointing this out

GbotemiB:
By CSP time series, do you mean if a renewable profile network file is being created for CSP?

If so, yes. At the moment, I can confirm that a CSP renewable profile network is created alongside other renewable profiles.

davide-f:
> By CSP time series, do you mean if a renewable profile network file is being created for CSP?
> 
> If so, yes. At the moment, I can confirm that a CSP renewable profile network is created alongside other renewable profiles.

Cool great work 👯‍♂️

ekatef:
Hello @GbotemiB! Great progress 😄

A couple of technical comments:
1) It looks like `costs.csv` has been modified more deeply as it has been probably intended. Could you please fix that?
2) Regarding the costs inputs, [this part](<https://github.com/PyPSA/technology-data/blob/6e5066bfd972ff8041891b04899c138d54da55a0/inputs/manual_input.csv#L242-L277>) of `technology-data` repo seems to be relevant. But there a need to fetch data for `2030` to match with PyPSA-Earth conventions for `data/costs.csv`.

scripts/add_electricity.py
```diff
@@ -385,6 +385,38 @@ def attach_wind_and_solar(
...
+                csp_buses_i = n.madd(
+                    "Bus",
+                    ds.indexes["bus"] + " " + tech,
+                    carrier=tech,
+                    unit="MW",
+                )
+
+                n.madd(
+                    "Store",
+                    csp_buses_i,
```
ekatef: The workflow expects to have here an index-like object, while `csp_bises_i` have been defined earlier as `PyPSA.Network` object which can lead to issues.

scripts/add_electricity.py
```diff
@@ -385,6 +385,38 @@ def attach_wind_and_solar(
                 efficiency=costs.at[suptech, "efficiency"],
             )
 
+            if tech == "csp":  # add store and link for CSP
+                csp_buses_i = n.madd(
+                    "Bus",
+                    ds.indexes["bus"] + " " + tech,
+                    carrier=tech,
+                    unit="MW",
+                )
```
ekatef: Is there a particular need for assignment? My feeling is that `n.madd()` should work fine to add the the buses to a network.

GbotemiB: I picked it up from the `extra_components` scripts. It seems to be leading to issues with attaching links to the network.

scripts/add_electricity.py
GbotemiB: Thanks for the clarity. I used indexes but switched them at some point. I would change It back.

GbotemiB:
Hi @ekatef, regarding the data in [this part](https://github.com/PyPSA/technology-data/blob/6e5066bfd972ff8041891b04899c138d54da55a0/inputs/manual_input.csv#L242-L277) which you mentioned. Is there an already existing method used to fetch this data into the PyPSA-Earth Workflow?

data/costs.csv
```diff
@@ -71,15 +71,15 @@ coal,2030,efficiency,0.464,per unit,DIW DataDoc http://hdl.handle.net/10419/8034
...
-csp,2030,FOM,1.1,%/year,ATB CSP data (https://atb.nrel.gov/electricity/2021/concentrating_solar_power)
-csp,2030,investment,108.37,"EUR/kW_th,dp",ATB CSP data (https://atb.nrel.gov/electricity/2021/concentrating_solar_power) and NREL SAM v2021.12.2 (https://sam.nrel.gov/).
-csp,2030,lifetime,30.0,years,ATB CSP data (https://atb.nrel.gov/electricity/2021/concentrating_solar_power)
-csp TES,2030,FOM,1.1,%/year,see solar-tower.
-csp TES,2030,investment,14.52,EUR/kWh_th,ATB CSP data (https://atb.nrel.gov/electricity/2021/concentrating_solar_power) and NREL SAM v2021.12.2 (https://sam.nrel.gov/).
-csp TES,2030,lifetime,30.0,years,see solar-tower.
-csp power block,2030,FOM,1.1,%/year,see solar-tower.
-csp power block,2030,investment,759.17,EUR/kW_e,ATB CSP data (https://atb.nrel.gov/electricity/2021/concentrating_solar_power) and NREL SAM v2021.12.2 (https://sam.nrel.gov/).
-csp power block,2030,lifetime,30.0,years,see solar-tower.
+csp-tower,2030,FOM,1.1,%/year,ATB CSP data (https://atb.nrel.gov/electricity/2021/concentrating_solar_power)
```
GbotemiB, line 196: I have renamed the csp rows here to the default from the technology repo in PyPSA.

scripts/add_electricity.py
```diff
@@ -385,6 +385,36 @@ def attach_wind_and_solar(
...
+                available_csp_buses_i = n.buses.index[
+                    n.buses.index.isin(ds.indexes["bus"])
+                ]
+
+                n.madd(
+                    "Link",
+                    ds.indexes["bus"],
+                    " " + tech,
+                    bus0=ds.indexes["bus"],
+                    bus1=available_csp_buses_i,
```
ekatef: Not sure I get it: aren't length values of `ds.indexes["bus"]` and `available_csp_buses_i` different?

GbotemiB: The length for `ds.indexes["bus"]` is 161, while the length of `n.buses.indexes` is 203. Using n.buses.indexes results in a broadcast error, so I first filter for only indexes available in both buses before applying it into the link.

ekatef:
@GbotemiB perfect, thanks a lot for the fix for `costs.csv`.

Have checked `add_extra_components` and it looks like are are absolutely right about the type of `n.mapp()` returns: [that is in fact an index](https://github.com/PyPSA/PyPSA/blob/4298c23ed625fa3e0e4cecfb84335b27b76e2ac6/pypsa/components.py#L1005-L1034). Not sure that is the safest approach, but looks quite elegant. 

Great that you have found a workaround for the buses specification. Probably, we can improve it a bit, and that is a moment where out drafting work may get useful. Currently, we have electrical buses only, while there is a need to have also heat buses. Which means the need to add a set of buses like it's done in `add_extra_components`: https://github.com/pypsa-meets-earth/pypsa-earth/blob/d3d718b8e22fab344991a73a9f9637bf6f22ecb4/scripts/add_extra_components.py#L108-L112

Regarding the underwater fraction, that is a parameter which needs to be defined for submarine HVDC lines. `Link` component may be used to represent HVDC lines in the network. There are also links which connect the stores, but those are added after `simplify_network`. To me, it looks like we need to implement a method to exclude links corresponding to CSP from treatment in this part: https://github.com/pypsa-meets-earth/pypsa-earth/blob/d3d718b8e22fab344991a73a9f9637bf6f22ecb4/scripts/simplify_network.py#L439-L441 As a suggestion, filtering by carrier may be a good approach. What do you think?

GbotemiB:
Thanks @ekatef for the review. 

Question regarding the electrical buses, are the buses in the renewable profile network an electrical bus? 
I thought the buses from the renewable_profile will be as a result of the profile type, in this case heat for the csp profile.

ekatef:
> Thanks @ekatef for the review.
> 
> Question regarding the electrical buses, are the buses in the renewable profile network an electrical bus? I thought the buses from the renewable_profile will be as a result of the profile type, in this case heat for the csp profile.

Buses used in `build_renewable_profiles` are `bus_regions`: https://github.com/pypsa-meets-earth/pypsa-earth/blob/d3d718b8e22fab344991a73a9f9637bf6f22ecb4/Snakefile#L457-L461

These regions are built in `build_bus_regions` as Voronoi polygons around nodes of power grid (or as administrative regions if alternative clustering is considered): https://github.com/pypsa-meets-earth/pypsa-earth/blob/d3d718b8e22fab344991a73a9f9637bf6f22ecb4/scripts/build_bus_regions.py#L198-L207

So, buses are a mapping of the power grid nodes into space. As an example of the bus polygons overlaid by the network nodes:

image

Each polygon aka bus area is used to collect the available renewable potential to simulate a renewable generator which can be commissioned at a corresponding grid node.

Does it clarify? Buses are one of the major concepts we are using for modeling, but not sure our current explanations are clear enough. So, feel free absolutely to ask further questions if needed.

GbotemiB:
Thanks @ekatef. Much clearer.

GbotemiB:
Hello @ekatef, I have added the heat buses. I defined the carrier for the heat bus as csp.

GbotemiB:
In the `simplify_network.py`, it seems inserting the links before simplify_electricity triggers other issues.
```bash
  File "/Users/gbotemi/Documents/code/PYPSA/CSP/pypsa-earth/.snakemake/scripts/tmpybgzrdcz.simplify_network.py", line 890, in <module>
    n, simplify_links_map = simplify_links(
  File "/Users/gbotemi/Documents/code/PYPSA/CSP/pypsa-earth/.snakemake/scripts/tmpybgzrdcz.simplify_network.py", line 413, in simplify_links
    for b, buses, dc_edges in split_links(labels.index[labels == lbl]):
  File "/Users/gbotemi/Documents/code/PYPSA/CSP/pypsa-earth/.snakemake/scripts/tmpybgzrdcz.simplify_network.py", line 380, in split_links
    if m not in nodes or m in seen or contains_ac(ls):
  File "/Users/gbotemi/Documents/code/PYPSA/CSP/pypsa-earth/.snakemake/scripts/tmpybgzrdcz.simplify_network.py", line 321, in contains_ac
    return any(list(map(lambda x: is_ac_branch(x), ls)))
  File "/Users/gbotemi/Documents/code/PYPSA/CSP/pypsa-earth/.snakemake/scripts/tmpybgzrdcz.simplify_network.py", line 321, in <lambda>
    return any(list(map(lambda x: is_ac_branch(x), ls)))
  File "/Users/gbotemi/Documents/code/PYPSA/CSP/pypsa-earth/.snakemake/scripts/tmpybgzrdcz.simplify_network.py", line 315, in is_ac_branch
    return n.links.loc[x[1]].dc != True
  File "/Users/gbotemi/miniconda3/envs/pypsa-earth-osm/lib/python3.10/site-packages/pandas/core/generic.py", line 6296, in __getattr__
    return object.__getattribute__(self, name)
AttributeError: 'Series' object has no attribute 'dc'
```

GbotemiB:
## A short summary of discussion on CSP task with @ekatef and @davide-f .
The CSP task would involve adding links, stores, and generators to the csp bus, which would be connected to the network.
But the current PyPSA Earth Workflow currently adds links and stores in the add_extra_components script while it adds just a generator in the add_electricity script.

There are currently two approaches to this. 
1. Add everything related to the csp in the add_electricity script. Going through this process will lead to a lot of adjustments of simplify_network like adding dc column to CSP links or filtering links by carrier. This will lead to a lot of changes in simplify_network script because it will disrupt the workflow.
2. Add the generators for the csp as electrical generators in the add_electricity script, then in add_extra_components script, switch the electrical generators to csp generators and also add the links and stores for csp.

To implement point 2, we can consider implementation in two steps, when first we add a simplified representation to add_electricity, and after improve it, also adding links and stores in add_extra_components.

ekatef:
Perfect @GbotemiB! Thanks for such a detailed account, and many thanks to @davide-f for the most insightful discussion.

As a little additional comment, we have also discussed that having `m` option for clustering means a need to have a dedicated bus for every CSP generator.

Citing the docs for `{clusters}` wildcard: `If an m is placed behind the number of clusters (e.g. 100m), generators are only moved to the clustered buses but not aggregated by carrier; i.e. the clustered bus may have more than one e.g. wind generator`. The implementation looks like: https://github.com/pypsa-meets-earth/pypsa-earth/blob/d3d718b8e22fab344991a73a9f9637bf6f22ecb4/scripts/cluster_network.py#L693-L694

scripts/add_electricity.py
```diff
@@ -368,54 +368,69 @@ def attach_wind_and_solar(
...
+                    weight=ds["weight"].to_pandas(),
+                    marginal_cost=costs.at[suptech, "marginal_cost"],
+                    capital_cost=capital_cost,
+                    efficiency=costs.at[suptech, "efficiency"],
+                )
 
-            if tech == "csp":  # add store and link for CSP
-                csp_buses_i = n.madd(
+            elif tech == "csp":  # add store and link for CSP
+                n.madd(
```
GbotemiB: Hi @davide-f and @katia,
Based on our last discussion to try to add CSP bus to network in `add_electricity`,  while doing this, I encountered an error with this section.

`ValueError: Length of values (42) does not match length of index (161)`
I investigated further. `ds["x"]` has 42 items, `ds["y"]` has 34 items, `ds["bus"]` has 161 items.

This is where I made mentioned that the cordinates were not tallying. I checked profiles for solar and onwind, it has similar values, it just that the workflow doesn't add them in `add_electricity`.

I think it is justifiable to go ahead with the second approach which is to add the generators first, then add the other components in `add_extra_components`.

What are you general thoughts on this?

ekatef: > Hi @davide-f and @Katia, Based on our last discussion to try to add CSP bus to network in `add_electricity`, while doing this, I encountered an error with this section.
> 
> `ValueError: Length of values (42) does not match length of index (161)` I investigated further. `ds["x"]` has 42 items, `ds["y"]` has 34 items, `ds["bus"]` has 161 items.
> 
> This is where I made mentioned that the cordinates were not tallying. I checked profiles for solar and onwind, it has similar values, it just that the workflow doesn't add them in `add_electricity`.
> 
> I think it is justifiable to go ahead with the second approach which is to add the generators first, then add the other components in `add_extra_components`.
> 
> What are you general thoughts on this?

Hello @GbotemiB! Great that you have localised the problem. So, we need to define `x` and `y` coordinates for buses ini some smart way.

Agree on the approach you suggest. I think it may be also more consistent as compared to adding links+stores in `add_electricity`.

GbotemiB:
Hello @davide-f @katia,

Is there a method in pypsa to reassign buses of network components?

I am open to thoughts on how to remove and attach csp buses to the electrical generators.

davide-f:
> Hello @davide-f @Katia,
> 
> Is there a method in pypsa to reassign buses of network components?
> 
> I am open to thoughts on how to remove and attach csp buses to the electrical generators.

You can do that by using:
`n.generators.loc[csp_index, "bus"] = new_csp_bus_vector`

For example:
`new_csp_bus_vector = csp_index + " csp"`

GbotemiB:
Hello @ekatef @davide-f,

Everything seems to work now. I also added option for `advanced or simple` in the config for csp.
I think the PR is ready for review.

GbotemiB:
I think the CI is failing because the cost file is missing csp rows

config.default.yaml
```diff
@@ -145,16 +145,16 @@ electricity:
     H2: 168
 
   extendable_carriers:
-    Generator: [solar, onwind, offwind-ac, offwind-dc, OCGT]
+    Generator: [solar, onwind, offwind-ac, offwind-dc, OCGT, csp]
```
ekatef: Probably, it could make sense to move `csp` in the list to have all the renewable carriers grouped together.

config.default.yaml
```diff
@@ -145,16 +145,16 @@ electricity:
...
     StorageUnit: []  # battery, H2
-    Store: [battery, H2]
+    Store: [battery, H2, csp]
```
ekatef: My feeling is that it should be `csp -> architecture` flag which triggers adding stores for CSP. So, no need to specify `csp` here.

config.default.yaml
```diff
@@ -145,16 +145,16 @@ electricity:
...
     Link: []  # H2 pipeline
 
   powerplants_filter: (DateOut >= 2022 or DateOut != DateOut)
   custom_powerplants: false  #  "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
 
   conventional_carriers: [nuclear, oil, OCGT, CCGT, coal, lignite, geothermal, biomass]
-  renewable_carriers: [solar, onwind, offwind-ac, offwind-dc, hydro]
+  renewable_carriers: [solar, onwind, offwind-ac, offwind-dc, hydro, csp]
```
ekatef: Would be great to re-arrange carriers.

config.default.yaml
```diff
@@ -308,6 +308,27 @@ renewable:
...
+    capacity_per_sqkm: 4.6 # From 1.7 to 4.6 addresses issue #361
+    # Determined by comparing uncorrected area-weighted full-load hours to those
+    # published in Supplementary Data to
+    # Pietzcker, Robert Carl, et al. "Using the sun to decarbonize the power
+    # sector: The economic potential of photovoltaics and concentrating solar
+    # power." Applied Energy 135 (2014): 704-720.
+    correction_factor: 0.854337
+    copernicus:
+      grid_codes: [20, 30, 40, 60, 90]
+      distancing_codes: [50]
```
ekatef: Could be great to cross-check the parameters and revise if needed to adjust according to `CSP` requirements.

scripts/add_extra_components.py
```diff
@@ -185,6 +185,40 @@ def attach_stores(n, costs, config):
             marginal_cost=costs.at["battery inverter", "marginal_cost"],
         )
 
+    if "csp" in config["renewable"].keys():
+        if (
+            config["renewable"]["csp"]["architecture"] == "advanced"
+            and "csp" in carriers
+        ):
```
ekatef: Nested `if`-s may be simplified to a single if with a joint condition.

scripts/add_extra_components.py
```diff
@@ -185,6 +185,40 @@ def attach_stores(n, costs, config):
...
+            csp_buses_i = n.madd("Bus", buses_i + " csp", carrier="csp", **bus_sub_dict)
+
+            # change bus of existing csp generators
+            old_csp_bus_vector = buses_i + " csp"
+            n.generators.loc[old_csp_bus_vector, "bus"] = csp_buses_i
```
ekatef: As we discussed, using `n.madd` output as an index is an elegant solution under the current implementation of PyPSA, but is risky in long-term perspective. I'd suggest to better use an index of `csp_buses_i` explicitly.

scripts/add_extra_components.py
```diff
@@ -232,7 +266,7 @@ def attach_hydrogen_pipelines(n, costs, config):
         from _helpers import mock_snakemake
 
         os.chdir(os.path.dirname(os.path.abspath(__file__)))
-        snakemake = mock_snakemake("add_extra_components", simpl="", clusters=10)
+        snakemake = mock_snakemake("add_extra_components", simpl="", clusters=6)
```
ekatef: What has been the reason on this change?

scripts/simplify_network.py
```diff
@@ -319,10 +319,10 @@ def simplify_links(
     exclude_carriers=[],
     aggregation_strategies=dict(),
 ):
-    ## Complex multi-node links are folded into end-points
+    # Complex multi-node links are folded into end-points
     logger.info("Simplifying connected link components")
 
-    if n.links.empty:
+    if n.links.empty or "csp" in renewable_config:
```
ekatef: Is it still needed in the updated implementation?

ekatef (changes requested):
Hello @GbotemiB! Thanks a lot for the discussion and agree that we are close to finalise :) Great work!

Added some technical comments. Could you please look into them?

Regarding CI failing, agree that something seems to get wrong in `_add_missing_carriers_from_costs` due to the missed `csp` technology in the cocsts data. Do you have any ideas on how to resolve it?

scripts/simplify_network.py
GbotemiB: We no longer need this. I will revert it to the default.

scripts/add_extra_components.py
GbotemiB: Oh, I was testing locally. I didnt intend to push this.

config.default.yaml
GbotemiB: We could have csp after solar.

ekatef: Agree with the idea. Feel free to implement it :)

config.default.yaml
```diff
@@ -154,7 +154,7 @@ electricity:
   custom_powerplants: false  #  "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
 
   conventional_carriers: [nuclear, oil, OCGT, CCGT, coal, lignite, geothermal, biomass]
-  renewable_carriers: [solar, onwind, offwind-ac, offwind-dc, hydro]
+  renewable_carriers: [solar, csp, onwind, offwind-ac, offwind-dc, hydro]
```
ekatef, line 158: I wonder if it's possible to use `csp-tower` everywhere, like we have different types of wind power? My feeling is that it would be more consistent.

config.default.yaml
```diff
@@ -308,6 +308,27 @@ renewable:
...
+    correction_factor: 0.854337
+    copernicus:
+      grid_codes: [20, 30, 40, 60, 90]
+      distancing_codes: [50]
+      distance_to_codes: 3000
+    natura: true
+    potential: simple # or conservative
+    clip_p_max_pu: 1.e-2
+    extendable: true
+    architecture: advanced # simple or advanced
```
ekatef: Could we adjust the naming to make more clear that this point relates to the code part, not to design of CSP itself? Like `implementation`, or `representation` or `csp_model`? Obviously, feel free to suggest other versions.

config.default.yaml
```diff
@@ -308,6 +308,27 @@ renewable:
...
+    resource:
+      method: csp
+      installation: SAM_solar_tower
+    capacity_per_sqkm: 4.6 # From 1.7 to 4.6 addresses issue #361
+    # Determined by comparing uncorrected area-weighted full-load hours to those
+    # published in Supplementary Data to
+    # Pietzcker, Robert Carl, et al. "Using the sun to decarbonize the power
+    # sector: The economic potential of photovoltaics and concentrating solar
+    # power." Applied Energy 135 (2014): 704-720.
+    correction_factor: 0.854337
```
ekatef: Would be great to cross-check these parameters for CSP, ideally both for solar power tower and parabolic trough. As discussed, starting points may be [NREL report ](https://www.osti.gov/biblio/1491726/) and [a fresh work ](http://dx.doi.org/10.2139/ssrn.4526901)on the global assessment of PV and CSP viability.

Totally agree on using [parameters from atlite](https://github.com/euronion/trace/blob/1b313271081a04e1fef6090ef3ec494489021035/config/config.default.yaml#L222-L245) as the first approximation, there is also [solar_multiple_estimate](https://github.com/euronion/trace/blob/1b313271081a04e1fef6090ef3ec494489021035/config/config.default.yaml#L252C5-L252C28) which can be probably helpful.

ekatef:
Perfect @GbotemiB! Congratulations with making the approach work 🎉 

My feeling is that the design is basically complete and we are very close to finalise it's implementation. Have added some technical comments which relate to polishing of the current approach.

ekatef:
The next step is data checking and creation of an example (which can also serve as a testing approach). 

In part of data:
1) we need to ensure that the Copernicus grid codes are appropriate;
2) would be great to cross-check values of `capacity_per_sqkm` and `correction_factor` against the published data, keeping in sight #362 and #361 may be useful for that [as a side note, I do not completely understand the meaning of `correction_factor`; could be a good idea to discuss it and improve documentation on that].

Regarding the example, we can probably add a section on CSP into [build_renewable_profiles](https://github.com/pypsa-meets-earth/documentation/blob/main/notebooks/build_renewable_profiles.ipynb), if CSP is viable for Nigeria (not sure about that), or create a dedicated example considering an area with higher direct radiation values.

As an advanced comment, we are considering only one type of CSP technologies, namely CSP tower. I wonder if it would it be possible to add also parabolic through. What do you think?

GbotemiB:
Thanks @ekatef for the support.

Regarding adding another csp technology, I can look into it. It shouldn't be too difficult to handle.

ekatef:
> Thanks @ekatef for the support.
> 
> Regarding adding another csp technology, I can look into it. It shouldn't be too difficult to handle.

My pleasure!

Perfect, thanks! It would be definitely great to have both CSP technologies implemented.

config.default.yaml
GbotemiB: Done...rather than use csp-tower everywhere, I used just csp. When we have another technology different from csp-tower, then it will be easy to generalize them as csp.

config.default.yaml
GbotemiB: I have also changed this to `csp_model`

ekatef:
A technical comment: I'm thinking that CSP are especially valuable thanks to the storage which is their part by design. So, CSP generation do not follow all the cloudiness variations immediately, but provides some constant output. That is a nice contrast to dynamics of PV or wind generation and facilitates operation of the power system a lot. 

Moreover, CSP can be used as a [dispatchable source](https://en.wikipedia.org/wiki/Dispatchable_generation), which can dramatically facilitate integration of RES into a power system, e.g. as described in [this report](https://d-nb.info/1054989648/34).

It would be great to ensure that the modelled behaviour of CSP demonstrates the mentioned features in fact. Ideally, CSP generation should more or less follow the demand.

I suspect that the simplified CSP representation doesn't satisfy this condition, as CSP potential is likely to be as intermittent, as one of PV generation, since we don't use any smoothing. Could be probably daily or multi-daily averaging be applied to CSP potential in the simplified representation to emulate smoothing effects of the storage?

@GbotemiB @davide-f what is your feeling about that?

ekatef:
An additional academic output which can probably help for quantitative analysis of our results is this [Copernicus  preprint](https://esd.copernicus.org/articles/15/307/2024/). It is focused on projections for the CSP potential, but there are also some suggestions for land usage and some corrections used to evaluate the potential.

config.default.yaml
```diff
@@ -313,6 +313,26 @@ renewable:
...
+    cutout: cutout-2013-era5
+    resource:
+      method: csp
+      installation: SAM_solar_tower
+    capacity_per_sqkm: 4.6 # From 1.7 to 4.6 addresses issue #361
+    # Determined by comparing uncorrected area-weighted full-load hours to those
+    # published in Supplementary Data to
+    # Pietzcker, Robert Carl, et al. "Using the sun to decarbonize the power
+    # sector: The economic potential of photovoltaics and concentrating solar
+    # power." Applied Energy 135 (2014): 704-720.
```
ekatef: @GbotemiB  as discussed, I do agree that this value should be revised for CSP. A citation from [Franke et al 2024]: `For CSP power plants, the highest capacity is 6.4 MW/km2 as this technology needs more area per plant`, while for PVs they report 12.26 MW/km2. Obviously, feel free to investigate that issue using also other sources.

data/costs.csv
ekatef: Great idea!

A formatting suggestion: it would be better to keep `CSP` capitalised, as for `HVDC` abbreviation. Could it be possible?

GbotemiB:
@ekatef @davide-f @pz-max 

To visualize the renewable potential of CSP, I have uploaded the notebooks [here](https://github.com/GbotemiB/csp-design/tree/validate_csp/notebooks)

To visualize the generation mix from the modeling output (CSP inclusive), I have uploaded the notebooks [here](https://github.com/GbotemiB/csp-design/tree/validate_csp/generation_mix_notebooks)

ekatef:
> @ekatef @davide-f @pz-max
> 
> To visualize the renewable potential of CSP, I have uploaded the notebooks [here](https://github.com/GbotemiB/csp-design/tree/validate_csp/notebooks)
> 
> To visualize the generation mix from the modeling output (CSP inclusive), I have uploaded the notebooks [here](https://github.com/GbotemiB/csp-design/tree/validate_csp/generation_mix_notebooks)

Fantastic work @GbotemiB!! I think, the outputs look pretty good. Some technical points can be probably improved, like adjusting of the plots to account for the energy balance and some parameter fixes. However, I agree with @davide-f that we should not to go into too much details with parameters investigation: it should be enough to cross-check `capacity_per_sqkm ` and the land codes.

What may be of concern are these CSP spikes in the negative area of the dispatch plots. It looks like CSP storage serves not only CSP needs, but also to the energy system in general. That is probably good from the system perspective, but not exactly realistic.

doc/index.rst
```diff
@@ -190,6 +190,7 @@ Documentation
 
 * :doc:`release_notes`
 * :doc:`how_to_contribute`
+* :doc:`how_to_docs`
```
ekatef: Not sure if omitting of `how_to_docs` was intentional previously. I have a feeling that it the instructions on the documentation have been moved into [how_to_contribute](https://pypsa-earth.readthedocs.io/en/latest/how_to_contribute.html#documentation).

scripts/add_electricity.py
GbotemiB: I have removed the conditional and refined the implementation here.

config.default.yaml
GbotemiB: Fixed

scripts/add_electricity.py
GbotemiB: fixed

scripts/add_electricity.py
GbotemiB: This has been removed as we are no longer handling this section in the `add_electricity.py` script

davide-f:
> > @ekatef @davide-f @pz-max
> > To visualize the renewable potential of CSP, I have uploaded the notebooks [here](https://github.com/GbotemiB/csp-design/tree/validate_csp/notebooks)
> > To visualize the generation mix from the modeling output (CSP inclusive), I have uploaded the notebooks [here](https://github.com/GbotemiB/csp-design/tree/validate_csp/generation_mix_notebooks)
> 
> Fantastic work @GbotemiB!! I think, the outputs look pretty good. Some technical points can be probably improved, like adjusting of the plots to account for the energy balance and some parameter fixes. However, I agree with @davide-f that we should not to go into too much details with parameters investigation: it should be enough to cross-check `capacity_per_sqkm ` and the land codes.
> 
> What may be of concern are these CSP spikes in the negative area of the dispatch plots. It looks like CSP storage serves not only CSP needs, but also to the energy system in general. That is probably good from the system perspective, but not exactly realistic.

Great @GbotemiB :D
Regarding the stores, in the plot, it is recommended to have the plot of the link rather than the store.
For internal check, on the other hand, we can have a plot where only the csp generator, the csp store and the csp link are plotted to guarantee its operation.
I think it makes sense the store to exchange sign, but the link must not.

davide-f (approved):
This PR seems to me good to go :)
@ekatef, feel free with @GbotemiB to crosscheck the csp time series parameters, e.g. the capacity per square km or so (I agree with you about not spending too much time), and then feel free to merge :)
Really amazing work!

GbotemiB:
Considering Algeria, 
Here is the result for the links.
image

image
image
The timeseries for the links actually cancel out.

This is the plot for the generator
image


This is the plot for the stores
image


This is the generation mix for Algeria. Initially, I used stores for `csp`, when it should have been `generator` just like it was done for solar pv.
image

ekatef:
Fantastic @GbotemiB! Great work and agree with @davide-f that we are very close to finalise 😄

Thanks a lot to you both for cross-checking operation of CSP unit. It looks like the model prefers to size CSP in a way which prevents the plant from running continuously. Not sure it's exactly realistic from the engineering perspective, but that is a rather advanced question which is clearly out of scope for the PR.

Currently going through the final review round mainly to ensure to ensure integrity of the implementation improvements.

test/config.landlock.yaml
```diff
@@ -9,7 +9,7 @@ retrieve_databundle: # required to be "false" for nice CI test output
 countries: ["BW"]
 
 electricity:
-  renewable_carriers: [solar, onwind, hydro]
+  renewable_carriers: [solar, csp, onwind, hydro]
```
ekatef, line 12: @GbotemiB Did you have a specific idea when adding CSP only to a landlock test? (Looks good for me, just double-checking)

ekatef (approved):

ekatef:
Hello @GbotemiB!

Looks perfect 😄 Thanks for the final touches.

There is a small question left on the test: have you been thinking to include CSP in CI, but wanted to keep it light, correct?

The only change I'm thinking about is adjusting `capacity_per_sqkm` to ensure that we capture the differences with PV. A citation from [Franke et al 2024]: `"For CSP power plants, the highest capacity is 6.4 MW/km2 as this technology needs more area per plant, while for PVs they report 12.26 MW/km2"`.  I'd probably take CSP/PV ratio from there and adjust our `capacity_per_sqkm` accordingly -> multiplying the current values for CSP on `6.4/12.26 = 0.52`. What do you think?

ekatef:
Two more advanced modeling points I'd track as features requests:
1. Landuse availability for CSP should include the slope, which can be implemented [using gebco dataset](https://github.com/euronion/trace/blob/1b313271081a04e1fef6090ef3ec494489021035/config/config.default.yaml#L235-L240).
2. Some additionally constraints should be applied for CSP model to make sure that its operation is more realistic from thermal engineering perspective.

But that is another story 🙂

test/config.landlock.yaml
GbotemiB: At the point, CI was failing due to missing csp in the renewable carriers. Adding csp solved the CI issue.

doc/index.rst
GbotemiB: I have reverted adding `how_to_docs`

test/config.landlock.yaml
ekatef: Ahh, I see! The config of the landlock test case is re-defining the renewable carriers, while this list should be consistent with the tutorial config (used as a base to run the tests). Thanks a lot for the explanation 🙂

GbotemiB:
> Hello @GbotemiB!
> 
> Looks perfect 😄 Thanks for the final touches.
> 
> There is a small question left on the test: have you been thinking to include CSP in CI, but wanted to keep it light, correct?
> 
> The only change I'm thinking about is adjusting `capacity_per_sqkm` to ensure that we capture the differences with PV. A citation from [Franke et al 2024]: `"For CSP power plants, the highest capacity is 6.4 MW/km2 as this technology needs more area per plant, while for PVs they report 12.26 MW/km2"`. I'd probably take CSP/PV ratio from there and adjust our `capacity_per_sqkm` accordingly -> multiplying the current values for CSP on `6.4/12.26 = 0.52`. What do you think?

About including CSP into the CI, I thought about it. I noticed that the rest of the renewable technology are already in the CI. All it took was adding csp to landlock config.

With regards to the `capacity_per_sqkm` for csp. I agree that we can reduce the `capacity_per_sqkm` with a factor of 0.52, which results to 2.392.

config.default.yaml
GbotemiB: Report from [here](https://www.renewableenergyworld.com/solar/calculating-solar-energys-land-use-footprint/) estimates the `capacity_per_sqkm` to 7.1MW/km2 which is close to the results from [Franke et al 2024].

GbotemiB: Based on the ratio, the new value for `capacity_per_sqkm` will be 2.392. I have set this in the config files.

davide-f:
To me this PR seems fine, to align @ekatef feel free to merge when appropriate :)

config.default.yaml
ekatef: Awesome, thanks a lot for the adjustment!

ekatef:
Merged! 😄 Thanks a lot for an amazing contribution @GbotemiB 🎉 🎉 🎉 
Thank you @davide-f and @pz-max for the outstanding support 🙂

pz-max:
Hi guys, I think we need a more generic version of CSP. Can someone raise an issue?

- people want to run several CSP technology version at the same time (trough, tower, etc)
- there is a lot of flexibility to pick different receiver, storage and power block technologies

Current implementation is quite static but of course a good start.

davide-f:
Good point @pz-max .
Issue added here: https://github.com/pypsa-meets-earth/pypsa-earth/issues/1038

Tough I'm unsure about the urgency, indeed the current implementation is a good start.
do we have some users on that?

PR #999, opened 2024-03-30
Generalize linetypes for AC and DC lines

### LINKED ISSUE #998: Generalize line_types

I'd like that our line-voltage management strategy would extend beyond the fixed 3-type voltage lines to accommodate an arbitrary list of voltage levels.

To do so, on one hand, we need to revise:
- the handling of the line-type handling in build_osm_network. In particular, we shall revise [this function](https://github.com/pypsa-meets-earth/pypsa-earth/blob/d3d718b8e22fab344991a73a9f9637bf6f22ecb4/scripts/base_network.py#L284) to make sure that each line is matched with a line_type that is the closest among the provided ones. The same shall be extended also for DC lines by voltage
- avoid the rebase to config values here: https://github.com/pypsa-meets-earth/pypsa-earth/blob/d3d718b8e22fab344991a73a9f9637bf6f22ecb4/scripts/base_network.py#L440 . In the latest pypsa-eur this function is not needed anymore

davide-f:
# Closes #998 

## Checklist

### DIFF
--- a/config.tutorial.yaml
+++ b/config.tutorial.yaml
@@ -181,11 +181,15 @@ electricity:
       Onshore: [onwind]
       PV: [solar]
 lines:
-  types:
+  ac_types:
+    110.: "490-AL1/64-ST1A 220.0"
     220.: "Al/St 240/40 2-bundle 220.0"
     300.: "Al/St 240/40 3-bundle 300.0"
     380.: "Al/St 240/40 4-bundle 380.0"
-  dc_type: "HVDC XLPE 1000"
+    500.: "Al/St 240/40 4-bundle 380.0"
+    750.: "Al/St 560/50 4-bundle 750.0"
+  dc_types:
+    500.: "HVDC XLPE 1000"
   s_max_pu: 0.7
   s_nom_max: .inf
   length_factor: 1.25
--- a/config.tutorial.yaml
+++ b/config.tutorial.yaml
@@ -182,7 +182,7 @@ electricity:
       PV: [solar]
 lines:
   ac_types:
-    110.: "490-AL1/64-ST1A 220.0"
+    132.: "243-AL1/39-ST1A 20.0"
     220.: "Al/St 240/40 2-bundle 220.0"
     300.: "Al/St 240/40 3-bundle 300.0"
     380.: "Al/St 240/40 4-bundle 380.0"
--- a/doc/configtables/lines.csv
+++ b/doc/configtables/lines.csv
@@ -1,5 +1,5 @@
 ,Unit,Values,Description
-ac_types,--,"Values should specify a `line type in PyPSA <https://pypsa.readthedocs.io/en/latest/components.html#line-types>`_ for AC lines. Keys should specify the corresponding voltage level (e.g. 220., 300. and 380. kV)", "Specifies line types to assume for the different voltage levels of the ENTSO-E grid extraction. Should normally handle voltage levels 220, 300, and 380 kV."
+ac_types,--,"Values should specify a `line type in PyPSA <https://pypsa.readthedocs.io/en/latest/components.html#line-types>`_ for AC lines. Keys should specify the corresponding voltage level (e.g. 220., 300. and 380. kV)", "Specifies line types to assume for the different voltage levels of the target region. Should normally handle voltage levels 220, 300, and 380 kV."
 dc_types, --, "Values should specify a `line type in PyPSA <https://pypsa.readthedocs.io/en/latest/components.html#line-types>`_ for DC-lines. Keys should specify the corresponding voltage level (e.g. 220., 300. and 380. kV)", "Specifies DC-line types."
 s_max_pu,--,"Value in [0.,1.]","Correction factor for line capacities (``s_nom``) to approximate :math:`N-1` security and reserve capacity for reactive power flows"
 s_nom_max,MW,"float","Global upper limit for the maximum capacity of each extendable line."
--- a/doc/configtables/lines.csv
+++ b/doc/configtables/lines.csv
@@ -1,6 +1,6 @@
 ,Unit,Values,Description
-types,--,"Values should specify a `line type in PyPSA <https://pypsa.readthedocs.io/en/latest/components.html#line-types>`_ in case AC lines or a custom linetype in case of DC lines modeled as Lines. Keys should specify the corresponding voltage level (e.g. 220., 300. and 380. kV)", "Specifies line types to assume for the different voltage levels of the ENTSO-E grid extraction. Should normally handle voltage levels 220, 300, and 380 kV."
-dc_type, --, "Values should specify a `line type in PyPSA <https://pypsa.readthedocs.io/en/latest/components.html#line-types>`_ for DC-lines.", "Specifies DC-line types."
+ac_types,--,"Values should specify a `line type in PyPSA <https://pypsa.readthedocs.io/en/latest/components.html#line-types>`_ for AC lines. Keys should specify the corresponding voltage level (e.g. 220., 300. and 380. kV)", "Specifies line types to assume for the different voltage levels of the ENTSO-E grid extraction. Should normally handle voltage levels 220, 300, and 380 kV."
+dc_types, --, "Values should specify a `line type in PyPSA <https://pypsa.readthedocs.io/en/latest/components.html#line-types>`_ for DC-lines. Keys should specify the corresponding voltage level (e.g. 220., 300. and 380. kV)", "Specifies DC-line types."
 s_max_pu,--,"Value in [0.,1.]","Correction factor for line capacities (``s_nom``) to approximate :math:`N-1` security and reserve capacity for reactive power flows"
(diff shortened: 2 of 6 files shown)


config.tutorial.yaml
```diff
@@ -181,11 +181,15 @@ electricity:
...
       PV: [solar]
 lines:
-  types:
+  ac_types:
+    110.: "490-AL1/64-ST1A 220.0"
     220.: "Al/St 240/40 2-bundle 220.0"
     300.: "Al/St 240/40 3-bundle 300.0"
     380.: "Al/St 240/40 4-bundle 380.0"
-  dc_type: "HVDC XLPE 1000"
+    500.: "Al/St 240/40 4-bundle 380.0"
```
ekatef, line 189: May it be worth to add a comment on why we think that a line type for nominal `380kV` is suitable alos for `500kV`?

If I understand properly, the reason is:
1) a list of line types in PyPSA is incomplete;
 2) a conductor diameter (and transmission capacity) in the same for `380kV` and `500kV` line types. Probably, the insulation should differ, but that is not particularly relevant for our analysis.
 
 Is it a correct understanding? :)

scripts/base_network.py
```diff
@@ -281,30 +272,60 @@ def _load_transformers_from_osm(fp_osm_transformers, buses):
...
+
+def _get_linetype_by_voltage(v_nom, d_linetypes):
+    """
+    Return the linetype based on the voltage.
+    """
+    v_nom_min, line_type_min = min(
+        d_linetypes.items(),
+        key=lambda x: abs(x[0] - v_nom),
+    )
+    return line_type_min
```
ekatef: If I get it right, a line type will be selected which has the minimal difference of `v_nom`. That means that `v_nom` of the line type may be also lower than the actual `v_nom` value. I suspect that it's perfectly appropriate, as OSM-extracted `v_nom` values  are likely to be some estimations, but just to be sure: given your expertise in the area, do you consider it a safe approach?

scripts/base_network.py
```diff
@@ -281,30 +272,60 @@ def _load_transformers_from_osm(fp_osm_transformers, buses):
...
 
-def _set_electrical_parameters_lines(lines_config, voltages_config, lines):
+def _get_linetypes(line_types, voltages):
+    """
+    Return the linetypes for selected voltages.
+    """
+    vnoms_diff = set(voltages).symmetric_difference(set(line_types.keys()))
+    if vnoms_diff:
+        logger.warning(
+            f"Voltages {vnoms_diff} not in the {line_types} or {voltages} list."
```
ekatef, line 290: Could it also make sense to explain what will be done to close this gap?

scripts/base_network.py
```diff
@@ -281,30 +272,60 @@ def _load_transformers_from_osm(fp_osm_transformers, buses):
...
+        )
+    return {k: v for k, v in line_types.items() if k in voltages}
+
+
+def _get_linetype_by_voltage(v_nom, d_linetypes):
+    """
+    Return the linetype based on the voltage.
+    """
```
ekatef: May we probably to add a bit more details to the docstring to make more clear distinction with `_get_linetypes`? Like: `Return the linetype for each line in the network by matching v_nom for the line against parameters of the standard line types.`

scripts/base_network.py
```diff
@@ -281,30 +272,60 @@ def _load_transformers_from_osm(fp_osm_transformers, buses):
     return transformers
 
 
-def _set_electrical_parameters_lines(lines_config, voltages_config, lines):
+def _get_linetypes(line_types, voltages):
```
ekatef: I wonder if it could be possible to adjust a bit naming to highlight that ` _get_linetypes` and `_get_linetype_by_voltage` do very different things... Probably, ` _get_linetypes` -> ` _lookup_linetypes`?

scripts/simplify_network.py
```diff
@@ -838,7 +838,7 @@ def merge_isolated_nodes(n, threshold, aggregation_strategies=dict()):
     n = pypsa.Network(snakemake.input.network)
 
     base_voltage = snakemake.params.electricity["base_voltage"]
-    linetype = snakemake.params.config_lines["types"][base_voltage]
+    linetype = snakemake.params.config_lines["ac_types"][base_voltage]
```
ekatef, line 841: What would happen if `base_voltage` in `electricity` section of the config doesn't correspond to any of the standard voltage values defined in `ac_types`? Could it be possible to add some back-up approach?

ekatef (commented):
Hello @davide-f!
Great work and an elegant implementation! :D That is a pleasure to review your PRs ;)

Have added some comments which mainly relate to adding some clarifications, cross-checking the assumptions and considering an edge case for `base_voltage`. As a little additional point: I have an impression that we also don't need voltage levels used previously for rebase, which means that:
- [this line](https://github.com/davide-f/pypsa-earth/blob/f5bdc322e6ef5c3fe2e1896f0fe0760e1c2e6c15/config.default.yaml#L129) should be removed from the config;
- `voltages_config` argument and [a voltage warning](https://github.com/davide-f/pypsa-earth/blob/f5bdc322e6ef5c3fe2e1896f0fe0760e1c2e6c15/scripts/base_network.py#L131-L135) are not anymore needed in `_load_buses_from_osm`, `_load_lines_from_osm` and `_load_links_from_osm`.

Could you please check it? Otherwise, looks perfect :) Looking forward to see the effect on the accuracy of the grid representation!

scripts/base_network.py
davide-f: Done

scripts/base_network.py
davide-f: Revised, what do you think?

scripts/base_network.py
davide-f: Done

scripts/simplify_network.py
davide-f: I think that may not be needed; if we receive complaints, we can revise

scripts/base_network.py
davide-f: Yeah; I think it is ok.
A bare conducture can be used in a wide range of voltages and its insulation is guaranteed by insulators rather than the conductor itself.
The ability of a conductor to be used at different voltages is indirect: higher voltages imply higher heights, thus higher wind speeds, thus higher mechanical stresses for example.

I prefer this approach as for non-standard voltages, we may miss the allocation thus leading to issues.
Moreover, even if "better standard values" could be used, the error may be small and acceptable.
I'd personally like to extend the standard values to lower values to better reflect global values

config.tutorial.yaml
davide-f: The values are copied from pypsa-eur XD
The description of the conductor-voltage is explained below.

scripts/base_network.py
ekatef: Agree, the incorrect allocation of the line type seems to be the major issue for now.

Thanks a lot for the detailed explanation!

config.tutorial.yaml
ekatef: Super, thanks for the clarification!

scripts/base_network.py
ekatef: Perfect!

scripts/simplify_network.py
ekatef: Agree, it may be better not to over-complicate :)

doc/configtables/lines.csv
```diff
@@ -1,6 +1,6 @@
 ,Unit,Values,Description
-types,--,"Values should specify a `line type in PyPSA <https://pypsa.readthedocs.io/en/latest/components.html#line-types>`_ in case AC lines or a custom linetype in case of DC lines modeled as Lines. Keys should specify the corresponding voltage level (e.g. 220., 300. and 380. kV)", "Specifies line types to assume for the different voltage levels of the ENTSO-E grid extraction. Should normally handle voltage levels 220, 300, and 380 kV."
-dc_type, --, "Values should specify a `line type in PyPSA <https://pypsa.readthedocs.io/en/latest/components.html#line-types>`_ for DC-lines.", "Specifies DC-line types."
+ac_types,--,"Values should specify a `line type in PyPSA <https://pypsa.readthedocs.io/en/latest/components.html#line-types>`_ for AC lines. Keys should specify the corresponding voltage level (e.g. 220., 300. and 380. kV)", "Specifies line types to assume for the different voltage levels of the ENTSO-E grid extraction. Should normally handle voltage levels 220, 300, and 380 kV."
```
ekatef: Ahh, sorry, have forgotten to add a comment: probably, we could also remove mentioning of ENTSO-E or add an explanation that it relates to Europe only. What do you think?

ekatef (approved):

ekatef:
Hello @davide-f! Thanks a lot for the explanations and introducing the changes.
Looks great :D

doc/configtables/lines.csv
davide-f: Done, agree :) Soon I'll then merge

PR #1006, opened 2024-04-19
Add option to adjust load scale for each individual countries

virio-andreyana:
# Closes # (if applicable).

Add option to individually adjust load scale for each country. The code format in the config is the same as focus_weights.

## Checklist

### DIFF
--- a/scripts/build_demand_profiles.py
+++ b/scripts/build_demand_profiles.py
@@ -184,8 +184,9 @@ def build_demand_profiles(
 
     if isinstance(scale, dict):
         logger.info(f"Using custom scaling factor for load data.")
+        DEFAULT_VAL = scale.get("DEFAULT", 1.0)
         for country in countries: 
-            scale.setdefault(country,1.0) 
+            scale.setdefault(country, DEFAULT_VAL) 
             
         for country, scale_country in scale.items():
             gegis_load.loc[
--- a/scripts/build_demand_profiles.py
+++ b/scripts/build_demand_profiles.py
@@ -186,9 +186,11 @@ def build_demand_profiles(
         if isinstance(scale, dict):
             logger.info(f"Using custom scaling factor for load data.")
             for country, scale_country in scale.items():
-                gegis_load.loc[gegis_load.region_code == country,"Electricity demand"] *= scale_country
-        
-        elif isinstance(scale, (int,float)):
+                gegis_load.loc[
+                    gegis_load.region_code == country, "Electricity demand"
+                ] *= scale_country
+
+        elif isinstance(scale, (int, float)):
             logger.info(f"Load data scaled with scaling factor {scale}.")
             gegis_load["Electricity demand"] *= scale
 
--- a/scripts/build_demand_profiles.py
+++ b/scripts/build_demand_profiles.py
@@ -181,8 +181,17 @@ def build_demand_profiles(
     gegis_load = gegis_load.to_dataframe().reset_index().set_index("time")
     # filter load for analysed countries
     gegis_load = gegis_load.loc[gegis_load.region_code.isin(countries)]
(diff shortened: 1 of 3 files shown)


scripts/build_demand_profiles.py
```diff
@@ -181,8 +181,19 @@ def build_demand_profiles(
     gegis_load = gegis_load.to_dataframe().reset_index().set_index("time")
     # filter load for analysed countries
     gegis_load = gegis_load.loc[gegis_load.region_code.isin(countries)]
-    logger.info(f"Load data scaled with scaling factor {scale}.")
-    gegis_load["Electricity demand"] *= scale
+
+    if scale is not None:
```
ekatef: Great that you have thought about edge cases! However, not sure this particular check makes sense, as if `scale` is not defined it would cause an error earlier, at https://github.com/pypsa-meets-earth/pypsa-earth/blob/d89a9a384ccff8b65adba6cae835a6df6ee53334/scripts/build_demand_profiles.py#L249 So, I'd remove this check which would also make the code easier to read. What do you think?

ekatef:
Hello @virio-andreyana! An amazing idea and very nice implementation. I have only one comment on it. Otherwise, looks perfect 🙂

scripts/build_demand_profiles.py
davide-f: I agree

scripts/build_demand_profiles.py
```diff
@@ -181,8 +181,19 @@ def build_demand_profiles(
...
+        if isinstance(scale, dict):
```
davide-f: It may be useful to have a "DEFAULT" code, as in some cases, we may like only to describe a subset of users.
In the case it is a dict, I'd read a "DEFAULT" code (with .get("DEFAULT", 1.0)) and use this value for all countries that are not in the list.
What do you think?

scripts/build_demand_profiles.py
```diff
@@ -246,7 +257,7 @@ def upsample(cntry, group):
     load_paths = snakemake.input["load"]
     countries = snakemake.params.countries
     admin_shapes = snakemake.input.gadm_shapes
-    scale = snakemake.params.load_options["scale"]
+    scale = snakemake.params.load_options.get("scale")
```
davide-f: We could do .get("scale", 1.0) to avoid the None case

davide-f (commented):
Great contributions @virio-andreyana :D and very interesting indeed!
I've added few comments.
It would be great to have your opinion on this little changes :)

scripts/build_demand_profiles.py
```diff
@@ -181,8 +181,21 @@ def build_demand_profiles(
     gegis_load = gegis_load.to_dataframe().reset_index().set_index("time")
     # filter load for analysed countries
     gegis_load = gegis_load.loc[gegis_load.region_code.isin(countries)]
-    logger.info(f"Load data scaled with scaling factor {scale}.")
-    gegis_load["Electricity demand"] *= scale
+
+    if isinstance(scale, dict):
+        logger.info(f"Using custom scaling factor for load data.")
+        for country in countries: 
+            scale.setdefault(country,1.0)
```
davide-f: Great @virio-andreyana :D
I think this PR is basically ready to merge; however, what do you think about whether having this "1.0" fixed,
we have it by config, like:
```suggestion
        logger.info(f"Using custom scaling factor for load data.")
        DEFAULT_VAL = scale.get("DEFAULT", 1.0)
        for country in countries: 
            scale.setdefault(country, DEFAULT_VAL) 
```
Basically the proposal is to optionally treat the country key "DEFAULT" as the default value, if present

pz-max:
@virio-andreyana is next me. PR is ready to merge.

davide-f (approved):

Issue #501, opened 2022-10-29
Move from os to pathlib


pz-max:
From Discord:
"os.path vs pathlib": https://builtin.com/software-engineering-perspectives/python-pathlib
*The os module represents paths as strings with which you cannot do much. The pathlib module represents paths as special objects with useful methods and attributes.*

Max: Let's use Pathlib by default. Tracking this here as a feature request so someone can work on the transition in future. I think it is a nice exercise for new contributors to get familiar with the code & the git workflow

finozzifa:
Hi all, I picked up this issue as a _good first issue_ to get familiar with the code :)

A few important notes:
1) pathlib.Path.walk is available starting from the Python version 3.12. Hence for the moment I am keeping the equivalent command from the os package (i.e. os.walk)
2) pathlib does not provide an equivalent command to os.chdir. Hence I kept the latter

ekatef:
Hey @finozzifa! Perfect, thanks a lot for taking this :)

@pz-max if you have any insights on that, feel free to share :D

PR #1017, opened 2024-04-30
Issue 501 os to pathlib

finozzifa:
# Closes # (if applicable).
Issue #501 

The PR aims at replacing the os module with the pathlib module

## Checklist

### DIFF
--- a/scripts/_helpers.py
+++ b/scripts/_helpers.py
@@ -96,12 +96,12 @@ def read_osm_config(*args):
     {"Africa": {"DZ": "algeria", ...}, ...}
     """
     if "__file__" in globals():
-        base_folder = pathlib.Path(__file__).parent
+        base_folder = get_dirname_path(__file__)
         if not pathlib.Path(base_folder, "configs").exists():
-            base_folder = pathlib.Path(base_folder).parent
+            base_folder = get_dirname_path(base_folder)
     else:
-        base_folder = pathlib.Path.cwd()
-    osm_config_path = str(get_path(base_folder, "configs", REGIONS_CONFIG))
+        base_folder = str(pathlib.Path.cwd())
+    osm_config_path = get_path(base_folder, "configs", REGIONS_CONFIG)
     with open(osm_config_path, "r") as f:
         osm_config = yaml.safe_load(f)
     if len(args) == 0:
@@ -132,7 +132,7 @@ def sets_path_to_root(root_directory_name):
     while n >= 0:
         n -= 1
         # if repo_name is current folder name, stop and set path
-        if repo_name == pathlib.Path(".").absolute().name:
+        if repo_name == get_basename_abs_path("."):
             repo_path = str(pathlib.Path.cwd())  # current_path
             os.chdir(repo_path)  # change dir_path to repo_path
             print("This is the repository path: ", repo_path)
@@ -170,7 +170,7 @@ def configure_logging(snakemake, skip_handlers=False):
     kwargs.setdefault("level", "INFO")
 
     if skip_handlers is False:
-        fallback_path = get_path(pathlib.Path(__file__).parent, "..", "logs", f"{snakemake.rule}.log")
+        fallback_path = get_path(get_dirname_path(__file__), "..", "logs", f"{snakemake.rule}.log")
         logfile = snakemake.log.get("python", snakemake.log[0] if snakemake.log else fallback_path)
         kwargs.update(
             {
@@ -502,7 +502,7 @@ def mock_snakemake(rulename, **wildcards):
     def make_accessable(*ios):
         for io in ios:
             for i in range(len(io)):
-                io[i] = str(pathlib.Path(io[i]).absolute())
+                io[i] = get_abs_path(io[i])
 
     make_accessable(job.input, job.output, job.log)
     snakemake = Snakemake(
@@ -521,7 +521,7 @@ def mock_snakemake(rulename, **wildcards):
 
     # create log and output dir if not existent
     for path in list(snakemake.log) + list(snakemake.output):
-        pathlib.Path(path).parent.mkdir(parents=True, exist_ok=True)
+        build_directory(get_dirname_path(path))
 
     os.chdir(script_dir)
     return snakemake
@@ -646,7 +646,7 @@ def read_csv_nafix(file, **kwargs):
     if "na_values" not in kwargs:
         kwargs["na_values"] = NA_VALUES
 
-    if pathlib.Path(file).stat().st_size > 0:
+    if get_path_size(file) > 0:
         return pd.read_csv(file, **kwargs)
     else:
         return pd.DataFrame()
@@ -694,7 +694,7 @@ def read_geojson(fn, cols=[], dtype=None, crs="EPSG:4326"):
         CRS of the GeoDataFrame
     """
     # if the file is non-zero, read the geodataframe and return it
-    if pathlib.Path(fn).stat().st_size > 0:
+    if get_path_size(fn) > 0:
         return gpd.read_file(fn)
     else:
         # else return an empty GeoDataFrame
@@ -818,13 +818,56 @@ def get_last_commit_message(path):
 
 def get_dirname_abs_path(path):
     """
-    It returns the directory name of a normalized and absolutized version of the
-    pathname path.
+    It returns the directory name of a normalized and absolutized version of the path.
     """
     return str(pathlib.Path(path).absolute().parent)
+
+
+def get_dirname_path(path):
+    """
+    It returns the directory name of the path.
+    """
+    return str(pathlib.Path(path).parent)
+
+
+def get_abs_path(path):
+    """
+    It returns the absolutized version of the path.
+    """
+    return str(pathlib.Path(path).absolute())
+
+
+def get_basename_abs_path(path):
+    """
+    It returns the base name of a normalized and absolutized version of the path
+    """
+    return pathlib.Path(path).absolute().name
+
+
+def get_basename_path(path):
+    """
+    It returns the base name of the path
+    """
+    return pathlib.Path(path).name
+
 
 def get_path(*args):
     """
     It returns a new PosixPath object.
     """
-    return pathlib.Path(*args)
+    return str(pathlib.Path(*args))
+
+
+def get_path_size(path):
+    """
+    It returns the size of a path (in bytes)
+    """
+    return pathlib.Path(path).stat().st_size
+
+
+def build_directory(path):
+    """
+    It creates recursively the directory and its leaf directories
+    """
+    pathlib.Path(path).mkdir(parents=True, exist_ok=True)
+
--- a/scripts/_helpers.py
+++ b/scripts/_helpers.py
@@ -861,7 +861,7 @@ def get_path(*args):
     """
     It returns a new path string.
     """
-    return str(pathlib.Path(*args))
+    return pathlib.Path(*args)
 
 
 def get_posixpath(*args):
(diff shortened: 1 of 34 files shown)


finozzifa:
mmm tests were ok when I executed them locally. I need to figure out what's the matter. Apologies for the failing pipelines.

finozzifa:
Hi all,
just as a side-note, every change has been tested locally in ad-hoc python script, in order to make sure that we get with the pathlib module what we got with the os module.

Moreover, I started on May 3rd to replace the code snippets with pathlib that repeat themselves often, with methods in _helpers.py. I unit test these methods, following this procedure:
1) I write the method using the os module code snippet present on the main branch
2) I build the unit test
3) I modify the method, replacing the os module with the corresponding pathlib code snippet
4) I make sure that the unit test still works

test/test_helpers.py
```diff
@@ -0,0 +1,90 @@
...
+    get_dirname_abs_path,
+    get_dirname_path,
+    get_path,
+    get_path_size,
+)
+
+path_cwd = str(pathlib.Path.cwd())
+
+
+def test_get_abs_path():
```
pz-max, line 29: A nice test would be to check if the `os` thing you are replacing is equal to the pathlib solution. If this tests passes it proves that the full implementation can be merged without any risk (like the CI crashes you experience :D)

scripts/_helpers.py
```diff
@@ -821,3 +818,68 @@ def get_last_commit_message(path):
...
+    It returns the size of a path (in bytes)
+    """
+    return pathlib.Path(path).stat().st_size
+
+
+def build_directory(path):
+    """
+    It creates recursively the directory and its leaf directories.
+    """
+    pathlib.Path(path).mkdir(parents=True, exist_ok=True)
```
pz-max: For consistency, just tagging that abstractions are missing for: 

- `str(pathlib.Path.cwd())`

- `str(pathlib.Path(str(config["destination"])).relative_to("."))` 

- `pathlib.Path(base_folder, "configs").exists()`

Maybe also a few more.

pz-max (commented):
Really a PR that goes through everything. Just added a comment on 1) abstraction consistency and a 2) tests that makes sure that the os to pathlib swap is faultless. On another note, I am not sure if the abstraction layer is super helpful as the pathlib library is quite readable.

This is just a comment. Davide/Katia have for sure more perspectives on this.

ekatef:
> Really a PR that goes through everything. Just added a comment on 1) abstraction consistency and a 2) tests that makes sure that the os to pathlib swap is faultless. On another note, I am not sure if the abstraction layer is super helpful as the pathlib library is quite readable.
> 
> This is just a comment. Davide/Katia have for sure more perspectives on this.

Thanks a lot for reviewing, @pz-max! 

And thank you @finozzifa for the great work :) As discussed, our `"good first issues"` are quite labor-intensive, but you manage it graciously. Also, thanks a lot for the PRs you have completed along the way! From our side, I fear that reviewing can take some time. Apologies for that, and please keep in mind that it's not your fault in any way.

Regarding testing, @finozzifa has mentioned that every change has been tested locally to ensure that the workflow output remained the same. [Windows CI currently crashes due to troubles with GADM server: `Exception: GADM server is down at https://geodata.ucdavis.edu/gadm/gadm4.1/gpkg/gadm41_NGA.gpkg. Data needed for building shapes can't be extracted.`] But I agree that it would be really handy to have some testing output to be sure the introduced changes work, even for some edge cases. @finozzifa do you see any possibility to share the testing infrastructure you have developed? Like creating and pushing a dedicated branch for testing? That is not your word in which I have doubts, rather our ability to detect potential troubles with just an intense look into the code... 🙃

Regarding the consistency, I wonder if a string representation of `Path` object does make sense (e.g. `str(pathlib.Path.cwd())`), as it basically gives the result as `os` methods. [The idea taken from @davide-f  :) ] May it be a better approach to revert the changes, if such a transformation is needed? That would also allow to keep the PR lighter...

@finozzifa @davide-f @pz-max what are your feeling about all that? 🙂

finozzifa:
Hi all,

thanks for your interest in my PR. I really enjoyed working on it :)

-) @pz-max:  I created separate methods that enclose pathlib invocations to enable an easier unit testing and to make the code more maintainable. Now, if we need to change one line in (say) method _get_path_, we just need to change one file, rather than the entire **scripts** folder. Finally, I believe that having this abstraction layer makes the code review easier.

-) @pz-max and @ekatef: I have created dedicated unit tests, in the _test_helpers.py_ file

-)@pz-max and @ekatef: The unit tests have been created as follows: 1) I create the method with the os module , 2) I build the unit test, 3) I re-write the method with pathlib, 4) I make sure that the corresponding unit test still works. 

-) Indeed @pz-max , I have not created methods for all invocations of pathlib. My goal would be do that and so having the _import pathlib_ just in the __helpers.py_ file. I am happy to do it in this issue, or create a new one and work on it after this PR is merged. Let me know what you guys prefer.

Next steps:
-) I will investigate further why the pipeline failed (thanks @ekatef) --> the windows pipeline is still failing because this website is not reachable https://geodata.ucdavis.edu/gadm/gadm4.1/gpkg/gadm41_BEN.gpkg --> it now works. Thanks @ekatef for restarting the pipeline
-) I will modify the unit tests, adding a further assert statement where I compare the output with the corresponding os invocation I replaced ----> this is done!

test/test_helpers.py
finozzifa: done! I have added extra assert statements in test/test_helpers.py :D

FabianHofmann (commented):
@finozzifa I am having just a very short look at this. I think the idea to move to pathlib Path is great. I also think there is a potential for even leaner code. 


In general, I don't think it is a good idea to introduce abstracted functions with function names nearly as long as the content. For example 

```
def get_posixpath(*args):
    """
    It returns a new PosixPath object.
    """
    return pathlib.Path(*args)

```

should definitely be avoided. Why? Python users (who may well be experienced in Python) don't know this function when they see their usage in the scripts. So on the first sight, they won't know things like

```
dirname = get_posixpath("path/to/my/dir")
```

but they would know operations like

```
dirname = Path("path/to/my/dir")
```

Another downside is that you lose all the doc-strings, argument refs (and potential meta data for static type checking) compared to calling `Path`. 

So, whenever there is very basic, short python code, I would propose to not abstracting it. 


Therefore, I would try to always work with the explicit Path objects in the scripts (Path super powerful!) and use operators like `.parent` or `/` . For example:

- Operations like: 
```
GADM_inputfile_gpkg = get_path(
        str(pathlib.Path.cwd()), "data", "gadm", GADM_filename, GADM_filename + ".gpkg")
```

would be 

```
GADM_inputfile_gpkg = Path.cwd() / "data/gadm" / GADM_filename / GADM_filename + ".gpkg"
```

this is much more readable and everyone knows whats happening. 

This would probably remove the need for most new defined get_ functions and avoid to cycle between `str` and `Path` most of the times.

For the often used `os.chdir(get_dirname_abs_path(__file__))`, that could perhaps be combined with a info, since this might otherwise confuse users?


```
import os
from pathlib import Path

def change_to_script_dir(script_path):
    """
    Change the current working directory to the directory containing the given script.
    
    Parameters:
    script_path (str or Path): The path to the script file.
    """
    # Ensure script_path is a Path object
    script_path = Path(script_path)
    
    # Get the directory containing the script
    script_directory = script_path.parent
    
    # Change the current working directory to the script directory
    os.chdir(script_directory)
    
    # Optionally, print the new working directory
    # potentially use logger.info instead
    print("Changed directory to: ", Path.cwd())

```
call it with `change_to_script_dir(__file__)`

These are just my thoughts, nothing mandatory. Hope that helps.

finozzifa:
Hi @FabianHofmann thanks for your comment :)

I think that I just follow a different coding style. Whenever I see a line of code which repeats itself several times, I try (where possible) to create methods that implement such line of code and that can be separately tested. 

This is because I want to ensure consistency and make future changes easier. In fact, instead of tracing throughout the code each invocation of (say) _os.path.basename()_, I just need to modify once the method where such line is implemented.

I am not sure why code readability would suffer following this approach, given the fact that people would just need to trace the method in the __helpers.py_ to check what it does.

I believe that in this way, code reviews, code merges and further code expansions will be easier, as we can rely on soundproof methods that we have unit-tested.

FabianHofmann:
I think that is super reasonable. It makes most sense to avoid code duplication. Just felt that some of the functions are not "abstract" enough. Perhaps a common ground is trying to avoid the back and forth between Path and string and trying to stick to Path wherever possible? Btw, f-strings nicely process `Path` objects, like 


```
p  = Path("path/to/file")
print(f"This is the path: {p}")
```

But as said, you have the full power :)

finozzifa:
Hi @FabianHofmann,

thanks again for your comment and your contribution :) (which I really appreciate)

I added the casting to string of the PosixPath object because otherwise some parts of code are failing. 

Leveraging on the possibilities offered by the abstract method _get_path_, I removed such casting and restarted the pipelines. As you see they are unfortunately failing. One error that I have found is:

_ERROR:_helpers:An error happened in module '/home/runner/work/pypsa-earth/pypsa-earth/scripts/build_natura_raster.py', function '<listcomp>': argument of type 'PosixPath' is not iterable_

It seems therefore that in some parts of the code, the casting is indeed needed.

I have created also the _get_posixpath_ method because, based on my tests, the PosixPath object is needed in the _build_test_configs.py_ script

How would you proceed in this case?

FabianHofmann:
Sorry for being late here. I am wondering where this happens. Could you point me to the exact spot? The string object is iterable but in case of a path it would not make sense ("path/to" would be divided into "p", "a", "t", ...).

finozzifa:
Hi @FabianHofmann, 
thanks a lot :)
I believe that my previous comment was a bit misleading. 
I just wanted to point out that apparently there are some parts of the code where PosixPath objects cannot be used and so their string casting is necessary.
Anyway, I once again remove the string casting from _get_path_ and I am letting the pipelines fail again so that I can perform further investigations.

finozzifa:
Hi @FabianHofmann ,
now get_path returns a posixPath object. I solved the occurrences where the casting to string was needed. Thanks :)

finozzifa:
Hi all,

please find below a recap of the changes I performed in this PR.

I tried my best to include all of your suggestions and recommendations. If I missed some, I apologize and I will include them as soon as possible.

**### Recap of the changes**

1) The _os_ module has been replaced with the _pathlib_ module. There are still a few _os-module-calls_ in _build_natura_raster.py_ (this is because there is no pathlib alternative for os.walk) and in the __helpers.py_

2) As mentioned before, I have introduced an abstraction of the path methods. Such methods are available in the __helpers.py_ script. In fact, whenever I see lines of code that repeat themselfs several times, I try to create methods that implement such lines of code and that can be separately unit-tested. In this was I can ensure consistency and make future changes easier. Instead of tracing throughout the code each invocation of (say) _os.path.basename()_, I just need to modify once the method where such line is implemented. I believe that in this way, code reviews, code merges and further code expansions will be easier, as we can rely on soundproof methods that we have unit-tested.

3) I have introduced dedicated unit tests for the path methods I coded in the _helpers.py. As of now, these unit tests are NOT executed as part of the CI/CD pipelines.

**### Possible spin-offs**

1) It would nice to push these modifications a bit further, so that the dependency on the _os_ and _pathlib_ modules is explicit just in the __helpers.py_ script.

2) It would nice to include the execution of the unit tests in the CI/CD pipeline, as already done for the framework repository

I thank you in advance for your comments in reviewing this PR :)

Fabrizio

PR #1019, opened 2024-05-01
Fix bug to run the cluster wildcard `min`

virio-andreyana:
Its a minor bugfix so that the cluster wildcard option `min` can be run on snakemake without error
- In Snakefile, `min` doesn't require any number, similar to `all`
- In cluster_network.py, the number of network after `determine_network_topology()` is always larger than the original n_cluster. A network copy is made to count that, but better suggestion is appreciated.

## Checklist

### DIFF
--- a/scripts/cluster_network.py
+++ b/scripts/cluster_network.py
@@ -691,6 +691,8 @@ if __name__ == "__main__":
         "exclude_carriers", []
     )
     aggregate_carriers = set(n.generators.carrier) - set(exclude_carriers)
+
+    n.determine_network_topology()
     if snakemake.wildcards.clusters.endswith("m"):
         n_clusters = int(snakemake.wildcards.clusters[:-1])
         aggregate_carriers = snakemake.params.electricity.get("conventional_carriers")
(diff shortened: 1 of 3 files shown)


scripts/cluster_network.py
```diff
@@ -699,7 +699,9 @@ def cluster_regions(busmaps, inputs, output):
     elif snakemake.wildcards.clusters == "all":
         n_clusters = len(n.buses)
     elif snakemake.wildcards.clusters == "min":
-        n_clusters = n.buses.groupby(["country", "sub_network"]).size().count()
+        m = n.copy()
+        m.determine_network_topology()
+        n_clusters = m.buses.groupby(["country", "sub_network"]).size().count()
```
davide-f: Thanks @virio-andreyana !
As a comment, is there a specific need to copy the network?
I'd simply do calculations on n; calculating the network topology doesn't harm the network.
After this, the PR can go :D
Remember to add a minor line to the release_note :)

davide-f (approved):
Great @virio-andreyana :D

Issue #687, opened 2023-04-16
Address pandas deprecations


ekatef:
The recent [pandas 2.0 release](https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html) at 3 April has removed append for both Series and Dataframes. There is a need to check where pandas and geopandas `append` are used in the code

Thanks a lot to @yerbol-akhmetov for reporting!

## Checklist

- [х] I am using the current [`main`](https://github.com/pypsa-meets-earth/pypsa-earth/tree/main) branch or the latest [release](https://github.com/pypsa-meets-earth/pypsa-earth/releases). Please indicate.
- [х] I am running on an up-to-date [`pypsa-earth` environment](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

Any attempt to call pandas `append` results in an error `AttributeError: 'DataFrame' object has no attribute 'append'`, if an environment is using pandas `version>2.0.0`.

## Error Message

Currently, one of the ways to reproduce an error is to run a workflow with `custom_powerplants: true` (will be fixed with #686):

```bash
Traceback (most recent call last):
  File "~/pypsa-earth/.snakemake/scripts/tmpl0i66xpm.build_powerplants.py", line 314, in <module>
    ppl = add_custom_powerplants(
  File "~/pypsa-earth/.snakemake/scripts/tmpl0i66xpm.build_powerplants.py", line 254, in add_custom_powerplants
    return ppl.append(add_ppls, sort=False, ignore_index=True, verify_integrity=True)
  File "~/miniconda3/envs/pypsa-earth-upd6/lib/python3.10/site-packages/pandas/core/generic.py", line 5989, in __getattr__
    return object.__getattribute__(self, name)
AttributeError: 'DataFrame' object has no attribute 'append'. Did you mean: '_append'?
```

ekatef:
NB using list `append` is perfectly alright. Actually, the reason why pandas `append` has been removed was avoiding confusion between list and pandas `append` methods

ekatef:
Quick scan of the workflow with tutorial (more or less standard) settings don't throw any problems. So, hopefully the issue is rather minor

ekatef:
Apart of `append` deprecation, it may be also good to fix other pandas deprecations, like `'axis' keyword in DataFrame.groupby`.

ekatef:
[...]

PR #1023, opened 2024-05-09
Issue 687 pandas deprecations

finozzifa:
# Closes # (if applicable).

Issue #687 

The changes are listed in the comment https://github.com/pypsa-meets-earth/pypsa-earth/issues/687#issuecomment-2100447481

## Checklist

- [x ] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [x ] I tested my contribution locally and it seems to work fine.
- [x ] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes, including reference to the requested PR.

### DIFF
--- a/scripts/build_osm_network.py
+++ b/scripts/build_osm_network.py
@@ -723,8 +723,9 @@ def fix_overpassing_lines(lines, buses, distance_crs, tol=1):
         lines = gpd.GeoDataFrame(df_to_add, crs=lines.crs)
     else:
         lines = gpd.GeoDataFrame(
-        pd.concat([lines, df_to_add], ignore_index=True).reset_index(drop=True),
-        crs=lines.crs)
+            pd.concat([lines, df_to_add], ignore_index=True).reset_index(drop=True),
+            crs=lines.crs,
+        )
 
     return lines, buses
 
--- a/scripts/build_osm_network.py
+++ b/scripts/build_osm_network.py
@@ -719,13 +719,9 @@ def fix_overpassing_lines(lines, buses, distance_crs, tol=1):
     # remove original lines
     lines.drop(lines_to_split, inplace=True)
 
-    if lines.empty:
-        lines = gpd.GeoDataFrame(df_to_add.reset_index(drop=True), crs=lines.crs)
-    else:
-        lines = gpd.GeoDataFrame(
-            pd.concat([lines, df_to_add], ignore_index=True).reset_index(drop=True),
-            crs=lines.crs,
-        )
+    lines = df_to_add if lines.empty else pd.concat([lines, df_to_add])
+
+    lines = gpd.GeoDataFrame(lines.reset_index(drop=True), crs=lines.crs)
 
     return lines, buses
 
--- a/scripts/build_osm_network.py
+++ b/scripts/build_osm_network.py
@@ -149,15 +149,15 @@ def set_lines_ids(lines, buses, distance_crs):
         )
         if distance_bus0 > 0.0:
             # the line does not start in the node, thus modify the linestring
-            lines.geometry.loc[i] = linemerge(
+            lines.loc[i, "geometry"] = linemerge(
                 [
                     LineString(
                         [
-                            buses.geometry.loc[bus0_id],
-                            lines.geometry.loc[i].boundary.geoms[0],
+                            buses.loc[bus0_id, "geometry"],
+                            lines.loc[i, "geometry"].boundary.geoms[0],
                         ]
                     ),
-                    lines.geometry.loc[i],
+                    lines.loc[i, "geometry"],
                 ]
             )
 
@@ -171,13 +171,13 @@ def set_lines_ids(lines, buses, distance_crs):
         )
         if distance_bus1 > 0.0:
             # the line does not end in the node, thus modify the linestring
-            lines.geometry.loc[i] = linemerge(
+            lines.loc[i, "geometry"] = linemerge(
                 [
-                    lines.geometry.loc[i],
+                    lines.loc[i, "geometry"],
                     LineString(
                         [
-                            lines.geometry.loc[i].boundary.geoms[1],
-                            buses.geometry.loc[bus1_id],
+                            lines.loc[i, "geometry"].boundary.geoms[1],
+                            buses.loc[bus1_id, "geometry"],
                         ]
                     ),
                 ]
@@ -720,7 +720,7 @@ def fix_overpassing_lines(lines, buses, distance_crs, tol=1):
     lines.drop(lines_to_split, inplace=True)
 
     if lines.empty:
-        lines = gpd.GeoDataFrame(df_to_add, crs=lines.crs)
+        lines = gpd.GeoDataFrame(df_to_add.reset_index(drop=True), crs=lines.crs)
     else:
         lines = gpd.GeoDataFrame(
             pd.concat([lines, df_to_add], ignore_index=True).reset_index(drop=True),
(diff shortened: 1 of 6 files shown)


finozzifa:
Hi all,
The PR apparently solves just three out of the four deprecations listed in Issue #687. The one in **build_osm_network** remains unfortunately still outstanding. Sincere apologies!

EDIT on May 10th 2024: all deprecations have been addressed!

scripts/build_osm_network.py
```diff
@@ -719,10 +719,13 @@ def fix_overpassing_lines(lines, buses, distance_crs, tol=1):
     # remove original lines
     lines.drop(lines_to_split, inplace=True)
 
-    lines = gpd.GeoDataFrame(
-        pd.concat([lines, df_to_add], ignore_index=True).reset_index(drop=True),
-        crs=lines.crs,
-    )
+    if lines.empty:
+        lines = gpd.GeoDataFrame(df_to_add, crs=lines.crs)
```
ekatef: Not sure about the index: it may be the case that it's safer to reset it, as it's done for the concatenated data frame bellow. It would be great to test.

ekatef:
> Hi all, The PR apparently solves just three out of the four deprecations listed in Issue #687. The one in **build_osm_network** remains unfortunately still outstanding. Sincere apologies!
> 
> EDIT on May 10th 2024: all deprecations have been addressed!

Perfect, thanks a lot for addressing this issue!! The PR looks good for me, have only some doubts regarding this special case of empty `lines` dataframe. It would be great to test that it really works as expected.

As an additional comment, I see that the CI logs became now much cleaner, which is the fantastic news 💚 But it appears that CI is reporting another pandas deprecation issue (haven't seen it locally): `lines.geometry.loc[i] = linemerge(
/home/runner/work/pypsa-earth/pypsa-earth/.snakemake/scripts/tmprqe2v9pu.build_osm_network.py:178: FutureWarning: ChainedAssignmentError: behaviour will change in pandas 3.0!`. Would you mind to look into it?

scripts/build_osm_network.py
finozzifa: Hi @ekatef , thanks :)

 I am not sure however how to perform this test. I thought that such test was included in our regression tests. How do you reckon we should proceed?

finozzifa:
Hi all, I have found extra deprecations that I have resolved. However there are a couple I would like your opinion about.

You can find them listed in this [comment](https://github.com/pypsa-meets-earth/pypsa-earth/issues/687#issuecomment-2100447481).

scripts/build_osm_network.py
ekatef: Unfortunately, not. CI tests include rather standard cases (more or less corresponding to the tutorial case which deals with Nigeria + Benin). If there is need to test more sophisticated cases (for example, if the PR may behave differently for different grid topologies like in our case), some additional test cases may be needed.

Not sure however for which countries can we expect to have all the lines to be splited... @davide-f by any chance, are you aware on the region, where all the original lines are removed by this fragment: https://github.com/pypsa-meets-earth/pypsa-earth/blob/cea0dc03bf6e718504eada9dbb1eff9a844f18f2/scripts/build_osm_network.py#L719-L720
?

davide-f: If lines is empty this entire function looses meaning:
This function aims to split line fragments where there is a substation on its path.
So this exception should never occur.
I expect that when lines is empty, so will lines_to_add. In that case the previous if not lines_to_add should shortcut the execution

ekatef: > If lines is empty this entire function looses meaning: This function aims to split line fragments where there is a substation on its path. So this exception should never occur.

@davide-f  thanks for looking into that! :)
I mean rather a situation when all lines should be split which means that the line `lines.drop(lines_to_split, inplace=True)` removes all the lines from the network. Is it not possible?

The point is, `pandas` complained about concatenating dataframes, while one of them is empty or all NAs. Looking into this check,

https://github.com/pypsa-meets-earth/pypsa-earth/blob/cea0dc03bf6e718504eada9dbb1eff9a844f18f2/scripts/build_osm_network.py#L706-L707

I have an impression that `lines_to_add` can't be empty, so the deprecation problem arises if there are no lines left in `n.lines`, after removing `lines_to_split`. Am I missing something?

davide-f: Ahhh I get it. Actually I don't know, it may be an edge case of small countries with very few lines.
The previous code never crashed though.
Rather then duplicating the creation of the dataframe, what about doing:
dfs=[b] if a.empty else [a, b]
And then create the geodf using concat applied to dfs?
Also the check on lines.empty is also ok

finozzifa: Hi all,
I modified the code with 

    lines = df_to_add if lines.empty else pd.concat([lines, df_to_add])

    lines = gpd.GeoDataFrame(lines.reset_index(drop=True), crs=lines.crs)

In this way the pipelines are still working and deprecation warning disappears.

scripts/prepare_network.py
```diff
@@ -225,18 +225,18 @@ def set_transmission_limit(n, ll_type, factor, costs, Nyears=1):
 
 
 def average_every_nhours(n, offset):
-    logger.info(f"Resampling the network to {offset}")
+    logger.info(f"Resampling the network to {offset.casefold()}")
     m = n.copy(with_time=False)
 
-    snapshot_weightings = n.snapshot_weightings.resample(offset).sum()
+    snapshot_weightings = n.snapshot_weightings.resample(offset.casefold()).sum()
```
ekatef, line 237: My feeling is that it's a perfectly appropriate solution, as keeping the user interfaces must have a priority over code elegance. I'm only thinking that probably it could make sense to add a comment to this line explaining that the case transformation is needed to keep `H` in the options definitions. What do you think @finozzifa?

scripts/build_osm_network.py
ekatef: Perfect!! 😄

scripts/prepare_network.py
```diff
@@ -225,18 +225,18 @@ def set_transmission_limit(n, ll_type, factor, costs, Nyears=1):
 
 
 def average_every_nhours(n, offset):
-    logger.info(f"Resampling the network to {offset}")
+    logger.info(f"Resampling the network to {offset.casefold()}")
```
ekatef: My feeling is that it'd be better to keep here the capital `H` to be consistent with the options definition in the file name. The point is, snapshots resampling is coded in the model name as something like `24H` (so called [opts wildcard](https://pypsa-eur.readthedocs.io/en/latest/wildcards.html#the-opts-wildcard)). So, that is something which is visible for a modeller, and as a such I'd expect that capitalisation would be kept also in the logs.

@finozzifa does it sound reasonable for you?

scripts/prepare_network.py
finozzifa: I agree. Done!

scripts/prepare_network.py
finozzifa: I agree! Done and thanks :)

ekatef (approved):

ekatef:
Thanks a lot for introducing the changes @finozzifa! In my opinion, the PR is ready for merge :)

@davide-f what is your feeling about that? :) May you have any further comments?

davide-f (approved):
The PR is fine for me :) Approving

ekatef:
Merged :D

Thanks a lot @finozzifa for the great contribution! 🎉 🎉 🎉 
Thanks to your PR, we are now ready to the future `pandas`, which is awesome :)))

@davide-f thank you for the fantastic support 🙂

PR #1053, opened 2024-07-03
Test routine: add Makefile; CI: consolidate ci.yaml files

### LINKED ISSUE #1052: Refactor test routine
## Checklist

## Problem Statement

Currently the test routine uses nested snakemake calls where snakemake is called via a bash command in the Snakefile. This architecture omits error streams and snakemake runs that fail are not reported as such, see https://github.com/pypsa-meets-earth/pypsa-earth/actions/runs/9667460555/job/26669336528#step:9:1739.

## Suggestion

I would propose to execute `snakemake` in the GH workflow and optionally keep the commands in a common file for easy local execution, like https://github.com/PyPSA/pypsa-eur/blob/master/test.sh

I would as well propose to collect all workflow CI's in one yaml that uses a `matrix` entry for setting the operation systems.

FabianHofmann:
Closes #1052 

This PR proposes four changes:

 - fix a bug in `add_extra_components`. CSP generators are not necessarily attached to every bus (this is probably related to a "missing" substation or shape assigment). This lead to an error in the CI before (as mentioned in #1052) in the alignment of `buses_i` being the AC-buses and the csp buses.
 - use a Makefile to setup and run tests. With `make test` all the snakemake tests can now be run locally. Why is that better? The workflow used a recursive snakemake call. This lead to errors not being reported back as errors. With the Makefile, errors are strightly reported back and the test command are explicit and easily accessible.
 - the `run_scenario` rule now streams errors which it did not do before. At some point, we might also think using something similar as for the tests to improve visibility of the command used. 
 -  Consolidate CI yaml. Use only one CI yaml instead of three. Use micromamba with it built in cashing functionality. 

In general, the workflow would profit from using `--configfile=<additional.config.yaml>` instead of the diff_config approach. Was there a specific reason to stick to the manual insertion of config diffs? Perhaps is that historically grown? Note that `--configfile` overrides existing and adds additonal keys to the config given in the Snakefile. So it does essentially the same.     

The CI will likely fail now, I need probably some more iterations. 

## Checklist


pz-max:
I can comment here a bit:
- @FabianHofmann , we split the CI to test everything in Linux (faster) & only a subset in other OS. Another reason was the need for a badge for each OS. This can be useful e.g. when you run on Linux where all tests are passing and you don't care about Windows too much.
- love the `make` solution 

Regarding this bit:
> In general, the workflow would profit from using --configfile=<additional.config.yaml> instead of the diff_config approach. Was there a specific reason to stick to the manual insertion of config diffs? Perhaps is that historically grown? Note that --configfile overrides existing and adds additonal keys to the config given in the Snakefile. So it does essentially the same

I worked on the config diff stuff. Snakemake uses the exact same config merging code as we do (realised that later). I think @davide-f possibly added a few other parts. We implemented it like this to be independent of Snakemake and more flexible for this functionality (e.g. we are using ruaml (as Yaml container), or smt like that, which also keeps comments after the config is merged). Not sure what other benefits/ disadvantages exist. @davide-f might have more thoughts. 

Feel free to decide what is the better solution for now :+1:

Snakefile
```diff
@@ -18,14 +18,10 @@ from pathlib import Path
 
 HTTP = HTTPRemoteProvider()
 
-if "config" not in globals() or not config:  # skip when used as sub-workflow
-    if not exists("config.yaml"):
-        copyfile("config.tutorial.yaml", "config.yaml")
-
-    configfile: "config.yaml"
-
```
davide-f: Are we sure about this?
I'm wondering that it may be good to preserve.
What do you think?
I like though specifying the three configfile: as proposed but probably within the if?

davide-f (commented):
Many thanks @FabianHofmann !!! Much needed!
Added a comment, happy to talk tomorrow if you plan to participate to the meeting [I understood so or I'm wrong?]

However the CI is not working; are you sure the configfiles get merged well?

Snakefile
FabianHofmann: Hey @davide-f, thanks for your reply! we are getting there :) This is indeed one change that I am not sure about. The reason why I put this in is that I think the workflow should also be tested against the `config.default.yaml` settings. Before, the CI was "only" testing the tutorial config as a basis. This leads to the artefact that one always has to secure manually that the default config is in line with the tutorial config. The proposed approach is the same as we have it in pypsa-eur and uses 

- config.default.yaml as level-zero defaults
- add configs/bundle_config.yaml which extends the default (no overrides)
- use incremental config changes from the user via `config.yaml`. Here the user can only set the keys which should be different from the `config.default.yaml`. In a python function, that would translate to setting a keyword-argument which already has a default to something else. So the `config.yaml` could be empty, and the workflow would still run.   

The structure ensure that the user always has a stable set of default configs. Whenever there is an update to the default, the user does not need to copy it to `config.yaml` and insert their custom settings manually, it happens automatically.  

However, it leads to one or two more needed fixes in the workflow. For example, the CI is breaking now as the solver_options are not fully compatible. Snakemake is currently merging the solver options from the default solver with the ones from the tutorial solver. 

Two options:

1. Go full in -> make an adoption of the solver_options as we have it in pypsa-eur (already happened and will be happening anyway through #796 )
2. Revert this setting and introduce it at a later point (which I would strongly encourage to do)

what to you think?

FabianHofmann:
@pz-max @davide-f this is ready for being merged, objective values are reproduced. A merge today would be cool, so I can look into #796 soon

Snakefile
ekatef: Hello @FabianHofmann! Sounds like a well-elaborated and quite an elegant solution.

However, the idea to automate introducing changes into `config.yaml` sounds quite alarming for me. If feels like we are making the user interface unstable (because `config.yaml` is actually the major UI of the model)  which is a famous architectural anti-pattern and must be avoided.

According to the user experience we have observed, the only use case when such automation may be done in a safe way, is when `config.yam` doesn't exist. Exactly that is a case addressed by the initial configuration. So, I agree with @davide-f that the changes would be better placed under `if` condition , as it was before. What do you think?

doc/release_notes.rst
```diff
@@ -48,6 +48,9 @@ E.g. if a new rule becomes available describe how to use it `snakemake -j1 run_t
 
 * Create files where the code outputs the value of the objective function. `PR #1033 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1033>`__
 
+* The workflow configuration now supports incremental changes to the default configuration in the `config.yaml` and configfiles passed to snakemake via `--configfile myconfig.yaml`. Therefore the user may now only include settings in their `config.yaml` which differ from the default configuration. `PR #1053 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1053>`_
+
```
ekatef, line 52: Actually, the workflow still has functionality which allows for incrementally changes in the configuration files. However, this functionality is implemented using a dedicated function which places it among advanced stuff which you are using only if you well understand what you are doing. That allows to avoid side effects for the users interested in basic modelling functionality. 

I recognise that there is an approach close to what is used in PyPSA-Eur, but according to out observations for the requests in `Support` channel, that is not a concept which is easy for use.

envs/environment.yaml
```diff
@@ -77,7 +77,7 @@ dependencies:
 
 # Default solver for tests (required for CI)
 - glpk
-- ipopt<3.13.3
+- ipopt
```
ekatef, line 80: As we discussed, that would lead to workflow crashes for Windows users, unfortunately.

config.default.yaml
```diff
@@ -408,18 +412,66 @@ solving:
...
+      run_crossover: "off"
+      small_matrix_value: 1e-6
+      large_matrix_value: 1e9
+      primal_feasibility_tolerance: 1e-5
+      dual_feasibility_tolerance: 1e-5
+      ipm_optimality_tolerance: 1e-4
+      parallel: "on"
+      random_seed: 123
+    gurobi-default:
+      threads: 4
```
ekatef, line 435: @FabianHofmann has it been a specific reason for adding these updates to the PR? 

The linopy-integration PR #796 introduces the same changes, so we'll have them in the workflow in any case. Do you see any advantages for testing from having these changes earlier? Just trying to understand the logic :)

ekatef (commented):
Hello @FabianHofmann, thanks a lot for the great contribution!

Have added a couple of questions related to some details. Could you please look into them? My general feeling is that it would be great to avoid breaking changes, especially those which relate to the behaviour of user interfaces.

Snakefile
FabianHofmann: hey @ekatef! Thanks for the review :) Perhaps I did not explain myself too well. So, there will be now breaking changes to the workflow. As well, the workflow will not automatically change the config.yaml. It is just the order and config updates that changes. Instead of one config files, the workflow now uses multiples to ensure a full set of defaults and allow the user to make use of config diffs.  The user can now use 

* the config.yaml  as a pure "diff" config to the defaults which only contains effective changes set by the user
* the `snakemake --configfile <myconfig.yaml>`  options

See for example the new [config.tutorial](https://github.com/open-energy-transition/pypsa-earth/blob/refactor-test-and-ci/config.tutorial.yaml). It only contains effective diffs to the default. This makes it much easier to maintain both the default and tutorial config and allows to actually see what are the effective changes to defaults coming from the config.tutorial.   

Old `config.yaml` file will still work. In case there is no `config.yaml` I added a function `copy_default_files()` which sets an empty config.yaml in place as a placeholder (see some lines below). The only exception is the settings of the solver options which anyway comes with the linopy integration which are necesary to allow for config updates.

If you want we can also chat about it later, if that suits :)

doc/release_notes.rst
FabianHofmann: I hope I understand the point correctly. Yes, the internal scenario config update routine (sorry for the long term) I did not touch. That is still the same as before. I only touched the "base" config to be the "config.default" but the scenario building is compatible with it. I mainly focused on the test routine and the allowing for incremental config changes (they are not necessarily required, old config behaviour still works)

envs/environment.yaml
FabianHofmann: yes, this was intended (but forgot to mention it somewhere), I referred to davides comment that we temporarily disable the windows stability until the linopy integration where we can kick out ipopt

config.default.yaml
FabianHofmann: Yes, these are needed to enable the incremental config changes (see above). But this will be easy to merge with the linopy PR

finozzifa:
Indeed a great PR @FabianHofmann! thanks a lot

FabianHofmann:
@davide-f @ekatef the PR is now ready. As discussed the `config.default.yaml` is ignored by now, and if no `config.yaml` is existent, the `config.tutorial.yaml` is copy to `config.yaml`

ekatef:
> @davide-f @ekatef the PR is now ready. As discussed the `config.default.yaml` is ignored by now, and if no `config.yaml` is existent, the `config.tutorial.yaml` is copy to `config.yaml`

Hey @FabianHofmann, thank you so much for the discussion and the adjustments! I love the solution with commenting-out the automated update using `config.default.yaml`. Absolutely agree that improvements are needed on management of the configuration files, and having this like in Snakemake should be an effective reminder! Could we probably add an issue to track the ideas and the progress on that, also accounting for the modularity idea @davide-f has suggested?

FabianHofmann:
@ekatef that is a good idea! thank you for your comment. I will raise an issue

davide-f (approved):
Great @FabianHofmann :D

davide-f:
Thanks @FabianHofmann !
Agree, let's keep it like this for now and edit after the merge :)

FabianHofmann:
mmh, annoying, as soon as passing the `--configfile` via snakemake, the condition 

```
if "config" not in globals() or not config:  # skip when used as sub-workflow
```

is false, so it will not load the `config.yaml`. It reverted the latest changes and have a look on monday, whether the if clause is really necessary for pypsa-earth-sec, or how we can adjust it.

davide-f:
> mmh, annoying, as soon as passing the `--configfile` via snakemake, the condition
> 
> ```
> if "config" not in globals() or not config:  # skip when used as sub-workflow
> ```
> 
> is false, so it will not load the `config.yaml`. It reverted the latest changes and have a look on monday, whether the if clause is really necessary for pypsa-earth-sec, or how we can adjust it.

Unfortunately it is necessary, I had to add that if clause exactly because of that unfortunately.
Worth pointing @hazemakhalek to agree on a solution. If we merge this, we cannot update -sec to latest main here
The merge will solve this issue...

davide-f:
Hello @FabianHofmann , just to confirm, we can close this PR as it is merged already into the merge right?

FabianHofmann:
Correct!

PR #1057, opened 2024-07-09
Fix hydro inflow overestimation for GADM

### LINKED ISSUE #1043: Overestimated hydro inflow
## Checklist

- [ ] I am using the current [`main`](https://github.com/pypsa-meets-earth/pypsa-earth/tree/main) branch or the latest [release](https://github.com/pypsa-meets-earth/pypsa-earth/releases). Please indicate. -> 58 commits behind
- [ ] I am running on an up-to-date [`pypsa-earth` environment](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`. -> Approx 2 months old

## Describe the Bug

In my model results, the hydro capacity factor is at 100%. This is probably due to extremely high inflow values, which are in the range of TWh within the first hour already:

grafik

This is tested for New Zealand. A possible source of error could be, that I have selected 
`normalization:method: hydro_capacities` in the config but didn't add the data for New Zealand in `data/hydro_capacities`. In this case, there should be a hint to do so in the config?

Currently running `normalization:method:eia` to see if it works.

energyLS:
Closes #1043. See the issue for details. 
Case separation is necessary, a more elaborate fix has not been found yet. Note: There are now a few case separations in the respective script.

### DIFF
--- a/scripts/build_renewable_profiles.py
+++ b/scripts/build_renewable_profiles.py
@@ -603,14 +603,13 @@ if __name__ == "__main__":
             columns={"x": "lon", "y": "lat", "country": "countries"}
         ).loc[bus_in_hydrobasins, ["lon", "lat", "countries", "shape_id"]]
 
-
         if snakemake.params.alternative_clustering == False:
             resource["plants"]["installed_hydro"] = [
                 True if (bus_id in hydro_ppls.bus.values) else False
                 for bus_id in resource["plants"].index
             ]
 
-         ### TODO: quickfix. above case and the below case should by unified
+        ### TODO: quickfix. above case and the below case should by unified
         if snakemake.params.alternative_clustering == True:
             resource["plants"]["installed_hydro"] = [
                 True if (bus_id in hydro_ppls.region_id.values) else False
@@ -649,7 +648,7 @@ if __name__ == "__main__":
                         )
                         * config.get("multiplier", 1.0)
                     )
-                    
+
                 elif method == "eia":
                     path_eia_stats = snakemake.input.eia_hydro_generation
                     normalize_using_yearly = get_eia_annual_hydro_generation(
--- a/scripts/build_renewable_profiles.py
+++ b/scripts/build_renewable_profiles.py
@@ -603,14 +603,13 @@ if __name__ == "__main__":
             columns={"x": "lon", "y": "lat", "country": "countries"}
         ).loc[bus_in_hydrobasins, ["lon", "lat", "countries", "shape_id"]]
 
+        # TODO: these cases shall be fixed by restructuring the alternative clustering procedure
         if snakemake.params.alternative_clustering == False:
             resource["plants"]["installed_hydro"] = [
                 True if (bus_id in hydro_ppls.bus.values) else False
                 for bus_id in resource["plants"].index
             ]
-
-        ### TODO: quickfix. above case and the below case should by unified
-        if snakemake.params.alternative_clustering == True:
+        else:
             resource["plants"]["installed_hydro"] = [
                 True if (bus_id in hydro_ppls.region_id.values) else False
                 for bus_id in resource["plants"].shape_id.values
--- a/scripts/build_renewable_profiles.py
+++ b/scripts/build_renewable_profiles.py
@@ -489,7 +489,7 @@ if __name__ == "__main__":
         from _helpers import mock_snakemake
 
         os.chdir(os.path.dirname(os.path.abspath(__file__)))
-        snakemake = mock_snakemake("build_renewable_profiles", technology="solar")
+        snakemake = mock_snakemake("build_renewable_profiles", technology="hydro")
         sets_path_to_root("pypsa-earth")
     configure_logging(snakemake)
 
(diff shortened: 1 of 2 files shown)


scripts/build_renewable_profiles.py
```diff
@@ -603,10 +603,18 @@ def create_scaling_factor(
...
+                True if (bus_id in hydro_ppls.bus.values) else False
+                for bus_id in resource["plants"].index
+            ]
+
+        ### TODO: quickfix. above case and the below case should by unified
+        if snakemake.params.alternative_clustering == True:
+            resource["plants"]["installed_hydro"] = [
+                True if (bus_id in hydro_ppls.region_id.values) else False
+                for bus_id in resource["plants"].shape_id.values
+            ]
```
ekatef: A great catch! On the first glance, it feels like the TODO can be addressed quite easily, using something like:
 
```suggestion
        if snakemake.params.alternative_clustering == True:
            resource["plants"]["installed_hydro"] = [
                True if (bus_id in hydro_ppls.region_id.values) else False
                for bus_id in resource["plants"].shape_id.values
            ]
            else:
                resource["plants"]["installed_hydro"] = [
                   True if (bus_id in hydro_ppls.bus.values) else False
                   for bus_id in resource["plants"].index
               ]
```
@energyLS does it make sense for you?

scripts/build_renewable_profiles.py
```diff
@@ -489,7 +489,7 @@ def create_scaling_factor(
         from _helpers import mock_snakemake
 
         os.chdir(os.path.dirname(os.path.abspath(__file__)))
-        snakemake = mock_snakemake("build_renewable_profiles", technology="solar")
+        snakemake = mock_snakemake("build_renewable_profiles", technology="hydro")
```
ekatef: Just to double-check: what has been the reason behind this change?

ekatef (commented):
Hello @energyLS, thanks a lot for the fix! The PR looks good to me. Have added a couple of comments, but they are quite minor.

Would you like to add a release note? The issue has been traced for quite a long, and that is a substantial improvement to have it resolved.

scripts/build_renewable_profiles.py
energyLS: Because I tested hydro locally; however I can revert it back to solar if desired

scripts/build_renewable_profiles.py
energyLS: @ekatef thanks for the suggestion, it indeed makes sense and feel free to commit your suggestion. However, I actually though of a even "deeper" unification, in the sense that 
`hydro_ppls.bus.values`and `hydro_ppls.region_id.values` are merged to one id/index and `resource["plants"].shape_id.values` and `resource["plants"].index` are merged to one id/index no matter wether the alternative clustering is true or false. But I think this might be an extra issue?

energyLS:
@ekatef thanks for the review, I have replied to your review and added a release note. Please let me know if any concerns persist, otherwise feel free to merge it.

scripts/build_renewable_profiles.py
ekatef: I see, thanks for the explanation! 🙂 I think, it's better to avoid changes which we actually don't need for the purpose of a particular PR. So, if you could please revert if, it would be great.

scripts/build_renewable_profiles.py
ekatef: Agree, it would be perfect to use the generalisation you suggest. Do you have any idea on how to implement it quickly? Probably we can use

```python
resource["plants"]["installed_hydro"] = [True if (bus_id in atlite_hydro_ids) else False for bus_id in ppl_hydro_ids] 
  ```
  
While `atlite_hydro_ids` and `ppl_hydro_ids` will be defined depending on `snakemake.params.alternative_clustering` (naming may be improved for sure!). What do you think?

**Comment by ekatef** [comment]
Perfect @energyLS, thank you for the answers and the release note!

Agree on your suggestion about finding a more general implementation, and have an impression that it can be implemented in quite a straightforward way.  Also, I'd avoid adding changes which are not necessary for the PR.

We are almost there 😄

File: scripts/build_renewable_profiles.py (hunk 1 above)
Comment (davide-f) [reply scripts/build_renewable_profiles.py]: Great reviews :D
Katia anticipated my comments.
As mentioned in the past call, I am ok with these types of temporary fixes.
The solid solution would be instead of dropping completely the inconsistency between alternative yes/no:
if an optional shapefile is passed to build_bus_regions and alternative is enabled, the index of that geojson will be used as index of the following shapes.
The source of all these issues are related to this I believe

File: scripts/build_renewable_profiles.py (hunk 2 above)
Comment (energyLS) [reply scripts/build_renewable_profiles.py]: I have reverted it.

File: scripts/build_renewable_profiles.py (hunk 1 above)
Comment (energyLS) [reply scripts/build_renewable_profiles.py]: @ekatef and @davide-f thanks for your inputs, I agree the clean solution would be dropping completely the inconsistency between alternative clustering. However I feel like this is a separate PR?

Comment (davide-f) [inline scripts/build_renewable_profiles.py]:
```suggestion
        # TODO: these cases shall be fixed by restructuring the alternative clustering procedure
        if snakemake.params.alternative_clustering == False:
            resource["plants"]["installed_hydro"] = [
                True if (bus_id in hydro_ppls.bus.values) else False
                for bus_id in resource["plants"].index
            ]
        else:
            resource["plants"]["installed_hydro"] = [
                True if (bus_id in hydro_ppls.region_id.values) else False
                for bus_id in resource["plants"].shape_id.values
            ]
```

What about this?
If @ekatef agrees we could go for it and close the PR :)

I'd avoid a duplicated if-if case for clarity but I think it is good to go for me :)

ekatef: @davide-f agree, and we have discussed with @energyLS that some slight improvements could make sense to avoid duplication. But I think the most important part is actually finding the reason for issues and having a fix for it. 

So, I'd go for the suggested changes if it doesn't require much time. Otherwise, feel free to merge as-is 🙂

energyLS:
@davide-f I have commited the changes and resolved the merge conflict, please merge once the CI runs through. Thank you!

davide-f:
Great @energyLS :D
Merging :)

PR #1058, opened 2024-07-09
Implement config versioning

ekatef:

Implements a check to ensure that the current local `config.yaml` is consistent with the version defined in `config.default.yaml`.

## Checklist

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -12,18 +12,13 @@ from shutil import copyfile, move
 
 from snakemake.remote.HTTP import RemoteProvider as HTTPRemoteProvider
 
-from _helpers import create_country_list, get_last_commit_message
+from _helpers import create_country_list, get_last_commit_message, match_config
 from build_demand_profiles import get_load_paths_gegis
 from retrieve_databundle_light import datafiles_retrivedatabundle
 from pathlib import Path
 
 
 HTTP = HTTPRemoteProvider()
-
-# using snakemake capabilities to deal with yanl configs
-with open("config.default.yaml", "r") as f:
-    actual_config = yaml.safe_load(f)
-actual_config_version = actual_config.get("version")
 
 if "config" not in globals() or not config:  # skip when used as sub-workflow
     if not exists("config.yaml"):
@@ -32,15 +27,7 @@ if "config" not in globals() or not config:  # skip when used as sub-workflow
     configfile: "config.yaml"
 
 
-current_config_version = config.get("version")
-
-if actual_config_version != current_config_version:
-    logger.warning(
-        f"The current version of 'config.yaml' doesn't match to the code version:\n\r"
-        f" {current_config_version} provided, {actual_config_version} expected.\n\r"
-        f"That can lead to the errors during execution of the workflow.\n\r"
-        f"Please update 'config.yaml' according to 'config.default.yaml' "
-    )
+match_config(config=config)
 
 
 configfile: "configs/bundle_config.yaml"
--- a/Snakefile
+++ b/Snakefile
@@ -16,7 +16,13 @@ from build_demand_profiles import get_load_paths_gegis
 from retrieve_databundle_light import datafiles_retrivedatabundle
 from pathlib import Path
 
+
 HTTP = HTTPRemoteProvider()
+
+# using snakemake capabilities to deal with yanl configs
+with open("config.default.yaml", "r") as f:
+    actual_config = yaml.safe_load(f)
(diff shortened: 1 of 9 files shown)


ekatef:
@davide-f have drafted a possible implementation. Could you please have a look? 

Have added one more field to the version number to track the changes in the config. Not sure it's the most elegant solution, though.

A possible improvement could be to add a check on the differences between the local config and `config.default.yaml`. Sometimes, it feels a good idea to have an opportunity to review the recent updates in the configuration options. However, not sure Snakemake traceback is the best place to place results of such a check. Probably, having an additional helper function would be a better idea. What is your feeling about that?

Snakefile
```diff
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
 import sys
+import yaml
```
davide-f: Great draft!

I'd propose to create an auxiliary function into helpers and place there all the code.
The function may by like: check_version(config)
Internally it loads the config.default and then does the magic.
What do you think?

config.default.yaml
```diff
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: CC0-1.0
 
-version: 0.3.0
+version: 0.3.0.0
```
davide-f: Please revert :)

config.tutorial.yaml
```diff
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: CC0-1.0
 
-version: 0.3.0
+version: 0.3.0.0
```
davide-f: Please revert

davide-f (commented):
Great @ekatef :D
Beyond the minor comments, please add a release doc :)

config.default.yaml
ekatef: Done!

config.tutorial.yaml
ekatef: Done :)

Snakefile
ekatef: Agree, revised.

ekatef:
> Great @ekatef :D Beyond the minor comments, please add a release doc :)

Done 🙂 Thanks a lot for the review!

I'm thinking what would be the best way to remind that a release is needed every time, when the config structure gets changed. Adding a point for that to PR template sounds like not a really good idea, but I don't have better ones at the moment. What is your perspective on that?

Snakefile
davide-f: Can we drop this line? it seems is not needed

scripts/_helpers.py
```diff
@@ -27,6 +27,28 @@
...
+def check_config_version(config):
+    """
+    Check that a version of the local config.yaml matches to the actual config
+    version as defined in config.default.yaml.
+    """
+
+    # using snakemake capabilities to deal with yanl configs
+    with open("config.default.yaml", "r") as f:
+        actual_config = yaml.safe_load(f)
+    actual_config_version = actual_config.get("version")
```
davide-f: I'd propose to have the file path as parameter here, like:

```suggestion
def check_config_version(config, fp_config="config.default.yaml"):
    """
    Check that a version of the local config.yaml matches to the actual config
    version as defined in config.default.yaml.
    """

    # using snakemake capabilities to deal with yanl configs
    with open(fp_config, "r") as f:
        actual_config = yaml.safe_load(f)
    actual_config_version = actual_config.get("version")
```

davide-f (commented):
Great @ekatef :D

I've added a minor comment. Moreover there is a minor merge conflict to fix.
This PR is ready to go afterwards. If you like to include these minor changes, that would be nice but the functionality is there.
Afterwards, if CI passes, feel free to merge yourself :)

Snakefile
ekatef: True, have forgotten to remove it. Thanks!

scripts/_helpers.py
ekatef: Agree, done!

ekatef:
> Great @ekatef :D
> 
> I've added a minor comment. Moreover there is a minor merge conflict to fix. This PR is ready to go afterwards. If you like to include these minor changes, that would be nice but the functionality is there. Afterwards, if CI passes, feel free to merge yourself :)

@davide-f fixed! Thank you 😄 Ok, then I'm merging it.

As a comment, the current implementation writes an error in logs in case the config is outdated, but don't terminate execution. Leaving that for now, while it can be improved further when introducing better management of the configs.

PR #1092, opened 2024-08-30
lifetime computation in add_electricity.py

danielelerede-oet:
Unlike the earlier version, this one introduces a check for NaN values directly within the apply function and replaces them with 0 instead of filling with np.inf after the operation (might be especially useful if working with a powerplants.csv originated from a custom_powerplants.csv file). It also doesn't simply replace spaces with 0 but explicitly ensures that any resulting value is numeric before the subtraction.

## Checklist

### DIFF
--- a/scripts/add_electricity.py
+++ b/scripts/add_electricity.py
@@ -432,8 +432,9 @@ def attach_conventional_generators(
         efficiency=ppl.efficiency,
         marginal_cost=ppl.marginal_cost,
         capital_cost=ppl.capital_cost,
-        build_year=ppl.datein.fillna(0).astype(int),
-        lifetime=(ppl.dateout - ppl.datein).fillna(np.inf),
+        build_year=ppl.datein.fillna(0).astype(int),        
+        lifetime=(ppl.dateout.replace(' ', '0')
+                  .apply(lambda x: int(x) if pd.notna(x) else 0) - ppl.datein),
     )
 
     for carrier in conventional_config:
--- a/scripts/add_electricity.py
+++ b/scripts/add_electricity.py
@@ -432,9 +432,11 @@ def attach_conventional_generators(
         efficiency=ppl.efficiency,
         marginal_cost=ppl.marginal_cost,
         capital_cost=ppl.capital_cost,
-        build_year=ppl.datein.fillna(0).astype(int),        
-        lifetime=(ppl.dateout.replace(' ', '0')
-                  .apply(lambda x: int(x) if pd.notna(x) else 0) - ppl.datein),
+        build_year=ppl.datein.fillna(0).astype(int),
+        lifetime=(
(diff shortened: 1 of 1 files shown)


scripts/add_electricity.py
```diff
@@ -433,7 +433,10 @@ def attach_conventional_generators(
         marginal_cost=ppl.marginal_cost,
         capital_cost=ppl.capital_cost,
         build_year=ppl.datein.fillna(0).astype(int),
-        lifetime=(ppl.dateout - ppl.datein).fillna(np.inf),
+        lifetime=(
+            ppl.dateout.replace(" ", "0").apply(lambda x: int(x) if pd.notna(x) else 0)
+            - ppl.datein
+        ),
```
davide-f, line 439: Thanks @danielelerede-oet for the proposal.
Could you explain why np.inf does not work?
Filling with 0 may lead to issues with multi-year optimization as zero lifetime should imply the capacity should be dropped.
It may be a better default approach to fill values according to the costs.csv values.
What do you think?

davide-f:
Hello @yerbol-akhmetov :D
My review is already attached, was it a refuse?

yerbol-akhmetov:
> Hello @yerbol-akhmetov :D My review is already attached, was it a refuse?

It is up to you and @danielelerede-oet, I guess. With a fix of custom_powerplants, it does not cause the issues.

You recommendation on taking a lifetime from `costs.csv` file is great suggestion. But I think it may cause issue in some scenarios for users who already using the current version. Generally, we used to give attention to `DateOut` when we exactly know when plant will be decommissioned from cited sources, otherwise we assume that powerplant will operate. So, in this case have indefinite decommission date convenient. Sometimes, also powerplants (nuclear or coal) are set to operate more than expected lifetime by the government. So taking directly from `costs.csv` might not be accurate, unless we provide option to enable or disable it into config.

davide-f:
@danielelerede-oet what do you think?
I belive the point by @yerbol-akhmetov is a fair point.

Just looking at the code again, doing any filling beyond np.inf in in add_electricity may seem a bit akward.
Numerical filling may be better placed into build_powerplants and I'd be happy to keep np.inf here.

I believe a missing value here may be a side effect of improper cleaning/data filtering in previous steps instead.

danielelerede-oet:
@davide-f @yerbol-akhmetov yes, I agree with both of you. I would keep things as they already are. I was proposing to @ekatef that it would be instead useful to have some checks on any custom_powerplants.csv to ensure that the basic rules  on e.g., capital letters or indications of DateIn/DateOut are respected. What do you think?

davide-f:
> @davide-f @yerbol-akhmetov yes, I agree with both of you. I would keep things as they already are. I was proposing to @ekatef that it would be instead useful to have some checks on any custom_powerplants.csv to ensure that the basic rules on e.g., capital letters or indications of DateIn/DateOut are respected. What do you think?

I agree that it would be interesting.
I believe that may not be a priority though before the merge, what do you all think?

PR #1109, opened 2024-09-25
Split inflow in ppl units + same hydro profile in alt. clustering

### LINKED ISSUE #759: Normalization issue when computing hydro resources
## Checklist

## Describe the Bug

*When running scenarios with custom powerplants for a specific country (Bolivia), hydro availability potential is overestimated compared to the fixed EIA data (normalization: method: eia).
The problem was tracked down to the normalization process, which is done by region/bus in which the normalized inflow is assigned to represent its availability. 
Given that each region might have several hydro power plants (Bolivia does, which is why we found the issue), the following computations assume the same normalized value for each of them. Afterward, when computing the potential of a region, the aggregated values are significatively larger (proportional to the number of hydropower plants existing in each region).
At the moment, to "solve" the issue, we decided to "manually" provide some adapted inflow availabilities to each powerplant. However, an idea that seems to have sense is to add an additional step (function) before the designation of the p_max_pu in the add_electricity rule and reassign inflow values divided based on the number of powerplants recognized in the base_network (still exploring the solution). 
If I'm not mistaken, @hazemful was tackling the issue in one of the periodic meetings however, I thought it might be useful to have the issue posted (just to keep track of it and/or at least corroborate that the hunch I was thinking of would be worth exploring).*

```bash
<paste here>
```

danielelerede-oet:
# Closes #759 .

`build_renewable_profiles`: the computation of `hydro_profile.nc` is not distinguished whether alternative clustering is applied or not.
`add_electricity`: the indexing of the different power plants is performed according to the bus either in case alternative clustering is applied or not and a `hydro_inflow_factor` is computed prior to the computation of `inflow_t` to split the inflow according to the capacity of each different unit of each power plant (beforehand, the same inflow value was assigned to all the units of a single power plant possibly causing an overestimation of the total generation).

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -58,7 +58,7 @@ repos:
 
 # Find common spelling mistakes in comments and docstrings
 - repo: https://github.com/codespell-project/codespell
-  rev: v2.2.1
+  rev: v2.3.0
   hooks:
   - id: codespell
     args: ['--ignore-regex="\b[A-Z]+\b"', '--ignore-words-list=fom,appartment,bage,ore,setis,tabacco,berfore,vor'] # Ignore capital case words, e.g. country codes


scripts/build_renewable_profiles.py
```diff
@@ -575,17 +554,10 @@ def create_scaling_factor(
...
-        ### TODO: quickfix. above case and the below case should by unified
-        if snakemake.params.alternative_clustering == False:
-            filter_bus_to_consider = regions.index.map(
-                lambda bus_id: config.get("extendable", False)
-                | (bus_id in hydro_ppls.bus.values)
-            )
+        filter_bus_to_consider = regions.index.map(
+            lambda bus_id: config.get("extendable", False)
+            | (bus_id in hydro_ppls.bus.values)
+        )
```
davide-f: This formulation still relies on using the inflow for the bus region rather than the specific powerplant, which I understand it is intended.
By the way it is a great improvement of the status quo as it also removes lines of code

davide-f (commented):
Great contribution @danielelerede-oet !
The functionality is neat and it improves the status quo.
Personally, I'd be curious on the implementation by powerplant.

Few comments:
- Have you checked that the indexing with and without alternative clustering is appropriate?
  The fact that the model works may not be sufficient as the index of the regions and powerplants are both numeric and they can be exchanged easily without noticing well
- Have you also checked that by dropping the customized "regions" override the regions are still the same? FYI: I checked the build bus region and it seems consistent and to confirm a numerical check would be great. [if you open the geojson file with vscode you can easily check that; you can install the Geo Data Viewer extension to do so if you don't have it]

I'd like a comment by @cpschau , @hazemakhalek or @energyLS as they have also worked on this.

My perspective is the following:
- this PR improves the status quo by (a) fixing the inflow bug, (b) dropping if/else of alternative clustering and (c) reducing the number of lines of code.
- the improved version would aim to estimate the inflow by powerplant as although this formulation fixes the inflow total, the numeric values are still affected quite a bit especially for large regions, however, the current implementation is worse.

As the above comments are addressed, we could merge this PR

Issue #1073, opened 2024-08-06
Add params config to retrieve_cost_data


davide-f:
## Describe the feature you'd like to see

We shall add the params keyword to the rule retrieve_cost_data.
Without it, the rule is not re-triggered if changes in the config file occur

PR #1120, opened 2024-09-27
get costs_year.csv instead of costs.csv

danielelerede-oet:
As the column year has been dropped from resources/costs.csv this PR proposes a similar approach to pypsa-eur to retrieve a costs.csv file also reporting the reference year for the relative techno-economic assumptions (as from the costs section in the config file).

## Checklist


davide-f:
Great @danielelerede-oet :D
This seems nearly done; probably we can slightly improve it further.
In particular, here:
https://github.com/pypsa-meets-earth/pypsa-earth/blob/3142b1f71a1f73bfe753fd0107aa03c47e6740af/Snakefile#L393-L420

we probably don't need anymore two versions of the same rule where only the output file path is changed.
with the revision of costs with "costs_{year}", the retrieve_cost_data_flexible should match the format of costs.
I'd propose to delete the retrieve_cost_data rule and rename retrieve_cost_data_flexible into retrieve_cost_data.

That should clean the code too.
Great contribution and don't forget a line in release note :)

danielelerede-oet:
Ciao @davide-f . Considering that in the config file the argument to be passed to `costs` is `year` and not `planning_horizons`, I think we should just leave the current `retrieve_cost_data` instead of using `retrieve_cost_data_flexible` and renaming it to `retrieve_cost_data`. Is that right or am I missing anything? :)

danielelerede-oet:
Hi @davide-f, I perfomed some tests and actually keeping `retrieve_cost_data_flexible` only and renaming it to `retrieve_cost_data` is not a viable solution: in that case `resources/costs_{year}.csv` is not generated at all. Using `planning_horizons` as wildcard always results in a KeyError. The only working solution is the one I proposed in the latest commit, but I'm open to discuss it :)

davide-f:
As we are addressing the retrieve_cost_data, it would be nice to add the "params:" snakemake option into the rule as:
```snakemake
params:
  version: config["costs"]["version"]
```
And update the default tech_data version to the latest: 0.9.2
https://github.com/pypsa-meets-earth/pypsa-earth/blob/8d93ddb30dd4f2e6522cac0864a86996789e69c6/config.default.yaml#L358
with version: 0.9.2

This ensures that if the tech_data version is changed, the rule is retriggered

Linked to https://github.com/pypsa-meets-earth/pypsa-earth/issues/1073

Otherwise we can address this in a new version.

davide-f:
Note: current CI passes, but the workflow would not capture scenarios like costs_2040 or costs_2050 unfortunately #1249 should fix that

PR #1134, opened 2024-10-09
Fix issue of adding H2 store in `add_extra_components` rule

yerbol-akhmetov:
Good day. Here I propose to fix the bug related to addition of H2 Store in `add_extra_components.py` script. The issue was found while simulating USA, which had both AC and DC buses. The following error appeared. Which means that for 10 AC buses, it tried to add 12 x, y, country data.
image

Based on the current code, `buses_i` only looks for AC buses, while `bus_sub_dict` considers all. In the US model, we had two additional DC buses (10 vs 12).
https://github.com/pypsa-meets-earth/pypsa-earth/blob/b64623d8d3d8127e186dbe3b9a899e326e1bcb72/scripts/add_extra_components.py#L107-L111

In the new version, I propose to add H2 Stores for AC buses only, if it is intended so.

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -26,6 +26,8 @@ E.g. if a new rule becomes available describe how to use it `make test` and in o
 
 * Include a dedicated cutout for Europe in bundle_config.yaml `PR #1125 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1125>`_
 
+* Fix the mismatch between buses and x, y locations while creating H2 Stores `PR #1134 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1134>`_
+
 PyPSA-Earth 0.4.1
 =================
 
--- a/scripts/add_extra_components.py
+++ b/scripts/add_extra_components.py
@@ -105,7 +105,7 @@ def attach_stores(n, costs, config):
     _add_missing_carriers_from_costs(n, costs, carriers)
 
     buses_i = n.buses.query("carrier == 'AC'").index
-    bus_sub_dict = {k: n.buses[k].values for k in ["x", "y", "country"]}
+    bus_sub_dict = {k: n.buses.loc[buses_i, k].values for k in ["x", "y", "country"]}
 
     if "H2" in carriers:
         h2_buses_i = n.madd("Bus", buses_i + " H2", carrier="H2", **bus_sub_dict)
--- a/scripts/add_extra_components.py
+++ b/scripts/add_extra_components.py
@@ -104,8 +104,8 @@ def attach_stores(n, costs, config):
 
     _add_missing_carriers_from_costs(n, costs, carriers)
 
-    buses_i = n.buses.query("carrier == 'AC'").index
-    bus_sub_dict = {k: n.buses.loc[buses_i, k].values for k in ["x", "y", "country"]}
+    buses_i = n.buses.index
+    bus_sub_dict = {k: n.buses[k].values for k in ["x", "y", "country"]}
 
     if "H2" in carriers:
         h2_buses_i = n.madd("Bus", buses_i + " H2", carrier="H2", **bus_sub_dict)


yerbol-akhmetov:
I have looked into the previously solved networks for US. `H2 Store` was placed to both AC and DC buses. So maybe I will change code accordingly. I will remove AC filter for bus.
```bash
def attach_stores(n, costs, config):
    elec_opts = config["electricity"]
    carriers = elec_opts["extendable_carriers"]["Store"]

    _add_missing_carriers_from_costs(n, costs, carriers)

    buses_i = n.buses.index
    bus_sub_dict = {k: n.buses[k].values for k in ["x", "y", "country"]}
```

davide-f (commented):
Hello @yerbol-akhmetov :D great spot!!!
From a technology point of view, I expect it is more realistic to have it installed only on the AC nodes, currently.
Would you mind force pushing to the previous commit?
The proposal looks good to me :D

ekatef:
Hello!

Great work @yerbol-akhmetov, and thanks a lot for the quick review @davide-f! 🙏🏽 

As a small clarification question, @davide-f could you please give a bit more details on the reasons why we don't need storages at DC buses? I think it could make sense to amend some explanation on that to the code, as H2 storages have been added to both AC and DC nodes both the previous implementation in PyPSA-Earth, and in [the current implementation ](https://github.com/PyPSA/pypsa-eur/blob/54020883ca162482ef20d1c60af7f8574b3fc040/scripts/add_electricity.py#L849-L852)of PyPSA-Eur. Which of course doesn't meant that it can't be improved 😉 

Do you mean that DC lines are normally magistral transmission backbones which are build to carry more or less constant power flow, like connecting a big hydro with a power demand node?

davide-f:
Good point Katia and thanks for the review.
For consistency with pypsa-eur, we can keep it this way.

In the current HVDC network there is nothing attached to the DC bus of hvdc lines but the converters to link the assets to the ac grid.
Electrolyzer cells operate at around 1V and to reach high voltage insulation becomes an issue.

Current technologies are lower scale and typically AC, that's while I proposed AC.
I believe HVDC electrolyzers may be viable only with relatively acceptable large scale adoption of DC, otherwise economies of scale are likely to scale poorly.
However, I expect limited differences in results whatever we choose.

Consistency with pypsa-eur may be advisable.

What do you think @ekatef ?

ekatef:
> Good point Katia and thanks for the review. For consistency with pypsa-eur, we can keep it this way.
> 
> In the current HVDC network there is nothing attached to the DC bus of hvdc lines but the converters to link the assets to the ac grid. Electrolyzer cells operate at around 1V and to reach high voltage insulation becomes an issue.
> 
> Current technologies are lower scale and typically AC, that's while I proposed AC. I believe HVDC electrolyzers may be viable only with relatively acceptable large scale adoption of DC, otherwise economies of scale are likely to scale poorly. However, I expect limited differences in results whatever we choose.
> 
> Consistency with pypsa-eur may be advisable.
> 
> What do you think @ekatef ?

Ahhh, of course! The issue is not DC polarity but the fact that you need to change the voltage level to feed an electrolyzer. Thanks for the explanation!

So, strictly speaking H2 stores should have a bit different parameters for an electrolyzer... Agree however that consistency with PyPSA-Eur is advisable. We also need that fix merged as soon as we can as it's a bug fix which is needed to run the model if DC nodes present in the grid. 

I'd suggest to merge the PR as is, and track the question with the electrolyzers at DC nodes. May it be a good idea to open an issue on that? Or add a point so some existing issue?

In any case, great catch @yerbol-akhmetov and thanks a lot for providing the fix

yerbol-akhmetov:
@davide-f , what did we have decided? Will we have H2 both in AC and DC nodes? Because I see further in sector-coupled scripts, AC nodes are selected for hydrogen building. If we make AC+DC, we will need to change several lines there too in `prepare_sector_network`, because it causes the issue there too. `SMR` and other hydrogen related techs uses AC buses in `prepare_sector_network`.

davide-f:
As also katia agree, let's merge it as is :)
Great and many thanks

PR #1135, opened 2024-10-09
Enable configfile specification for mock_snakemake

yerbol-akhmetov:
Good day. Here I propose to enable specification of configfile while using `mock_snakemake`. It is particularly useful when debugging the scenario with its own run name. Currently, the files are searched in `resources/` or `networks/ ` folders as example. If user had already prepared files in folders under run name, it is not visible. Overwriting config file is enabled in snakemake<8 with overwrite_config parameter, which takes dictionary.

## Checklist

### DIFF
--- a/scripts/_helpers.py
+++ b/scripts/_helpers.py
@@ -574,8 +574,6 @@ def mock_snakemake(
         if isinstance(configfile, str):
             with open(configfile, "r") as file:
                 configfile = yaml.safe_load(file)
-        else:
-            configfile = None
 
         workflow = sm.Workflow(
             snakefile,
--- a/scripts/_helpers.py
+++ b/scripts/_helpers.py
@@ -522,7 +522,7 @@ def get_aggregation_strategies(aggregation_strategies):
     return bus_strategies, generator_strategies
 
 
-def mock_snakemake(rulename, root_dir=None, submodule_dir=None, **wildcards):
+def mock_snakemake(rulename, root_dir=None, submodule_dir=None, configfile=None, **wildcards):
     """
     This function is expected to be executed from the "scripts"-directory of "
     the snakemake project. It returns a snakemake.script.Snakemake object,
@@ -534,6 +534,8 @@ def mock_snakemake(rulename, root_dir=None, submodule_dir=None, **wildcards):
     ----------
     rulename: str
         name of the rule for which the snakemake object should be generated
+    configfile: str
+        path to config file to be used in mock_snakemake
     wildcards:
         keyword arguments fixing the wildcards. Only necessary if wildcards are
         needed.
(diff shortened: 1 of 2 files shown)


scripts/_helpers.py
```diff
@@ -566,9 +570,19 @@ def mock_snakemake(rulename, root_dir=None, submodule_dir=None, **wildcards):
             if os.path.exists(p):
                 snakefile = p
                 break
+
+        if isinstance(configfile, str):
+            with open(configfile, "r") as file:
+                configfile = yaml.safe_load(file)
+        else:
```
davide-f: Rather than else, maybe "elif not isinstance(configfile, dict)" or drop the if case?
By default it is none, so the else case may not be necessary but a dictionary may be provided as well as input.
What do you think?

Almost ready to merge :D

davide-f (commented):
Great again @yerbol-akhmetov !!!
I've added a minor comment; I'd merge the PRs with squash and merge if it is ok for you as it would leave flexibility for the merge issue that is open.

scripts/_helpers.py
yerbol-akhmetov: @davide-f, thanks for the feedback. I have initially meant to omit else case, because default if None. But then I thought what if the user gives something other than str, it can cause error. Therefore I have added additional else with None. Similar thing I have noticed in PyPSA-Eur. Maybe it is better then drop else case?

davide-f: > @davide-f, thanks for the feedback. I have initially meant to omit else case, because default if None. But then I thought what if the user gives something other than str, it can cause error. Therefore I have added additional else with None. Similar thing I have noticed in PyPSA-Eur. 

Let's drop it :) that function is not used by the normal user.
If provided a wrong input, probably it is safe for the code to through an error rather than silently fix it.
Great contribution!

yerbol-akhmetov: @davide-f, ok, thanks, understood.

yerbol-akhmetov: I have dropped `else` condition, while keeping `if`, so to make sure, config is read only if the path is given as string.

davide-f:
@yerbol-akhmetov apologies, by merging the other PR this has let to conflicts. Can you fix it?
As you do that, I squash and merge this too.
I don't want to do that to avoid being co-author of the PR while you worked on it.

yerbol-akhmetov:
> @yerbol-akhmetov apologies, by merging the other PR this has let to conflicts. Can you fix it? As you do that, I squash and merge this too. I don't want to do that to avoid being co-author of the PR while you worked on it.

Thanks, @davide-f. I have resolved the merge conflicts and updated the branch with main.

davide-f (approved):

PR #1137, opened 2024-10-10
Use absolute path for `config.default.yaml` in `_helpers.py`

yerbol-akhmetov:
Good day. Here I propose to use absolute path to `config.default.yaml` in `_helpers.py` script. Such change is important while importing PyPSA-Earth rules elsewhere and running then in external directories within another Snakemake. This is helpful for setting workflow. 

## Checklist

### DIFF
--- a/scripts/retrieve_databundle_light.py
+++ b/scripts/retrieve_databundle_light.py
@@ -93,7 +93,7 @@ from _helpers import (
     create_country_list,
     create_logger,
     progress_retrieve,
-    PREFIX
+    BASE_DIR
 )
 from google_drive_downloader import GoogleDriveDownloader as gdd
 from tqdm import tqdm
@@ -144,7 +144,7 @@ def download_and_unzip_zenodo(config, rootpath, hot_run=True, disable_progress=F
     """
     resource = config["category"]
     file_path = os.path.join(rootpath, "tempfile.zip")
-    destination = os.path.join(PREFIX, config["destination"])
+    destination = os.path.join(BASE_DIR, config["destination"])
     url = config["urls"]["zenodo"]
 
     if hot_run:
@@ -189,7 +189,7 @@ def download_and_unzip_gdrive(config, rootpath, hot_run=True, disable_progress=F
     """
     resource = config["category"]
     file_path = os.path.join(rootpath, "tempfile.zip")
-    destination = os.path.join(PREFIX, config["destination"])
+    destination = os.path.join(BASE_DIR, config["destination"])
     url = config["urls"]["gdrive"]
 
     # retrieve file_id from path
@@ -267,7 +267,7 @@ def download_and_unzip_protectedplanet(
     """
     resource = config["category"]
     file_path = os.path.join(rootpath, "tempfile_wpda.zip")
-    destination = os.path.join(PREFIX, config["destination"])
+    destination = os.path.join(BASE_DIR, config["destination"])
     url = config["urls"]["protectedplanet"]
 
     def get_first_day_of_month(date):
@@ -439,7 +439,7 @@ def download_and_unzip_direct(config, rootpath, hot_run=True, disable_progress=F
     True when download is successful, False otherwise
     """
     resource = config["category"]
-    destination = os.path.join(PREFIX, config["destination"])
+    destination = os.path.join(BASE_DIR, config["destination"])
     url = config["urls"]["direct"]
 
     file_path = os.path.join(destination, os.path.basename(url))
@@ -493,7 +493,7 @@ def download_and_unzip_hydrobasins(
     True when download is successful, False otherwise
     """
     resource = config["category"]
-    destination = os.path.join(PREFIX, config["destination"])
+    destination = os.path.join(BASE_DIR, config["destination"])
     url_templ = config["urls"]["hydrobasins"]["base_url"]
     suffix_list = config["urls"]["hydrobasins"]["suffixes"]
 
@@ -544,7 +544,7 @@ def download_and_unzip_post(config, rootpath, hot_run=True, disable_progress=Fal
     True when download is successful, False otherwise
     """
     resource = config["category"]
-    destination = os.path.join(PREFIX, config["destination"])
+    destination = os.path.join(BASE_DIR, config["destination"])
 
     # load data for post method
     postdata = config["urls"]["post"]
@@ -793,8 +793,8 @@ def datafiles_retrivedatabundle(config):
 
 
 def merge_hydrobasins_shape(config_hydrobasin, hydrobasins_level):
-    basins_path = os.path.join(PREFIX, config_hydrobasin["destination"])
-    output_fl = os.path.join(PREFIX, config_hydrobasin["output"][0])
+    basins_path = os.path.join(BASE_DIR, config_hydrobasin["destination"])
+    output_fl = os.path.join(BASE_DIR, config_hydrobasin["output"][0])
 
     files_to_merge = [
         "hybas_{0:s}_lev{1:02d}_v1c.shp".format(suffix, hydrobasins_level)
--- a/scripts/retrieve_databundle_light.py
+++ b/scripts/retrieve_databundle_light.py
@@ -89,11 +89,11 @@ import geopandas as gpd
 import pandas as pd
 import yaml
 from _helpers import (
+    BASE_DIR,
     configure_logging,
     create_country_list,
     create_logger,
     progress_retrieve,
-    BASE_DIR
 )
 from google_drive_downloader import GoogleDriveDownloader as gdd
 from tqdm import tqdm
--- a/scripts/retrieve_databundle_light.py
+++ b/scripts/retrieve_databundle_light.py
@@ -93,6 +93,7 @@ from _helpers import (
     create_country_list,
     create_logger,
     progress_retrieve,
+    PREFIX
 )
 from google_drive_downloader import GoogleDriveDownloader as gdd
 from tqdm import tqdm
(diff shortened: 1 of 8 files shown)


yerbol-akhmetov:
We are still re-testing all rules execution with @GbotemiB. Unfortunately, these small changes were not sufficient for some rules, while it worked in one machine. We will let know when it is 100% ready. 
@davide-f @ekatef

yerbol-akhmetov:
With recent commit, prefix was identified for `retrieve_databundle_light` rule. Generally, a prefix is added to input and outputs of the imported rules automatically when prefix is defined as:
```bash
module pypsa_earth:
    snakefile:
        "submodules/pypsa-earth/Snakefile"
    config:
        config
    prefix:
        "submodules/pypsa-earth"
```
Those inputs and outputs (with prefix) are accessed properly when input and outputs of rules are defined explicitly as in almost all rules, except `retrieve_databundle_light` rule. Therefore, prefix is added inside of the `retrieve_databundle_light.py` script. It helps to match the inputs of following rule to outputs of preceding rule if Snakefile import is done. If snakemake is run in pypsa-earth directory (as regularly), then prefix is an empty string `` and rules are run regularly.

scripts/retrieve_databundle_light.py
```diff
@@ -143,7 +143,7 @@ def download_and_unzip_zenodo(config, rootpath, hot_run=True, disable_progress=F
     """
     resource = config["category"]
     file_path = os.path.join(rootpath, "tempfile.zip")
-    destination = os.path.relpath(config["destination"])
+    destination = os.path.join(PREFIX, config["destination"])
```
davide-f: Why do you propose PREFIX here instead of BASE_DIR as in the other cases?

yerbol-akhmetov: Thanks, @davide-f. That is the valid point. I have used BASE_DIR as the first try. And it worked fine as standalone for rule execution. But to make sure rules see each other and trigger, outputs and inputs needs to match. So `submodules/pypsa-earth/data/eez/eez_v11.gpkg` is not equal to absolute path `/home/yerbol/efuels/.../submodules/pypsa-earth/data/eez/eez_v11.gpkg` when it comes to input outputs of snakemake. 

With BASE_DIR instead `build_shapes` were not triggering `retrieve_databundle_light`. But `retrieve_databundle_light` worked just fine separately when calling it. CI test actually helped me to figure that out (https://github.com/pypsa-meets-earth/pypsa-earth/actions/runs/11314101704/job/31463655978)

BASE_DIR is good for handling files inside of the script, that are not tight to input/outputs of snakemake. If it is input/output, only PREFIX makes the full workflow intact and preserve rule connections.

davide-f: Do you mind talking on thursday about the prefix?
As PREFIX is taken out from the log inside the retrieve function, we do PREFIX = BASE_DIR, the result should be equivalent.

Moreover, don't have they the same meaning? I'd avoid a different naming unless necessary

yerbol-akhmetov: @davide-f, thanks, yes, we can discuss it in details in Thursday. In short, when importing the rules and defining prefix, all rule's inputs and outputs are appended with prefix, if they are explicitly set as input and output. 

I have performed several tests with `retrieve_databundle_light` and `build_shapes` rules using BASE_DIR and PREFIX. I have deleted `data/eez` which is used by `build_shapes`, then I run `build_shapes` rule. 
1) When using BASE_DIR inside `retrieve_databundle_light.py`, `build_shapes` were throwing error of missingInputException. Although `retrieve_databundle_light` rule can be executed separately just fine. Its outputs were absolute path. `build_shapes` could then be executed fine, but workflow is broken. Input of `build_shapes` is not exact to output of `retrieve_databundle_light`, although they are technically same file (one relative path PREFIX, and absolute path BASE_DIR).
2) When using PREFIX, outputs of `retrieve_databundle_light` perfectly matches with input of `build_shapes`. As a result, missing `data/eez` easily triggers `retrieve_databundle_light` rule when `build_shapes` are executed.

If having two names (BASE_DIR and PREFIX) is not preferred. We should have one: PREFIX. I will think about defining it in `_helpers.py` and replacing where BASE_DIR is used.

davide-f: Ok, but the content of variable PREFIX is the same as BASE_DIR in this PR now right?
How would the content of variable PREFIX change in the tests you are performing?
Do you have a branch somewhere where I can test and see what you are trying?

What operating system are you using for the tests? in windows there were known issues also on recognizing paths due to snakemake.

yerbol-akhmetov: > Ok, but the content of variable PREFIX is the same as BASE_DIR in this PR now right? How would the content of variable PREFIX change in the tests you are performing? Do you have a branch somewhere where I can test and see what you are trying?
> 
> What operating system are you using for the tests? in windows there were known issues also on recognizing paths due to snakemake.

Hi, @davide-f, the content of PREFIX is `submodules/pypsa-earth`, while the content of BASE_DIR is `/home/user/.../some_project/submodule/pypsa-earth`. PREIX does not changes it is static relative path to pypsa-earth.

Regarding testing, I am testing on Windows and Linux in parallel, @GbotemiB was testing in Mac OS. The procedure and errors were consistent across all platforms. The issue of using BASE_DIR not triggering `retrieve_databundle_light` by following `build_shapes` rule had appeared in Linux and Windows as I tested.

To test the procedure, you can clone our `efuels-supply-potential` repository (https://github.com/open-energy-transition/efuels-supply-potentials) and checkout to `snakemake_workflow` branch. After submodule init and update of `submodule/pypsa-earth`, you will need to checkout to my branch `absolute_path_fixes`. You can also test with `BASE_DIR` by going back into commits. Please enable retrieving databundle and then you can run in `efuels-supply-potential` directory:
```bash
snakemake -call build_shapes --configfile configs/calibration/config.base.yaml
```
If you look into inputs and outputs of rules, you will see `submodule/pypsa-earth/...`. Log files directory is also changes from `log/build_shapes` to `submodules/pypsa-earth/log/...`. By splitting by `log` and taking first part [0], makes us to take `submodule/pypsa-earth` as PREFIX. If rules are run in pypsa-earth directory, splitting by log and taking initial part results in `` empty. So PREFIX is empty and everything runs as usual. 

As I have mentioned, I am planning to define PREFIX in `_helpers.py` by taking difference between where snakemake is run (RUN_DIR) and pypsa-earth base directory (BASE_DIR). It will give difference as relative path of `submodules/pypsa-earth`.

If you want to try with BASE_DIR, I would suggest to delete one of the data (eg. `data/eez` ) and go back several commits back. Then running `build_shapes` will not trigger r`etrieve_databundle_light`. But if `PREFIX` is used then it can trigger.

yerbol-akhmetov: @davide-f, I am planning to get rid of using BASE_DIR in scripts, instead PREFIX will be used in each rule for consistency. The PREFIX will be defined `_helpers.py` and will be determined using BASE_DIR. What do you think?

davide-f: I've tested the repo and trying the procedure you suggested.
I still had issues with the workflow and I believe the reason is not really BASE_DIR/PREFIX. In this case we probably can always use BASE_DIR and it should be fine.

The problem lays into the function `datafiles_retrivedatabundle` that creates the vector of the files that retrieve_databundle shall produce.
The paths are by construction relative paths assuming the cwd is pypsa-earth, because they rely directly on the bundle_configs settings.

Probably the best solution lays in including in such function what you now call prefix.
I'd recommend however to find another name as prefix in pypsa-eur means another thing.
e.g. path_to_parent ? feel free to propose an alternative; it may also be a function.

yerbol-akhmetov: Thanks for the review, @davide-f. This "PREFIX" (path_to_parent) need to be added not only to `retrieve_databudle_light`, but other scripts, such as `build_shapes` and `build_demand_profiles`. I will look for `datafiles_retrivedatabundle` function to include path_to_parent.

yerbol-akhmetov: > I've tested the repo and trying the procedure you suggested. I still had issues with the workflow and I believe the reason is not really BASE_DIR/PREFIX. In this case we probably can always use BASE_DIR and it should be fine.
> 
> The problem lays into the function `datafiles_retrivedatabundle` that creates the vector of the files that retrieve_databundle shall produce. The paths are by construction relative paths assuming the cwd is pypsa-earth, because they rely directly on the bundle_configs settings.
> 
> Probably the best solution lays in including in such function what you now call prefix. I'd recommend however to find another name as prefix in pypsa-eur means another thing. e.g. path_to_parent ? feel free to propose an alternative; it may also be a function.

Can you please show what type of issue do you have? Have you tried only power model? If you can show the error message, then I can figure that out. Everything worked fine in my PCs (linux and windows) and Emmanuel's one. @davide-f

davide-f: [test_yerbol.txt](https://github.com/user-attachments/files/17488224/test_yerbol.txt)
Attached is the output in linux for simplicity.
Have you tested from a fresh repo?

For transparency, I haven't added any manual changes into the snakefile. This should not be needed.

yerbol-akhmetov: @davide-f, I have removed PREFIX and replaced with BASE_DIR everywhere. Also added BASE_DIR `build_renewable_profiles`. Then tested the full power workflow from subworkflow and pypsa-earth directory itself. Both worked as intended both in Linux and Windows.

davide-f (commented):
Great @yerbol-akhmetov :D
This PR is ready to go functionality-wise! :D

Would you mind to fix the merge conflict? Than this is merged :)
Many thanks!

yerbol-akhmetov:
> Great @yerbol-akhmetov :D This PR is ready to go functionality-wise! :D
> 
> Would you mind to fix the merge conflict? Than this is merged :) Many thanks!

Thanks, @davide-f, for great help in testing and suggestions. I have resolved the conflicts and updated the branch.

davide-f (approved):

davide-f:
Thanks @yerbol-akhmetov :D

Issue #1051, opened 2024-07-01
Simplifying polygons in build_shapes.py leads to holes and overlaps


jome1:
- [x] I am using the current [`main`](https://github.com/pypsa-meets-earth/pypsa-earth/tree/main) branch or the latest [release](https://github.com/pypsa-meets-earth/pypsa-earth/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-earth` environment](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

The script `build_shapes.py` downloads the GADM layers from https://geodata.ucdavis.edu/gadm/gadm4.1/gpkg/{GADM_filename}.gpkg. 
When inspecting the created `gadm_shapes.geojson` closely, one can identify holes and overlaps between the polygons describing GADM layer 1 (here called "governorates"). See the following picture:

![image](https://github.com/pypsa-meets-earth/pypsa-earth/assets/49280197/4c04eb8c-00b0-43d8-8a7a-7390b71f6fd6)

I think this is because within the script `build_shapes.py` the downloaded polygons are [simplified](https://github.com/pypsa-meets-earth/pypsa-earth/blob/bae5586a3186955f7788a8ab81b26e3f3a7a70a1/scripts/build_shapes.py#L236C1-L252C47). 

The following picture compares the original boarders of the governorates with the simplified boardes. Thereby the newly created and simplified polygons at some locations either overlap or leave holes in between. 

![image](https://github.com/pypsa-meets-earth/pypsa-earth/assets/49280197/bf681075-6e1e-4c35-8f95-b70fbfe5cae6)

In other words: The polygons are all simplified seperately. So shared boarders (edges) are not considered. 

This problem might also play a role when using GADM layers as the fundamental shapes as for example mentioned by @pz-max in https://github.com/pypsa-meets-earth/pypsa-earth/issues/382 and when building the bus regions.

## Possible solution

Polygons can only be simplified where they are not sharing the line with another polygon (so skip the simplification for shared boarders) OR the conterminous (neighbouring) polygons have to be simplified to the same line.

I did a little internet search.
Maybe [shapely.snap](https://shapely.readthedocs.io/en/stable/reference/shapely.snap.html) could help to solve this problem.

helpful mini threads about this phenomenon:
- https://gis.stackexchange.com/q/188352
- https://gis.stackexchange.com/q/178 

[...]

PR #1138, opened 2024-10-10
Introducing an option for the user to decide on simplifying GADM shapes

SermishaNarayana:
# Closes # (if applicable).

## Checklist


config.default.yaml
```diff
@@ -106,6 +106,7 @@ cluster_options:
 
 build_shape_options:
   gadm_layer_id: 1 # GADM level area used for the gadm_shapes. Codes are country-dependent but roughly: 0: country, 1: region/county-like, 2: municipality-like
+  simplify_gadm: false # When true, shape polygons are simplified else no
```
davide-f: Great @SermishaNarayana :D
What about we make this option numeric? Like we can rename it as simplify_tolerance, that by default it is 0.01 (current default value) and if the value is False or <=0, then the simplification does not occur?

SermishaNarayana:
Screenshot 2024-10-10 at 6 03 39 PM
@ekatef Here, I have plotted the difference in the shape files with and without simplifying the GADM shapes for the US. The differences lie mostly in the consideration of some small islanded lands and in the borders of the US states

davide-f (commented):
Great contribution :D added a comment, please also add a release note.
We are very close I believe :D

config.default.yaml
SermishaNarayana: @davide-f Yes, I am planning to add it and it is already in progress. I was trying to understand the reason for the explosion of the regions_onshore.geojson file without simplification in the meantime.

ekatef: @SermishaNarayana @davide-f thanks a lot for taking care about that!

I'm afraid that it may not work with having a numerical option to simplify due to some `_simplify_polys( )` function itself... 

@jome1 did [a great investigation](https://github.com/pypsa-meets-earth/pypsa-earth/issues/1051#issue-2383439654) on behaviour of `_simplify_polys( )` which has demonstrated that all the polygons are being simplified independently of each other. That can lead to emerging a number of "enclaves" across the border of the regions.

I suspect explosion of the polygons you observed can be related to that: once we call `_simplify_polys( )` it results in emerging of large amount of such enclaves because the geometry is quite complex. The good news is that the issue is likely to be resolved with the next release of `shapely` which should contain an improved simplification algorithm

So, I'd probably leave a boolean flag for now and returned to the idea to add a numeric parameter, once an advanced simplification option will be available in `shapely`. What do you think?

SermishaNarayana: @ekatef 

If I understand it right, the problem you mention occurs when simplifying the polygon and using the _simplify_polys() function in `build_shapes` . But in the current observation of the explosion of regions_onshore.geojson filesize, the issue occurs when we have the GADM simplification turned off. The code in this case skips _simplify_polys(). Is the issue somehow still related then?

Please correct me if I understood it wrong

ekatef: @SermishaNarayana you are absolutely right, that is `_simplify_polys( )` which creates issues in the #1051

Have misunderstood that you are observing the file expansion in case `_simplify_polys( )` is bypassed. Not sure if that's really related to #1051, then. Thank you so much for the explanation!

Just to be sure that I get the problem: you are also observing some strange geometry effects, right? If that is the case could you please post a picture there? Would be great to understand what is going on with geometries

SermishaNarayana: @ekatef Thanks for clarifying :)

The geometric effects that I observed are very similar to what has been mentioned as the issue with #1051 in
 _simplify_polys(). I was observing overlaps and holes as mentioned in the issue.
 
Screenshot 2024-10-15 at 8 11 05 AM

SermishaNarayana: @davide-f @ekatef 

The other issue that I observed is in the clustered shape file (regions_onshore_elec_s_{clusters}.geojson)

In the runs for US, I have used 50 clusters (to be equal to number of US states). The observation is that there is an overlap in the statewise shapes itself. Some of the states are extending much beyond their actual borders

Screenshot 2024-10-15 at 1 56 03 PM

In the picture, the geometry of the region "US.26_1_AC" which refers to the state of Florida as tagged in the regions_onshore_elec_s_50.geojson file has been plotted. But the geometry of the state is projected to cover many more states of the US

jome1: I am not sure but it might be indeed related to https://github.com/pypsa-meets-earth/pypsa-earth/issues/1051. I explain it myself, that due to the holes and overlaps in the fundamental/basis shapes, the clustering has some erros. The clustering happens on the basis of the shapes of the US states in this case, so the shapes of the US states itself should be flawless to receive flawless clusters.

davide-f: Thanks everyone.
Indeed this is weird and the combination of multiple issues:
1. the tolerance used for the gadm (when 0 or negative or False, no simplification is applied)
2. the algorithm in simplify polys that sometimes may lead to "normal" overlaps though not wanted
3. the same algorithm that sometimes lead to weird clustered shapes

In this PR we probably shall focus on 1 while tracking the other issues in the other PRs, otherwise it feels a bit overwhelming and hopefully along the way shapely improves the algorithm too.

I'd propose to use a numeric (or false) value instead of a pure boolean as it is an inexpensive feature that only improves the existing also giving quite flexibility.
With the boolean we only allow (a) enable the feature with predefined tolerance or (b) disabling the feature entirely that is a bit extreme.
With the number, the tolerance itself can be calibrated.

The tolerance has also implications into the geospatial comparison of subsequent rules, therefore the higher the simplification the better the performances, at the cost of accuracy though.
That's why customizing the number may be handy

davide-f: > In the picture, the geometry of the region "US.26_1_AC" which refers to the state of Florida as tagged in the regions_onshore_elec_s_50.geojson file has been plotted. But the geometry of the state is projected to cover many more states of the US

@SermishaNarayana could you share the geojson?
Do I understand well that the gadm file of florida contains the broader US? What is the large black shape you are showing? I understand the borders but not much the large states though

The original build_shapes should not overlap theoretically, but in practise it may be.

having slight overlap across boundaries may not lead to issues; alteratively, we can ensure no overlap in build_shapes with a heuristic, but not sure that is needed.

SermishaNarayana: @davide-f 

I have attached the regions_onshore_elec_s_50.geojson file when simplifying the GADM shapes with a tolerance of 0.01 (default)
[regions_onshore_elec_s_50_simplify_gadm.geojson.zip](https://github.com/user-attachments/files/17391828/regions_onshore_elec_s_50_simplify_gadm.geojson.zip)

The big black shape in the previous comment shows the geometric shape of the state of Florida as classified in this particular geojson file

davide-f: @SermishaNarayana can you share also the gadm_shapes and the whole shapes folder? simplify_polys apply on gadm_shapes first and with alternative_clustering I expected that to be used.
The output of bus_regions is regions_onshore; then other rules apply and they edit them further.
You may have found another bug later on down the chain, but it is very likely that it is not linked to this PR

davide-f: > @SermishaNarayana can you share also the gadm_shapes and the whole shapes folder? simplify_polys apply on gadm_shapes first and with alternative_clustering I expected that to be used.
The output of bus_regions is regions_onshore; then other rules apply and they edit them further.
You may have found another bug later on down the chain, but it is very likely that it is not linked to this PR

SermishaNarayana: @davide-f 

Sure, I have attached all the files in the shape folder
[shape_files.zip](https://github.com/user-attachments/files/17394720/shape_files.zip)

Also, with respect to using a numerical option for tolerance, I am planning to raise it as a separate PR. Is that alright? And also to clarify (a tolerance value of 0 has the same issues as not allowing the GADM shapes to be simplified )

davide-f: The gadm_shapes here are fine, so this PR should be ok. The problem is likely later on.

Let's also keep the flag true/false and open an issue for custom tolerances and minarea.

Feel free to turn this PR as ready for review whenever ready. Please add a line in release note :)

SermishaNarayana: @davide-f Apologies, The files I had uploaded earlier were the ones with simplify_gadm option turned ON. I am attaching another set of files with simplify_gadm option turned OFF.

[shape_files_non_simplified.zip](https://github.com/user-attachments/files/17395648/shape_files_non_simplified.zip)

Can you please verify if these files look alright? 
I shall add the release note :)

davide-f: Great! At first glance it looks good .
Here we have a great comparison and justification on why to add custom tolerances.
The feedback I see is the following:
1. the number of shapes in the two dataset is the same: 55 shapes in both. This means that the shapes are conserved.
2. the "simplified" one is 0.7Mb, while the non-simplified is 73 Mb. This is a good reason on why to use simplification. For some applications, however, the default value may be too much and lower simplication may be accepted.

The shapes are consistent and the bug you found later on in the simplification is not visible here. This suggests that the PR is ready to go functionality-wise :)

davide-f: By default, let's enable the simplification. This on average speeds up computation and is desirable

davide-f (commented):
Hello @SermishaNarayana :)
Thank for the contribution and the PR is ready to merge.

Unfortunately git history is not the best; for future PRs, it is a great idea if you update your main and clean it.

Moreover, please fix merge conflicts into the README; I see the documenter is leading to some weird effects to be fixed in the future.

davide-f:
Hello @SermishaNarayana :D
Just commentng to check your expectations on this PR. If you wish I can cherry-pick this and finalize or feel free to resolve the comments above and we can finalize :)

SermishaNarayana:
> Hello @SermishaNarayana :D Just commentng to check your expectations on this PR. If you wish I can cherry-pick this and finalize or feel free to resolve the comments above and we can finalize :)

Hi @davide-f ! Apologies, I somehow missed the comment on the merge conflicts with READme earlier. I have resolved it now.

davide-f:
Thanks! @SermishaNarayana :D merging

PR #1165, opened 2024-11-02
Convert conventional ppls from generators to link with existing capacity

yerbol-akhmetov:
Good day. I am proposing to convert conventional generators (e.g. coal, lignite, biomass, oil and etc) from generators to link with stores and generators. Currently it is implemented for OCGT in `add_generation` function of `prepare_sector_networks` and oil in `add_oil` functions. However, the current implementation is not fully correct as existing OCGT generators from elec network is not removed when added as a link. 

This PR aims to convert technologies listed in `conventional_generation` section of `sector` section of config to links, and remove from generators. Having stores and links for conventional carriers is practical when defining new technologies that uses those carriers (eg. carbon capture). In addition, it will be possible to keep existing capacities and transfer them to links based on the option in config.

## Checklist

### DIFF
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -664,6 +664,8 @@ def define_spatial(nodes, options):
         spatial.coal.locations = ["Africa"]
         spatial.coal.industry = ["Africa coal for industry"]
 
+    spatial.coal.df = pd.DataFrame(vars(spatial.coal), index=spatial.nodes)
+
     # lignite
 
     spatial.lignite = SimpleNamespace()
@@ -674,6 +676,8 @@ def define_spatial(nodes, options):
     else:
         spatial.lignite.nodes = ["Africa lignite"]
         spatial.lignite.locations = ["Africa"]
+
+    spatial.lignite.df = pd.DataFrame(vars(spatial.lignite), index=spatial.nodes)
 
     return spatial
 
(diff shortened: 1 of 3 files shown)


yerbol-akhmetov:
Similar to `spatial_oil` and `spatial_gas`, I have added such options for lignite and coal. I have set `spafial_coal: true` and `spatial_lignite: false` by default in `config.default.yaml`. But that can be discussed.
I have a question: Why do we have distributed oil by default (`spatial_oil: true`)? We also have oil reserves as 100 TWh in the config (`fossil_reserves`). In the code, currently 100TWh is set as `e_initial` at all distributed stores. So each store will have 100TWh of oil equivalent available, making total volume depend on number of stores. Should that number split by buses? I suppose so. Method of split is another question. Maybe it is better to set spatial_oil: false by default?

yerbol-akhmetov:
Another thing that I want to discuss is the possibility to rename spatial bus names from `Africa` to `Region` (or Regional). I think it can be more neutral. Eg. we have `Africa solid biomass` node, it can be `Regional solid biomass`.

yerbol-akhmetov:
Hi, @davide-f. It would be great to hear your opinion on this PR.

scripts/prepare_sector_network.py
```diff
@@ -703,6 +644,68 @@ def define_spatial(nodes, options):
...
+    # oil
+
+    spatial.oil = SimpleNamespace()
+
+    if options["oil"]["spatial_oil"]:
+        spatial.oil.nodes = nodes + " oil"
+        spatial.oil.locations = nodes
+    else:
+        spatial.oil.nodes = ["Africa oil"]
+        spatial.oil.locations = ["Africa"]
```
davide-f: We can take the change to replace this into World/Earth oil instead of Africa.
The same applies below

davide-f (commented):
Hello @yerbol-akhmetov :D
The PR looks quite nice to me.

Beyond the comments here; I'd recommend to check at least a test config file, to ensure the new code is well tested and included in the CI. For example the spatial_lignite: false.

Great contribution!!!

Update: In agreement to the -sec meeting, we may have also a view of this PR by hazem

yerbol-akhmetov:
Thanks, @davide-f, for suggestion. I have renamed all Africa instances to Earth.

davide-f:
Merging :D Congratulations!

PR #1166, opened 2024-11-04
Drop non string entries in country column of CO2_emissions data in `prepare_transport_data_input`

yerbol-akhmetov:
Good day. Here I propose to update `download_co2_emissions` function in `prepare_transport_data_input.py` script to drop entries where country column contains not string. This issue appeared very recently, as I have noticed it in the CI check. Great thanks for @finozzifa, who pointed out the error location and proposed the solution.

## Checklist

### DIFF
--- a/scripts/prepare_transport_data_input.py
+++ b/scripts/prepare_transport_data_input.py
@@ -103,6 +103,13 @@ def download_CO2_emissions():
     # Drop region names that have no ISO2:
     CO2_emissions = CO2_emissions[CO2_emissions.country != "not found"]
 
+    # Drop region names where country column contains list of countries
+    CO2_emissions = CO2_emissions[
+        CO2_emissions.apply(
+            lambda x: False if not isinstance(x.country, str) else True, axis=1
+        )
+    ]
+
     return CO2_emissions
 
 
--- a/scripts/prepare_transport_data_input.py
+++ b/scripts/prepare_transport_data_input.py
@@ -104,7 +104,7 @@ def download_CO2_emissions():
 
     # Drop region names where country column contains list of countries
     CO2_emissions = CO2_emissions[
-        CO2_emissions.country.apply(lambda x: isinstance(x, str), axis=1)
+        CO2_emissions.country.apply(lambda x: isinstance(x, str))
     ]
     return CO2_emissions
 
--- a/scripts/prepare_transport_data_input.py
+++ b/scripts/prepare_transport_data_input.py
@@ -104,11 +104,8 @@ def download_CO2_emissions():
 
     # Drop region names where country column contains list of countries
     CO2_emissions = CO2_emissions[
-        CO2_emissions.apply(
-            lambda x: False if not isinstance(x.country, str) else True, axis=1
-        )
+        CO2_emissions.country.apply(lambda x: isinstance(x, str), axis=1)
     ]
-
     return CO2_emissions
 
 
--- a/scripts/prepare_transport_data_input.py
+++ b/scripts/prepare_transport_data_input.py
@@ -95,9 +95,8 @@ def download_CO2_emissions():
     # Add ISO2 country code for each country
     CO2_emissions = CO2_emissions.rename(columns={"Country Name": "Country"})
     cc = coco.CountryConverter()
-    Country = pd.Series(CO2_emissions["Country"])
-    CO2_emissions["country"] = cc.pandas_convert(
-        series=Country, to="ISO2", not_found="not found"
+    CO2_emissions.loc[:, "country"] = cc.pandas_convert(
+        series=CO2_emissions["Country"], to="ISO2", not_found="not found"
     )
 
     # Drop region names that have no ISO2:
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -42,6 +42,8 @@ E.g. if a new rule becomes available describe how to use it `make test` and in o
 
 * Fix bugs in `prepare_sector_network.py` related to links with H2 buses and bug of re-addition of H2 and battery carriers in present `PR #1145 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1145>`_
 
+* Drop entries that contain non-string elements in country column of `CO2_emissions_csv` data in `prepare_transport_data_input.py` script `PR #1166 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1166>`_
+
 PyPSA-Earth 0.4.1
 =================
 


yerbol-akhmetov:
Hi, @hazemakhalek. Can you please review the bug fix provided in this PR? Currently CI fails due to that too.

hazemakhalek:
Hi Yerbol, this part I believe is legacy code that is not used anymore. Can you please confirm that in prepare_sector_network.py (this should be the only script we use this output file) and then we can omit the part of the emissions here altogether?

yerbol-akhmetov:
> Hi Yerbol, this part I believe is legacy code that is not used anymore. Can you please confirm that in prepare_sector_network.py (this should be the only script we use this output file) and then we can omit the part of the emissions here altogether?

Hi, @hazemakhalek. I have checked the `prepare_transport_data_input` rule. Its output is used in `prepare_transport_data`. This rule has an output of transport data (`resources/demand/transport_...`), which is used in `prepare_sector_network` rule to define the demand for land transport (using `add_land_transport`). I think it is not redundant, no?

hazemakhalek:
Hi Yerbol, thanks for taking the time to check it. I meant only the CO2 emission part, not the whole file. Is it still being used in prepare_sector_network?

yerbol-akhmetov:
> Hi Yerbol, thanks for taking the time to check it. I meant only the CO2 emission part, not the whole file. Is it still being used in prepare_sector_network?

Sorry @hazemakhalek for not mentioning, `CO2_emissions_csv` data have `average_fuel_efficiency` column, which is used in further rule `prepare_transport_data`:
https://github.com/pypsa-meets-earth/pypsa-earth/blob/107545678b43ce8304da1dae7b6fecc986849924/scripts/prepare_transport_data.py#L99-L112

scripts/prepare_transport_data_input.py
```diff
@@ -103,6 +103,13 @@ def download_CO2_emissions():
     # Drop region names that have no ISO2:
     CO2_emissions = CO2_emissions[CO2_emissions.country != "not found"]
 
+    # Drop region names where country column contains list of countries
```
davide-f, line 103: This is an interesting issue. What line fails?

I see it is linked to the code above:
```
    CO2_emissions = CO2_emissions.rename(columns={"Country Name": "Country"})
    cc = coco.CountryConverter()
    Country = pd.Series(CO2_emissions["Country"])
    CO2_emissions["country"] = cc.pandas_convert(
        series=Country, to="ISO2", not_found="not found"
    )
```
It is a bit surprising the problem you find as pandas_convert should thorugh "not found" when the conversion fails.
This code could be improved; for example, there is no need to re-create the pd.Series of a Series object.
Moreover, the use of CO2_emissions["country"] = serie may lead to unexpected behavior when index does not match; plus the format df[col] = ... may change behavior, conversely to the use f loc[:, col]
These may be the reason of the issue you experience and lead to improper data matching.

If that is not sufficient, also discarding further unexpected values is clearly fine

davide-f (commented):
Added a comment; what do you think?

scripts/prepare_transport_data_input.py
yerbol-akhmetov: > This is an interesting issue. What line fails?
> 
> I see it is linked to the code above:
> 
> ```
>     CO2_emissions = CO2_emissions.rename(columns={"Country Name": "Country"})
>     cc = coco.CountryConverter()
>     Country = pd.Series(CO2_emissions["Country"])
>     CO2_emissions["country"] = cc.pandas_convert(
>         series=Country, to="ISO2", not_found="not found"
>     )
> ```
> 
> It is a bit surprising the problem you find as pandas_convert should thorugh "not found" when the conversion fails. This code could be improved; for example, there is no need to re-create the pd.Series of a Series object. Moreover, the use of CO2_emissions["country"] = serie may lead to unexpected behavior when index does not match; plus the format df[col] = ... may change behavior, conversely to the use f loc[:, col] These may be the reason of the issue you experience and lead to improper data matching.
> 
> If that is not sufficient, also discarding further unexpected values is clearly fine

Hi, @davide-f. Currently CIs are failing in `prepare_transport_data_input` rule. In particular, `country` column contains not only two letter codes (eg. `AU` or `DE`). But it also contains list `[IN, CN]`. It is because underlying data have such merged data. Here is the error:
image

yerbol-akhmetov: > This is an interesting issue. What line fails?
> 
> I see it is linked to the code above:
> 
> ```
>     CO2_emissions = CO2_emissions.rename(columns={"Country Name": "Country"})
>     cc = coco.CountryConverter()
>     Country = pd.Series(CO2_emissions["Country"])
>     CO2_emissions["country"] = cc.pandas_convert(
>         series=Country, to="ISO2", not_found="not found"
>     )
> ```
> 
> It is a bit surprising the problem you find as pandas_convert should thorugh "not found" when the conversion fails. This code could be improved; for example, there is no need to re-create the pd.Series of a Series object. Moreover, the use of CO2_emissions["country"] = serie may lead to unexpected behavior when index does not match; plus the format df[col] = ... may change behavior, conversely to the use f loc[:, col] These may be the reason of the issue you experience and lead to improper data matching.
> 
> If that is not sufficient, also discarding further unexpected values is clearly fine

Conversions are fine, in my mind. The only thing is column may contain several countries. And it will return it as a list.

yerbol-akhmetov: > This is an interesting issue. What line fails?
> 
> I see it is linked to the code above:
> 
> ```
>     CO2_emissions = CO2_emissions.rename(columns={"Country Name": "Country"})
>     cc = coco.CountryConverter()
>     Country = pd.Series(CO2_emissions["Country"])
>     CO2_emissions["country"] = cc.pandas_convert(
>         series=Country, to="ISO2", not_found="not found"
>     )
> ```
> 
> It is a bit surprising the problem you find as pandas_convert should thorugh "not found" when the conversion fails. This code could be improved; for example, there is no need to re-create the pd.Series of a Series object. Moreover, the use of CO2_emissions["country"] = serie may lead to unexpected behavior when index does not match; plus the format df[col] = ... may change behavior, conversely to the use f loc[:, col] These may be the reason of the issue you experience and lead to improper data matching.
> 
> If that is not sufficient, also discarding further unexpected values is clearly fine

Hi, @davide-f. I have addressed your comments on removing redefinition of Series and using `.loc[:, col]`. Can you please check the PR? It used to be a simple PR to fix one of the current issues of CI and general workflow. Otherwise sector is does not work.

scripts/prepare_transport_data_input.py
```diff
@@ -95,14 +95,20 @@ def download_CO2_emissions():
...
+    CO2_emissions = CO2_emissions[
+        CO2_emissions.apply(
+            lambda x: False if not isinstance(x.country, str) else True, axis=1
+        )
+    ]
+
```
davide-f:
```suggestion
    CO2_emissions = CO2_emissions[
        CO2_emissions.country.apply(lambda x: isinstance(x, str), axis=1)
    ]
```
Can you check if this passes the pre-commit?
Happy to merge right after :)

Great check!

davide-f (commented):
Great @yerbol-akhmetov and thanks for the clarification.
A minor stylistic comment, ready to merge right after

scripts/prepare_transport_data_input.py
yerbol-akhmetov: Thank you, @davide-f, for suggestion. I have made a minor change there. Just removed (axis=1). As inner part is now Series.

davide-f:
Many thanks :D merged :)

PR #1168, opened 2024-11-04
Update PyPSA & enable linopy

ekatef:
The PRs is intended to put together work on linopy implementation for both power-only and cross-sectoral model


ekatef:
Closing in favour #1169

PR #1065, opened 2024-07-21
Update PyPSA version

ekatef:
Disentangles update of PyPSA version in #796 from changes needed to implement `linopy` into model construction.

Covers #787 and includes adaptation of the changed PyPSA implications for `n.buses` and `n.lines` structure to our current data structures.

As a note: the strategy towards processing "non-standard" columns has changed in the recent PyPSA versions. Non-standard means names which are not included into columns of components attributes (see [lines attributes](https://github.com/PyPSA/PyPSA/blob/master/pypsa/component_attrs/lines.csv), for example).

## Changes proposed in this Pull Request

The PR follows the approach of PyPSA-Eur, but introduces additional functionality to define custom aggregation strategies in a flexible way.

TODO It would be nice to have a more formalised definition of the components data structures in PyPSA. Currently, the custom columns are implicitly allowed, but lead to troubles during clustering.

## Checklist

- [x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [x] I tested my contribution locally and it seems to work fine.
- [ ] Code and workflow changes are sufficiently documented.
- [ ] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`.
- [ ] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [ ] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [ ] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes, including reference to the requested PR.

Issue #968, opened 2024-02-07
unable to install Ipopt on Mac


nikhil550:
## Checklist

- [x] I am using the current [`main`](https://github.com/pypsa-meets-earth/pypsa-earth/tree/main) branch or the latest [release](https://github.com/pypsa-meets-earth/pypsa-earth/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-earth` environment](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

Installing pypsa earth on Mac M1, mamba was unable to install Ipopt dependecy.

## Error Message

`mamba env create -f envs/environment.yaml` yielded the following error:

```
Could not solve for environment specs
The following package could not be installed
└─ ipopt <3.13.3  does not exist (perhaps a typo or a missing channel).
```

with conda env update:
```
PackagesNotFoundError: The following packages are not available from current channels:

  - ipopt[version='<3.13.3']
```

davide-f:
Hello @nikhil550 !
Thanks for notifying.
That is weird as generally CI passes, it may be really OS-specific.
You may try to drop the dependency in the env/environment file, just to create the environment.
Then you may install ipopt in mac using an alternative approach. That should work.

It is worth keeping this issue open so that we track the problem and later revise the environment as the #494 issue is solved, which enables to upgrade pypsa and support the solving capabilities using linopy.

GbotemiB:
I faced the same issue. Removing the constraint on ipopt in the `environment.yml` file should solve the issue.

nikhil550:
I was able to install by removing removing Ipopt and installing locally per the recommendation above.

mfripp:
This occurs because `envs/environment.yaml` requires `ipopt <3.13.3` but the first version available for Apple silicon (arm/M1 or later) is 3.14.5. You can see this by running the following commands:

```
# search for macOS arm architecture
CONDA_SUBDIR=osx-arm64 conda search ipopt

# search for macOS x86 architecture
CONDA_SUBDIR=osx-64 conda search ipopt
```

It is possible to work around this by running `CONDA_SUBDIR=osx-64 mamba env create -f envs/environment.yaml`, which will create an x86 environment. But pypsa-earth will then in emulation mode, which reduces performance. Or users can change the ipopt version dependency in `envs/environment.yaml`, with unknown side effects.

[...]

Issue #1170, opened 2024-11-04
Harmonise line endings across the scripts


ekatef:
Currently, line endings differ from script to script. [Can be checked for example with `git ls-files --eol`]

That is an artifact which appeared after the merge and originated from different conventions used previously in the power and sectoral models. Pre-commit doesn't mind as far as line endings are same across each file.

## Describe the feature you'd like to see

Would be great to harmonise it.

The most straight-forward approach is using `.gitattributes`, though finding some pre-commit hooks may be an interesting option.

PR #1172, opened 2024-11-05
Linopy integration for Power and Sector-Coupled Modelling

### LINKED ISSUE #494: Integrate Linopy

ekatef:
# Closes #494 
(and hopefully #968)

Integrated changes needed to update PyPSA version and use linopy. 

To have a better overview, the PR has been implemented in three steps:
1. Update PyPSA version (follows #1065)
2. Integrate `linopy` into power part of the model (follows #796, the diffs can be checked in [this local PR](https://github.com/ekatef/pypsa-earth/pull/3))
3. Integrate `linopy` into cross-sectoral part of the model (implemented currently as [a dedicated PR](https://github.com/pypsa-meets-earth/pypsa-earth-sec/pull/351) to the Sec- repo)

## Checklist


ekatef:
The PR has been tested against `main`, and the change in the objective function has been found to be less than 1%.

Codewise, the implementation is same as in #1065. Manual reimplementation has been needed because of #1170

ekatef:
@GbotemiB As discussed, we can use a branch of this PR (`pypsa_linopy_update` located in upstream directly) to accommodate [your linopy PR](https://github.com/pypsa-meets-earth/pypsa-earth-sec/pull/351) from PyPSA-Earth-Sec. What do you think?

GbotemiB:
> @GbotemiB As discussed, we can use a branch of this PR (`pypsa_linopy_update` located in upstream directly) to accommodate [your linopy PR](https://github.com/pypsa-meets-earth/pypsa-earth-sec/pull/351) from PyPSA-Earth-Sec. What do you think?

Thanks @ekatef. I can start working on the linopy PR for the sec aspect.

scripts/solve_network.py
```diff
@@ -220,165 +258,235 @@ def add_EQ_constraints(n, o, scaling=1e-1):
...
 
 
 def add_battery_constraints(n):
+    """
+    Add constraint ensuring that charger = discharger, i.e.
+    1 * charger_size - efficiency * discharger_size = 0
+    """
     nodes = n.buses.index[n.buses.carrier == "battery"]
-    if nodes.empty or ("Link", "p_nom") not in n.variables.index:
+    # TODO Check if the second part of the condition can make sense
```
davide-f: Does this still apply?

scripts/solve_network.py
```diff
@@ -220,165 +258,235 @@ def add_EQ_constraints(n, o, scaling=1e-1):
...
+        vars_link.sel({"Link-ext": nodes + " discharger"}) * -eff[0],
     )
-    define_constraints(n, lhs, "=", 0, "Link", "charger_ratio")
+    n.model.add_constraints(lhs == 0, name="link_charger_ratio")
 
 
-def add_RES_constraints(n, res_share):
+def add_RES_constraints(n, res_share, config):
     lgrouper = n.loads.bus.map(n.buses.country)
+    # TODO drop load
```
davide-f: Is this for a later stage?

scripts/solve_network.py
```diff
@@ -388,103 +496,68 @@ def add_RES_constraints(n, res_share):
...
+    gens_i = n.generators.query("carrier in @renew_techs").index
+    stores_i = n.storage_units.query("carrier in @renew_techs").index
+
     charger_i = n.links.query("carrier in @charger").index
     discharger_i = n.links.query("carrier in @discharger").index
 
+    stores_t_weights = n.snapshot_weightings.stores
+
     # Generators
+    # TODO restore grouping by countries un-commenting calls of groupby()
```
davide-f: Does this not work? for later?

scripts/solve_network.py
```diff
@@ -992,10 +1066,11 @@ def solve_network(n, config, solving={}, opts="", **kwargs):
     else:
         n = pypsa.Network(snakemake.input.network)
 
-    if snakemake.params.augmented_line_connection.get("add_to_snakefile"):
-        n.lines.loc[n.lines.index.str.contains("new"), "s_nom_min"] = (
-            snakemake.params.augmented_line_connection.get("min_expansion")
-        )
+    # TODO Double-check handling the augmented case
```
davide-f: Why comment this? Does it trigger an issue?
It seems look it generates issues, what do you think?

scripts/solve_network.py
```diff
@@ -1016,15 +1091,15 @@ def solve_network(n, config, solving={}, opts="", **kwargs):
     else:
         n_ref = None
 
-    n = prepare_network(n, solving["options"])
+    # needed to get `n.model` property
```
davide-f: needed?

scripts/solve_network.py
```diff
@@ -159,6 +153,25 @@ def prepare_network(n, solve_opts):
 
 
 def add_CCL_constraints(n, config):
```
davide-f, line 155: In your experience, does the ci test these functions [not just this one]?
Indeed unit testing may by quite great to have or at least the validator similar to -eur

davide-f (commented):
Great :)
Very minor things, basically ready to merge :)

scripts/solve_network.py
ekatef: Yes, my feeling is that the enclosing `add_RES_constraints( )` will need some additional revisions which should go beyond this PR. 

In our current upstream implementation, the function throws [a warning](https://github.com/pypsa-meets-earth/pypsa-earth/blob/d9a364cec2df1866c77d53226e509275795f2dbd/scripts/solve_network.py#L377-L381) stating that it's only a preliminary implementation. I think, it may be a good idea to open an issue to track further improvements

scripts/solve_network.py
ekatef: Yeah, `linopy.groupby( )` has been not in the best shape when I have been debugging this part. So, I have left it out for future improvements of `add_RES_constraints( )`

scripts/solve_network.py
ekatef: A great catch! That is just an old comment needed to debug the main functionality. Now it must be lifted

Testing it

scripts/solve_network.py
ekatef: It has been needed to use linopy methods as creating PyPSA.Network object doesn't enable them. I'll check it I may be missing something

scripts/solve_network.py
ekatef: CI definitely omits at least some of those functions. Agree that it would be a great application for CI. Happy to discuss the details

scripts/solve_network.py
davide-f: Maybe we should add a warning&issue then?

scripts/solve_network.py
ekatef: Have get this part back, and testing worked well

ekatef:
Update on testing with different versions of PyPSA:
- works with `0.28`
- fails with `0.30` under Windows due to some environment conflicts
- currently testing with the latest PyPSA version

scripts/solve_network.py
ekatef: My understanding is that this second part addresses some special case when the network doesn't contain any batteries, despite of the fact that there are nodes with `battery` carrier. Not sure when exactly can it be the case and just leaving it as is for now, as it seems not to relate to interfacing with linopy. Does it make sense for you or am I missing something?

ekatef:
CI has been fixed successfully thanks to the encoding fix for `countrycode` which @yerbol-akhmetov has introduced into it's upstream [thanks a lot, incredibly helpful!!]

scripts/solve_network.py
ekatef: Have fixed the functionality. It has been more straight-forward ;)

scripts/solve_network.py
ekatef: In the updates version of `add_RES_constraints( )`, this TODO is not needed anymore

scripts/solve_network.py
ekatef: After cross-checking, have found that initialisation is not needed. Not sure why it was needed before. Probably I tested extra-functionalites fetched in some exotic way. Or it was issue fixed with the later releases of linopy and/or PyPSA

ekatef:
CI is currently failing due to the need to upgrade `add_h2_network_cap( )`. Locally, the issue has been resolved by using the updated version of `add_h2_network_cap( )` from #1182.

It feels like it's time to merge #1182 😄

doc/release_notes.rst
```diff
@@ -81,6 +81,8 @@ PyPSA-Earth 0.4.0
 
 * Add an option to use csv format for custom demand imports. `PR #995 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/995>`__
 
+* Implement changes in processing network topology to use the updated PyPSA version. `PR #1065 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1065>`__
+
```
davide-f: This needs to be moved upwards

envs/environment.yaml
```diff
@@ -89,6 +90,8 @@ dependencies:
   - git+https://github.com/pypsa/atlite@master
   - git+https://github.com/davide-f/google-drive-downloader@master  # google drive with fix for virus scan
   - git+https://github.com/FRESNA/vresutils@master  # until new pip release > 0.3.1 (strictly)
+  - git+https://github.com/vincentarelbundock/pycountrycode@main # until the fix will be available via https://github.com/vincentarelbundock/pycountrycode/pull/13
```
davide-f: The PR has been merged; why do we need this dependency?

scripts/solve_network.py
```diff
@@ -220,271 +259,290 @@ def add_EQ_constraints(n, o, scaling=1e-1):
...
+        .sum()
+        .sum("snapshot")
     )
-    lhs_spill = (
-        linexpr(
-            (
-                -n.snapshot_weightings.stores * scaling,
-                get_var(n, "StorageUnit", "spill").T,
-            )
+    # TODO: double check that this is really needed, why do have to subtract the spillage
```
davide-f: I believe we need it as we consider the full inflow; however, there is the assumption in the inflow that the efficiency_dispatch is 1, which is not in general. I've added an issue in -earth and -eur.
This comment may be dropped I believe; no need to fix this function in this PR

scripts/solve_network.py
```diff
@@ -220,271 +259,290 @@ def add_EQ_constraints(n, o, scaling=1e-1):
...
+    conventional_carriers = config["electricity"]["conventional_carriers"]
+    ext_gens_i = n.generators.query(
+        "carrier in @conventional_carriers & p_nom_extendable"
+    ).index
+    capacity_variable = n.model["Generator-p_nom"]
+    p_nom = n.model["Generator-p_nom"].loc[ext_gens_i]
+    lhs = p_nom.sum()
     exist_conv_caps = n.generators.query(
-        "~p_nom_extendable & carrier in @conv_techs"
+        "~p_nom_extendable & carrier in @entional_carriers"
```
davide-f: This was not captured; we shall add an issue to ensure the testing of these optional functions.
```suggestion
        "~p_nom_extendable & carrier in @conventional_carriers"
```

scripts/solve_network.py
```diff
@@ -220,271 +259,290 @@ def add_EQ_constraints(n, o, scaling=1e-1):
...
-    cgrouper = n.links.bus0.map(n.buses.country)
-
+def add_RES_constraints(n, res_share, config):
     logger.warning(
-        "The add_RES_constraints functionality is still work in progress. "
+        "The add_RES_constraints() is still work in progress. "
         "Unexpected results might be incurred, particularly if "
         "temporal clustering is applied or if an unexpected change of technologies "
-        "is subject to the obtimisation."
+        "is subject to future improvements."
```
davide-f, line 494: Do you remember/know why? We have no issue about this

scripts/solve_network.py
davide-f: can we drop this then?

davide-f (commented):
Contribution looks quite interesting and nearly finalized.

It is advisable to update the branch too and check the new results of the run.
I've compared the results of the objective values of the CI and they seem a bit different:
- the first power test: 3809010949.0 [this PR] vs 2540477210.0 [main](https://github.com/pypsa-meets-earth/pypsa-earth/actions/runs/12332228654/job/34419788552)
- the sec test: 7.5107178e+11 [this PR] vs 7.270497e+11[main]

I remember you crosschecked them already; do you recall?

doc/release_notes.rst
ekatef: Done!

envs/environment.yaml
ekatef: Definitely not 🙂 
Cross-checked and updated the environment specification

scripts/solve_network.py
ekatef: Agree, and thanks for taking care about the issues. Have replaced TODO with the comment with a link to the issue

scripts/solve_network.py
ekatef: Ouch... Fixed, thanks a lot!

Agree, and it has been still noted that extra-functionality functions may not be working as expected. To my understanding, testing it is a great application for unit testing. Have opened #1203 to track that

scripts/solve_network.py
ekatef: That is kind-of legacy comment from KZ project. Not sure what exactly have been meant: I don't see any problems with change of technologies, while advanced temporal clustering can lead to problems also with other technologies. Have added a brief docstring to make the purpose more clear

So, not sure about the issue though don't want to remove the comment to avoid losing some hints. What do you think?

ekatef:
> Contribution looks quite interesting and nearly finalized.
> 
> It is advisable to update the branch too and check the new results of the run. I've compared the results of the objective values of the CI and they seem a bit different:
> 
> * the first power test: 3809010949.0 [this PR] vs 2540477210.0 [main](https://github.com/pypsa-meets-earth/pypsa-earth/actions/runs/12332228654/job/34419788552)
> * the sec test: 7.5107178e+11 [this PR] vs 7.270497e+11[main]
> 
> I remember you crosschecked them already; do you recall?

@davide-f  thanks for the review! 😄 

I confirm that me and Emmanuel have checked that the PR doesn't change the objective function. Though, the may be some pitfalls with testing extra-functionality functions. Have to check it.

Currently, CI is failing as it's using pinned environments which must be re-build using the updated `environment.yaml`. Happy to look into it but if you have some experience on that, some hints would be much appreciated 🙂

scripts/solve_network.py
```diff
@@ -220,271 +258,303 @@ def add_EQ_constraints(n, o, scaling=1e-1):
...
 
     p_max_pu = get_as_dense(n, "Generator", "p_max_pu")
 
-    lhs = linexpr((1, dispatch), (1, reserve))
+    lhs = merge(
+        dispatch * 1,
+        reserve * 1,
+    )
 
     if not ext_i.empty:
```
davide-f, line 421: In pypsa-eur, there is no need for this if case, though I don't think is the reason of the issue

scripts/solve_network.py
```diff
@@ -220,271 +258,303 @@ def add_EQ_constraints(n, o, scaling=1e-1):
...
-        ]
-        lhs += linexpr(
-            (-EPSILON_VRES * capacity_factor, renewable_capacity_variables)
-        ).sum(1)
+        renewable_capacity_variables = (
+            n.model["Generator-p_nom"]
+            .loc[vres_i.intersection(ext_i)]
+            .rename({"Generator-ext": "Generator"})
+        )
+        lhs = merge(
```
davide-f: I see you have made extensive use of the merge function.
I'm dropping a comment hoping it may help.
The documentation of merge informs that when a list of expressions is merged, it assumes they share the same index basically: in case of a list, the first index overrides the others.
https://linopy.readthedocs.io/en/latest/generated/linopy.expressions.merge.html

Not sure if it may have unintended consequences.
In pypsa-eur, I've seen a lower usage of merge but it may not be the cause of this issue

scripts/solve_network.py
```diff
@@ -220,271 +258,303 @@ def add_EQ_constraints(n, o, scaling=1e-1):
...
+            lhs,
+            (renewable_capacity_variables * (-EPSILON_VRES * capacity_factor)).sum(
```
davide-f: if we rename p_nom_vres the renewable capacity variables, maybe we shorten a bit the text, but minor comment

davide-f: In pypsa-eur, they also include a xr.DataArray(capacity_factor), not sure if it has implications into the calculations

davide-f (commented):
Added few comments related to the reserve constraints, hoping to help

ekatef:
Update: re-testing of the objective function is working well in some testing cases, but leads to the differences in others. In particular, the **tutorial-based test** is not working properly. I'm able to reproduce the issue locally, and it appears to be linked with the changes introduced before `solve_network` script. 

Basically, the value of the objective function does not depend on the solving script but on the procedure before: solving the network prepared by `main` workflow with the `solve_network` of this PR leads to the same objective value as in this PR. Visa versa is true as well: solving the prepared network of this PR with `solve_network` of `main` gives the same objective value as in this PR.

A similar issue has been observed before for power-only model, and has been resolved by setting `v_nom` needed to resolve the negative objective issue #961. It looks like a similar pitfall has appeared due to the changes introduced by some cross-sectoral related changes.

davide-f:
> Update: re-testing of the objective function is working well in some testing cases, but leads to the differences in others. In particular, the **tutorial-based test** is not working properly. I'm able to reproduce the issue locally, and it appears to be linked with the changes introduced before `solve_network` script.
> 
> Basically, the value of the objective function does not depend on the solving script but on the procedure before: solving the network prepared by `main` workflow with the `solve_network` of this PR leads to the same objective value as in this PR. Visa versa is true as well: solving the prepared network of this PR with `solve_network` of `main` gives the same objective value as in this PR.
> 
> A similar issue has been observed before for power-only model, and has been resolved by setting `v_nom` needed to resolve the negative objective issue #961. It looks like a similar pitfall has appeared due to the changes introduced by some cross-sectoral related changes.

Many thanks for investigating. What is your insight on this issue and status of the PR therefore?

can we track what has been merged that may have led to the issue?

davide-f:
Let's remind to change the dependencies also in the file doc/requirements.txt.
The API reference was not compiling because of that

scripts/solve_network.py
ekatef: Agree, aligned with pypsa-eur implementation

scripts/solve_network.py
ekatef: Agree and revised, except for `add_CCL_constraints` where it's not straght-forward (the current design implies forming an expression as a list, while pypsa-eur implementation has changed significantly)

scripts/solve_network.py
ekatef: True. Probably `p_max_pu[ext_i]` is safe for empty `ext_i`. Have added a TODO to revise this part

davide-f:
I've been testing the model and to me, it seems like the change in the config.tutorial.yaml are not due to changes in solve_network but something else.
In particular, when debugging the solve_network in this PR and in the main, only 3 functions are actually activated add_battery_constraints, add_chp_constraints (but irrelevant and skipped) and add_co2_sequestration_limit.
Only add_battery_constraint actually changes the model but they seem fine

The checks on the network seem to suggest that the input networks match but something changes for some reason.

At deeper analysis, it seems like the onwind (and marginally offwind) are different.
In particular, the new formulation has half onwind than the base case, which may lead to different outputs

Update: the onwind p_max_pu is expected to be the same after elec.nc [add_electricity]; probably there is something going on in the clustering steps that is different

ekatef:
@davide-f many thanks for testing (and providing the fantastic infrastructure for inter-comparing the networks!). The issues indeed are quite weird and it has been precious to have your independent workflow to be sure that they are reproducible. Writing down the results of the testing below to possibly improve further both implementations 🙂 

It has been found that the network structure and parameters are exactly the same for both implementations. The reason of the differences in the objective function are linked with the difference in `p_max_pu` time series for  wind generators (a quick way to reproduce: `n_gen_onwind = n.generators[n.generators.carrier == "onwind"].index; n.generators_t.p_max_pu[n_gen_onwind].sum(axis=1).sum()`). This difference stems from the fact that `aggregation_strategies` are defined in the different way for the `main` and the updated implementation.

In the older version of PyPSA, a default strategy have has been used for `p_max_pu` time-series, while now we must use custom definitions for aggregation strategies. If we change `aggregation_strategies:generators:p_max_pu` between `mean` and `max`, the objective value changes between ~4.5 [financial units] and ~1.7 [financial units] for the updated implementation with ~2.5 [financial units] for the implementation in main.

**Points of action**
Given a dramatic change in the underlaying computational approaches, it doesn't make any sense trying to keep the objective. What must be done instead is tracing across the code values affected by the aggregation in `simplify_network` and `cluster_network` and check that conservation laws are respected. Potentially, it means that checks or improvements can be also needed to be added into the current main implementation which will remain as a stable backward compatible version after this PR will be merged.

@davide-f great that we have been able to discover the reason, as it appears to be a potentially major issue also for the current implementation 🎉 🎉 🎉  Thank you so much for that. I think it's a great result, and we are very close to both finalise this PR and make the current implementation really stable 😄

ekatef:
It appears, the trick was to use an explicit definition of an aggregation strategy as `p_max_pu: weighted_average`. That brings the objective function to `2512675751.0` which is quite close to `2512675455.0` from `main`.

The full testing results looks like follows:

### main
INFO:__main__:Objective function: 2512675455.0
INFO:__main__:Objective function: 2341159599.0
INFO:__main__:Objective function: 2321096575.0
INFO:__main__:Objective function: 336330443.3
INFO:__main__:Objective function: 4248565071.0
INFO:__main__:Objective function: 4602716893.0
INFO:__main__:Objective function: 5723516039.0
INFO:__main__:Objective function: 476366499.4
INFO:__main__:Objective function: 10926432310.0
INFO:__main__:Objective function: 1461452593.0
INFO:__main__:Objective function: 0.211751626
INFO:__main__:Objective function: 4376884296.0
INFO:__main__:Objective function: -0.006962630898
INFO:__main__:Objective function: 722220262793.1165

### this PR
INFO:__main__:Objective function: 2512675751.0
INFO:__main__:Objective function: 2341214199.0
INFO:__main__:Objective function: 2321096682.0
INFO:__main__:Objective function: 476365695.9
INFO:__main__:Objective function: 10747113970.0
INFO:__main__:Objective function: 1461452491.0
INFO:__main__:Objective function: 4324387954.0
INFO:__main__:Objective function: 0.001945018768
INFO:__main__:Objective function: 5723515621.0
INFO:__main__:Objective function: 336330259.2
INFO:__main__:Objective function: 4503925925.0
INFO:__main__:Objective function: 4186016571.0
INFO:__main__:Objective function: 5.44154318e-07
INFO:__main__:Objective function: 722265045983.7013

That gives the following differences:
#### config.custom.yaml
1.178027e-07 
#### config.landlock.yaml
2.332123e-05 
#### config.sector.yaml
4.609890e-08
#### config.test_myopic.yaml
6.200382e-05

For Monte-Carlo test, the differences may be very high (`2.939659e-01  6.046785e-01 -2.149413e+00 -3.235436e-01 -2.449161e+11 -9.090421e-01
[10] -3.345290e+00  1.000000e+00 -4.559650e-02`) but probably it's intentional.

ekatef:
@davide-f I think we have done it 😄

davide-f (commented):
Great achievement @ekatef ! we have narrowed down the issue!
We have some large differences in few iterations.
A sorted comparison is better than it looks at first though; differences within 1-2% generally or limited in absolute terms.

Do you have 2 folders with the results of the two monte carlo executions? it would be amazing to compare if you have availability; I'm trying them locally in case with #1276 
image

The monte carlo should have the seed implemented by looking at the code; @GbotemiB can you confirm?

GbotemiB:
@davide-f, Yes I can confirm that the monte carlo has a seed option implemented.

davide-f:
Update: I've been investigating the differences and the "sorted comparison" is confirmed, thus differences shown in pu in the diff column are actually small.
image

The larger difference is about 2.2% related to the "m3" case, the reason being a different resourse allocation between offwind and solar induced by noisy costs on node mapping that is not identical across the two iterative execution.

An "issue" identified is that the capital_cost and marginal_costs of offshore-dc technology are not to be fully reproducible with differences in the order of 1e-3-1e-2.
The root cause of this is that the order of the nodes of the file base.nc is not fully reproducibile. As the order of the nodes is not univoque, the final result of the clustering is not fully exactly repeatible, that summed with the use of noisy costs make slight deviations in the objective function.

In summary, this PR for me is fine as is and experienced changes are not due to the changes proposed here.
Worth tracking this point. An interesting option would be to potentially sort the rows to improve repeatability.

davide-f (approved):

ekatef:
Fantastic @davide-f! 😄 A really brilliant investigation. Totally agree that it would be a great feature to maintain the network structure after clustering.

Just out of curiosity: I see that the objective values for `m6` are very low, and the relative difference in the objective are high which may be not meaningful, though. By any chance, do you have an explanation for that?

ekatef:
I think this PR should be closed in favour of #1276.

For the full transparency, the issues we have found along the way are summarised in [this comment](https://github.com/pypsa-meets-earth/pypsa-earth/pull/796#issuecomment-2286545655). For now, point 1 seems to be resolved thanks to some environment updates (as CI also tests fetching isolated networks), points 2 and 3 are tracked in the issues #1283 and #1245, while point 4 addressed by adding a comment.

Eddy-JV:
Please adapt the new constraint in PR https://github.com/pypsa-meets-earth/pypsa-earth/pull/1192 before merging.

davide-f:
This PR is merged into: https://github.com/pypsa-meets-earth/pypsa-earth/pull/1276, moving the message there.
As #1276 is merged, this PR automatically closes as all commits are fully integrated into the otherone

ekatef:
@GbotemiB I suggest to use this PR to review the docstrings in `solve_network` script. All the commits of this PR are still incorporated into #1276 which will be merged. While we can use this PR as a playground for #1297. What do you think?

davide-f:
Merged with #1276 :D Congratulations!!!

ekatef:
> Merged with #1276 :D Congratulations!!!

Thank you so much @davide-f! 😄 

Fantastic work of you and @GbotemiB 🙂 It has been a long way, but it has been definitely worth it

PR #1175, opened 2024-11-07
Revise export ports data 

GbotemiB:
# Closes # (if applicable).

The changes in this PR include
- use port.csv data as input to add_export script
- preprocessing steps for ports.csv data in add_export script

## Checklist


davide-f:
Hello @GbotemiB :)
Thanks for the proposal to address this point.

Currently, the "export_ports" list is expected to be filled by the user; export_ports is indeed a file available in the data folder.
However, I agree that it is unconfortable to be filled manually and this can lead to issues especially when an export potential is requested.

The fact is that not all ports in ports.csv can actually be suitable for exports.
Ideally, the user shall propose either to use the custom value or a file calculated by the workflow.

I believe @hazemakhalek, @energyLS and/or @Eddy-JV may have some insights here.
Feel free to propose something by first writing in here, before coding.

Long-term, we may have prepare_ports to have 2 outputs: 1 ports.csv and 1 export.csv.
Moreover, if relevant for the -sec team, we also keep export.csv and have an option about whether using the export produced by ports or the custom file.

For the short term, we can have the option about either using ports.csv or export.csv

davide-f (commented):
Hello @GbotemiB :)
Thanks for the proposal to address this point.

Currently, the "export_ports" list is expected to be filled by the user; export_ports is indeed a file available in the data folder.
However, I agree that it is unconfortable to be filled manually and this can lead to issues especially when an export potential is requested.

The fact is that not all ports in ports.csv can actually be suitable for exports.
Ideally, the user shall propose either to use the custom value or a file calculated by the workflow.

I believe @hazemakhalek, @energyLS and/or @Eddy-JV may have some insights here.
Feel free to propose something by first writing in here, before coding.

Long-term, we may have prepare_ports to have 2 outputs: 1 ports.csv and 1 export.csv.
Moreover, if relevant for the -sec team, we also keep export.csv and have an option about whether using the export produced by ports or the custom file.

For the short term, we can have the option about either using ports.csv or export.csv

GbotemiB:
Thanks @davide-f. I will await comments from the -sec team on the best suitable approach.

But running the US model, what do you recommend at the moment to solve the model?

davide-f:
> Thanks @davide-f. I will await comments from the -sec team on the best suitable approach.
> 
> But running the US model, what do you recommend at the moment to solve the model?

The simplest approach is copying the ports.csv file and rename it as export_ports.csv, or selecting a subset of such ports; not that it is likely that only large ports can actually be exporting ones so it makes sense to select a subset of them.
To make the model run, the procedure above should work in the meantime

Eddy-JV:
Thank you @davide-f for already replying earlier and thank you @GbotemiB for rasing this.

First let me understand the context:

- Currently the `export_ports.csv` is kept in data because usually the user has in mind very specific export ports for Hyrdrogen export. That was actually always the case till now, that export countries have already selected certain hydrogen export ports. What I understand, is that you want the model to decide which export port to invest in randomly. Please correct  me if am wrong.

I that is the case, then I agree with @davide-f . We should then allow the user to choose in the `config` file under `custom_data` as **True or False**:

1. **Use custom data for export ports (True):** which entales using `export_ports.csv` file that we place in `data_custom` instead of `data` folder.
2. **Use wrokflow (False):** Then we add another output for `prepare_ports.py` , where we select export ports to be only "Large" ports for example and then save it in `resources `folder.

What do you think?

energyLS:
> Thank you @davide-f for already replying earlier and thank you @GbotemiB for rasing this.
> 
> First let me understand the context:
> 
>     * Currently the `export_ports.csv` is kept in data because usually the user has in mind very specific export ports for Hyrdrogen export. That was actually always the case till now, that export countries have already selected certain hydrogen export ports. What I understand, is that you want the model to decide which export port to invest in randomly. Please correct  me if am wrong.
> 
> 
> I that is the case, then I agree with @davide-f . We should then allow the user to choose in the `config` file under `custom_data` as **True or False**:
> 
>     1. **Use custom data for export ports (True):** which entales using `export_ports.csv` file that we place in `data_custom` instead of `data` folder.
> 
>     2. **Use wrokflow (False):** Then we add another output for `prepare_ports.py` , where we select export ports to be only "Large" ports for example and then save it in `resources `folder.
> 
> 
> What do you think?

@Eddy-JV I support this idea, seems reasonable to me.

GbotemiB:
> Thank you @davide-f for already replying earlier and thank you @GbotemiB for rasing this.
> 
> First let me understand the context:
> 
> * Currently the `export_ports.csv` is kept in data because usually the user has in mind very specific export ports for Hyrdrogen export. That was actually always the case till now, that export countries have already selected certain hydrogen export ports. What I understand, is that you want the model to decide which export port to invest in randomly. Please correct  me if am wrong.
> 
> I that is the case, then I agree with @davide-f . We should then allow the user to choose in the `config` file under `custom_data` as **True or False**:
> 
> 1. **Use custom data for export ports (True):** which entales using `export_ports.csv` file that we place in `data_custom` instead of `data` folder.
> 2. **Use wrokflow (False):** Then we add another output for `prepare_ports.py` , where we select export ports to be only "Large" ports for example and then save it in `resources `folder.
> 
> What do you think?

Hi @Eddy-JV, Thank you very much for this contribution. I have a clarification question. There is a possibility that not all countries have large ports. Incases like this, we can use the medium ports, and if there is no medium ports we can use small ports. This way every country is covered.

Eddy-JV:
> > Thank you @davide-f for already replying earlier and thank you @GbotemiB for rasing this.
> > First let me understand the context:
> > 
> > * Currently the `export_ports.csv` is kept in data because usually the user has in mind very specific export ports for Hyrdrogen export. That was actually always the case till now, that export countries have already selected certain hydrogen export ports. What I understand, is that you want the model to decide which export port to invest in randomly. Please correct  me if am wrong.
> > 
> > I that is the case, then I agree with @davide-f . We should then allow the user to choose in the `config` file under `custom_data` as **True or False**:
> > 
> > 1. **Use custom data for export ports (True):** which entales using `export_ports.csv` file that we place in `data_custom` instead of `data` folder.
> > 2. **Use wrokflow (False):** Then we add another output for `prepare_ports.py` , where we select export ports to be only "Large" ports for example and then save it in `resources `folder.
> > 
> > What do you think?
> 
> Hi @Eddy-JV, Thank you very much for this contribution. I have a clarification question. There is a possibility that not all countries have large ports. Incases like this, we can use the medium ports, and if there is no medium ports we can use small ports. This way every country is covered.

Yeah that sounds good. Thank you for moving this forward.

config.default.yaml
```diff
@@ -478,6 +478,7 @@ custom_data:
   add_existing: false
   custom_sectors: false
   gas_network: false # If "True" then a custom .csv file must be placed in "resources/custom_data/pipelines.csv" , If "False" the user can choose btw "greenfield" or Model built-in datasets. Please refer to ["sector"] below.
+  export_data: false # If "True" then a custom .csv file must be placed in "data/custom/export_ports.csv"
```
Eddy-JV: I would suggest to name it: `export_ports`. I think it is more intuitive for the users.

data/custom/export_ports.csv
Eddy-JV, line 1: Here it is better to have an example file (not empty) where the needed columns are shown and at least an example row is also available. This way it is easier for the users to insert their custom data in the right format.

Eddy-JV: Ah we should also delete the export_ports.csv from data folder as it is not needed anymore. Maybe just move it here as it is.

Snakefile
```diff
@@ -1154,9 +1155,11 @@ rule add_export:
         export_profile=config["export"]["export_profile"],
         snapshots=config["snapshots"],
         costs=config["costs"],
+        custom_export=config["custom_data"]["export_data"],
     input:
         overrides="data/override_component_attrs",
-        export_ports="data/export_ports.csv",
+        custom_export_ports="data/custom/export_ports.csv",
```
Eddy-JV: I personally prefer that all files in resources to be the ones that are actually used in the workflow. This way the user can track a file easily if the file is not in resources then the workflow used the file from the data folder.

Therefore, I suggest to have this rule read only  `export_ports="resources/" + SECDIR + "export_ports.csv",` 
and the decision on using custom data or workflow generated one stays in `prepare_ports.py`. 

Thus in `prepare_ports.py`, either:

-  the file is copied from data/custom to resources if the config["custom_data"]["export_data"] is True
- or filter_ports function is activated and the output is saved in resources.

Please note that all my comments are subject to discussion if needed.

scripts/prepare_ports.py
```diff
@@ -102,3 +132,6 @@ def download_ports():
     ports["fraction"] = ports["Harbor_size_nr"] / ports["Total_Harbor_size_nr"]
 
     ports.to_csv(snakemake.output[0], sep=",", encoding="utf-8", header="true")
+    filter_ports(ports).to_csv(
```
Eddy-JV: Here we can put an if clause that i suggested in my Snakefile comment.

Eddy-JV (commented):
Hi @GbotemiB @davide-f . I took the liberty to review this PR. All comments are subject to discussion if you do not agree.

data/custom/export_ports.csv
GbotemiB: What I have done is to move the export_ports.csv file to the custom folder which should serve as an example. Let me know your thoughts on this.

Eddy-JV (approved):
From my side, the PR looks good now from the scripts changes point.

@davide-f @GbotemiB I can do the merge if the failing run checks are not important.

scripts/prepare_ports.py
```diff
@@ -31,12 +32,43 @@ def download_ports():
...
+    return filtered_ports
+
+
 if __name__ == "__main__":
     if "snakemake" not in globals():
         from _helpers import mock_snakemake
 
         snakemake = mock_snakemake("prepare_ports")
 
+    config = snakemake.config
```
davide-f, line 71: Is this needed? if not better removed.
Parameters shall be passed using "params" key in snakemake so that snakemake can better handle re-triggering when appropriate.

scripts/prepare_ports.py
```diff
@@ -102,3 +134,11 @@ def download_ports():
     ports["fraction"] = ports["Harbor_size_nr"] / ports["Total_Harbor_size_nr"]
 
     ports.to_csv(snakemake.output[0], sep=",", encoding="utf-8", header="true")
+
+    if snakemake.params.custom_export:
+        custom_export_path = Path.joinpath("data", "custom", "export_ports.csv")
+        shutil.move(custom_export_path, snakemake.output[1])
+    else:
+        filter_ports(ports).to_csv(
+            snakemake.output[1], sep=",", encoding="utf-8", header="true"
```
davide-f, line 143: Maybe header=True? [without apexes]

doc/release_notes.rst
```diff
@@ -57,6 +57,8 @@ E.g. if a new rule becomes available describe how to use it `make test` and in o
 
 * Adds CI to update keep pinned environment files up to date. `PR #1183 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1183>`__ and  `PR #1210 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1210>`__
 
+* Revise ports data for export in `add_export.py` related to sector model `PR #1175 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1175>`_
+
```
davide-f, line 61: This sounds a new feature; I'd recommend to place it under "new features" list above

davide-f (commented):
Great new development!

PR #1183, opened 2024-11-13
ci: adds fixed env ci

lkstrp:
Adds pinned env ci and runs test on pinned env instead, as implemented in pypsa-eur.
Removes `envs/environment.fixed.yaml` and adds `envs/<os>-pinned.yaml` instead which are automatically updated.
This is useful for the devcontainer implementation by @drifter089 and @GbotemiB

See https://github.com/PyPSA/pypsa-eur/pull/1379
- updates fixed env ci to
    - Create env for each os (ubuntu, windows, linux). They are not compatible across systems.
    - Runs weekly/ on manual trigger/ env changes
- Run tests on fixed env instead
    - This way we don't test dependency changes in the ci and only test the code or if data retrieval fails. 
    - Dependency updates are checked in the weekly trigger above.
    - As pypsa-eur is not a library, there is no big reason to keep the dependency constraints wide. And they are still via the normal environment file


lkstrp:
Tests will fail since the env files need to be created initially. But it looks like they are not running through anyways?

.github/workflows/update-pinned-env.yml
```diff
@@ -0,0 +1,85 @@
+name: Update pinned envs
+
+on:
+  push:
+    branches:
+    - master
```
davide-f: Please, revise to main

davide-f (commented):
Many thanks and this seems a great contribution!
This simplifies a lot the maintenance of the model indeed

I'd also recommend adding here a release note also to summarize the great three contributions you proposed :) you can find it in doc/release_notes.rst.
Many thanks :D

lkstrp:
Thank you and no worries! I added release notes for all three PRs.

The Tests CI fails right now, so this needs another look once this one is merged

lkstrp:
I added a fix to trigger the PR only by changes in env file on master branch. 
The tests will still fail, since the pinned env version is not created yet. So this needs to be merged first.

davide-f:
Understandable, great contribution @lkstrp :D Merging :)

PR #1192, opened 2024-11-14
Lossy Bidirectional Links

Eric-Nitschke:
Summary:
- fixed an existing bug regarding bidirectional lossy links that affected hydrogen pipelines.
- the fix primarily utilizes code already in use by PyPSA-Eur.

Changes:
- added a function to scripts/add_extra_components.py that "splits" bidirectional links into two unidirectional links. This is based upon the similar PyPSA-Eur implementation, which is slightly more advanced to accommodate PyPSA-Eurs more advanced efficiency calculations.
- added a function to scripts/solve_network.py that adds an additional constraint to the network, which ensures that both unidirectional links share the same nominal power. This is a modified version of the PyPSA-Eur function, as PyPSA-Eur uses linopy while the base PyPSA-Earth version operates on lopf.

Reason:
- without this commit, lossy bidirectional links are able to "generate" power by operating at a negative power level.
- this stems from the general link implementation, which applies p1 = -eta*p0 to links, regardless whether power flows from bus0 to bus1 or from bus1 to bus0.
- the current hydrogen pipeline implementation adds them to the network as lossy bidirectional links and is thus able to "generate" hydrogen.
- this might affect other lossy bidirectional components added by the sector network as well.

## Checklist


davide-f:
Many thanks @Eric-Nitschke :D That's a quite neat application!
Would you like to add a release note?

Eric-Nitschke:
@davide-f Thanks :)
I've added a relase note to the pull request!

README.md
```diff
@@ -210,87 +210,29 @@ The documentation is available here: [documentation](https://pypsa-earth.readthe
...
-            <td align="center">
-                <a href="https://github.com/Justus-coded">
-                    <img src="https://avatars.githubusercontent.com/u/44394641?v=4" width="100;" alt="Justus-coded"/>
-                    <br />
-                    <sub><b>Justus Ilemobayo</b></sub>
-                </a>
-            </td>
-            <td align="center">
-                <a href="https://github.com/mnm-matin">
-                    <img src="https://avatars.githubusercontent.com/u/45293386?v=4" width="100;" alt="mnm-matin"/>
```
ekatef: I guess those are changes from an automated run of github workflow? 😄 Don't think we need them in the PR. Could you please remove them?

The recent PR #1210 changes a mode of the automated updates of the contributors list to a pre-scheduled one. So we don't loose track of contributions but PRs won't have those irrelevant additions anymore.

doc/release_notes.rst
```diff
@@ -57,6 +57,8 @@ E.g. if a new rule becomes available describe how to use it `make test` and in o
 
 * Adds CI to update keep pinned environment files up to date. `PR #1183 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1183>`__
 
+* Fix lossy bidirectional links, especially H2 pipelines, which would sometimes gain H2 instead of losing it.  `PR #1192 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1192>`__
+
```
ekatef: Just to be sure: I understand H2 pipelines can gain energy not H2 substance, right? So, basically the flow can accelerate instead of being hampered by friction losses.

scripts/add_extra_components.py
```diff
@@ -265,6 +265,36 @@ def attach_hydrogen_pipelines(n, costs, config):
         carrier="H2 pipeline",
     )
 
+    # setup pipelines as bidirectional and lossy
+    lossy_bidirectional_links(n, "H2 pipeline")
+
+
+def lossy_bidirectional_links(n: pypsa.components.Network, carrier: str):
+
+    "Split bidirectional links into two unidirectional links to include transmission losses."
```
ekatef: Usually, we are using `"""` to write docstrings

scripts/solve_network.py
```diff
@@ -854,6 +854,43 @@ def add_existing(n):
             n.generators.loc[tech_index, tech] = existing_res
 
 
+def add_lossy_bidirectional_link_constraints(n: pypsa.components.Network) -> None:
+    """
```
ekatef, line 858: Types hinting is generally a great idea, but we don't use. I'd suggest to remove it here for consistency. 

Happy to discuss an overall revision to implement type hinting across the whole codebase if feels handy

scripts/solve_network.py
```diff
@@ -854,6 +854,43 @@ def add_existing(n):
...
+def add_lossy_bidirectional_link_constraints(n: pypsa.components.Network) -> None:
+    """
+    Ensures that the two links simulating a bidirectional_link are extended the same amount.
+    """
+
+    if not n.links.p_nom_extendable.any() or "reversed" not in n.links.columns:
+        return
+
+    n.links["reversed"] = n.links.reversed.fillna(0).astype(bool)
+    carriers = n.links.loc[n.links.reversed, "carrier"].unique()  # noqa: F841
```
ekatef, line 867: I understand that `noqa: F841` follows PyPSA-Eur implementation. Haven't you tried to understand what is the reason of it's inclusion into the code?

Looking into [flake8 error codes](https://flake8.pycqa.org/en/latest/user/error-codes.html), it appears that `# noqa: F841` suppresses warnings generated when `local variable name is assigned to but never used`. Frankly speaking, I'm not even sure it's working in our case as `flake` seems to be not included into our environment. Can we probably just remove it?

scripts/solve_network.py
```diff
@@ -854,6 +854,43 @@ def add_existing(n):
...
+        return pd.Index(
+            [
+                (
+                    re.sub(r"-(\d{4})$", r"-reversed-\1", s)
+                    if re.search(r"-\d{4}$", s)
+                    else s + "-reversed"
+                )
+                for s in forward_i
+            ]
+        )
```
ekatef, line 886: Just for better understanding, what is a particular reason to apply here regular expressions? Again, I understand that implementation follows PyPSA-Eur and I'm perfectly fine with leaving it there. Just trying to draft some approach to further improvements which we could probably add as TODO if it would make sense 🙂

I guess `sub` is used to extract the identifications from the links indices and build corresponding names for the reversed links. However, regexpr are not exactly readable, and the implementation can easily break if PyPSA conventions would change. I wonder if some pandas string functions can provide a good replacement?

Regular expressions are

ekatef (commented):
Hello @Eric-Nitschke, thanks a lot for investigating the issue and contributing the fix!

Have added some technical comments mainly aimed at better understanding and maintainability improvements. I recognise that the PR follows PyPSA-Eur implementation which is more or less state-of-the-art, but it would be definitely great to have deeper understanding of possible further improvements on some weak points. Your insights in this regard would be very helpful.

Personally, I think that this PR can be probably merged as-is as it feels like a much needed bug-fix. It would be great to have a review of @Eddy-JV who has been working on implementing similar solutions in PyPSA-Eur and definitely has a better overview

README.md
Eddy-JV: I agree.

scripts/add_extra_components.py
```diff
@@ -265,6 +265,36 @@ def attach_hydrogen_pipelines(n, costs, config):
         carrier="H2 pipeline",
     )
 
+    # setup pipelines as bidirectional and lossy
+    lossy_bidirectional_links(n, "H2 pipeline")
+
+
+def lossy_bidirectional_links(n: pypsa.components.Network, carrier: str):
```
Eddy-JV: I think that it is better to move `lossy_bidirectional_links` to `helpers.py`, beacuse it will be used more than once throughout the workflow. 

Additionally, I would prefer to apply directly the length based pipeline efficiency that I saw in PR #1215. Consequently changing the function to the following similarly to PyPSA-Eur but without the .location at the moment:


```
  def lossy_bidirectional_links(n, carrier, efficiencies={}):
  
      ``` Split bidirectional links into two unidirectional links to include transmission losses. ```
      
      carrier_i = n.links.query("carrier == @carrier").index
  
      if (
          not any((v != 1.0) or (v >= 0) for v in efficiencies.values())
          or carrier_i.empty
      ):
          return
  
      efficiency_static = efficiencies.get("efficiency_static", 1)
      efficiency_per_1000km = efficiencies.get("efficiency_per_1000km", 1)
      compression_per_1000km = efficiencies.get("compression_per_1000km", 0)
  
      logger.info(
          f"Specified losses for {carrier} transmission "
          f"(static: {efficiency_static}, per 1000km: {efficiency_per_1000km}, compression per 1000km: {compression_per_1000km}). "
          "Splitting bidirectional links."
      )
  
      n.links.loc[carrier_i, "p_min_pu"] = 0
      n.links.loc[carrier_i, "efficiency"] = (
          efficiency_static
          * efficiency_per_1000km ** (n.links.loc[carrier_i, "length"] / 1e3)
      )
      rev_links = (
          n.links.loc[carrier_i].copy().rename({"bus0": "bus1", "bus1": "bus0"}, axis=1)
      )
      rev_links["length_original"] = rev_links["length"]
      rev_links["capital_cost"] = 0
      rev_links["length"] = 0
      rev_links["reversed"] = True
      rev_links.index = rev_links.index.map(lambda x: x + "-reversed")
  
      n.links = pd.concat([n.links, rev_links], sort=False)
      n.links["reversed"] = n.links["reversed"].fillna(False).infer_objects(copy=False)
      n.links["length_original"] = n.links["length_original"].fillna(n.links.length)
  
      # do compression losses after concatenation to take electricity consumption at bus0 in either direction
      carrier_i = n.links.query("carrier == @carrier").index
      if compression_per_1000km > 0:
          n.links.loc[carrier_i, "bus2"] = n.links.loc[
              carrier_i, "bus0"
          ].str.removesuffix(' ')
          # TODO: use these lines to set bus 2 instead, once n.buses.location is functional and remove bus_suffix.
          """
          n.links.loc[carrier_i, "bus2"] = n.links.loc[carrier_i, "bus0"].map(
              n.buses.location
          )  # electricity
          """
          n.links.loc[carrier_i, "efficiency2"] = (
              -compression_per_1000km * n.links.loc[carrier_i, "length_original"] / 1e3
          )
```

As for the `config` file, the following change is advised in `sector` :

```
  transmission_efficiency:
    H2 pipeline:
      efficiency_per_1000km: 0.983 # DEA technology data. Energy losses, lines 5000-20000 MW, [%/1000 km] and year 2050
      compression_per_1000km: 0.015 # DEA technology data. year 2050
```

If this is done as above, then PR #1215 is not needed anymore and issue #1213 can be closed.

scripts/add_extra_components.py
```diff
@@ -265,6 +265,36 @@ def attach_hydrogen_pipelines(n, costs, config):
         carrier="H2 pipeline",
```
Eddy-JV, line 270: `efficiency=costs.at["H2 pipeline", "efficiency"],` to be removed as it is set now in `lossy_bidirectional_links` as per below comments.

doc/release_notes.rst
Eric-Nitschke: With the current barebone pipeline implementation, the H2 pipeline does not require any energy to operate (PR #1215 would change this). Instead, the pipeline receives an inflow of hydrogen at one bus (p0 if inflow at bus0 or p1 if inflow at bus1) and depending on the efficiency eta outputs are certain hydrogen flow at the other bus (p1 if inflow at bus0 or p0 if inflow at bus1).
p0 and p1 are linked as follows: p1 = -eta\*p0
As a result the pipeline loses hydrogen substance when operating from bus0 to bus1, and gains hydrogen substance when operating from bus1 to bus0.
This fix addresses the latter as it is definitely problematic, that pipelines can generate hydrogen. Specifically, the fix splits the pipeline into two pipelines, that share the same p_nom_opt (ensured by a new constraint). Both pipelines can only operate from bus0 to bus1 and the new pipeline has bus0 and bus1 switched to enable flow from the original bus1 to the original bus0. To maintain the orginal pipeline cost, the capex of the new pipeline are set to 0.

Arguably, it is suboptimal as well, that the pipelines lose hydrogen substance instead of requiring electricity for compression. This would be addressed by PR #1215. With PR #1215 the pipeline behavior would look as follows:
p0 and p1 are still linked by p1 = -eta\*p0, though eta is now calculated based on a static and a length dependend component.
In theory this means that H2 pipelines could still lose H2 substance, but in practice the efficiency values are setup in the config to result in eta=1 (same in PyPSA-Eur) so it doesn't lose any hydrogen. I think it's nice to still have option to considere pipeline leakage by changing the config.
In addition, the hydrogen inflow p0 and the electricity demand for compression p2 are linked via p2 = -eta2\*p0, where eta2 is the compression energy per H2 inflow and is calculated based on the pipeline length.

scripts/add_extra_components.py
```diff
@@ -265,6 +265,36 @@ def attach_hydrogen_pipelines(n, costs, config):
...
+    rev_links["reversed"] = True
+    rev_links.index = rev_links.index.map(lambda x: x + "-reversed")
+
+    n.links = pd.concat([n.links, rev_links], sort=False)
+    n.links["reversed"] = n.links["reversed"].fillna(False).infer_objects(copy=False)
+    n.links["length_original"] = n.links["length_original"].fillna(n.links.length)
+
 
 if __name__ == "__main__":
     if "snakemake" not in globals():
```
Eddy-JV, line 281: Please accomodate for override components by:

Removing:
 `n = pypsa.Network(snakemake.input.network)`

Adding: 
`overrides = override_component_attrs(snakemake.input.overrides)`
              `n = pypsa.Network(snakemake.input.network, override_component_attrs=overrides)`

Eddy-JV: @davide-f @ekatef please check if it is ok to override components in the electricity only model.

scripts/add_extra_components.py
```diff
@@ -265,6 +265,36 @@ def attach_hydrogen_pipelines(n, costs, config):
         carrier="H2 pipeline",
     )
 
+    # setup pipelines as bidirectional and lossy
+    lossy_bidirectional_links(n, "H2 pipeline")
```
Eddy-JV, line 274: Based on all my suggestions, this would be change to:

```
    # Add option to specify losses for bidirectional links, e.g. pipelines or HVDC links, 
    # in configuration file under sector: transmission_efficiency:. 
    # Users can specify static or length-dependent values as well as a length-dependent electricity demand for compression,
    #  which is implemented as a multi-link to the local electricity buses. 
    # The bidirectional links will then be split into two unidirectional links with linked capacities

    for k, v in snakemake.params.transmission_efficiency.items():
        if k == "H2 pipeline":
            lossy_bidirectional_links(n, k, v)
```

And in the `snakefile` please add to the `rule add_extra_components`:
```
       params:
              transmission_efficiency=config["sector"]["transmission_efficiency"],
```

scripts/solve_network.py
Eric-Nitschke: Regular expression enables it us to easily change indices tha end in a four digit number to not end in reversed but have it in the middle instead. I.e. "H2 pipeline DZ0 0-DZ0-1234" to "H2 pipeline DZ0 0-DZ0-reversed-1234" instead of "H2 pipeline DZ0 0-DZ0-1234-reversed".
That being said, no component in PyPSA-Earth ends in a four digit number and the H2 pipelines definitely don't. I'm not even sure if PyPSA-Eur does anymore, since it never looks for a r"-(\d{4})$" ending.

If we assume that the four digit ending can be ignored, the whole thing could be replaced with something like backward_i = forward_i + "-reversed" or backward_i = forward_i.apply(lambda x: f"{x}-reversed").

scripts/solve_network.py
Eric-Nitschke: Short answer: carriers is used in the lines just below, but the interpreter doesn't realize it, because its used in with .query(). Thats why noqa: F841 is used.

I'll write a long answer tomorrow.

scripts/solve_network.py
```diff
@@ -854,6 +854,43 @@ def add_existing(n):
             n.generators.loc[tech_index, tech] = existing_res
 
 
+def add_lossy_bidirectional_link_constraints(n: pypsa.components.Network) -> None:
```
Eddy-JV, line 857: I did this slightly different (See below). Feel free to check and comment if you do not agree:

```
def add_lossy_bidirectional_link_constraints(n):
    """
        Ensures that the two links simulating a bidirectional_link are extended the same amount.
    """
    if not n.links.p_nom_extendable.any() or "reversed" not in n.links.columns:
        return

    # Ensure 'reversed' column is boolean
    n.links["reversed"] = n.links.reversed.fillna(0).astype(bool)
    carriers = n.links.loc[n.links.reversed, "carrier"].unique()

    # Get forward link indices (non-reversed)
    forward_i = n.links.query(
        "carrier in @carriers and ~reversed and p_nom_extendable"
    ).index

    # Function to get backward (reversed) indices corresponding to forward links
    def get_backward_i(forward_i):
        return pd.Index(
            [
                re.sub(r"-(\d{4})$", r"-reversed-\1", s)
                if re.search(r"-\d{4}$", s)
                else s + "-reversed"
                for s in forward_i
            ]
        )

    backward_i = get_backward_i(forward_i)

    # Get the p_nom optimization variables for the links using the get_var function
    links_p_nom = get_var(n, "Link", "p_nom")

    # Only consider forward and backward links that are present in the optimization variables
    subset_forward = forward_i.intersection(links_p_nom.index)
    subset_backward = backward_i.intersection(links_p_nom.index)

    # Ensure we have a matching number of forward and backward links
    if len(subset_forward) != len(subset_backward):
        raise ValueError("Mismatch between forward and backward links.")

    # For each forward index, find the corresponding backward index and define the constraint
    for fwd, bwd in zip(subset_forward, subset_backward):
        lhs = linexpr((1, links_p_nom.loc[bwd]))  # LHS (backward link)
        rhs = linexpr((-1, links_p_nom.loc[fwd]))  # RHS (forward link with -1 coefficient)

        # Define the constraint for this pair of forward-backward link
        define_constraints(
            n,
            lhs + rhs,  # LHS - RHS = 0 (equality constraint)
            "=",
            0,  # Right-hand side is 0 because we want them to be equal
            name="Link",
            attr=f"bidirectional_sync_{fwd}",  # Custom name for this constraint (per pair)
            axes=[pd.Index([fwd])]  # The axes refer to the forward link
        )
```

scripts/solve_network.py
Eddy-JV: Agree. Maybe we can put this as a seperate issue if needed.

Eddy-JV (changes requested):
Thank you @Eric-Nitschke for this very important contribution. 

I took the liberty to add some comments in order to cover this PR #1192 as well as PR #1215 and Issue #1213 at once. Feel free to discuss, if you think things can be done differently. Thanks again for the great work.

scripts/solve_network.py
Eric-Nitschke: Slightly longer answer:
```
forward_i = n.links.query(
        "carrier in @carriers and ~reversed and p_nom_extendable"
    ).index
```
uses carriers, but because pandas.DataFrame.query() takes a string as an input, the python interpreter doesn't realize that carriers is used via the "@carriers".

I don't think it is necessary to use .query() here. Instead, we could use .loc[]:
```
forward_i = n.links.loc[
        (n.links.carrier in carriers) and
        ~n.links.reversed and
        n.links.p_nom_extendable
].index
```
In general, I think we should discuss whether we should phase out the use of .query() and replace it with .loc[] for all of PyPSA-Earth to make it more consistent and readable.

scripts/add_extra_components.py
Eric-Nitschke: Great idea, to move the function to helpers!

I think it's nice to implement both PR's together, but I'd prefer to leave the functions seperate (eventhough PyPSA-Eur combines them) to make them more reusable. I think it's quite plausible, that some new components might be bidirectional and lossy, but without a length-based efficiency, or have a length-based efficiency, but are only unidirectional.

Eric-Nitschke: @Eddy-JV regarding the efficiency_per_1000km you provided:
Is it intentional, that the pipeline loses hydrogen substance (which results from an efficiency_per_1000km < 1) or should it just require more electricity per distance, which is depicted by compression_per_1000km.

scripts/solve_network.py
ekatef: Thanks a lot for the explanations. Do you see any advantages in having `"reversed"` in the middle of the link name?

scripts/solve_network.py
ekatef: Ahh, ok! Thanks a lot for investigating this! Agree with the proposal on replacing here `query` with `loc`

Regarding the overall code refactoring, totally agree that there is a room for improvement and you are very welcome to contribute there. Frankly speaking, I'm not ready to confirm that `query` is less readable than `loc` for all the cases. But you may be right that it could be over-used for sure, and happy to discuss you propositions!

doc/release_notes.rst
ekatef: Ohh... Many thanks for the explanation. Haven't been aware of the details on how we are using electrical analogy (because `Link` component has been initially developed to represent a power flow). Frankly speaking, it sounds to me that it it would have more physical ground if we would treat `p` as pressure is using `Link` to represent the pipelines. Not sure however we can do it straight away 😄 

Thank you so much for explaining this!

ekatef:
Hello @Eric-Nitschke and @Eddy-JV! Great work on reviewing the PR.

@Eric-Nitschke do you need any support in revisions? If they look heavy, could it make sense to breakdown the task? That is a precious contribution, and it would be great to have it merged into the model 🙂

scripts/solve_network.py
Eric-Nitschke: @ekatef I don't see any advantage to it.

Eric-Nitschke:
Hey @ekatef,

in general I think the revisions look quite doable and I'm planning on finishing them before the holidays!
Still I'd like to make sure that we all agree on what the changes should look like:

Major changes:
- [x] move the lossy_bidirectional_links to _helpers
- [x] combine the lossy_bidirectional_links with the length based link efficiency. I would add them as two seperate functions to helpers.
- [x] maybe change the constraint implementation as suggested by @Eddy-JV.  I'm not yet fully sure on this, as I didn't have time yet to fully review it.

Minor changes:
- [x] fix the readme
- [x] use only """ for docstring
- [x] change query to loc in the constraint implementation (foreward_i)
- [x] replace the regular expression in the constraint implementation (backward_i)
- [x] add the transmission efficiencies provided by @Eddy-JV to the config
- [x] include override_components to allow for the length based efficiency implementation (cf. PR #1215)
- [x] add better comments (use the ones suggested by @Eddy-JV)

Discussions for another time:
- [ ] type hints, yay or nay?
- [ ] should the pipelines be implemented based on pressure instead?
- [ ] remove unnecessary regular expressions.

Did I miss anything?

Tests:
- [x] run with both lossy bidirectional pipelines and length based pipeline efficiencies
- [x] run with length based pipeline efficiencies
- [x] run with lossy bidirectional pipelines

scripts/add_extra_components.py
Eddy-JV: @Eric-Nitschke Thank you for catching that. In the Hydrogen case the efficiency should be `efficiency_per_1000km: 1 `.
In [DEA ](https://ens.dk/en/our-services/technology-catalogues/technology-data-transport-energy)they consider the energy needs for compressors within the efficiency per 1000km for different classes of pipeline sizes. So I considered the mean of the largest two classes: 5000-20000 MW and >20000 MW for 2020, 2030 and 2050. Hence, the config will look as follows:

```
  transmission_efficiency:
    H2 pipeline:
      efficiency_per_1000km: 1 
      compression_per_1000km: 0.017 # DEA technology data. Mean of  Energy losses, lines 5000-20000 MW and lines >20000 MW for 2020, 2030 and 2050, [%/1000 km]
```

Eddy-JV:
> Hey @ekatef,
> 
> in general I think the revisions look quite doable and I'm planning on finishing them before the holidays! Still I'd like to make sure that we all agree on what the changes should look like:
> 
> Major changes:
> 
> * [ ]  move the lossy_bidirectional_links to _helpers
> * [ ]  combine the lossy_bidirectional_links with the length based link efficiency. I would add them as two seperate functions to helpers.
> * [ ]  maybe change the constraint implementation as suggested by @Eddy-JV.  I'm not yet fully sure on this, as I didn't have time yet to fully review it.
> 
> Minor changes:
> 
> * [ ]  fix the readme
> * [ ]  use only """ for docstring
> * [ ]  change query to loc in the constraint implementation (foreward_i)
> * [ ]  replace the regular expression in the constraint implementation (backward_i)
> * [ ]  add the transmission efficiencies provided by @Eddy-JV to the config
> * [ ]  include override_components to allow for the length based efficiency implementation (cf. PR [Length based pipeline efficiency #1215](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1215))
> * [ ]  add better comments (use the ones suggested by @Eddy-JV)
> 
> Discussions for another time:
> 
> * [ ]  type hints, yay or nay?
> * [ ]  should the pipelines be implemented based on pressure instead?
> * [ ]  remove unnecessary regular expressions.
> 
> Did I miss anything?

Hey @Eric-Nitschke @ekatef . I think the PR is ready for merge once the above checklist provided by @Eric-Nitschke is done. I also replied to @Eric-Nitschke's comment regarding the config file. Thank you @Eric-Nitschke for the great work and looking forward for future contributions.

scripts/add_extra_components.py
ekatef: Hey @Eddy-JV, frankly speaking I'm not quite aware of the proper approach to overwrite attributes. I assume the need arises from the need to translate a power-only model into the sector-coupled version, but having a bit more details would be appreciated 🙂

scripts/solve_network.py
Eric-Nitschke: Hey @Eddy-JV, even though the additional checks you proposed here are not neccessary right now, I think it's great to future-proof the code right now. I'm going to include the additional checks in the final version.
However, I wouldn't add the constraints one by one for each component, since no other constraint is added that way.

scripts/add_extra_components.py
Eric-Nitschke: Hey @ekatef, in this case the need arises from adding a compression electricity demand to the pipelines. This requires a third bus (bus2) for the link and a corresponding efficiency (efficiency2). The override_component_attr function adds these fields (and more) to the network.

ekatef:
Hello 😄 
Just wanted to ask if any support may be needed 🙂

@Eric-Nitschke thanks a lot for all the responses and creating a check-list, thank you so much @Eddy-JV for the in-depth review and providing insights into the architecture 🙏🏽

I wonder what is the plan to proceed? @Eric-Nitschke I see that you have ticked the boxes but it appears that the changes have not been pushed to github. Do you experience any issues with git? [I have no ideas why but git is normally provides you with a lot of surprises when you start working in a repo] Happy to assist in resolving them if needed. Also, I see that there are some conflicts with the main. They result from come changes we introduced recently inside main and can be resolved relatively easily. Let me know if any hints on that would be helpful

@Eddy-JV  on the architectural level, I understand the idea is to polish the pipelines representation to a level when the model is truly realistic. Though, I have a feeling that the work needed may be considerable. @Eric-Nitschke do you see how to proceed with the advanced implementation proposed by @Eddy-JV? May be any break-down helpful?

Eric-Nitschke:
Hey @ekatef,

I had some issues with testing the changes I made since I wasn't at my usual setup and I didn't want to merge untestes changes into the pull request. However you can find the changes I made in a sperate branch here: https://github.com/Eric-Nitschke/pypsa-earth-bidirectional_lossy_links/tree/lossy_length_based

Eric-Nitschke:
Hey @ekatef, hey @Eddy-JV,
I've now run some tests and merged the result with the pull request.
Using the override_components function from _helpers required some changes, as on Windows solve_networks uses a "copy-minimal" shadow repository to avoid administration issues. As a result the shadow directory does not have access to the files in the override_components file. I've resolved this by telling it explicitely where to look with "os.getcwd()" when using Windows.
Could you maybe check, if this doesn't have any negative implications on other operating systems?

Aside from this, I'll remove the merge conflicts and then everything should be good to go!

Eric-Nitschke:
As I've mentioned during the discord call:
Between the creation of this pull request and now, PR #1221 added transmission efficiencies to the config for the energy distribution grid. Initially I thought we should modify the use of this in efficiency in prepare_sector_network.py to use the length based efficiency calculation introduced here as well. However, after checking now, I don't think it is nesseccary, as the energy distribution network only uses static efficiencies and not length based ones.

Snakefile
```diff
@@ -818,6 +819,12 @@ if config["monte_carlo"]["options"].get("add_to_snakefile", False) == False:
             solving=config["solving"],
             augmented_line_connection=config["augmented_line_connection"],
         input:
+            overrides=(
+                os.getcwd() + "/data/override_component_attrs"
+                if os.name == "nt"
+                else "/data/override_component_attrs"
+            ),
+            # on Windows os.getcwd() is required because of the "copy-minimal" shadow directory
```
ekatef: Just to be sure I get this point properly: is shadow directory the one which is created by the solver to store the temporary files? I wonder if this behaviour may also depend on a solver...

Regarding the files management, it has been found previously that `os.getcwd( )` may lead to some troubles if pypsa-earth model is being used as snakemake submodule. In particular #1137 has fixed this replacing `os.getcwd( )` by a custom directory path. May it be a way to go also in our case?

config.default.yaml
```diff
@@ -171,7 +171,7 @@ electricity:
     Generator: [solar, onwind, offwind-ac, offwind-dc, OCGT]
     StorageUnit: [] # battery, H2
     Store: [battery, H2]
-    Link: [] # H2 pipeline
+    Link: [H2 pipeline] # H2 pipeline
```
ekatef: Just for better understanding: what is a particular idea of changing the default configuration parameters?

Agree that it may it be a good idea to make sure that the improvements introduced by this PR are being tested. It can be done by adding `H2 pipeline` as `extendable_carriers`:

https://github.com/pypsa-meets-earth/pypsa-earth/blob/ef525ee237b27916caf9a94a2d2157d25bd962f3/test/config.sector.yaml#L21-L24

What do you think?

ekatef (commented):
Hello @Eric-Nitschke @Eddy-JV, thank you for a fantastic collaborative work!

@Eric-Nitschke I see that you have implemented the massive improvements, and they look very good to me. Have added two rather minor comments comments which relate to the integration of the contribution into the codebase. Another point is CI which is currently complaining about `/data/override_component_attrs` missed. Do you have any ideas how to fix that?

On the side of the domain-related functionality, @Eddy-JV could you please take a look? Given your expert knowledge on H2 systems, it would be perfect to have your review. My personal impression is that we are closing to finalise 🙂

scripts/solve_network.py
Eric-Nitschke: @ekatef I realized that the myopic implementation names their components ending in a four letter number.
```rename[assets] += f"-{str(baseyear)}"```
It would be great, if someone working on the myopic implementation could confirm, whether it is important, that the year is the last part of the name. If it is important, I think we should revert back to the original implementation even if it isn#t that readable.

File: Snakefile (hunk 12 above)
Comment (Eric-Nitschke) [reply Snakefile]: In short:
The shadow directory is a Snakemake thing and not solver related.

With the "shadow" rule Snakemake creates an isolated temporary directory, in which it runs the corresponding rule (in this case "solve_network"). With ```shadow: "shallow"```, Snakemake symlinks all top level files and directory so that all relative file paths provided will still reach the right directories. However with ```shadow: "copy-minimal"```, the inputs of the rule are copied instead of symlinked. In my tests this led to Snakemake not having access to the files within the "data/override_component_attrs" folder. My guess is, that copy minimal only copies files directly, but not the content of directories. PR #790 is the reason, why "copy-minimal" is needed on Windows.

You can check the [Snakemake documentation](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#shadow-rules) for more details, though it is not that extensive.

Comment (Eric-Nitschke) [reply Snakefile]: However, I think I found the issue:
`else "/data/override_component_attrs"` should be
`else "data/override_component_attrs"`

Comment (Eric-Nitschke) [reply Snakefile]: Generally, itmight be nice to move away from `os.getcwd()` to the changes in #1137. However, I am not confident in implementing these changes myself.

File: config.default.yaml (hunk 13 above)
Comment (Eric-Nitschke) [reply config.default.yaml]: Yes testing was the main reason for adding it. I've moved it to test/config.sector.yaml.

**Comment by Eric-Nitschke** [comment]
The failed readthedocs relates to issue #1291.

File: Snakefile (hunk 12 above)
Comment (Eddy-JV) [reply Snakefile]: @davide-f This definitely needs your attention as you already made PR https://github.com/pypsa-meets-earth/pypsa-earth/pull/790. And @Eric-Nitschke if this approved by Davide, then probably the solution should be done for all Rules that use the solve_network.py script.

File: scripts/add_extra_components.py [hunk 14]
```diff
@@ -261,10 +267,15 @@ def attach_hydrogen_pipelines(n, costs, config):
...
         capital_cost=costs.at["H2 pipeline", "capital_cost"] * h2_links.length,
-        efficiency=costs.at["H2 pipeline", "efficiency"],
         carrier="H2 pipeline",
     )
 
+    # split the pipeline into two unidirectional links to properly apply transmission losses in both directions.
+    lossy_bidirectional_links(n, "H2 pipeline")
+
+    # set the pipelines efficiency and the electricity required by the pipeline for compression
+    set_length_based_efficiency(n, "H2 pipeline", " H2", config)
```
Eddy-JV: @Eric-Nitschke please add a variable like `transmission_efficiency = snakemake.params.transmission_efficiency`  this means cahnging this to: `set_length_based_efficiency(n, "H2 pipeline", " H2", transmission_efficiency )`

and in the snakefile please add to the rule add_extra_components:

```
       params:
              transmission_efficiency=config["sector"]["transmission_efficiency"],
```
in order to trigger the rule in case the efficiencies were changed in the config file by the user.

scripts/solve_network.py
```diff
@@ -854,6 +854,48 @@ def add_existing(n):
...
+    n.links["reversed"] = n.links.reversed.fillna(0).astype(bool)
+    carriers = n.links.loc[n.links.reversed, "carrier"].unique()  # noqa: F841
+
+    # get the indices of all forward links (non-reversed), that have a reversed counterpart
+    forward_i = n.links.loc[
+        n.links.carrier.isin(carriers) & ~n.links.reversed & n.links.p_nom_extendable
+    ].index
+
+    # get the indices of all backward links (reversed)
+    backward_i = forward_i + "-reversed"
```
Eddy-JV: @Eric-Nitschke as you already said somewhere in this PR that the 4 digits at the end are needed for the myopic. Hence I prefer to keep the function:

```
    # Function to get backward (reversed) indices corresponding to forward links
    def get_backward_i(forward_i):
        return pd.Index(
            [
                re.sub(r"-(\d{4})$", r"-reversed-\1", s)
                if re.search(r"-\d{4}$", s)
                else s + "-reversed"
                for s in forward_i
            ]
        )

    backward_i = get_backward_i(forward_i)
```

Eddy-JV (commented):
@Eric-Nitschke Thank you once again for the great work. I only have few comments shown above. After that the PR can be merged without my review. 

@davide-f  please check the `os.getcwd` comment.

@davide-f @ekatef I guess PR #1172 will have to be adapted to the new contraint function added here.

scripts/solve_network.py
Eric-Nitschke: Hey @Eddy-JV, thank you for the review!
I've implemented the changes you suggested. Additionally I've noticed that the other rules using solve_network.py currently do not have a Windows specific shadow directory setting. I've created PR #1295 to address this.

Snakefile
davide-f: os.getcwd() must be avoided as it does not track the use of submodules and can lead to issues.
Have you tried with directory(data/...)? [although the directory command should be optional for inputs].

Moreover, not sure if copy-minimal is strictly needed here; have you tested?

davide-f (commented):
Hello!
I've added the comment about os.cwd; we must avoid that unfortunately.

Regarding the linopy implementation of the new constraint:
- when is this PR ready?
- does this contraint must be added all the time or optionally?

Ideally, we should ideally not delay further too much the merge of the linopy PR [expected basically in these days after long preparation...]

Eddy-JV (approved):

Snakefile
Eric-Nitschke: @davide-f I've tried running it on Windows with "shallow" instead, which led to issues with admin rights similar to PR #790.

I've not tried directory(data/...) yet. Would that look something like this?
```
input:
            overrides=(
                directory("/data/override_component_attrs")
                if os.name == "nt"
                else "/data/override_component_attrs"
            ),
```

Removing the different handling (as done by @Eddy-JV) will most likely mean, that the code will not work properly on Windows since it will not find the proper files when running with "copy-minimal" (an issue that will not be solved by PR #1295).
However, I'm in favor of doing it this way, if it means that we'll merge this PR somewhat soon and create a new issue for the Windows issues.

Eric-Nitschke:
@davide-f technically the constraint is only required, if at least one lossy bidirectional component is used.
However, as the code is written right now (which is mostly taken from PyPSA-Eur), the constraint checks first if it is needed and if it isn't simply returns without changing anything, so there is no harm in calling it every run.

Adapting the constraint to PR #1172 should be somewhat easy, as the PyPSA-Eur implementation already used the newer PyPSA version including linopy.
To do so, simply change
```
lhs = linexpr(
    (1, get_var(n, "Link", "p_nom")[backward_i].to_numpy()),
    (-1, get_var(n, "Link", "p_nom")[forward_i].to_numpy()),
)

# add the constraint to the PySPA model
define_constraints(n, lhs, "=", 0, "Link-bidirectional_sync")
```
to
```
lhs = n.model["Link-p_nom"].loc[backward_i]
rhs = n.model["Link-p_nom"].loc[forward_i]

n.model.add_constraints(lhs == rhs, name="Link-bidirectional_sync")
```
which is the section I changed to make it work with linopf.

Eddy-JV:
> Hello! I've added the comment about os.cwd; we must avoid that unfortunately.
> 
> Regarding the linopy implementation of the new constraint:
> 
> * when is this PR ready?
> * does this contraint must be added all the time or optionally?
> 
> Ideally, we should ideally not delay further too much the merge of the linopy PR [expected basically in these days after long preparation...]

@Eric-Nitschke Thank you for swiftly addressing my comments and @davide-f  Thanks for checking this up. Now that this PR is mature enough to be merged and very much needed ASAP for the Linopy adaptation, I took the liberty to change this:

```
            overrides=(
                os.getcwd() + "/data/override_component_attrs"
                if os.name == "nt"
                else "data/override_component_attrs"
            ),
            # on Windows os.getcwd() is required because of the "copy-minimal" shadow directory

```
 back to 
 ```
          overrides="data/override_component_attrs",
 ```      
       
 from the current PR and address it seperately in PR #1295 or another PR. Awesome work @Eric-Nitschke , I will merge now.

ekatef:
@Eric-Nitschke, @Eddy-JV, @davide-f fantastic work!! 😄 Great to see the contribution completed and merged 🎉

@Eric-Nitschke if you wish to go ahead with adapting the constraint for linopy, feel free to open a PR on that 🙂

ekatef:
@Eric-Nitschke just flagging that the constraint update has been implemented #1276 as it's crucial for the model update. Apologies if I have been misleading and thanks a lot for the major contribution in resolving the folder management issue for windows!

PR #1195, opened 2024-11-14
Revision locate bus

davide-f:
# Closes # (if applicable).

This PR proposes a revision of the procedure used in locate_bus and adopt geospatial comparisons of geopandas. This methodology is heavier and has deep code implications; therefore, I quickly merged a fix to solve the CI first.

A similar approach can be also included into build_powerplants and cluster_network.
Currently, it is adopted in the sector scripts where more outliers have been founded

## Checklist


scripts/_helpers.py
```diff
@@ -1200,118 +1200,83 @@ def download_GADM(country_code, update=False, out_logging=False):
...
-        # convert country name representation of the main country (GID_0 column)
-        geodf_temp["GID_0"] = [
-            three_2_two_digits_country(twoD_c) for twoD_c in geodf_temp["GID_0"]
-        ]
-
-        # create a subindex column that is useful
-        # in the GADM processing of sub-national zones
-        geodf_temp["GADM_ID"] = geodf_temp[f"GID_{code_layer}"]
+def _get_shape_col_gdf(path_to_gadm, co, gadm_level, gadm_clustering):
+    from build_shapes import get_GADM_layer
```
ekatef: Are there any specific reasons to have a local import for `get_GADM_layer`? 

That is absolutely consistent with all the other code in `_helpers`, so the comment is not quite relevant for the PR. I'm just wondering if we are not over-using local imports inside the functions. What do you think?

scripts/_helpers.py
```diff
@@ -1200,118 +1200,83 @@ def download_GADM(country_code, update=False, out_logging=False):
...
-    geodf_GADM = gpd.GeoDataFrame(pd.concat(geodf_list, ignore_index=True))
-    geodf_GADM.set_crs(geodf_list[0].crs, inplace=True)
+    col = "name"
+    if not gadm_clustering:
+        gdf_shapes = gpd.read_file(path_to_gadm)
+    else:
+        if path_to_gadm:
+            gdf_shapes = gpd.read_file(path_to_gadm)
+            if "GADM_ID" in gdf_shapes.columns:
+                col = "GADM_ID"
```
ekatef, line 1250: Not sure it's really easy this part with four nested `if`-conditions. Agree that it's not the point to be addressed now, as the fix is quite urgent, but could we probably add a TODO for that?

scripts/_helpers.py
```diff
@@ -1200,118 +1200,83 @@ def download_GADM(country_code, update=False, out_logging=False):
...
-        if path_to_gadm:
-            gdf = gpd.read_file(path_to_gadm)
-            if "GADM_ID" in gdf.columns:
-                col = "GADM_ID"
+    if col_out is None:
+        col_out = "gadm_{}".format(gadm_level)
+    df = df[df.country.isin(countries)]
+    df[col_out] = None
+    for co in countries:
+        gdf_shape, col = _get_shape_col_gdf(
```
ekatef, line 1299: I recognise that it may be opening Pandora box, but wouldn't it be possible to replace naming for `df` and `col` with something a bit more specific? To my perception, it could probably facilitate a bit reading of the code. What do you think?

scripts/build_industrial_distribution_key.py
```diff
@@ -30,19 +30,10 @@ def map_industry_to_buses(df, countries, gadm_level, shapes_path, gadm_clusterin
     Only cement not steel - proof of concept.
     Change hotmaps to more descriptive name, etc.
     """
```
ekatef: A docstring of `map_industry_to_buses( )` seems to be a bit outdated. May it be a good idea to fix it?

scripts/build_industrial_distribution_key.py
```diff
@@ -123,9 +114,10 @@ def match_technology(df):
         snakemake = mock_snakemake(
             "build_industrial_distribution_key",
             simpl="",
-            clusters=12,
+            clusters=4,
```
ekatef: Are changes in `mock_snakemake( )` intentional?

ekatef (commented):
Brilliant work @davide-f!!

Have gone trough the changes and added some comments which are mainly quite minor. The only potentially crucial point is changing arguments of `mock_snakemake`, but I assume there has been a reason for that. All the other points can be also addressed in future PRs if you feel that improvements would take too much time for now.

Thanks a lot for investigating that, and looking forward to see CI running! 😄

scripts/_helpers.py
davide-f: Yes, otherwise there is a circular import: helpers imports build_shapes and build_shapes imports helpers and it does not work actually.
Tbh, in the future, we shall avoid the use of get_gadm_layer here potentially and use the shapes themself.
This PR however aims to revise the existing code, to avoid adding too many changes together.

scripts/_helpers.py
davide-f: Completely agree; this is to maintain the features of the previous code.
To revise this I believe there is the need to revise this together with the revision of the alternative clustering procedure.
Good to add a TODO

scripts/_helpers.py
davide-f: Ok, I'll try :)

scripts/build_industrial_distribution_key.py
davide-f: test1 of the sector-coupled has 4 clusters; not really intentional, though there is no consistency in the wildcards for testing. Good to rise an issue about that

scripts/_helpers.py
ekatef: Ahh, ok, thanks for explaining! Absolutely agree that it's better to keep PR focused

scripts/build_industrial_distribution_key.py
davide-f: Function has been dropped entirely as it was just a placeholder

davide-f:
Objective function changes by about 1% in sec model.
Honestly I hoped less, bit it doesn't look too bad. The 
Should we merge @ekatef?

scripts/_helpers.py
ekatef: Good to go as is :)

scripts/add_export.py
```diff
@@ -219,6 +208,7 @@ def create_export_profile():
             discountrate="0.071",
             demand="AB",
             h2export="120",
+            # configfile="test/config.test1.yaml",
```
ekatef, line 211: Is this comment needed?

scripts/_helpers.py
```diff
@@ -1241,104 +1238,82 @@ def get_GADM_layer(country_list, layer_id, update=False, outlogging=False):
         When the requested layer_id is greater than the last available layer, then the last layer is selected.
         When a negative value is requested, then, the last layer is requested
     """
```
ekatef, line 1240: Just to make sure: do we still maintain this functionality to have a safe fallback if an inappropriate layer id is requested? Not sure the check really needed here but probably worth to revise the docsting.

ekatef (approved):
@davide-f have gone with a final check and have only very minor points which probably need some attention. Generally, I think the PR is ready to go :)

Great work!

ekatef:
> Objective function changes by about 1% in sec model.
> 
> Honestly I hoped less, bit it doesn't look too bad. The 
> 
> Should we merge @ekatef?

Agree, 1% sounds reasonable 🙂

Have reviewed the implementation once again and it looks great! Let me know please if you need any assistance with merging!

scripts/add_export.py
davide-f: We can drop this but it is handy when testing tutorial or test configurations

scripts/_helpers.py
davide-f: This function is an utility to select the appropriate shapefile and correct the id when necessary.
Honestly, I believe we should drop it when we normalize the GADMs

davide-f:
Thanks, also for the comments :) I'd merge, many thanks!

PR #1198, opened 2024-11-15
Remove H2 and battery related techs from elec-only model before re-addition

yerbol-akhmetov:
Good day. This PR aims to help organization of H2 and battery addition to sector network. Currently, `H2 electrolysis`, `H2 Fuel cell`, `battery charger` and `battery discharger` links are added in `add_extra_components` and then re-added in `prepare_sector_network`. As a result, the second time it shows pypsa warning (error that components already exist). This PR aims to remove H2 and battery related components from electricity-only network, so that addition in sector script will be smooth. 

One option what I thought was addition of `if/else` statements to ensure adding buses, links, and stores if it does not exist. But that approach is not clean in terms of code, as a lot of `if`s need to be added. Also addition of H2 and battery related techs (links) will be accomplished in different scripts and functions, which might make it slightly difficult to debug.

Also removing them first from electricity network, and adding as a whole function will make it more consistent. As everything will be added with `add_hydrogen` and `add_storage` functions.
@davide-f 

## Checklist


scripts/prepare_sector_network.py
```diff
@@ -488,22 +488,9 @@ def add_links_elec_routing_new_H2_pipelines():
...
-            #     .apply(lambda id: three_2_two_digits_country(id[0]) + "." + id[1])
-            # )
-            # h2_links["bus1"] = (
-            #     h2_links["bus1"]
-            #     .str.split(".")
-            #     .apply(lambda id: three_2_two_digits_country(id[0]) + "." + id[1])
-            # )
+        # Appending string for carrier specification '_AC', because hydrogen has _AC in bus names
+        h2_links["bus0"] = buses_ordered.str[0] + "_AC"
+        h2_links["bus1"] = buses_ordered.str[1] + "_AC"
```
davide-f, line 493: Can you explain this proposal? Have you tested the feature with and without force_ac in case?

davide-f (commented):
Hello @yerbol-akhmetov ,

Many thanks! To me this approach seems reasonable. A quick, simple but robust solution

scripts/prepare_sector_network.py
yerbol-akhmetov: Hi, @davide-f. Thanks for the comment. Currently, hydrogen buses are added for `spatial.node`:
https://github.com/pypsa-meets-earth/pypsa-earth/blob/e386102552dc4471115432b5980e08125238b5f7/scripts/prepare_sector_network.py#L193-L200
It comes from `pop_layout.index`, which have buses names taken from `regions_onshore_elec_s_10.geojson`. So buses have names in form of `US.1_1_AC`. In other words, all buses are in form of `US.10_1_AC {technology}` (hydrogen buses are `US.1_1_AC H2`). The problem is that by default `bus0` and `bus1` of `h2_links` (and `buses_ordered`) are in a form of `US.1_1` without any `_AC` part. As a result, there were errors of having links with no defined buses (since those links are not attached to H2 buses with _AC parts in names). 

I have tested it with `force_ac: false`. I can test for `force_ac: true` too.

yerbol-akhmetov: The warning will be something like this. These links are connected to non-existing buses.
image

yerbol-akhmetov: For NG, both force_ac false and true worked fine. But I did not compared the networks.

davide-f: Great @yerbol-akhmetov :D Thanks!

davide-f (approved):
Contributions are great and useful.
CI on sector coupled model changes slightly due to the intended feature that is added.
Merging. Many thanks @yerbol-akhmetov :D

davide-f (commented):
@yerbol-akhmetov sorry, last point. Can you add a major contribution to the release_note?

davide-f (approved):

PR #1200, opened 2024-11-18
Add coal and gas CC technologies

yerbol-akhmetov:
Good day. This PR aims to make coal and gas CC technologies available for selection in the config. Currently, it is a draft PR. 
@davide-f, @danielelerede-oet, @GbotemiB 

I am planning to add gas CC technologies and also test with costs.

## Checklist


scripts/prepare_sector_network.py
```diff
@@ -157,6 +157,113 @@ def add_generation(
...
 
+    # add coal CC technologies
+    if options["coal"].get("CC", False):
+        coal_CC_techs = options["coal"].get("coal_CC_techs", list())
+
+        coal_CC_names = {
+            "coal CC-95": "Coal-95%-CCS",
+            "coal CC-99": "Coal-99%-CCS",
+            "coal IGCC CC-90": "Coal-IGCC-90%-CCS",
+        }
```
ekatef: I wonder if we can use this "nice" version of names directly in the config and avoid having an additional dictionary in the code?

In the current implementation, it's needed to make changes both in the config and `prepare_sector_network` is `coal_CC_techs` require changes, while it's very easy to forget about `prepare_sector_network` when introducing new technologies into the config

scripts/prepare_sector_network.py
```diff
@@ -157,6 +157,113 @@ def add_generation(
...
+                carrier="coal",
+                efficiency=costs.at[coal_CC_names[cc_tech], "efficiency"],
+                efficiency2=costs.at["coal", "CO2 intensity"]
+                * (1 - costs.at[coal_CC_names[cc_tech], "capture_rate"]),
+                efficiency3=costs.at["coal", "CO2 intensity"]
+                * costs.at[coal_CC_names[cc_tech], "capture_rate"],
+                lifetime=costs.at[coal_CC_names[cc_tech], "lifetime"],
+            )
+
+    # add coal IGCC if enabled
```
ekatef: May it be a good idea to explain `IGCC` abbreviation to facilitate reading the code?

scripts/prepare_sector_network.py
```diff
@@ -157,6 +157,113 @@ def add_generation(
...
+        )
+
+    # add natural gas CC technologies
+    if options["gas"].get("CC", False):
+        gas_CC_techs = options["gas"].get("gas_CC_techs", list())
+
+        gas_CC_names = {
+            "gas NGCC CC-95": "NG 2-on-1 Combined Cycle (F-Frame) 95% CCS",
+            "gas NGCC CC-97": "NG 2-on-1 Combined Cycle (F-Frame) 97% CCS",
+        }
```
ekatef: Same comment as above: can use this "nice" version of names directly in the config?

scripts/prepare_sector_network.py
```diff
@@ -157,6 +157,113 @@ def add_generation(
...
+
+    # add Natural Gas Combined Cycle plants if enabled
+    if options["gas"].get("gas_NGCC", False):
+        n.madd(
+            "Link",
+            spatial.nodes + " " + "gas NGCC",
+            bus0=spatial.gas.nodes,
+            bus1=spatial.nodes,
+            bus2="co2 atmosphere",
+            marginal_cost=costs.at["NG 2-on-1 Combined Cycle (F-Frame)", "efficiency"]
```
ekatef, line 242: Would be probably better to move `"NG 2-on-1 Combined Cycle (F-Frame)"` to the configuration file for consistency and to give a better overview. What do you think?

scripts/prepare_sector_network.py
```diff
@@ -157,6 +157,113 @@ def add_generation(
...
+        for cc_tech in coal_CC_techs:
+            n.madd(
+                "Link",
+                spatial.nodes + " " + cc_tech,
+                bus0=spatial.coal.nodes,
+                bus1=spatial.nodes,
+                bus2="co2 atmosphere",
+                bus3=spatial.co2.nodes,
+                marginal_cost=costs.at[coal_CC_names[cc_tech], "efficiency"]
+                * costs.at[coal_CC_names[cc_tech], "VOM"],  # NB: VOM is per MWel
```
ekatef: Just to double-check: variable costs relate to rated power, not to the overall amount of fuel consumed, right? [which is the reason why we multiply by efficiency, not divide]

scripts/prepare_sector_network.py
```diff
@@ -157,6 +157,113 @@ def add_generation(
...
+                # NB: fixed cost is per MWel
+                capital_cost=costs.at[coal_CC_names[cc_tech], "efficiency"]
+                * costs.at[coal_CC_names[cc_tech], "fixed"],
```
ekatef: I see that PyPSA-Eur is also [adding here capital costs](https://github.com/PyPSA/pypsa-eur/blob/920ffb444497c9767fade5f2a9bdaa457a1b627b/scripts/prepare_sector_network.py#L1687-L1689C1) for biomass-fuelled CHP. Not sure we have to do the same as it appears to be a region-specific assumption for Europe, but worth noticing the difference. What is your understanding of that?

config.default.yaml
```diff
@@ -358,6 +358,7 @@ costs:
   year: 2030
   version: v0.6.2
   discountrate: [0.071] #, 0.086, 0.111]
+  scenario: "Moderate" # "Advanced", "Moderate", or "Conservative"
```
ekatef: Would be great to add a comment on what exactly is meant by `scenario` here, as we have quite many of them in the config 😄 Probably, adjustment of the naming can help, as well: like `tech_scenario` or `costs_scenario`?

config.default.yaml
```diff
@@ -508,6 +509,9 @@ sector:
     network: false # ALWAYS FALSE for now (NOT USED)
     network_data: GGIT # Global dataset -> 'GGIT' , European dataset -> 'IGGIELGN'
     network_data_GGIT_status: ["Construction", "Operating", "Idle", "Shelved", "Mothballed", "Proposed"]
+    gas_NGCC: false # if true, Natural Gas 2-on-1 Combined Cycle (F-Frame) plant is added
+    CC: false # enable gas carbon capture technologies
```
ekatef, line 515: My feeling is that it would be great to keep harmonise using upper and lowercase across the code: we have lower-cased `cc` for industry and co-generation [here](https://github.com/pypsa-meets-earth/pypsa-earth/blob/d727d39e95a04a1c1fe7d5935921ad060ad3fdfc/config.default.yaml#L674-L675), while upper-case is used for [steam methane reforming](https://github.com/pypsa-meets-earth/pypsa-earth/blob/d727d39e95a04a1c1fe7d5935921ad060ad3fdfc/config.default.yaml#L673) and thermal power plants in this PR. Can we align that somehow?

ekatef (changes requested):
Hello @yerbol-akhmetov, great work! 😄

Have added some comments which mainly relate to work with the configuration parameters. Could you please have a look? (sorry for jumping in, Davide has just notified that he is busy this week)

config.default.yaml
yerbol-akhmetov: Thanks for the review @ekatef. I will probably remove these changes. As it needs to be applied with @danielelerede-oet and @finozzifa PR on cost assumption integration. Being under costs section scenario is plan to mean `cost scenario`. But if it can lead to misunderstanding then I agree we can name it as `cost_scenario`.

config.default.yaml
yerbol-akhmetov: Actually, my thought was to change everywhere to uppercase `CC`. As it is strange that in `SMR CC` it is uppercase, while just `cc` is lowercase. I can make those changes within this PR if you agree.

scripts/prepare_sector_network.py
yerbol-akhmetov: Thanks for the insightful comments. Initially, I planned to provide full name, but I doubted to do, because they are typically too long. So I thought there is a some chance that user misspells something (of course we provide names as comments, but still), so I provided shorter versions which are explained in the config file. I agree having a dictionary inside `prepare_sector_network` is not nice. I will change to directly nice names, thus it will remove unnecessary mapping.

scripts/prepare_sector_network.py
yerbol-akhmetov: Thanks, that was actually a question that bothered me earlier, because in all other technologies are implemented that way. Therefore, I raised a question in PyPSA-Earth meeting. @davide-f explained the reason well. I think, yes, you are right. Marginal cost are for MWh in electrical side, therefore we multiply by efficiency.

scripts/prepare_sector_network.py
yerbol-akhmetov: Yes, this is what we have discussed with @danielelerede-oet. In PyPSA-Eur, they consider capital cost of coal CC constitute from investment costs for coal part and capture part. For capture, the only available capture technology is costs is `biomass CHP capture`. However, here we directly used cost parameters of coal CC technologies, because NREL ATB already contains investment costs for technologies with capture (no need to separately account for coal and carbon capture, it is combined).

scripts/prepare_sector_network.py
yerbol-akhmetov: Thanks for the comment. I will add explicit name for abbreviation.

scripts/prepare_sector_network.py
yerbol-akhmetov: Ok, I will do so.

scripts/prepare_sector_network.py
yerbol-akhmetov: We already have it, named gas_NGCC. Also there is a comment which explains that it is `NG 2-on-1 Combined Cycle (F-Frame)`. Do you think what is enough? I did not added it to CC section because list in CC contains plants with carbon capture which are added with `for loop`.

yerbol-akhmetov (commented):
Thanks for the review, @ekatef. I will apply your suggestions into the code. However, this PR will still be draft as it will require new costs assumptions first being added to PyPSA/technology data and PyPSA-Earth.

scripts/prepare_sector_network.py
ekatef: Sounds absolutely reasonable. Agree that having a short name may be a good idea... I wonder if we can use `nice_names` argument in [the config](https://github.com/pypsa-meets-earth/pypsa-earth/blob/c12ce1add1ffbaafba4ff92dd2c821d00a6d73d5/config.default.yaml#L977-L989) to store those if they are needed? Then, in the code we could use those nice names if available and full names otherwise. Does it sound reasonable for you?

scripts/prepare_sector_network.py
ekatef: Agree with you comment on having long names. Feel free to revise according to what you feel more usable

scripts/prepare_sector_network.py
ekatef: Ahhh, I see. My major concern is linked with having the technologies parameters kind-of hard-coded in the script. That is also connected with what we have discussed for `gas NGCC CC` technologies above. My feeling is that it would be great to find a way to place for all the "advanced" technologies the names and "nice names" in the config for facilitate reviewing and modifying them. Does it make sense for you?

scripts/prepare_sector_network.py
ekatef: Perfect, we are on the same page then 🙂

scripts/prepare_sector_network.py
ekatef: Thanks a lot for looking into these details. As far as I'm aware, biomass co-firing in CHPs is a "soft" way to decrease CO2 emissions for coal-fired CHPs while keeping the costs low. However, there are some controversies around that (in particular to what relates for food-energy nexus and CO2 capture rates), so PyPSA-Eur implementation seems to be not a very general one, and most likely linked with some particular project

So, completely support the design you have selected

config.default.yaml
ekatef: Totally agree 🙂

ekatef:
> Thanks for the review, @ekatef. I will apply your suggestions into the code. However, this PR will still be draft as it will require new costs assumptions first being added to PyPSA/technology data and PyPSA-Earth.

Thanks for going trough the comments @yerbol-akhmetov. I just remembered that we have promised a preliminary review. So, I have tried to provide some initial feedback, though I absolutely agree that the final design depends also on the particular implementation of the technologies parameters in the technologies database.

My general impression is that PR introduces very interesting features into the model. Great work, and looking forward to see it being completed.

scripts/prepare_sector_network.py
yerbol-akhmetov: Thanks, @ekatef, for suggestion. I have implemented you suggestions of removing dictionary inside `prepare_sector_script`. What I have changed is for list of `coal_CC_techs` and `gas_CC_techs`, I have used full CC technologies as you have suggested. However, technologies without carbon capture such as `gas_NGCC` and `coal_IGCC` I have left with shorter name, because they are keys of config not values. But to make it clear I have added explicit name as the comment both in the config and `prepare_sector_networks.py`. I think it should prevent misunderstanding.

ekatef: Sounds like a good solution for now, I think 🙂 Agree that it doesn't make to overcomplicate the implementation for `gas_NGCC` and `coal_IGCC` since there are single versions of the technologies

ekatef (approved):

ekatef:
> Thanks for the review, @ekatef. I will apply your suggestions into the code. However, this PR will still be draft as it will require new costs assumptions first being added to PyPSA/technology data and PyPSA-Earth.

Thanks for the revisions @yerbol-akhmetov. I recognise that the PR is still work in progress, but I think it's on a good track. 

Have lifted `changes requested` status to prevent blocking the merge of this PR, but it has been interpreted by GitHub as approval (which is a funny status for a draft). Obviously, no pressure to complete the PR immediately, and feel free to complete it as you feel best.

davide-f (commented):
I'm commenting here as I've been pinged; I've scanned the contribution and looks quite promising!
Do you envision updates?

As minor comments, I'd support Katia's comment to remove the line "scenario" in config.default.

Moreover, to test the functionality of the model, I'd advise to revise the file test/config.test1.yaml to enable the sector-coupled technologies you proposed to ensure they get tested in the CI

yerbol-akhmetov:
> I'm commenting here as I've been pinged; I've scanned the contribution and looks quite promising! Do you envision updates?
> 
> As minor comments, I'd support Katia's comment to remove the line "scenario" in config.default.
> 
> Moreover, to test the functionality of the model, I'd advise to revise the file test/config.test1.yaml to enable the sector-coupled technologies you proposed to ensure they get tested in the CI

Thanks for revising it, @davide-f. The only change that I would like to make is to make sure these technologies are added only is costs file contains information about these techs, and also provide log that tech is not in costs file is not present. This way we make sure the error will not occur and user will need to use proper costs. The costs PR is not yet in pypsa-earth though. Adding if condition can help to prevent errors. What do you think? In short, I mean adding if conditions:
```
    if options["coal"].get("CC", False):
        coal_CC_techs = options["coal"].get("coal_CC_techs", list())

        if cc_tech in costs.index:
            for cc_tech in coal_CC_techs:
                n.madd(
                    "Link",
                    spatial.nodes + " " + cc_tech,
                    bus0=spatial.coal.nodes,
                    bus1=spatial.nodes,
                    bus2="co2 atmosphere",
                    bus3=spatial.co2.nodes,
                    marginal_cost=costs.at[cc_tech, "efficiency"]
                    * costs.at[cc_tech, "VOM"],  # NB: VOM is per MWel
                    # NB: fixed cost is per MWel
                    capital_cost=costs.at[cc_tech, "efficiency"]
                    * costs.at[cc_tech, "fixed"],
                    p_nom_extendable=True,
                    carrier="coal",
                    efficiency=costs.at[cc_tech, "efficiency"],
                    efficiency2=costs.at["coal", "CO2 intensity"]
                    * (1 - costs.at[cc_tech, "capture_rate"]),
                    efficiency3=costs.at["coal", "CO2 intensity"]
                    * costs.at[cc_tech, "capture_rate"],
                    lifetime=costs.at[cc_tech, "lifetime"],
                )
```

yerbol-akhmetov:
> I'm commenting here as I've been pinged; I've scanned the contribution and looks quite promising! Do you envision updates?
> 
> As minor comments, I'd support Katia's comment to remove the line "scenario" in config.default.
> 
> Moreover, to test the functionality of the model, I'd advise to revise the file test/config.test1.yaml to enable the sector-coupled technologies you proposed to ensure they get tested in the CI

Thanks, @davide-f. Regarding inclusion to test CI, I will revise it to include added techs there. But it will require costs file by NREL ATB to fully test it. Otherwise if we implement `if case` as I have mentioned above, it will not build those techs as they will not be in regular PyPSA-Earth cost file.

yerbol-akhmetov:
I am closing this PR, because new CC technologies did not get to technology-data.

PR #1221, opened 2024-12-04
Add electricity distribution grid

yerbol-akhmetov:
Good day. This PR aims to add electricity distribution grid to sector network. The PR is inspired by implementation of distribution grid in PyPSA-Eur. I think this will be useful. What do you think @davide-f ?

## Checklist

### DIFF
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -2629,6 +2629,137 @@ def add_residential(n, costs):
     )
 
 
+def add_electricity_distribution_grid(n, costs):
+
+    nodes = pop_layout.index
+
+    n.madd(
+        "Bus",
+        nodes + " low voltage",
+        location=nodes,
+        carrier="low voltage",
+        unit="MWh_el",
+    )
+
+    n.madd(
+        "Link",
+        nodes + " electricity distribution grid",
+        bus0=nodes,
+        bus1=nodes + " low voltage",
+        p_nom_extendable=True,
+        p_min_pu=-1,
+        carrier="electricity distribution grid",
+        efficiency=1,
+        lifetime=costs.at["electricity distribution grid", "lifetime"],
+        capital_cost=costs.at["electricity distribution grid", "fixed"],
+    )
+
+    # deduct distribution losses from electricity demand as these are included in total load
+    # https://nbviewer.org/github/Open-Power-System-Data/datapackage_timeseries/blob/2020-10-06/main.ipynb
+    if (
+        efficiency := options["transmission_efficiency"]
+        .get("electricity distribution grid", {})
+        .get("efficiency_static")
+    ):
+        logger.info(
+            f"Deducting distribution losses from electricity demand: {np.around(100*(1-efficiency), decimals=2)}%"
+        )
+        n.loads_t.p_set.loc[:, n.loads.carrier == "AC"] *= efficiency
+
+    # move AC loads to low voltage buses
+    ac_loads = n.loads.index[n.loads.carrier == "AC"]
+    n.loads.loc[ac_loads, "bus"] += " low voltage"
+
+    # move industry, rail transport, agriculture and services electricity to low voltage
+    loads = n.loads.index[n.loads.carrier.str.contains("electricity")]
+    n.loads.loc[loads, "bus"] += " low voltage"
+
+    bevs = n.links.index[n.links.carrier == "BEV charger"]
+    n.links.loc[bevs, "bus0"] += " low voltage"
+
+    v2gs = n.links.index[n.links.carrier == "V2G"]
+    n.links.loc[v2gs, "bus1"] += " low voltage"
+
+    hps = n.links.index[n.links.carrier.str.contains("heat pump")]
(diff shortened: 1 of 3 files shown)


scripts/prepare_sector_network.py
```diff
@@ -2629,6 +2629,141 @@ def add_residential(n, costs):
...
+        )
+        n.loads_t.p_set.loc[:, n.loads.carrier == "AC"] *= efficiency
+
+    # move AC loads to low voltage buses
+    ac_loads = n.loads.index[n.loads.carrier == "AC"]
+    n.loads.loc[ac_loads, "bus"] += " low voltage"
+
+    # move industry, rail transport, agriculture and services electricity to low voltage
+    loads = n.loads.index[n.loads.carrier.str.contains("electricity")]
+    n.loads.loc[loads, "bus"] += " low voltage"
```
ekatef, line 2675: It appears to me that PyPSA-Eur is using naming conventions which are specific for the dataset they are using (according to [this comment](https://github.com/PyPSA/pypsa-eur/blob/920ffb444497c9767fade5f2a9bdaa457a1b627b/scripts/prepare_sector_network.py#L1229-L1231)). Have you checked if it works also in our case?

scripts/prepare_sector_network.py
```diff
@@ -2629,6 +2629,141 @@ def add_residential(n, costs):
     )
 
 
+def add_electricity_distribution_grid(n, costs):
+    logger.info("Adding electricity distribution network")
```
ekatef, line 2633: Great that you thought about adding an info message

scripts/prepare_sector_network.py
```diff
@@ -2629,6 +2629,141 @@ def add_residential(n, costs):
...
+    n.links.loc[hps, "bus0"] += " low voltage"
+
+    rh = n.links.index[n.links.carrier.str.contains("resistive heater")]
+    n.links.loc[rh, "bus0"] += " low voltage"
+
+    mchp = n.links.index[n.links.carrier.str.contains("micro gas")]
+    n.links.loc[mchp, "bus1"] += " low voltage"
+
+    if options.get("solar_rooftop", False):
+        logger.info("Adding solar rooftop technology")
```
ekatef, line 2693: Just out of curiousity: do I understand properly that PyPSA-Eur adds (nodes + links) for distribution grid for all "distribution-relevant" technologies even they are not being actually used in the model?

ekatef (commented):
Hello @yerbol-akhmetov, looks like great and nicely implemented contribution. My only concern relates to the naming conventions. Could you please check it?

Otherwise, it appears to be an advancement of PyPSA-Eur approach 🙂

scripts/prepare_sector_network.py
yerbol-akhmetov: Hi, @ekatef. Thanks for the comment. Here `loads` represent loads for `industry electricity`, `rail transport electricity`, `agriculture electricity` and `services electricity`. All these contain electricity in names and exist in PyPSA-Earth. I have checked that with `n.loads.carrier.unique()`. I have also taken there `ac_loads` and converted to low voltage because in PyPSA-Earth AC loads are named as carrier `AC` (rather than `electricity` as in PyPSA-Eur).

scripts/prepare_sector_network.py
yerbol-akhmetov: Can you please clarify that? If I understand properly, you ask about if conditions for solar rooftop and home battery right? In PyPSA-Eur there is no if conditions, so they are forced to model. Here I have added if condition so you can turn them off, if not needed for some scenario (we have actually come across with such scenario). However, by default `home battery` and `solar rooftop` are added into the model.

yerbol-akhmetov (commented):
Thanks for the feedback @ekatef

davide-f (approved):
Hello @yerbol-akhmetov :D
Thanks for the contribution! 
To me the contribution looks great and is tested in the CI as the technology is added by default.
The log does not have warnings related to the CI and contribution does not affect the rest of the code.
If katia is satisfied, on my side we can merge this :D

scripts/prepare_sector_network.py
ekatef: Sounds like a great approach

scripts/prepare_sector_network.py
ekatef: Perfect, thanks a lot for the double-check. Also, a great point on AC loads.

ekatef (approved):

ekatef:
> Hello @yerbol-akhmetov :D Thanks for the contribution! To me the contribution looks great and is tested in the CI as the technology is added by default. The log does not have warnings related to the CI and contribution does not affect the rest of the code. If katia is satisfied, on my side we can merge this :D

I'd say that not only CI is working but the implementation is also perfectly correct and well-elaborated 😄 Great work @yerbol-akhmetov! 

Absolutely agree that we must merge it. @yerbol-akhmetov would you mind to resolve the merge conflict? Then, the PR is ready to go 🚀

ekatef:
Merged 😄 Thanks a lot for the great contribution @yerbol-akhmetov!!

PR #1226, opened 2024-12-05
update FT to account for Hydrogen and Electricity input

GbotemiB:
# Closes # (if applicable).

This PR aims to revise the implementation of FT to account for hydrogen and electricity input. The changes is adapted from [pypsa-eur](https://github.com/PyPSA/pypsa-eur/blob/920ffb444497c9767fade5f2a9bdaa457a1b627b/scripts/prepare_sector_network.py#L1143-L1160)

@ekatef @davide-f @hazemakhalek 

## Checklist


yerbol-akhmetov (commented):
Thanks @GbotemiB for the PR. What was the reason of changing `efficiency` and efficiency2`? In PyPSA-Eur, efficiencies were similar as it is in PyPSA-earth previously.
```
        efficiency=costs.at["Fischer-Tropsch", "efficiency"],
        efficiency2=-costs.at["oil", "CO2 intensity"]
        * costs.at["Fischer-Tropsch", "efficiency"],
```

scripts/prepare_sector_network.py
```diff
@@ -171,13 +171,13 @@ def H2_liquid_fossil_conversions(n, costs):
...
+        efficiency=1 / costs.at["Fischer-Tropsch", "electricity-input"],
+        efficiency2=costs.at["Fischer-Tropsch", "hydrogen-input"]
+        / costs.at["Fischer-Tropsch", "electricity-input"],
         capital_cost=costs.at["Fischer-Tropsch", "fixed"]
-        * costs.at[
-            "Fischer-Tropsch", "efficiency"
-        ],  # Use efficiency to convert from EUR/MW_FT/a to EUR/MW_H2/a
-        efficiency2=-costs.at["oil", "CO2 intensity"]
-        * costs.at["Fischer-Tropsch", "efficiency"],
+        / costs.at["Fischer-Tropsch", "electricity-input"],
```
yerbol-akhmetov: Also, I think capital_cost should be multiplication of `fixed` with `efficiency` as previously defined.

scripts/prepare_sector_network.py
```diff
@@ -171,13 +171,13 @@ def H2_liquid_fossil_conversions(n, costs):
...
+        marginal_cost=costs.at["Fischer-Tropsch", "VOM"]
+        / costs.at["Fischer-Tropsch", "electricity-input"],
```
yerbol-akhmetov: The same is true for `marginal_cost`. Efficiency should be multiplied. @davide-f, can you please comment? Also generally `efficiency` is used rather than `electricity-input`.

davide-f: Indeed the multiplication with efficiency is appropriate because of the unit of measurement used for the calculation.
Regarding this PR, I'm adding a comment below also

scripts/prepare_sector_network.py
davide-f: Agree, see comment below

davide-f (commented):
Mmm could you explain the proposal?
The referenced link to pypsa-eur refers to the haber-bosch process that converts electricity and hydrogen into ammonia, while here the edited code is Fischer-Tropsch that converts hydrogen and CO2 into oil products.
It feels to me that we are mixing things, could you please check?


Note that fischer trops is also availabel in pypsa-eur with a similar formulations as in -earth:
https://github.com/PyPSA/pypsa-eur/blob/920ffb444497c9767fade5f2a9bdaa457a1b627b/scripts/prepare_sector_network.py#L3610-L3616

GbotemiB:
Hi @yerbol-akhmetov @davide-f @danielelerede-oet, I will be closing this PR for now. The purpose of this PR was to add electricity and hydrogen input for FT. But this is no longer needed as the current implementation is sufficient for now.

GbotemiB:
Hi @davide-f @hazemakhalek @yerbol-akhmetov @danielelerede-oet 
I have remodified the `H2_liquid_fossil_conversions` function, by adding electricity bus to bus3. I also made change to how the efficiency is calculated. 

Here is an explanation. Big thanks to @yerbol-akhmetov 
image

yerbol-akhmetov (commented):
Thanks, @GbotemiB, for the proposal. I agree with using `1 / costs.at["Fischer-Tropsch", "hydrogen-input"]` as an efficiency as `hydrogen-input` has units of `MWh_H2/MWh_FT`. So it maps output to input. Note, `1/hydrogen-input` is very close to efficiency value in costs. It is interesting how `efficiency` is estimated and what it refers to?
Another question is: if we use `1 / costs.at["Fischer-Tropsch", "hydrogen-input"]` as efficiency from inpu to output, then should we change `costs.at["Fischer-Tropsch", "efficiency"]` elsewhere? What do you think @davide-f ?

davide-f (commented):
Thanks @GbotemiB and @yerbol-akhmetov :D

The proposal is quite neat; I'd advise for a minor release note also.

Regarding the point by @yerbol-akhmetov , I agree that it is advisable to use -input parameters or efficiency, but not a mix of them for consistency.
For reproducibility considerations, I'd be more prone in merging a contribution where we keep using efficiency instead of hydrogen-input for now, for consistency with pypsa-eur and reproducibility considerations.
Using hydrogen-input sounds appealing, but given its potential breaking changes may require some alignment with others.

The potential issue I see is that with this change, the results of the model with the same cost configuration will change because the code selects a different input data.
The current efficiency and the efficiency with hydrogen-input differ by >5% that may have implications; on the other hand, the electricity-input is likely to have a small impact.
Current CI changes only marginally [727049... vs 727069...], but the CO2 constraint is limited so the impact of FT is not big.

Is there a specific need to adopt hydrogen-input here for @yerbol-akhmetov  and @GbotemiB ?

I ping also @FabianHofmann as it would be good to keep an alignment with pypsa-eur and @energyLS who worked on -sec PRs about it.
For fabian and -sec modellers, using hydrogen-input and electricity-input seems interesting: the data source of *-input in technology data is newer. Are you aware of reasons for not using it?
Personally, using hydrogen-input and electricity-input seems appealing but need to align before acting here

davide-f:
Hello @yerbol-akhmetov @GbotemiB,
I'm just writing here to avoid stalling, is there a reason to adopt hydrogen-input instead of efficiency for your application?
Could efficiency be used instead?

GbotemiB:
Hi @davide-f, We can just use efficiency.

davide-f:
This looks good to me.
To clarify, @GbotemiB have you discussed this contribution with the sec team during a -sec meeting?

davide-f:
Ok, I understand you have discussed it and given its consistency, we can merge. If issues will be found, we can change in the future.
Great :D

davide-f (approved):

Issue #1231, opened 2024-12-08
H2 export loads are not satisfied due to duplicate `hydrogen_buses_ports`


yerbol-akhmetov:
## Checklist

- [x] I am using the current [`main`](https://github.com/pypsa-meets-earth/pypsa-earth/tree/main) branch or the latest [release](https://github.com/pypsa-meets-earth/pypsa-earth/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-earth` environment](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

Good day, I have noticed a bug that leads an infeasibility of sector network. The issue was noticed when simulating US. It is related to `H2 export`, where H2 load is added in `add_export.py` script. However, the links that will connect external `H2 export bus` containing H2 load and internal H2 buses are not added. It is because the names are duplicate for `hydrogen_buses_ports` used to define links, as a result, the links are not added (which cause infeasibility as demand is not satisfied). I have found the cause of the issue by making `load_shedding: true`. So probably we need to remove duplicates to solve the issue. @ekatef @davide-f 

![image](https://github.com/user-attachments/assets/262bb286-cb9e-49ed-a345-531c973f2911)

![image](https://github.com/user-attachments/assets/9383114f-e6ab-4888-9957-749c56fdb332)

## Error Message

![image](https://github.com/user-attachments/assets/13dcc9d3-1cf3-4e93-9e9c-e445b5e4c76d)

Having links `repurposed` without buses connected is another topic that I will study.

ekatef:
Hello @yerbol-akhmetov, thanks for the reporting and working on the fix! Has the issue been resolved with #1233?

yerbol-akhmetov:
Thanks, @ekatef, for pointing that out. Yes, the issue was resolved with that PR. I will close the issue.

ekatef:
> Thanks, @ekatef, for pointing that out. Yes, the issue was resolved with that PR. I will close the issue.

Many thanks for fixing that! 😄

PR #1233, opened 2024-12-10
Drop duplicate hydrogen ports in `add_export.py`

yerbol-akhmetov:
Good day. This PR aims to minor a bug related to duplicate entries of H2 export ports which prevents addition of links that connect `H2 export bus` and H2 buses. This issue happens particularly for US, where multiple ports are available. The PR is linked to issue #1231. Dropping the duplicates solve the issue.
image

## Checklist

### DIFF
--- a/scripts/add_export.py
+++ b/scripts/add_export.py
@@ -66,6 +66,9 @@ def select_ports(n):
     hydrogen_buses_ports = n.buses.loc[ports.index + " H2"]
     hydrogen_buses_ports.index.name = "Bus"
 
+    # drop duplicate ports if exists
+    hydrogen_buses_ports = hydrogen_buses_ports.drop_duplicates()
+
     return hydrogen_buses_ports
 
 
--- a/scripts/add_export.py
+++ b/scripts/add_export.py
@@ -60,14 +60,14 @@ def select_ports(n):
         axis=1,
     )
 
-    ports = ports.set_index("gadm_{}".format(gadm_level))
+    # TODO: revise if ports quantity and property by shape become relevant
+    # drop duplicated entries
+    gcol = "gadm_{}".format(gadm_level)
+    ports_sel = ports.loc[~ports[gcol].duplicated(keep="first")].set_index(gcol)
 
     # Select the hydrogen buses based on nodes with ports
-    hydrogen_buses_ports = n.buses.loc[ports.index + " H2"]
+    hydrogen_buses_ports = n.buses.loc[ports_sel.index + " H2"]
     hydrogen_buses_ports.index.name = "Bus"
-
-    # drop duplicate ports if exists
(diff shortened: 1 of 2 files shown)


GbotemiB:
I have also tested this and it works properly. Before this, the US model was not feasible, but with this solution the model is now feasible.

ekatef:
Thank you @yerbol-akhmetov! Great to hear that the fix works, and a great investigation of the reasons  #1231. 

Just to be sure, I wonder if we could drop some available shipping capacity when dropping duplicated buses in the current implementation? From the way how the `hydrogen_buses_ports` are being formed, it appears that the duplicates correspond to the same GADM index:

https://github.com/pypsa-meets-earth/pypsa-earth/blob/5d02b2920e1839f3c325093d53a728936351f4c2/scripts/add_export.py#L63-L66

May it be the case that duplicates in `hydrogen_buses_ports` are different ports but located in the same GADM region?

scripts/add_export.py
```diff
@@ -66,6 +66,9 @@ def select_ports(n):
     hydrogen_buses_ports = n.buses.loc[ports.index + " H2"]
     hydrogen_buses_ports.index.name = "Bus"
 
+    # drop duplicate ports if exists
+    hydrogen_buses_ports = hydrogen_buses_ports.drop_duplicates()
```
davide-f: Thanks a lot @yerbol-akhmetov for this great proposal; I'd propose just a minor revision that may make the code more robust.

If I get the issue right, the problem in the current workflow is that the column "gadm_{}".format(gadm_level) used as index may contain duplicates.
For your case, however, drop_duplicates works.
The function drop_duplicates() by default drops only duplicated rows, if rows have the same index but different content they are not dropped.

To ensure robustness and keep the proposed formulation, a possible proposal would be to move the drop_duplicates few lines up with something like:
``` python
    gcol = "gadm_{}".format(gadm_level)
    ports_sel = ports.loc[~ports[gcol].duplicated(keep='first')].set_index(gcol)
```
May that work? Feel free to revise and improve as you feel best

I'd also recommend to add a comment like "TODO: revise if ports quantity and property by shape become relevant" just to keep track of this approximation

yerbol-akhmetov: Thanks @davide-f, @ekatef. I have applied the latest changes that were suggested and tested it locally. It worked as intended. Also, the comment about TODO was added too.

ekatef (approved):

scripts/add_export.py
ekatef: Ahh, true!! The check for duplicates relate to the whole lines, not only indices. Agree then that data loss is highly unlikely.

I think the implementation looks great, and if @davide-f doesn't have further comments, happy to merge 😄

davide-f:
Merged, many thanks @yerbol-akhmetov :D

PR #1244, opened 2024-12-19
add wiki table to complement number of vehicles per country in transport_data csv

ljansen-iee:
Currently, the number of registered vehicles per country (part of resources/transport_data.csv) is downloaded from the WHO Global Health Observatory data repository. A few countries are missing in the WHO list (e.g. South Africa, Algeria). 

* Fill in the missing countries with the number of vehicles per country from Wikipedia 'https://en.wikipedia.org/wiki/List_of_countries_and_territories_by_motor_vehicles_per_capita'
* The proposed concatenation of downloaded vehicle data will prioritise the WHO list -> existing model instances for previously available countries will not be affected.
* Refactoring and restructuring of the code in the prepare_transport_data_input.py script

## Checklist

- [x] I tested my contribution locally and it seems to work fine:
    - With the changes, the code behaves as expected: Number of vehicles for a few more countries are added. However, it seems that the World Bank's CO2_emission API changed in November and the requested data, CO2 emissions from transport in % of total fuel combustion, is no longer available. Since this is not part of the proposed contribution of this pull request, I'll write a separate issue about it. 

### DIFF
--- a/scripts/prepare_transport_data_input.py
+++ b/scripts/prepare_transport_data_input.py
@@ -14,72 +14,84 @@ from _helpers import BASE_DIR
 
 logger = logging.getLogger(__name__)
 
+
 def _add_iso2_code_per_country_and_clean_data(df):
-    
+
     cc = coco.CountryConverter()
     df["country"] = cc.pandas_convert(
-        series=pd.Series(df["Country"]), to="ISO2", not_found="not found")
-    
+        series=pd.Series(df["Country"]), to="ISO2", not_found="not found"
+    )
+
     df = df[df.country != "not found"]
-    
+
     # Drop region names where country column contains list of countries
     df = df[df.country.apply(lambda x: isinstance(x, str))]
-    
+
     df = df.drop_duplicates(subset=["country"])
-    
+
     return df
+
 
 def download_number_of_vehicles():
     """
-    Downloads and returns the number of registered vehicles 
+    Downloads and returns the number of registered vehicles
     as tabular data from WHO and Wikipedia.
-    
-    The csv data from the WHO website is imported 
-    from 'https://apps.who.int/gho/data/node.main.A995'. 
-    A few countries are missing in the WHO list (e.g. South Africa, Algeria). 
-    Therefore, the number of vehicles per country table from Wikipedia 
+
+    The csv data from the WHO website is imported
+    from 'https://apps.who.int/gho/data/node.main.A995'.
+    A few countries are missing in the WHO list (e.g. South Africa, Algeria).
+    Therefore, the number of vehicles per country table from Wikipedia
     is also imported for completion (prio 2):
     'https://en.wikipedia.org/wiki/List_of_countries_and_territories_by_motor_vehicles_per_capita'.
     """
-    
+
     def _download_vehicles_data_from_gho():
-        url = ("https://apps.who.int/gho/athena/data/GHO/RS_194?filter=COUNTRY:*&ead=&x-sideaxis=COUNTRY;YEAR;DATASOURCE&x-topaxis=GHO&profile=crosstable&format=csv")
+        url = "https://apps.who.int/gho/athena/data/GHO/RS_194?filter=COUNTRY:*&ead=&x-sideaxis=COUNTRY;YEAR;DATASOURCE&x-topaxis=GHO&profile=crosstable&format=csv"
         storage_options = {"User-Agent": "Mozilla/5.0"}
         df = pd.read_csv(url, storage_options=storage_options, encoding="utf8")
-        
-        df.rename(columns={"Countries, territories and areas": "Country",
-                           "Number of registered vehicles": "number cars"},
-                  inplace=True)
-        
+
+        df.rename(
+            columns={
+                "Countries, territories and areas": "Country",
+                "Number of registered vehicles": "number cars",
+            },
+            inplace=True,
+        )
+
         df["number cars"] = df["number cars"].str.replace(" ", "").replace("", np.nan)
-        
+
         df = df.dropna(subset=["number cars"])
-        
+
         return df[["Country", "number cars"]]
 
     def _download_vehicles_data_from_wiki():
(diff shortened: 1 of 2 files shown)


scripts/prepare_transport_data_input.py
```diff
@@ -12,56 +12,87 @@
 import pandas as pd
 from _helpers import BASE_DIR
 
-# logger = logging.getLogger(__name__)
+logger = logging.getLogger(__name__)
+
+
+def _add_iso2_code_per_country_and_clean_data(df):
+
```
davide-f, line 23: Would you mind adding a minor string or drop this line?

scripts/prepare_transport_data_input.py
```diff
@@ -12,56 +12,87 @@
...
+    df["country"] = cc.pandas_convert(
+        series=pd.Series(df["Country"]), to="ISO2", not_found="not found"
+    )
+
+    df = df[df.country != "not found"]
+
+    # Drop region names where country column contains list of countries
+    df = df[df.country.apply(lambda x: isinstance(x, str))]
+
+    df = df.drop_duplicates(subset=["country"])
```
davide-f, line 34: Here we drop, should we sum instead? is there a reason?
It can be a TODO also, though a log info may be advisable if that happens

scripts/prepare_transport_data_input.py
```diff
@@ -71,40 +102,38 @@ def download_CO2_emissions():
...
+
+        CO2_emissions.loc[:, "average fuel efficiency"] = CO2_emissions[
+            "average fuel efficiency"
+        ].round(3)
+
+        return CO2_emissions[["country", "average fuel efficiency"]]
+
+    except Exception as e:
+        logger.warning("Failed to read the file:", e)
+        return pd.DataFrame()
```
davide-f: Moving the exception at the end may mask some errors that happen along the procedure due to change in the data format, rather than access to the source.
That may be less desirable, is there a reason for the proposal?

davide-f (commented):
Great contribution @ljansen-iee :D Many thanks and it is great to have data filling :)

I've added some minor comments; plus I'd advise to add a major contribution into the file doc/release_note.rst mentioning this PR, for example:
"* Add wikipedia source to transport_data \`PR #1244 \<https://github.com/pypsa-meets-earth/pypsa-earth/pull/1244>`__"

I'd also kindly ask @Eddy-JV to look at this PR as he contributed significantly to this section and his opinion would be extremely valuable.

scripts/prepare_transport_data_input.py
ljansen-iee: Sure, added a minor doc string.

scripts/prepare_transport_data_input.py
ljansen-iee: Hey, good point, I thought it would be more desirable, but hadn't given it much thought. I've limited the exception handling to the download part again - as originally introduced by @Eddy-JV and @GbotemiB.

scripts/prepare_transport_data_input.py
ljansen-iee: The drop_duplicates function with keep='first' is only used to keep the WHO GHO list entries and delete the duplicate entries per country in the Wikipedia source. -> I have moved the operation to the vehicle function.

Would you still sum the values based on this information? Why would we do that?
Indeed, the values per country are different more often than expected. For example, WHO-GHO reports 50.6 million registered vehicles in Vietnam, while Wikipedia reports 4.18 million. But I can't say which sources (secondary or primary) are more valid or appropriate.
What would you recommend?

ljansen-iee:
Hey @davide-f, thanks for welcoming the contribution! :) I'm happy to follow the advice to add it to the release notes! I have followed your suggestion.

scripts/prepare_transport_data_input.py
davide-f: I now understand the process; it is a bit misleading but works: this function does not clean each source but the whole dataframe.
It implicitly assumes that as WHO are always first, then those inputs are kept first.
It would be better to select those inputs from Wikipedia that are missing in the first dataframe for stability/code robustness, yet currently it works.
The keep='first' is not explicit, so even the change in the default implementation of the function can lead to change of the results that would be completely silent in the workflow.

Have you discussed this method with the others?

scripts/prepare_transport_data_input.py
```diff
@@ -12,56 +12,93 @@
...
+        url = "https://en.wikipedia.org/wiki/List_of_countries_and_territories_by_motor_vehicles_per_capita"
+        df = pd.read_html(url)[0]
+
+        df.rename(
+            columns={"Location": "Country", "Vehicles": "number cars"}, inplace=True
+        )
+
+        return df[["Country", "number cars"]]
+
     try:
```
davide-f, line 81: A nice-to-have would be moving this try/catch inside the _download function exactly to filter the pd.read_html/pd.read_csv functions.
That allows to well capture the main issue of the problem, what do you think?
To ensure that all countries of interest are available, we can add a final check and if countries are missing, an error can be thrown [or an empty dataframe]

The use of try/catch can easily silent issues that may be relevant to tackle, such as data changes in the workflow.
That especially in the case of the filtering that relies on keep='first'.
Moreover, if just one of the 2 sources is offline, the whole method is no longer used.

davide-f (commented):
Hello @ljansen-iee :D
Great contribution and thanks for the comments.

I've added few comments to align with you; some points may be left for follow-up if deemed intensive.
Have you discussed this implementation with others? Apologies but I couldn't attend many sec meetings recently, by next year the situation is expected to improve :)

Have you discussed the method with others?

ljansen-iee:
Hi @davide-f , sorry for the delay. I created a new branch and a new pull request today for both problems with the prepare_transprot_data_input script (missing countries and discontinued World Bank source. See #1410 and #1402. I have taken your comments into account in the new pull request. It would be great if I could get your feedback there again.

PR #1315, opened 2025-01-21
Align structure of the components with PyPSA consistency checks

ekatef:
A possible solution of the issues caused by AssertionErrors in PyPSA, like #1308.

The updated PyPSA version has more strict requirements towards the component structure which can easily lead to errors in case when `n.lines` and `n.buses` contains any "extra" fields. The PR aims to drop all the columns after `base_network` to make the workflow more robust. 

## Checklist


scripts/base_network.py
```diff
@@ -492,6 +492,10 @@ def base_network(
     lines_ac = lines[lines.tag_frequency.astype(float) != 0].copy()
     lines_dc = lines[lines.tag_frequency.astype(float) == 0].copy()
 
+    # avoiding the conflicts in processing the network model
```
davide-f: Instead of dropping tag_frequency here, we could drop it in build_osm_nwtwork and rely on the "dc" column in base_networl instead.
What do you think?

In build osm, we have both the dc and tag_frequency columns that are passed to base_network.
Probably, we can drop tag frequency in build and rely on dc column

davide-f (commented):
Thanks for taking care of this!
A possible comment and we are ready to go!
Please add a line in the release note

scripts/base_network.py
ekatef: Ahh, I see! We now have calculation of `dc` column directly in `clean_osm` script, so it's basically a duplication in `base_network`! Agree, no need then to keep `tag_frequency` up to `base_network`. Thanks a lot for checking it!

Revised 🙂

davide-f (approved):

davide-f:
Thanks @ekatef :D
Merged

PR #1319, opened 2025-01-23
Universal currency conversion

danielelerede-oet:
Hi, @finozzifa and I have been working on universal currency conversion.

In this PR, universal currency conversion is introduced taking advantage of the Currency_Converter package. 
Currently, the only allowed conversion adopts a fixed USD to EUR exchange rate for the year 2013. Here, any currency among those in the list at https://github.com/alexprengere/currencyconverter/blob/master/currency_converter/eurofxref.csv may be introduced in the PyPSA-Earth input dataset and converted into the `output_currency` specified in the config file under `costs` according to the specified `currency_year` for each investment, FOM and VOM cost (the adopted conversion factor will correspond to the yearly average for the specified `currency_year`, according to the `get_yearly_currency_exchange_average` in `_helpers.py`).

Additionally, in this way results can be obtained in any currency the user prefers (according to the specification of `output_currency`) - EUR is set as default as it is in the current PyPSA-Earth.

The only caveat at the moment is that cost units must be specified via code (i.e., USD, EUR, etc. instead of $, €). That is not an issue as the current `technology_data` output files only contain codes, but a PR will be also raised to `technology_data` to address the conversion from symbol to code, if necessary (something similar is already there now, but only acts on some input data).

## Checklist

- [X] Code and workflow changes are sufficiently documented.
- [X] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`.
- [X] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.


config.default.yaml
```diff
@@ -358,10 +358,9 @@ renewable:
 # Costs Configuration
 costs:
   year: 2030
-  version: v0.10.0
+  version: v0.11.0
```
ekatef: My feeling is that `v0.10.0 -> v.0.11.0` update is worth an additional testing. I recognise that may be just a single line of code but the improvements in `technology-data` are quite significant, and it would be great to test all the implications also going beyond the "standard" testing geography

doc/release_notes.rst
```diff
@@ -23,6 +23,8 @@ This part of documentation collects descriptive release notes to capture the mai
 
 * Update workflow to geopandas >= 1.0 `PR #1276 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1276>`__
 
+* Introduce universal currency conversion
+
```
ekatef: It would be great to adjust the note formatting to match the format of the others for consistency

envs/environment.yaml
```diff
@@ -89,3 +89,4 @@ dependencies:
   - git+https://github.com/davide-f/google-drive-downloader@master  # google drive with fix for virus scan
   - chaospy  # lastest version only available on pip
   - fake_useragent
+  - currencyconverter
```
ekatef, line 81: Have there been any particular reasons to use `pip` distribution? As discussed, `pip` can easily lead to the dependencies conflicts and must be avoided

envs/linux-pinned.yaml
```diff
@@ -575,6 +575,7 @@ dependencies:
 - zstd=1.5.6
 - pip:
   - chaospy==4.3.17
+  - currencyconverter==0.17.30
```
ekatef: Note please that pinned environments are being automatically updated (e.g. #1321 is an example of an automated PR which fixes this issue). So, luckily there is no need for manual changes

envs/macos-pinned.yaml
```diff
@@ -503,6 +503,7 @@ dependencies:
 - zstd=1.5.6
 - pip:
   - chaospy==4.3.17
+  - currencyconverter==0.17.30
```
ekatef: Same comment as above: no need for manual changes

envs/windows-pinned.yaml
```diff
@@ -502,6 +502,7 @@ dependencies:
 - zstd=1.5.6
 - pip:
   - chaospy==4.3.17
+  - currencyconverter==0.17.30
```
ekatef: Same comment as above: no need for manual changes

ekatef (commented):
Hello @danielelerede-oet, thank you for the great contribution! 

Looks really nice, and will be definitely very helpful. Have added a couple of quick comments which mainly relate to the maintenance part. Could you please have a look?

config.default.yaml
danielelerede-oet: Hi @ekatef ! Happy to respond on this: `v0.11.0` was placed here because, in addition to `standard` data, it should also contain US data. It's not like that at the moment because some data were overwritten by the updates on the US ones, but we're already working on bringing back the situation to the previous state. This given, any new version adopted here will just fetch the exact same data as `v0.10.0`, and unless any specific input dataset is called (e.g., the US one, that will be object of another PR), it won't bring any change to the currently used data.

doc/release_notes.rst
danielelerede-oet: Sure, done! :)

config.default.yaml
ekatef: @danielelerede-oet thanks a lot for the explanation and the discussion today. Sounds perfectly reasonable. Just for completeness, adding also the point you mentioned that the version update has been also needed to use the fix implemented in [this PR](https://github.com/PyPSA/technology-data/pull/168) to `technology-data` included into `v.0.11.0`.

envs/environment.yaml
ekatef: As a follow-up of our discussion, a requirement to have `conda` installation is by no means ultimative and can be also managed on the maintenance side if there is no other way around 🙂

danielelerede-oet:
The changes in data/costs.csv are needed to align results of the tests with the previous version. I added the "further description" header to uniform with the standard structure of cost files from `technology-data` while `currency_year` is set to 2013 as it's only used for the conversion of USD2013 to EUR2013, thus it only acts on `hydrogen storage tank`, `battery inverter` and `battery storage`.

danielelerede-oet:
Hi @ekatef sorry is there anything still needed to adjust here? :)

doc/release_notes.rst
ekatef: Perfect, thank :)

doc/requirements.txt
```diff
@@ -40,3 +40,4 @@ pyDOE2
 graphviz
 
 fake_useragent
+currencyconverter
```
ekatef: Do we need the `currencyconverter `also here? The idea is to use this specification to generate an environment needed only for generating the documentation locally. [Agree though that it's not the most consistent approach]

ekatef:
> Hi @ekatef sorry is there anything still needed to adjust here? :)

Hey @danielelerede-oet and sorry for missing the updates.
To me, the implementation looks good have added only a very minor comment.

I think two major points we need to address:
1) Fix the pre-commit which is currently complaining on some file endings. Do you have any ideas what could go wrong?
2) Run CI test. To do so, there is a need to update pinned environment first as they are specifications used to build environments in a CI run. Currently CI is failing as the PR adds a new dependency which is not included into pinned environments. It can be fixed by a manual run of `Update pinned environment` in Actions:

image

I'm currently doing that for another PR, so happy to catch-up on that if you prefer 🙂 

3) It would be great to cross-check we have a mitigation strategy for `CurrencyConverter`. Just to make sure, it's maintained in [this repo](https://github.com/alexprengere/currencyconverter), right? 

If that is the case, we should be safe as it appears to be very active. If it'll be the case also in future, the major risk is inconsistency of `pip` and `conda` solvers: it has been seen that `pip` can break an environment generated by conda. Ideally, it would be good to contact maintainers of `CurrencyConverter` and support them in releasing also conda distribution. @davide-f has done that for two other dependencies but it looked like a highly non-trivial effort so not sure how much effort it would take in our case.

As a simpler solution, may it be possible to have some default currency option for a case when `CurrencyConverter` is not available for any reason?

danielelerede-oet:
Hi @ekatef ! thanks for the review!
Concerning point 1, I fixed the pre-commit!
Regarding point 2, I think I don't have the permissions to update the pinned environment, but I'm not sure about that.
For point 3, the repo is indeed very active (and it's the one you provided the link for), especially due to new data on daily currency conversion constantly fed to the dataset. I like your idea of having a default option as it was before, I can work on that!

danielelerede-oet:
Hi @ekatef I added a fallback default exchange rate in case of failure in the use of currencyconverter!

ekatef:
> Hi @ekatef I added a fallback default exchange rate in case of failure in the use of currencyconverter!

Hey @danielelerede-oet, perfect! We are moving to finalise 😄 


The only thing left is running CI tests. Generation of the environment [has failed](https://github.com/open-energy-transition/pypsa-earth/actions/runs/15190421275) for windows which has been expected. So, some temporary changes are needed if we want to have results quickly. 

If you wish, I can introduce the temporal fix into the PR branch by switching-off windows and re-do generation of the pinned environments. Otherwise, we can wait until the environment will be fixed in the upstream (@davide-f and @GbotemiB are working on that currently). Let me know your preferences!

danielelerede-oet:
Hi @ekatef thank you! Not sure if the environment issue has been completely fixed with #1458 . In that case maybe we can proceed with the CI tests :)

GbotemiB:
Hello @danielelerede-oet, just adding to this. It looks like CI is complaining of currencyconverter. 

`No module named 'currency_converter'`

ekatef:
@danielelerede-oet I confirm that the environment issues are now resolved in the upstream 😄 So, it's now possible to generate environment files for CI in a straightforward way.

@GbotemiB you are completely right! The currency-converter package has been added to the environment specification file, but CI is relays for environment definition on the lock files which must be updated to include the new dependency.

FYI [here](https://github.com/open-energy-transition/pypsa-earth/actions/runs/15327862416) are outputs of the actions workflow which contains the updated lock files but fails for some reason contribute them into a branch of this PR (it should open a supplementary PR for that). Not sure if there are some permissions which should be adjusted.

danielelerede-oet:
Hi @ekatef after the solution of environment issues the CI was failing again for all of the OS. 
The problem was that `currencyconverter` was not included in the lock files when generated, but in a new `conda-lock.yml` file as apparently the generation of the lock file is not recognizing the pip package `currencyconverter` (I guess this could happen in the future with other pip packages). I'm not really sure how the lock files including `googledrivedownloader` were generated, but if the same strategy can be used to also include `currencyconverter` I can definitely try. Otherwise we can simply keep `conda-lock.yml` and it should be sufficient to avoid problems

BTW now it still fails for Windows, for reasons apparently not related to `currencyconverter` itself. How could we proceed?

ekatef:
> Hi @ekatef after the solution of environment issues the CI was failing again for all of the OS. The problem was that `currencyconverter` was not included in the lock files when generated, but in a new `conda-lock.yml` file as apparently the generation of the lock file is not recognizing the pip package `currencyconverter` (I guess this could happen in the future with other pip packages). I'm not really sure how the lock files including `googledrivedownloader` were generated, but if the same strategy can be used to also include `currencyconverter` I can definitely try. Otherwise we can simply keep `conda-lock.yml` and it should be sufficient to avoid problems
> 
> BTW now it still fails for Windows, for reasons apparently not related to `currencyconverter` itself. How could we proceed?

Hello @danielelerede-oet, interesting! Regardless of the particular output it looks like you are gaining expertise in managing CI & environment which will be useful in any case. In Python, dependencies management has unfortunately a bit of magic so having experience there is precious.

Very useful insights on `pip` and `conda` interactions. They are based on different logic, and some side effects have been noticed previously. It appears though we are not fully aware of all the applications and good to have deeper understanding on that.

On the way to move forward, I can suggest to push `add_costs_usa` branch to the upstream and open a PR from it. (Normally, a preferable way is work from forks but using branches is acceptable in some cases, we just need to make sure that we'll clean-up those after merging) That should allow to overcome permission limitations and use a usual automated approach for generation of the conda-lock files. What do you think?

danielelerede-oet:
Hi @ekatef yes, yesterday I had a very immersive experience in understanding what was going on with the CI :laughing: a bit frustrating sometimes but definitely good to build some experience.
Anyway I don't have the permission to push directly upstream, if you could grant me that (even temporarily)  I'll proceed as you suggest! :smile:

danielelerede-oet:
@ekatef looks like the latest update to `workflow/tests.yml` solved the CI failure for Windows too (I really didn't expect that)

ekatef:
> Hi @ekatef yes, yesterday I had a very immersive experience in understanding what was going on with the CI 😆 a bit frustrating sometimes but definitely good to build some experience. Anyway I don't have the permission to push directly upstream, if you could grant me that (even temporarily) I'll proceed as you suggest! 😄

I can't change the permission on my own as all the while steering committee [Hazem, Leon, Davide and me] must be aligned on that. Personally, I think you definitely need some code-maintenance permissions and have requested the inputs from the colleagues on that. In the meantime, I can support with this maintenance bit. 

Have pushed the PR branch to the upstream, and [a run](https://github.com/pypsa-meets-earth/pypsa-earth/actions/runs/15508204152) of the locked environments has been successful and the content is merged to the branch. Could you please check if you can work from that and open a PR? If it won't agree to collaborate as-is, a workaround may be to fetch this branch to your personal fork.

ekatef:
> @ekatef looks like the latest update to `workflow/tests.yml` solved the CI failure for Windows too (I really didn't expect that)

Yeah, it appears that the checks are successful 😄 CI for windows failed due to some troubles with connection to the remote sources.

ekatef:
@danielelerede-oet I think the only point left is to remove the changes which have been needed to make the locked conda files to work. Sorry to back-and-forth with the environment and obviously feel free to contribute any changes which you have found useful when tackling the issue with the environment update.

danielelerede-oet:
Sure @ekatef ! The only needed addition at this point should be the `conda-lock.yml` file, useful to manage those pip dependencies that cannot be managed by the lock files related to each specific OS

ekatef:
> Sure @ekatef ! The only needed addition at this point should be the `conda-lock.yml` file, useful to manage those pip dependencies that cannot be managed by the lock files related to each specific OS

@danielelerede-oet looking into automatically generated `linux-64.lock.yaml` `pip` seems to worked properly installing `currencyconverter` (e.g. [this line](https://github.com/pypsa-meets-earth/pypsa-earth/blob/8d43ba85a46c3a32c574af0d22bc676ee35b0927/envs/linux-64.lock.yaml#L541)). 

Just to clarify: I have pushed `add_costs_usa` into the upstream and manually triggered generation of the locked environment by [this run](https://github.com/pypsa-meets-earth/pypsa-earth/actions/runs/15508204152) of github workflow and it appeared to have worked properly. So, I think `conda-lock.yml` may be not needed for this PR. Am I missing anything?

danielelerede-oet:
Hi @ekatef Thanks for the clarification — you're right, `currencyconverter` is present in the `linux-64.lock.yaml` generated by the CI, but not in my local version.

I’ll use the artifact from the CI to align locally and commit that, removing `conda-lock.yaml`. Likely just a small difference in how the environment was locked on my end.

danielelerede-oet:
@ekatef I generated the locked env files using the strategy mentioned in #1473 and  triggered the update of locked environments [here](https://github.com/danielelerede-oet/pypsa-earth/actions/runs/15519024425)  and it seems to finally work fine

ekatef (approved):
I love the changes! Great work 😄

ekatef:
> @ekatef I generated the locked env files using the strategy mentioned in #1473 and triggered the update of locked environments [here](https://github.com/danielelerede-oet/pypsa-earth/actions/runs/15519024425) and it seems to finally work fine

Perfect, congratulations! Great that you have found a way to make CI work without swapping the source branch 😄 

To me, the PR looks ready to be merged.

danielelerede-oet:
Thanks @ekatef , as soon as we merge I would proceed with checking #1448 too!

ekatef:
> Thanks @ekatef , as soon as we merge I would proceed with checking #1448 too!

Merged 😄 
Thank you for the great contribution! 🎉 🎉 🎉

PR #1344, opened 2025-02-07
ci: improve update pins job

lkstrp:
- Run biweekly (too many prs otherwise)
- Remove default channels

### DIFF
--- a/.github/workflows/update-pinned-env.yml
+++ b/.github/workflows/update-pinned-env.yml
@@ -4,16 +4,16 @@ on:
   push:
     paths:
     - envs/environment.yaml
-  # Run every Sunday at 5:00 UTC
   schedule:
-  - cron: "0 5 * * 0"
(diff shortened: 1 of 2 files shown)


davide-f (commented):
Thanks! agree, indeed I was closing useless PRs.
what about monthly? Note that we added the manual "workflow_dispatch" that may be useful also in -eur.

Additionally, what if we include also a run of the pre-commit?
It is a bit annoying that everytime this runs, the pre-commit fixes the file as they have a different formatting.

lkstrp:
You can also switch to monthly if you like. Or just run it manually.

I added an ignore to the pre-commit, so that the pinned files are not touched

davide-f (approved):

davide-f:
Many thanks! Merged
I've just restore the workflow_dispatch option to enable the execution of the procedure also in local branches that are not main

PR #1393, opened 2025-04-02
Fix AL emissions

yerbol-akhmetov:
Good day. This PR aims to fix the calculation part of the emissions from aluminum industry. Aluminum is generally emission intensive industry. I have noticed that `AL_prod_tom` is given in ktons. Finally, when we calculate emissions, the current code just multiplies by `emission_factors["non-ferrous metals"]`, please see: https://github.com/pypsa-meets-earth/pypsa-earth/blob/f5c362d08bb334ecb0b515d1e96e0b19c2c5d646/scripts/build_industry_demand.py#L229 
However, we need to also convert from ktons to tons, similar as we do for other industry emissions. This is a minor bug fix. Additionally, I have seen that function `country_to_nodal` uses global parameter `dist_keys` instead of given parameter `keys`. This was also fixed. What do you think, @davide-f, @hazemakhalek ?

image

## Checklist

### DIFF
--- a/scripts/build_industry_demand.py
+++ b/scripts/build_industry_demand.py
@@ -226,7 +226,8 @@ if __name__ == "__main__":
         # Reindex and fill missing values with 0.0
         AL_prod_tom = AL_prod_tom.reindex(countries_geo, fill_value=0.0)
 
-        AL_emissions = AL_prod_tom * emission_factors["non-ferrous metals"]
+        # Estimate emissions for aluminum production and converting from ktons to tons
+        AL_emissions = AL_prod_tom * emission_factors["non-ferrous metals"] * 1000
 
         Steel_emissions = (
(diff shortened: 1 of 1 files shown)


yerbol-akhmetov:
Hi, @davide-f, I have checked upstream pypsa-earth-sec and the PRs there. I think fix on aluminum emission is not yet done there.

hazemakhalek:
Hey Yerbol, good catch.

According to the JRC, AL production emissions are:

EU+EFTA countries: 1.6 tCO₂ per ton of aluminium produced​

Rest of the world: 1.79 tCO₂ per ton of aluminium produced

Ideally, we should turn this into a config parameter. I can start an issue for this

hazemakhalek (approved):
Looks good

PR #1399, opened 2025-04-04
Fix path to costs file

yerbol-akhmetov:
Good day. This PR aims to fix the path to `costs_{planning_horizon}.csv` in the `add_brownfield` rule of the Snakefile. It was causing the error while trying myopic runs.

## Checklist


ekatef (approved):
Hello @yerbol-akhmetov, and thank you for the fix! Agree with the implementation since it's perfectly consistent with the approach currently used to define the costs file corresponding to the `planning_horizon`.

As an additional comment, it appears there is a duplication in the configuration parameters which can potentially lead to troubles. Have created an issue #1407 to track that.

ekatef:
Just to be sure, has checked the reason why this error hasn't been captured by CI to find that `add_existing_baseyear` has precedence over `add_brownfiled`. Since the fix adjusts the path to the costs file according to `add_existing_baseyear` I think we can safely merge this PR.

Many thanks for the fix @yerbol-akhmetov!

PR #1401, opened 2025-04-05
Fix bus names for gas network used in H2 pipeline 

yerbol-akhmetov:
Good day. This PR aims to fix the bus names in `gas_networks/gas_network_elec_s{simpl}_{clusters}.csv` to prevent undefined bus names for `H2 pipeline` links. While simulating the sector model for US, I have noticed that `H2 pipeline` linked to non-existing buses:
image

Looking into gas network, it was found that bus names are not found. By doing debug I have found that, the issue was due to conversion of the bus name from three letter country code to two letter code. So before conversion from three letter to two letter:
image
After:
image
It should not be converted from three letter to two letter, because it is already in two letter format.

Previously, we have used this conversion, because gas_network used to have three letter bus names. We have included this change in PR #1198. 

## Checklist


ekatef:
@yerbol-akhmetov Thank you for submitting the fix and apologies for taking a while to review. I totally agree that we need a solution for this `not found` issue just wondering if we can enhance a bit the implementation to make the model implementation a bit cleaner. 

In particular, I wonder if it may be feasible to exclude `_AC` suffix from the names of H2 pipelines. As discussed with @energyLS today, `_AC` is there for some historical reasons which may be not quite relevant for now. However, it would be great to better understand the implications of removing the suffix from the bus names as there may be some side effects. What do you think about that?

energyLS:
Dear @yerbol-akhmetov thank you for setting up the PR! Out of couriosity, are you on the latest PyPSA-Earth version and does your solver find a solution? Because mine does not: #1404 and I somehow think it might be the pipelines which cause some issues/infeasibilities. 
@AlexanderMeisinger and @doneachh FYI.

ekatef:
@yerbol-akhmetov @energyLS Many thanks for the discussion yesterday! 

As mentioned, the need to fix this part for the second time may mean that some architectural adjustments may be needed. That may be the next step after this PR is completed. 

For now I think we must we must merge the PR as it's a much needed bug fix. @davide-f could you please have a look since you have been working on #1198 together with @yerbol-akhmetov?

yerbol-akhmetov:
> @yerbol-akhmetov @energyLS Many thanks for the discussion yesterday! 
> 
> As mentioned, the need to fix this part for the second time may mean that some architectural adjustments may be needed. That may be the next step after this PR is completed. 
> 
> For now I think we must we must merge the PR as it's a much needed bug fix. @davide-f could you please have a look since you have been working on #1198 together with @yerbol-akhmetov?
> 

Hi, @ekatef, I will get back when the simulation with upstream will be finished.

yerbol-akhmetov:
Hi, @ekatef, @energyLS. I have tried to simulate the sector network for US, however faced the issues with `prepare_sector_network`. In particular, with upstream commit, the `s_threshold_fetch_isolated` does not work well. Also I think there might be a problem with H2 export as noted in the issue #1404.

yerbol-akhmetov:
Hi, @ekatef, @energyLS. Sorry for the previous message and bothering. I have successfully solved the sector model for US with the upstream. Previously, I forgot to update the environment to the latest version. 
image
It was feasible even without H2 pipeline fixes.
image
With the H2 pipeline fixes, it was also optimal:
image

ekatef:
> Hi, @ekatef, @energyLS. Sorry for the previous message and bothering. I have successfully solved the sector model for US with the upstream. Previously, I forgot to update the environment to the latest version. 

Hello @yerbol-akhmetov, no worries and many thanks for the update. Great to hear that the model is feasible. Just out of curiosity, do the H2 pipeline fixes change a value of the objective function?

yerbol-akhmetov:
> > Hi, @ekatef, @energyLS. Sorry for the previous message and bothering. I have successfully solved the sector model for US with the upstream. Previously, I forgot to update the environment to the latest version.
> 
> Hello @yerbol-akhmetov, no worries and many thanks for the update. Great to hear that the model is feasible. Just out of curiosity, do the H2 pipeline fixes change a value of the objective function?

Hi, @ekatef. The objective has changed very slightly. Before fix: 1.58259e12, after fix: 1.58266e12.

davide-f (commented):
This PR looks good to me, given the tests by Yerbol too :) thanks for having solved this!
I'd be in favor of merging

ekatef:
Hello! 🙂 
@yerbol-akhmetov Perfect! Many thanks for checking the objective and running the tests. Looks like a great fix!
@davide-f many thanks for checking the PR and totally agree that it's ready to merge 😄

PR #1403, opened 2025-04-09
Fix load shedding

energyLS:
# Closes # (if applicable).

If load shedding is turned on, the generators used for load shedding are connected to *all* busses. However, they should either

1.  only be attached to the AC buses or
2. to all busses, but if that is the case the marginal price needs adjustments.

This PR proposes the solution to 1.), however if load shedding should be applied to all busses, a more differentated solution is required. Without the fix proposed here, e.g. hydrogen export is not working properly since the generators for load shedding supply the carrier hydrogen, this is not intended.

## Checklist


scripts/solve_network.py
```diff
@@ -111,11 +111,12 @@ def prepare_network(n, solve_opts, config):
 
     if solve_opts.get("load_shedding"):
         n.add("Carrier", "Load")
+        ac_buses = n.buses[n.buses.carrier == "AC"].index
```
ekatef: Completely agree that it would be great to have load shedding switched-on in a more flexible way. However, it would be nice to keep an option when load shedding is used for all carriers. Infeasibility seems to be quite a frequent problem, while load shedding is very effective to mitigate a significant part of these troubles.

Can we probably use `load_shedding` configuration parameter to determine to which carriers load shedding should it be applied? The logic can be as follows:
- `load_shedding: true` -> apply to all carriers
- `load_shedding: ["AC"]` -> apply to `AC` only (as in the current )
- `load_shedding: ["AC", "H2"]` -> apply to `AC` and `H2`
- `load_shedding: ["something weird"]` -> issue a `not-implemented` warning and apply to all carriers 
- `load_shedding: false` -> do not use load shedding at all

What do you think?

ekatef (commented):
Thank you @energyLS! Actually, we are struggling with a similar problem for another project. So, many thanks for the fix!

My feeling is that we must keep also an option to apply load shedding to all the carriers as it can be crucial for debug. Would you mind to add a modification for that?

Have also added a suggestion on a possible implementation, feel free to suggest alternatives! 🙂

energyLS:
> Thank you @energyLS! Actually, we are struggling with a similar problem for another project. So, many thanks for the fix!
> 
> My feeling is that we must keep also an option to apply load shedding to all the carriers as it can be crucial for debug. Would you mind to add a modification for that?
> 
> Have also added a suggestion on a possible implementation, feel free to suggest alternatives! 🙂

Dear @ekatef thanks for the quick feedback. I fully support your ideas and I am happy to implement them. We need to be careful though: the marginal costs currently set for AC are 100 €/MWh, a price so low that it incentives the load shedding generator to supply a certain carrier even if there is no constraint. E.g. Fischer-Tropsch fuel prices are higher, but would not be generated the ordinary way, since this would involve marginal prices higher than that. How about setting a price of 1000 €/MWh across all carriers?

Also, could you clarify what this option exactly does?
`load_shedding: ["something weird"] -> issue a not-implemented warning and apply to all carriers`

ekatef:
> > Thank you @energyLS! Actually, we are struggling with a similar problem for another project. So, many thanks for the fix!
> > My feeling is that we must keep also an option to apply load shedding to all the carriers as it can be crucial for debug. Would you mind to add a modification for that?
> > Have also added a suggestion on a possible implementation, feel free to suggest alternatives! 🙂
> 
> Dear @ekatef thanks for the quick feedback. I fully support your ideas and I am happy to implement them. We need to be careful though: the marginal costs currently set for AC are 100 €/MWh, a price so low that it incentives the load shedding generator to supply a certain carrier even if there is no constraint. E.g. Fischer-Tropsch fuel prices are higher, but would not be generated the ordinary way, since this would involve marginal prices higher than that. How about setting a price of 1000 €/MWh across all carriers?
> 
> Also, could you clarify what this option exactly does? `load_shedding: ["something weird"] -> issue a not-implemented warning and apply to all carriers`

Ahh, a great point regarding the shedding costs! We actually have inherited the value from PyPSA-Eur. That means a heavy overestimation of the costs for many regions of the world when a real load shedding represents real outages. Can we probably add shedding costs as a configuration parameter to avoid for more flexible customisation?

For `"something weird"` case, I just mean that it could be great to address also a situation when a wrong carrier name is defined in the configuration file. My first idea was to address this case by applying load shedding to all the carriers but feel free to suggest other options 🙂

davide-f:
Hello @energyLS and great discussion :)

About this PR what is the plan?

As comments: supporting load curtailment for all carriers should ensure feasibility of the model that can be very useful to debug it and understand where the problem is.
On the other hand, having 100€/MWh sticked in the code leads to the issues that you have also mentioned.

Having the value from the configuration file would be very much needed for sure as katia mentioned.
The exclusion by carrier may be inline with the spirit of this PR, and can be left for further developments if you are scarse in resources.
We could also have a load curtailment value by carrier, but that is another level. Like:
```
load_shedding:
   AC: 1000
   hydrogen: 3000
   default: 1000
```

energyLS:
Dear @ekatef and @davide-f thank you for your valuable suggestions. I have been looking deeper into this, and actually I think there is no/not much action necessary.

I was falsely assuming that the load shedding price is 100€/**MWh**, but after having a closer look it is actually 100€/**kWh** = 100.000€/**MWh**, a price certainly sufficient and not competitive with any alternative generation (for any carrier). My motivation for this PR was based on this false assumption, but after doublechecking now it makes sense as it is. In my case, I could not see any electrolyzer operation because there was/is an issue/infeasibility, which was correctly detected by the load shedding.

In conclusion, the load shedding works fine. As you have suggested it could be refined with carrier-specific prices, but actually I do not see any need for action and want to avoid overcomplication. What do you think @ekatef and @davide-f?

A straightforward way to check load shedding in a solved network:
```
n.statistics.energy_balance(comps=["Generator"]).loc[:,"load",:].div(1e3)
```

davide-f:
> Dear @ekatef and @davide-f thank you for your valuable suggestions. I have been looking deeper into this, and actually I think there is no/not much action necessary.
> 
> I was falsely assuming that the load shedding price is 100€/**MWh**, but after having a closer look it is actually 100€/**kWh** = 100.000€/**MWh**, a price certainly sufficient and not competitive with any alternative generation (for any carrier). My motivation for this PR was based on this false assumption, but after doublechecking now it makes sense as it is. In my case, I could not see any electrolyzer operation because there was/is an issue/infeasibility, which was correctly detected by the load shedding.
> 
> In conclusion, the load shedding works fine. As you have suggested it could be refined with carrier-specific prices, but actually I do not see any need for action and want to avoid overcomplication. What do you think @ekatef and @davide-f?
> 
> A straightforward way to check load shedding in a solved network:
> 
> ```
> n.statistics.energy_balance(comps=["Generator"]).loc[:,"load",:].div(1e3)
> ```

Agree that for now that overcomplication by carrier does not seems necessary.
Having the single number accessible and customizable from the config could be a good idea though, what do you think?

energyLS:
> Agree that for now that overcomplication by carrier does not seems necessary. Having the single number accessible and customizable from the config could be a good idea though, what do you think?

Dear @davide-f I fully support the idea and have implemented it. Just for understanding: The parameter

```
sign=1e-3,  # Adjust sign to measure p and p_nom in kW instead of MW
```

is used to avoid large numbers and increase solving stability?

Ready to merge from my side, unless there are questions remaining.

scripts/solve_network.py
```diff
@@ -118,10 +118,7 @@ def prepare_network(n, solve_opts, config):
             bus=n.buses.index,
             carrier="load",
             sign=1e-3,  # Adjust sign to measure p and p_nom in kW instead of MW
```
davide-f: To answer your question, this is used to specify the marginal cost below in terms of €/kWh as it is then divided 1e-3.

To be honest, it is best to avoid the use of this option. If we get rid of this line, below we can directly specify the value in €/MWh explicitly. The sign option may also be dropped in the future as it is not so used.
Would you mind doing that?

config.default.yaml
```diff
@@ -704,7 +704,7 @@ sector:
 solving:
   options:
     formulation: kirchhoff
-    load_shedding: true
+    load_shedding: 100000 # Set to "false" or willingness to pay in €/MWh, e.g. 100000 €/MWh (intersect between macroeconomic and surveybased willingness to pay http://journal.frontiersin.org/article/10.3389/fenrg.2015.00055/full)
```
davide-f: In the config file, personally I believe having the parameter in €/kWh that is more readable [easy to miss zeros with €/MWh], what do you think?

davide-f:
Great @energyLS :D amazing, much needed.

I have the reply to your question above, if you had time to account for it great otherwise I can too.

scripts/solve_network.py
energyLS: Done :) I also adjusted p_nom accordingly.

config.default.yaml
energyLS: I have adjusted it to €/kWh. I agree, it is much more readable... however, this was the very initial reason for my confusion, because I have mixed up €/kWh with €/MWh since it is a much more realistic cost measure in terms of values we usually deal with. But this one is on me, I have reverted it back to €/kWh

energyLS:
> Great @energyLS :D amazing, much needed.
> 
> I have the reply to your question above, if you had time to account for it great otherwise I can too.

@davide-f thanks! I took care and have implemented your suggestions. Let me know what you think.

davide-f (approved):

davide-f:
Great :D Merging :)

PR #1427, opened 2025-05-06
Fix params for prepare_sector_network script

Eddy-JV:
# Closes # (if applicable).

This PR is straightforward and aims at putting all params for prepare_sector_network.py in snakefile.

## Checklist


Snakefile
```diff
@@ -1077,6 +1077,27 @@ rule prepare_sector_network:
...
+            "biomass_transport_default_cost"
+        ],
+        cc=config["sector"]["cc"],
+        gadm_layer_id=config["build_shape_options"]["gadm_layer_id"],
+        alternative_clustering=config["cluster_options"]["alternative_clustering"],
+        international_bunkers=config["sector"]["international_bunkers"],
+        h2_policy=config["policy_config"]["hydrogen"],
+        sector_options=config["sector"],
+        foresight=config["foresight"],
+        water_costs=config["custom_data"]["water_costs"],
```
davide-f: We have `sector_options=config["sector"]` that importw the whole sector block.
To have the snakemake lighted maybe we can load the other options inside the script rather than having them explicit here.
What do you think?

scripts/prepare_sector_network.py
```diff
@@ -65,7 +65,7 @@ def add_carrier_buses(n, carrier, nodes=None):
     n.madd("Bus", nodes, location=location, carrier=carrier)
 
     # initial fossil reserves
-    e_initial = (snakemake.config["fossil_reserves"]).get(carrier, 0) * 1e6
+    e_initial = (snakemake.params.fossil_reserves).get(carrier, 0) * 1e6
```
davide-f, line 68: I'm just commenting but no action needed I believe.

Personally, I don't like to have "snakemake" used in the scripts as global variables.
It is quite common and widely used now so, it is not worth addressing this, but I'm commenting to share this and we can align.

The reason of the above is that the use of global variables, instead of having the proper argument leads to have a hidden dependency of the function that can lead to side effects.
For example we have a function like `add_hydrogen()` that internally has the use of snakemake block.
This means that anywhere add_hydrogen is used, the snakemake object needs to be prepared beforehand accordingly.
For the current intended behavior, that is ok, but can be obscure.
Having the dependency explicit, such as `add_hydrogen(options)` makes the code robust and more explicit.

Said this; this PR is definitely not the place to address that, even if you agree on the above.

davide-f (commented):
Great @Eddy-JV :D
My comments have been added :)

Snakefile
Eddy-JV: I totally agree to that. I even thought of doing it already. Anyways, **adjusted and done** now.

scripts/prepare_sector_network.py
Eddy-JV: I agree to that too. Maybe We can do a seperate PR for this throughout the whole model code.

I was also thinking of adding to all functions the type of each dependecy similar to this:

```
def add_heat(
    n: pypsa.Network,
    costs: pd.DataFrame,
    cop_profiles_file: str,
    direct_heat_source_utilisation_profile_file: str,
    hourly_heat_demand_total_file: str,
    ptes_e_max_pu_file: str,
    district_heat_share_file: str,
    solar_thermal_total_file: str,
    retro_cost_file: str,
    floor_area_file: str,
    heat_source_profile_files: dict[str, str],
    params: dict,
    pop_weighted_energy_totals: pd.DataFrame,
    heating_efficiencies: pd.DataFrame,
    pop_layout: pd.DataFrame,
    spatial: object,
    options: dict,
    investment_year: int,
):
```



But I wanted to discuss with you and the group if it makes sense.

Eddy-JV:
@davide-f 
The test workflow is failing due to a change on geofrabrik webiste that happened today for Botswana. The files below are both not present anymore:

image

Eddy-JV:
Geofabrik added the file again. Test runs except Windows are successful. Merging now.

Issue #1431, opened 2025-05-14
KeyError in prepare_sector_network when modeling Ethiopia with alternative clustering


doneachh:
## Checklist

- [x] I am using the current [`main`](https://github.com/pypsa-meets-earth/pypsa-earth/tree/main) branch or the latest [release](https://github.com/pypsa-meets-earth/pypsa-earth/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-earth` environment](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

Clustering seems not to work when modeling sector-coupled Ethiopia with alternative clustering. The Error seems to already lay in the clustering. When modeling it without alternative clustering / voronoi, we dont run into this error.

## Error Message

*If applicable, paste any terminal output to help illustrating your problem.*
*In some cases it may also be useful to share your list of installed packages: `conda list`.*

```bash
Traceback (most recent call last):
  File "/mnt/d/Git/AP2/Ethiopia/pypsa-earth/.snakemake/scripts/tmpv9kfsxvh.prepare_sector_network.py", line 2947, in <module>
    n.loads.loc[nodes, "carrier"] = "AC"
  File "/home/aca39878/anaconda3/envs/pypsa-earth/lib/python3.10/site-packages/pandas/core/indexing.py", line 907, in __setitem__
    indexer = self._get_setitem_indexer(key)
  File "/home/aca39878/anaconda3/envs/pypsa-earth/lib/python3.10/site-packages/pandas/core/indexing.py", line 774, in _get_setitem_indexer
    return self._convert_tuple(key)
  File "/home/aca39878/anaconda3/envs/pypsa-earth/lib/python3.10/site-packages/pandas/core/indexing.py", line 989, in _convert_tuple
    keyidx = [self._convert_to_indexer(k, axis=i) for i, k in enumerate(key)]
  File "/home/aca39878/anaconda3/envs/pypsa-earth/lib/python3.10/site-packages/pandas/core/indexing.py", line 989, in <listcomp>
    keyidx = [self._convert_to_indexer(k, axis=i) for i, k in enumerate(key)]
  File "/home/aca39878/anaconda3/envs/pypsa-earth/lib/python3.10/site-packages/pandas/core/indexing.py", line 1522, in _convert_to_indexer
    return self._get_listlike_indexer(key, axis)[1]
[...]

PR #1432, opened 2025-05-14
fix: Error when no load in cluster

doneachh:
# Closes # (if applicable).
#1431 
bugfix for Issue when there is no load in a cluster

## Checklist

- [X] I tested my contribution locally and it seems to work fine.
- [X] Code and workflow changes are sufficiently documented.
- [X] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`.
- [X] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [X] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [X] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.


doneachh:
Hi @ekatef!
I recently came over an Issue, where there are no loads for some buses and we run into an error. I'm not sure, if we should tackle this somewhere else/earlier, but this was my "quick" solution for this issue. Feel free to review and suggest possible solutions. I don't know, what is the best way of fixing this, but this seems to work

davide-f (commented):
Hello and great catch!

What about we fill transport dataframe to match nodes? Filling with zeros?

doneachh:
> Hello and great catch!
> 
> What about we fill transport dataframe to match nodes? Filling with zeros?

I'm not sure how easy the whole thing is afterwards. The complete load for this cluster does not exist, so you would have to fill a complete data series with names and co?
If we are filling it, I think it might make sense to tackle the whole thing directly during clustering, but where?
Skipping, as now, also seems to work - at least I have already solved several networks with it.
What do you think?

scripts/prepare_sector_network.py
```diff
@@ -2945,7 +2945,12 @@ def remove_carrier_related_components(n, carriers_to_drop):
     n.buses.location = n.buses.index
 
     # Set carrier of AC loads
-    n.loads.loc[nodes, "carrier"] = "AC"
+    existing_nodes = [node for node in nodes if node in n.loads.index]
+    if len(existing_nodes) < len(nodes):
+        print(
+            "fWarning: For {len(nodes) - len(existing_nodes)} of {len(nodes)} nodes there were no load nodes found in network and were skipped."
+        )
+    n.loads.loc[existing_nodes, "carrier"] = "AC"
```
davide-f, line 2953: I see why this line can throw error. I'm wondering why we need this though.
This line that is the source of the problem just specifies the carrer of the load; maybe we can fix that by adding the carrier in add_electricity

Probably, a simple fix is to add the carrier name accounting for the bus type (AC/DC) here?
https://github.com/pypsa-meets-earth/pypsa-earth/blob/43cd627adcaa752e81790326517354fff1129138/scripts/add_electricity.py#L269

scripts/prepare_sector_network.py
```diff
@@ -1882,7 +1882,7 @@ def add_land_transport(n, costs):
                 carrier="land transport fuel cell",
                 p_set=fuel_cell_share
                 / options["transport_fuel_cell_efficiency"]
-                * transport[nodes],
+                * transport[[node for node in nodes if node in transport.index]],
```
davide-f: It would be highly beneficial to adopt good data standards as otherwise the code can become quite obscure and these ad-hoc solutions may explode easily.

I'm commenting this as I see likely that the same root cause will percolate and appear in many different ways in different countries and options, leading to more efforts when not tackled directly.

Agree that understanding why transport does not have the full expected buses sounds a good idea and filling them even during the clustering phase sounds quite interesting. It may also be due to the population layout that may not have full entries.
Alternatively, having a utility function that fills the index

What do you think?

davide-f (commented):
Many thanks with the availability and commitment!
I'm adding more on-point comments as we are dealing with two linked topics:
- the n.loads = AC
- the missing nodes in transport

definitely aligning with your availability in contributing :)

I'm commenting as I see that we have some root causes that often percolate and lead to different issues that take effort multiple times. I'd not be surprised if you may encounter them again in different forms with different carriers and options, hence leading to more time into reviewing and debugging that can be spared

davide-f (commented):
@doneachh I'm in favor of the fix for the AC carrier; I've attempted the other solution but the workflow is not ready for that and other issues appear [unfortunately]

On the transport nodes, could we place the fix [here instead](https://github.com/pypsa-meets-earth/pypsa-earth/blob/bf81d000ffb86a1c1327b6c03537ce66b6de5dc8/scripts/prepare_sector_network.py#L2991)?
The proposal would be to fill the transport dataframe with zeros to match the nodes list.

In the future we may want consistency; having missing columns is likely an issue that may need deeper investigation. Actually it could be great to understand why and add an issue.
What do you think?

scripts/prepare_sector_network.py
davide-f: Update: this is tested in #1461 but not so trivial as originally thought; so I'm resolving this comment

doneachh:
@davide-f Agree! I’ve implemented it that way and hope it works as you had in mind. Let me know if anything needs adjusting!

davide-f (commented):
Many thanks @doneachh.

Only a minor thing, would you mind adding a release note?
If you have no capacity I can do.

On the technical side, good for me and thanks for the patience :)

doneachh:
@davide-f added some release notes for the last fixes! Feel free to merge :)

davide-f:
Great! Many thanks :) merging

PR #1436, opened 2025-05-15
Fix_scripts_for_custom_data

Eddy-JV:
# Closes # (if applicable).

This PR fixes some issues when using custom demand and custom shapes data. 
Addtionally, removing none necessary config items that are no longer used.

## Checklist


Snakefile
```diff
@@ -1204,15 +1204,15 @@ rule override_respot:
         countries=config["countries"],
     input:
         **{
-            f"custom_res_pot_{tech}_{planning_horizons}_{discountrate}": "resources/"
+            f"custom_res_pot_{tech}_{planning_horizons}_{discountrate}": "data/custom/"
```
ekatef, line 1207: Totally support moving the custom data from `resources` to `data`. Two questions:
1) Could it make sense to add a run-specific folder in `data` to keep the inputs in a more structured way? Even now it can be misleading to have custom data files directly in `data` folder and I fear that things can easily go out of hand once we'll have more custom data there. Can we probably create a folder with {run}  name in `data` folder, copy there templates and look for custom data there if applicable?
2) Not sure it's a good idea to use wildcards in custom files which are should be added manually. Wildcards are perfect if they are used in an automated workflow but are not particularly user-friendly. It it possible to be more explicit here defining a list of files which are expected by the workflow?

Snakefile
```diff
@@ -1318,6 +1318,8 @@ rule build_cop_profiles:
 
 
 rule prepare_heat_data:
+    params:
+        sector_options=config["sector"],
```
ekatef, line 1322: It would be probably safer to define a more narrow sub-set of `config["sector"]` parameters to avoid triggering unneeded re-runs of a workflow due to a change of some parameters which are actually not relevant for `prepare_heat_data`. What do you think?

Snakefile
```diff
@@ -1386,6 +1388,7 @@ rule prepare_energy_totals:
         countries=config["countries"],
         base_year=config["demand_data"]["base_year"],
         sector_options=config["sector"],
+        energy_demand=config["custom_data"]["sectoral_energy_demand"],
```
ekatef, line 1391: If I get it properly, `energy_demand` stands in fact for `custom_energy_demand`. If that is the case, the parameter name must be revised as it can be misleading right now

config.default.yaml
```diff
@@ -476,15 +477,16 @@ export:
...
+  # heat_demand: false
   industry_demand: false
   industry_database: false
-  transport_demand: false
+  # transport_demand: false
   water_costs: false
   h2_underground: false
   add_existing: false
-  custom_sectors: false
+  # custom_sectors: false
```
ekatef, line 489: From the previous experience, commented-out lines tend to stay in the code forever. The reason of commenting-out gets forgotten quite quickly, and you can't be sure for which particular purpose the commented-out stuff is needed and if it can be removed safely.

Would you mind to remove the unneeded parameters instead commenting-out them?

data/custom/TEMPLATE_industry_demand_AB_2030.csv
```diff
@@ -6,4 +6,4 @@ MA,coal,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,6875175.2,0.0,0.0,0.0,0.
 MA,heat,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0
 MA,biomass,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,171879.38,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,276211.484037
 MA,hydrogen,0.0,0.0,1158286.08,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0,276211.484037
-MA,process emission,0.0,0.0,50065.41247790604,40962.59763502796,0.0,0.0,0.0,0.0,67875.68113871076,0.0,1385.228977747275,8.72479576360429e-05,9323308.507746331,0.03203319690797162,0.0,0.0,0.0,0.0,1658616.7007389446,0.0,0.18963180469598598,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0
+MA,process emissions,0.0,0.0,50065.41247790604,40962.59763502796,0.0,0.0,0.0,0.0,67875.68113871076,0.0,1385.228977747275,8.72479576360429e-05,9323308.507746331,0.03203319690797162,0.0,0.0,0.0,0.0,1658616.7007389446,0.0,0.18963180469598598,0.0,0.0,0.0,0.0,0.0,0.0,0.0,0.0
```
ekatef, line 9: A great catch! Shouldn't be also the column name fixed in the parts of the code where it's used? It appears to be the case for `create_dummy_data` in `_helpers`, but there may be also other fragments. Could you please double-check that?

data/custom/custom_renewables/TEMPLATE_csp_2050_0.091_installable.csv
```diff
@@ -0,0 +1,36 @@
+Generator,p_nom_max,annualcostEuroPMW,fixedomEuroPKW,installedcapacity,lifetime
+TR.11_1_AC,1,280787.0,40.0,0.0,30.0
+TR.12_1_AC,1,280787.0,40.0,0.0,30.0
+TR.13_1_AC,1,280787.0,40.0,0.0,30.0
```
ekatef, line 4: I wonder if we can be more generic in the templates? Would be good to keep all of them as light as possible similar as it's done in `custom_powerplants.csv` or in `custom_res_installable.csv`. Do you think it's possible to shorten the templates files keeping only a line or a few and as few values as possible?

scripts/override_respot.py
```diff
@@ -102,15 +102,16 @@ def override_values(tech, year, dr):
...
+        # if snakemake.params.custom_data["elec_demand"]:
+        #     for country in countries:
+        #         n.loads_t.p_set.filter(like=country)[buses] = (
+        #             (
+        #                 n.loads_t.p_set.filter(like=country)[buses]
+        #                 / n.loads_t.p_set.filter(like=country)[buses].sum().sum()
+        #             )
+        #             * energy_totals.loc[country, "electricity residential"]
+        #             * 1e6
+        #         )
```
ekatef, line 115: As mentioned, feel free to remove this fragment completely if you think it's not needed.

As far as I can see, this fragment is commented-out in `prepare_sector_network`. Is it intentional or is there a need to uncomment in `prepare_sector_network`?

scripts/prepare_energy_totals.py
```diff
@@ -61,237 +61,259 @@ def fill_country_data(df, country, default_key="DEFAULT", label="", logger=_logg
...
-            fuel_shares["biomass to elec heat share"]
-            * fuel_shares["biomass residential heat share"]
-            * (fuel_shares["space to water heat share"])
-            * base_energy_totals["residential biomass"]
-            * efficiency_heat_biomass_to_elec
+    if snakemake.params.energy_demand:
+        _logger.info(
+            "Fetching custom sectoral energy demand data.. expecting file at 'data/custom/energy_totals_{0}_{1}.csv'".format(
+                snakemake.wildcards["demand"], snakemake.wildcards["planning_horizons"]
+            )
```
ekatef, line 68: What is the meaning of `{0}` and `{1}` and what is a purpose of having them?

scripts/prepare_sector_network.py
```diff
@@ -3085,6 +3085,12 @@ def remove_carrier_related_components(n, carriers_to_drop):
     if options.get("electricity_distribution_grid", False):
         add_electricity_distribution_grid(n, costs)
 
+    if options["dac"]:
+        add_dac(n, costs)
+
+    if snakemake.params.water_costs:
+        add_custom_water_cost(n)
+
```
ekatef, line 3093: Has there been a particular reason for this change?

ekatef (commented):
Hello @Eddy-JV and thank you for taking care about the custom data.
Totally support the overall idea and have added some comments on the particular implementation. Could you please go through them? Happy to discuss in a chat if it can make things easier.

Snakefile
ljansen-iee: Hi all, I support the idea of moving all optional custom data all to the custom folder and adding a ‘run’ folder to be able to specify custom data for each run.

ljansen-iee: Since the custom renewable data is also defined per discountrate wildcard, I have a follow-up question: As I understand it, the discountrate (not discount rate) only has an influence on the custom_res_ins_{tech}_{year}_{dr}. Is that correct? If so, we should consider generalising this and let the user define scenarios for the custom renewables by run name. What do you think?

Snakefile
ljansen-iee: My take on the naming convention: I suggest using the same names in the config under custom_data as the variable names in the script or the script names are. This works in the case that the custom option only refers to one data set. In this case, this means that it should be custom_energy_totals = config[‘custom_data’][‘energy_totals’].

data/custom/TEMPLATE_industry_demand_AB_2030.csv
ljansen-iee: Hi, shouldn't the header column be this?: ,carrier,chemical and petrochemical,construction,food and tobacco,iron and steel,machinery,mining and quarrying,non-ferrous metals,non-metallic minerals,other,paper pulp and print,textile and leather,transport equipment,wood and wood products

PR #1227, opened 2024-12-06
Add H2 production technologies

yerbol-akhmetov:
## Changes proposed in this Pull Request
Good day. This PR aims to integrate new H2 production technologies into the model. Technologies that are being add are: alkaline electrolyzer, PEM electrolyzer, SOEC electrolyzer, natural gas steam reforming (with/without CC), coal gasification (with/without CC), solid biomass steam reforming, solid biomass gasification (with/without CC), and heavy oil partial oxidation technologies. 

In addition, currently existing H2 production technologies, such as H2 Electrolysis, SMR, and SMR CC, are re-organized and placed to `add_hydrogen` function of `prepare_sector_network.py`. Moreover, all H2 production technologies are grouped and listed in config file in a single place:
```bash
hydrogen:
    production:
        h2_electrolyzer: true
        SMR: true
        ....
```

## Checklist

- [x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [x] I tested my contribution locally and it seems to work fine.
- [ ] Code and workflow changes are sufficiently documented.
- [ ] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`.
- [ ] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [ ] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [ ] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes, including reference to the requested PR.

PR #1448, opened 2025-05-20
Add US-specific cost configurations

danielelerede-oet:
This PR, authored by @finozzifa and myself, builds on #1319 and addresses the possibility to fetch US-specific data from https://github.com/PyPSA/technology-data/tree/master/outputs/US instead of https://github.com/PyPSA/technology-data/tree/master/outputs in case 2 conditions are satisfied (to avoid any possible mistake in fetching the "wrong" data):
1) US is in the only country in the country list
2) `technology_data_US` option is set to `true` in config.default.yaml (it is `false` by default)
It also adds a dedicated script for `retrieve_cost_data` to deal with the new options in the config file.

I deem this as the safest approach as also the US-specific cost files in technology-data do not delete any technology that is present in the "generic" folder, but simply overwrite some of them with US-specific cost values and add new technologies to be hopefully integrated in the model (not just for the US) e.g., PEM, alkaline and SOEC electrolyzers instead of a generic `electrolysis` process.

cc @yerbol-akhmetov @euronion @hazemakhalek @ekatef @davide-f 

## Checklist


scripts/add_electricity.py
```diff
@@ -138,17 +142,34 @@ def load_costs(tech_costs, config, elec_config, Nyears=1):
...
+
+        # both financial case and scenario are NULL
+        query_string_part_two = "financial_case.isnull() | scenario.isnull()"
+
+        query_string = " | ".join([query_string_part_one, query_string_part_two])
+
+        costs = costs.query(
+            query_string,
+            local_dict={"f": wished_financial_case, "s": wished_cost_scenario},
+        )
```
yerbol-akhmetov: Hi, @danielelerede-oet, @finozzifa. Thanks for the great PR.
Why not to use here more simpler/concise code like:
```
    costs = costs[costs.scenario.isin([wished_cost_scenario , np.nan])]
    costs = costs[costs.financial_case.isin([wished_financial_case , np.nan])]
```
or 
```
    costs = costs.query("scenario.str.casefold() in @wished_cost_scenario or scenario.isnull()")
    costs = costs.query("financial_case.str.casefold() in @wished_financial_case or financial_case.isnull()")
```

Snakefile
```diff
@@ -401,21 +402,17 @@ if config["enable"].get("retrieve_cost_data", True):
...
-            HTTP.remote(
-                f"raw.githubusercontent.com/PyPSA/technology-data/{config['costs']['version']}/outputs/"
-                + "costs_{year}.csv",
-                keep_local=True,
-            ),
+            countries=config["countries"],
+            costs=config["costs"],
         output:
-            "resources/" + RDIR + "costs_{year}.csv",
+            output_path=directory(pathlib.Path("resources", RDIR)),
```
yerbol-akhmetov: Because the output is not explicitly defined as `costs_{year}.csv`, it cannot download it. Snakemake gives error:
image

scripts/retrieve_cost_data.py
```diff
@@ -0,0 +1,48 @@
...
+        url_to_use = "/".join((base_url_to_use, "US", cost_file_name))
+    else:
+        # take technology-data/outputs/costs_yyyy.csv
+        url_to_use = "/".join((base_url_to_use, cost_file_name))
+        output_path = pathlib.Path(base_output_path, cost_file_name)
+
+    response = requests.get(url_to_use)
+    with open(output_path, mode="wb") as output_cost_file:
+        output_cost_file.write(response.content)
+    output_cost_file.close()
```
yerbol-akhmetov: Not sure if we need explicitly define `retrieve_cost_data.py`, because the only thing it makes downloading needed cost file for US if only US is selected and `technology_data_US: true`. I think we should be able to give the wildcard for url as this:
```
    if config["countries"] == ["US"] and config["costs"]["technology_data_US"]:
        cost_directory = "US"
    else:
        cost_directory = ""


    rule retrieve_cost_data:
        params:
            version=config["costs"]["version"],
        input:
            HTTP.remote(
                f"raw.githubusercontent.com/PyPSA/technology-data/{config['costs']['version']}/outputs/{cost_directory}/"
                + "costs_{year}.csv",
                keep_local=True,
            ),
        output:
            "resources/" + RDIR + "costs_{year}.csv",
        log:
            "logs/" + RDIR + "retrieve_cost_data_{year}.log",
        resources:
            mem_mb=5000,
        run:
            move(input[0], output[0])
```
This just worked fine. Of course, we need to roll back and use `version: v0.12.0`. Thus, wildcard will be added to input of the rule.

config.default.yaml
```diff
@@ -362,10 +362,13 @@ renewable:
 # Costs Configuration
 costs:
   year: 2030
-  version: v0.10.0
+  technology_data_url: "https://raw.githubusercontent.com/PyPSA/technology-data/"
+  technology_data_version: v0.12.0
```
yerbol-akhmetov: Hi, @danielelerede-oet, maybe we can leave as `version: v0.12.0`. Probably, we will not need `technology_data_url` if we use exiting `retrieve_cost_data` rule with wildcard that was suggested.

config.default.yaml
```diff
@@ -362,10 +362,13 @@ renewable:
...
+  technology_data_US: false # true if you want to use US-specific costs from technology-data/outputs/US instead of technology-data/outputs
+  cost_scenario: "moderate" # only used if `technology_data_US: true`; can be "moderate", "advanced" or "conservative"
+  financial_case: "market" # only used if `technology_data_US: true`; can be "market" or "r&d"
```
yerbol-akhmetov: Is it quite needed to have lowercase for `cost_scenario` and `financial_case`? I thought that `technology-data` for US have those starting with capital letter (e.g. Moderate, Market). If all scenario and financial case entries follow the same format, we can skip lowercase conversion later: `financial_case.str.casefold()`. But we can keep as it is.

yerbol-akhmetov (commented):
Hi, @danielelerede-oet, @finozzifa. Great thanks for the PR. I have left few comments.

Snakefile
danielelerede-oet: Really? This is so weird... it works for me! Then better to switch back to the older format, thanks for noticing @yerbol-akhmetov !

yerbol-akhmetov: > Really? This is so weird... it works for me! Then better to switch back to the older format, thanks for noticing @yerbol-akhmetov !

It seems like the following rules need `costs_2020.csv`, but `retrieve_cost_data` does not explicitly provides this file as output. As a result, it seems like it cannot be found.

scripts/add_electricity.py
```diff
@@ -138,17 +142,34 @@ def load_costs(tech_costs, config, elec_config, Nyears=1):
...
     costs = pd.read_csv(tech_costs, index_col=["technology", "parameter"]).sort_index()
 
-    # correct units to MW and EUR
+    # correct units to MW and output_currency
     costs.loc[costs.unit.str.contains("/kW"), "value"] *= 1e3
     costs.unit = costs.unit.str.replace("/kW", "/MW")
-    costs.loc[costs.unit.str.contains("USD"), "value"] *= config["USD2013_to_EUR2013"]
-
+    costs = convert_currency_and_unit(costs, config["output_currency"])
     costs = costs.value.unstack().fillna(config["fill_values"])
```
yerbol-akhmetov: Also, I think `costs = costs.value.unstack().fillna(config["fill_values"])` line needs to be moved after `scenario` and `financial_case` selection. Otherwise, I was getting error of duplicate entries. After moving downward, the error disappears.
image

scripts/add_electricity.py
danielelerede-oet: @yerbol-akhmetov thanks for the suggestion. Concerning your proposal to use a simpler code, I would propose a third option to avoid issues with possible missing `scenario` or `financial_case` fields in cost files and to manage cells containing `NaN` values, adding a check on empty cells in the cost file:

```
    # apply filter on financial_case and scenario, if they are contained in the cost dataframe
    wished_cost_scenario = config["costs"]["cost_scenario"]
    wished_financial_case = config["costs"]["financial_case"]
    for col in ["scenario", "financial_case"]:
        if col in costs.columns:
            costs[col] = costs[col].replace("", pd.NA)

    if "scenario" in costs.columns:
        costs = costs[
            (costs["scenario"].str.casefold() == wished_cost_scenario.casefold()) |
            (costs["scenario"].isnull())
            ]

    if "financial_case" in costs.columns:
        costs = costs[
            (costs["financial_case"].str.casefold() == wished_financial_case.casefold()) |
            (costs["financial_case"].isnull())
            ]
```

scripts/add_electricity.py
```diff
@@ -138,17 +142,34 @@ def load_costs(tech_costs, config, elec_config, Nyears=1):
...
 
+    # apply filter on financial_case and scenario, if they are contained in the cost dataframe
+    wished_cost_scenario = config["costs"]["cost_scenario"]
+    wished_financial_case = config["costs"]["financial_case"]
```
yerbol-akhmetov: I think here should be just:
```
    wished_cost_scenario = config["cost_scenario"]
    wished_financial_case = config["financial_case"]
```
Because config is already config["costs"] as defined in line 838:
```
    costs = load_costs(
        snakemake.input.tech_costs,
        snakemake.params.costs,
        snakemake.params.electricity,
        Nyears,
    )
```
image

scripts/retrieve_cost_data.py
danielelerede-oet: Hi @yerbol-akhmetov , the `retrieve_cost_data.py` script was developed in view of a possible integration of unit tests, but it's not needed anymore. We can work with the version you propose!

config.default.yaml
danielelerede-oet: Thanks @yerbol-akhmetov , indeed you're right, but we decided to go with this to avoid any possible issues with future `technology-data` updates that may  imply coming back here and take actions. My suggestion is to keep it like this

ekatef:
Update: have tested update of the lock-environment files, and I confirm that it works perfectly in the `upstream` ([an example](https://github.com/pypsa-meets-earth/pypsa-earth/actions/runs/15477035876) of a successful run). Running same [github actions](https://github.com/open-energy-transition/pypsa-earth/actions/runs/15477164067/job/43575528019) workflow in fork throws a permission error.

The full error traceback looks like that:
`Pushing pull request branch to 'origin/update-locked-environment'
  /usr/bin/git push --force-with-lease origin update-locked-environment:refs/heads/update-locked-environment
  remote: Permission to open-energy-transition/pypsa-earth.git denied to github-actions[bot].
  fatal: unable to access 'https://github.com/open-energy-transition/pypsa-earth/': The requested URL returned error: 403`
  
 It appears to me that some permissions must be adjusted for the fork.

Snakefile
```diff
@@ -396,26 +397,31 @@ if not config["enable"].get("build_natura_raster", False):
...
+                version=config["costs"]["technology_data_version"],
+            input:
+                HTTP.remote(
+                    f"raw.githubusercontent.com/PyPSA/technology-data/{config['costs']['technology_data_version']}/outputs/{cost_directory}/"
+                    + "costs_{year}.csv",
+                    keep_local=True,
+                ),
+            output:
+                "resources/" + RDIR + "costs_{year}.csv",
+            log:
```
yerbol-akhmetov: Hi, @danielelerede-oet. These two if conditions occasionally intended, as a result we see not `retrieve_cost_data` rule.

danielelerede-oet:
@ekatef @davide-f I updated this after changes in #1319 and solved the mystery behind failure of workflow tests thanks to @yerbol-akhmetov , now it should be ready. As mentioned in a previous comment, the new configuration doesn't impact the current workflow. US-specific data are triggered only if `country` is US AND `technology_data_us` is `true` as a safety measure to avoid fetching unintended data.

This PR also includes update to `technology_data_version` v0.13.1 (including reviewed electrolyzer data and small fixes to DAC values)

yerbol-akhmetov:
Hi, @danielelerede-oet, should we change the cost version to `v0.13.2`? What do you think?

danielelerede-oet:
> Hi, @danielelerede-oet, should we change the cost version to `v0.13.2`? What do you think?

@yerbol-akhmetov that's definitely needed in #1227  but I can do it already here

yerbol-akhmetov:
No worries, @danielelerede-oet. I just forgot that I have already updated cost version in #1227.

Snakefile
```diff
@@ -397,14 +398,20 @@ if not config["enable"].get("build_natura_raster", False):
             shutil.copyfile(input[0], output[0])
 
 
+if config["countries"] == ["US"] and config["costs"]["technology_data_US"]:
+    cost_directory = "US"
```
davide-f: Some users may be willing to use US costs even for another region; is there a praticular us-specific thing for ensuring this?

When the country is not US only [there is at least another country that is not US], it would be best to raise a log/warning

Snakefile
```diff
@@ -397,14 +398,20 @@ if not config["enable"].get("build_natura_raster", False):
...
 if config["enable"].get("retrieve_cost_data", True):
 
     rule retrieve_cost_data:
         params:
-            version=config["costs"]["version"],
+            version=config["costs"]["technology_data_version"],
         input:
             HTTP.remote(
-                f"raw.githubusercontent.com/PyPSA/technology-data/{config['costs']['version']}/outputs/"
+                f"raw.githubusercontent.com/PyPSA/technology-data/{config['costs']['technology_data_version']}/outputs/{cost_directory}/"
```
davide-f: Is this back-compatible?
Could be best to use something like:
```suggestion
                f"raw.githubusercontent.com/PyPSA/technology-data/{config['costs']['technology_data_version']}/outputs/{cost_directory}"
```
and add the "/" in cost_directory when relevant.

config.default.yaml
```diff
@@ -362,7 +362,11 @@ renewable:
 # Costs Configuration
 costs:
   year: 2030
-  version: v0.12.0
+  technology_data_version: v0.13.2
+  technology_data_US: false # true if you want to use US-specific costs from technology-data/outputs/US instead of technology-data/outputs (also requires country: ["US"])
```
davide-f: Instead of having the flag true/false, what about specifying directly the folder here? e.g. country_specific_data= "" [default] and = "US" and we use this as folder?
Is it worth generalizing or not really? I guess you have talked with Johannes on this; if it is not useful to generalize now we can ignore this comment

scripts/_helpers.py
```diff
@@ -1026,6 +1025,7 @@ def convert_currency_and_unit(
 
 def prepare_costs(
     cost_file: str,
+    config: str,
```
davide-f: should config be a dict?

davide-f (commented):
Great PR :D few thoughts :)

Snakefile
yerbol-akhmetov: Hi, @davide-f. Thanks for the great comment. We can leave only `config["costs"]["technology_data_US"]` in the `if` statement and raise warning if country is not US as you have suggested. Great suggestion.

Snakefile
yerbol-akhmetov: > raw

Thank you, @davide-f. It is back compatible because `cost_directory` will be empty string ``. And `//` is treated as single `/`. We can have / removed as suggested, but then for cost_directory, we need to have not "US", but "US/".

config.default.yaml
yerbol-akhmetov: Great comment, Davide. I like the suggestion. Maybe we can do so and highlight in the comment that US only available.

Snakefile
danielelerede-oet: @davide-f @yerbol-akhmetov actually that was the initial configuration. The only condition was that US was the only country in the `countries` list, so we can definitely remove the first condition and add a Warning :)

danielelerede-oet: @davide-f what about something like this?

```
if config["countries"] == ["US"] and config["costs"]["technology_data_US"]:
    cost_directory = "US"
elif config["costs"]["technology_data_US"]:
    cost_directory = ""
    warnings.warn(
        "Please note that you're using US-specific costs. If that is not intentional, "
        "set the config parameter \"technology_data_US\" to false"
    )
else:
    cost_directory = ""
```

So if you're using the model for the US you're actually using the best possible data, otherwise you can choose but you get a warning if `country` is not US

yerbol-akhmetov: Hi, @danielelerede-oet. I thought there should be `cost_directory = "US"` for the second option. Because what if the user wants to use US costs but for another country then we need to fetch US costs.

danielelerede-oet: Indeed, thanks Yerbol, I changed the logic a bit more also to comply with the other comment below to generalize cost directories

config.default.yaml
danielelerede-oet: Great suggestion @davide-f , implemented (and Snakefile changed accordingly)!

scripts/_helpers.py
danielelerede-oet: Thanks for spotting it Davide! Changed!

yerbol-akhmetov:
Hi, @davide-f. I have applied your suggestions regarding `/` and comment in config file. Could you please review it in your free time?

config.default.yaml
```diff
@@ -362,11 +362,15 @@ renewable:
 # Costs Configuration
 costs:
   year: 2030
-  version: v0.12.0
+  technology_data_version: v0.13.2
+  country_specific_data: "" # (optional) country-specific data are currently available only for "US". Use "US" to apply US-specific costs.
```
davide-f:
```suggestion
  country_specific_data: "" # (optional) Reference to the desired technology-data directory for techno-economic input data; Only "" and "US" supported, for other values check the technology-data output directory
```

What do you think?

doc/configtables/costs.csv
```diff
@@ -1,15 +1,22 @@
...
-version,--,"vX.X.X; e.g. 'v0.1.0'","Version of ``technology-data`` repository to use."
-rooftop_share,--,float,"Share of rooftop PV when calculating capital cost of solar (joint rooftop and utility-scale PV)."
-USD2013_to_EUR2013, --, "float", "US Dollar-Euro exchange rate. Source: `ECB <ECB: https://www.ecb.europa.eu/stats/exchange/eurofxref/html/eurofxref-graph-usd.en.html>`_."
-fill_values,--,float,"Default values if not specified for a technology in ``resources/costs.csv``."
+year,,"YYYY; e.g. '2030'","Year for which to retrieve cost assumptions of ``resources/costs.csv``."
+technology_data_version,,"vX.X.X; e.g. 'v0.1.0'","Version of ``technology-data`` repository to use."
+country_specific_data,bool,"""{"""",""US""}""",Reference to the desired `technology-data` directory for techno-economic input data
+cost_scenario,,"{"moderate","Only used if `country_specific_data`: ""US""` to select the desired NREL financial case",,
+financial_case,,"{"market","Only used if `country_specific_data: ""US""` to select the desired NREL financial case",
+discountrate,,"{0.071,0.086,0.111}","Rate of return used to discount future cash flows back to their present value"
```
davide-f: for discount rate we support also other values not limited to {0.071,0.086,0.111}, we can say

```suggestion
discountrate,,float,"Rate of return used to discount future cash flows back to their present value"
```

doc/configtables/costs.csv
```diff
@@ -1,15 +1,22 @@
...
+output_currency,,"""{""EUR"",""USD""}""",Reference currency for all costs,
```
davide-f: do we now support currencies beyond EUR/USD?

davide-f (commented):
Super minor things, then good to go for me! Many thanks!

doc/configtables/costs.csv
danielelerede-oet: Yes, this is the list of supported currencies for the conversion from EUR:
USD, JPY, BGN, CZK, DKK, GBP, HUF, PLN, RON, SEK, CHF, ISK, NOK, TRY, AUD, BRL, CAD, CNY, HKD, IDR, ILS, INR, KRW, MXN, MYR, NZD, PHP, SGD, THB, ZAR

davide-f: Then it may be good to revise the """{""EUR"",""USD""}""" with 1/2 entries more and add in the description the reference to the supported ones.
Is the info available in the package used for conversion?

danielelerede-oet: Added the information in configtables/costs.csv with a few currencies under values and the link to the table where all the available currencies are shown

davide-f (approved):
Many thanks :D merging

PR #1444, opened 2025-05-17
Fix missing comma for other energy in build_energy_totals calculation

yerbol-akhmetov:
## Changes proposed in this Pull Request
Good day. This PR aims to fix the missing comma for `other_energy` in `build_energy_totals.py` script. I have noticed that other energy source for US was not accounted while still being high (191 TWh). Making closer looked showed that we missed comma, as a result proper rows were not selected. Below I show the entry for other energy.

image

## Checklist

- [x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [x] I tested my contribution locally and it seems to work fine.
- [ ] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes, including reference to the requested PR.

PR #1468, opened 2025-05-30
Fill missing efficiency gain and growth rates with 0

yerbol-akhmetov:
Good day. This PR aims to fill missing columns in efficiency gain and growth rates with 1. After we fixed the comma and made the retrieval of consumption by other unspecified sources possible in #1444, it was noticed that `efficiency_gain` and `growth_rate` files have missing columns for `other electricity`, `other biomass` and so on. As a result, multiplication to calculate `energy_base_totals` result in `NaN`. As a temporary fix, it is suggested to fill missing columns with 1, so that those consumption is at least preserved rather than becoming `NaN` and finally 0. Below figure shows other energy consumption before introduction of change with `NaN`s and after filling with 0. 

image

I have one question also which is related to this topic. Currently, we are extracting consumption by `other` from UN data, but it is not further added in the workflow. Do you think adding other consumption to upstream workflow would benefit? @davide-f 

## Checklist


davide-f (commented):
Many thanks @yerbol-akhmetov !
I'm wondering: rather than having this code solution, what about adding the columns to the data files?

The reason is that this is a silent change that would silently fill columns that are not available in the data but actually needed.
Moreover, if the user improperly fills the values, this code may silently fills the holes leading to unexpected behaviors.

What do you think?

yerbol-akhmetov:
> Many thanks @yerbol-akhmetov ! I'm wondering: rather than having this code solution, what about adding the columns to the data files?
> 
> The reason is that this is a silent change that would silently fill columns that are not available in the data but actually needed. Moreover, if the user improperly fills the values, this code may silently fills the holes leading to unexpected behaviors.
> 
> What do you think?

Hi, @davide-f. I agree with your suggestion. I thought adding just 1.0 for `efficiency_gain` and `growth_factor` might be not good, while investigating projected values for already presented countries, I thought might take too much time. Do you think adding columns as 1.0 for those csv files for `other` energy sources are fine?

davide-f:
Agree

I believe you can make an initial good guess from the other values being present, by selecting values similar for other sectors.
Worst case, rather than having fill 1, it is better to have 1 in the csv as it would have the same effect but it is more explicit.

For example for other electricity, we could leverage on values of electricity columns for example.

For the growth values of previous regions, probably @hazemakhalek or @energyLS for MA and @danielelerede-oet for US could give insights if they have time.

yerbol-akhmetov:
> Agree
> 
> I believe you can make an initial good guess from the other values being present, by selecting values similar for other sectors. Worst case, rather than having fill 1, it is better to have 1 in the csv as it would have the same effect but it is more explicit.
> 
> For example for other electricity, we could leverage on values of electricity columns for example.
> 
> For the growth values of previous regions, probably @hazemakhalek or @energyLS for MA and @danielelerede-oet for US could give insights if they have time.

Hi, @davide-f, thanks for the review. I have added entries for `efficiency_gains_cagr.csv` and `growth_factors_cagr.csv` based on the existing entries. In particular, `other electricity`, `other oil`, `other gas`, `other biomass`, and `other heat`. For electricity, oil, gas, and biomass, I have taken residential as a reference (e.g. efficiency gains for other electricity as residential electricity and so on). For `other heat`, I have left with empty entries, because files did not contain heat related gains and factors. This is just a proposal. Happy to hear any more suggestions and recommendations.

davide-f (commented):
Great!
To me other heat could be filled as other heat columns.

Do you mind filling the missing values with 0.0, that hopefully fixes some challenges. Moreover, I'd say "DEFAULT" could be best filled with 0.0 rather than other values. Though this is not related to you.
If you'd be willing to include that it would be great :)

To me, filling other heat is enough to have this PR ready; I'd wait to hear @hazemakhalek , @energyLS or @danielelerede-oet 's opinion as they recently delved deeper into that.
Happy to merge afterwards :)

Pleare remind the release_note once finalized with changes as this would be nice to track. Many thanks :)

yerbol-akhmetov:
> Great! To me other heat could be filled as other heat columns.
> 
> Do you mind filling the missing values with 0.0, that hopefully fixes some challenges. Moreover, I'd say "DEFAULT" could be best filled with 0.0 rather than other values. Though this is not related to you. If you'd be willing to include that it would be great :)
> 
> To me, filling other heat is enough to have this PR ready; I'd wait to hear @hazemakhalek , @energyLS or @danielelerede-oet 's opinion as they recently delved deeper into that. Happy to merge afterwards :)
> 
> Pleare remind the release_note once finalized with changes as this would be nice to track. Many thanks :)

Hi, @davide-f, thank you for the suggestions. I have not filled heat because we do not have heat by itself. Instead, we have:
image
image
We have `residential heat oil`, `residential heat biomass`, and `residential heat gas`, not pure heat. I was not sure which one to choose. Therefore, I have filled `other heat` with 0.0 to be on the safe side.

davide-f (commented):
@yerbol-akhmetov souns great! :)
Do you mind filling the nan values too? that may allow fixing possible issues along the way

danielelerede-oet:
Thanks @yerbol-akhmetov for the PR. I checked the EIA projections for the US and the only thing they have regarding other unspecified consumption is this:
image
Personally, I see 2 options:
1) we consider a 0 growth for all the "other energy consumption" fields
2) we assign a −0.00228 CAGR to all of the "other energy consumption" fields
I would go for option 1 just to be more conservative as the single value provided by EIA is not that detailed, but I'd like to know what you think too :)

yerbol-akhmetov:
> Thanks @yerbol-akhmetov for the PR. I checked the EIA projections for the US and the only thing they have regarding other unspecified consumption is this: image Personally, I see 2 options:
> 
> 1. we consider a 0 growth for all the "other energy consumption" fields
> 2. we assign a −0.00228 CAGR to all of the "other energy consumption" fields
>    I would go for option 1 just to be more conservative as the single value provided by EIA is not that detailed, but I'd like to know what you think too :)

Thanks, @danielelerede-oet. I have filled `other sources` with 0 for the US. Also I have filled NaNs with 0 as suggested by @davide-f. I think now it should be ready.

davide-f (commented):
Looks good to me! :D
Do you mind adding a release note, to track when the file has changed? Then it is ready to fly :)

danielelerede-oet:
@yerbol-akhmetov @davide-f I have noticed another thing: the `default` values are usually referred to MA, but the updates for other energy uses are using 0 values, which are instead adopted for the US. Although I don't really like the fact that the `default` corresponds to a very optimistic situation, maybe would be better to align with the previous strategy

yerbol-akhmetov:
> @yerbol-akhmetov @davide-f I have noticed another thing: the `default` values are usually referred to MA, but the updates for other energy uses are using 0 values, which are instead adopted for the US. Although I don't really like the fact that the `default` corresponds to a very optimistic situation, maybe would be better to align with the previous strategy

Hi, @danielelerede-oet. Generally, default one tracks MA as you noted. However, @davide-f suggested to put 0.0 for other energy entries to be on the safe side. Anyway, the share of other energy is quite small.

danielelerede-oet:
@yerbol-akhmetov totally fine for me then!

davide-f:
> > @yerbol-akhmetov @davide-f I have noticed another thing: the `default` values are usually referred to MA, but the updates for other energy uses are using 0 values, which are instead adopted for the US. Although I don't really like the fact that the `default` corresponds to a very optimistic situation, maybe would be better to align with the previous strategy
> 
> Hi, @danielelerede-oet. Generally, default one tracks MA as you noted. However, @davide-f suggested to put 0.0 for other energy entries to be on the safe side. Anyway, the share of other energy is quite small.

In full transparency, the "DEFAULT" value was not chosen with a specific criterion; for efficiency gains and growth rates it would make absolute sense to have a default 0: no growth as it is conservative.

If you wish, we can include this change here too.
I didn't want to bother you, but I'm highly pro this in case

danielelerede-oet:
> > > @yerbol-akhmetov @davide-f I have noticed another thing: the `default` values are usually referred to MA, but the updates for other energy uses are using 0 values, which are instead adopted for the US. Although I don't really like the fact that the `default` corresponds to a very optimistic situation, maybe would be better to align with the previous strategy
> > 
> > 
> > Hi, @danielelerede-oet. Generally, default one tracks MA as you noted. However, @davide-f suggested to put 0.0 for other energy entries to be on the safe side. Anyway, the share of other energy is quite small.
> 
> In full transparency, the "DEFAULT" value was not chosen with a specific criterion; for efficiency gains and growth rates it would make absolute sense to have a default 0: no growth as it is conservative.
> 
> If you wish, we can include this change here too. I didn't want to bother you, but I'm highly pro this in case

Actually @davide-f my idea would be in the middle. I like your conservative view in the case of efficiency gains, but if we're performing future model runs I would expect a generalized increase in energy consumption due to increase in population/improved living conditions/economic development, reflected in non-zero growth rates. Maybe this is not the PR where we discuss this, but I would prefer the default value to fall back to US data (which can be associated with developed countries, thus depicting a significant but not dramatic increase in energy consumption), while I agree with keeping default efficiency gains to 0.

davide-f:
> Actually @davide-f my idea would be in the middle. I like your conservative view in the case of efficiency gains, but if we're performing future model runs I would expect a generalized increase in energy consumption due to increase in population/improved living conditions/economic development, reflected in non-zero growth rates. Maybe this is not the PR where we discuss this, but I would prefer the default value to fall back to US data (which can be associated with developed countries, thus depicting a significant but not dramatic increase in energy consumption), while I agree with keeping default efficiency gains to 0.

I see your perspective and I'm fine also to identify a non-zero default, however I wouldn't like to select a value for a developed country as the standard as it would be highly biased.
PyPSA-Earth is also likely used by various for regions that are not developed and I'd think a bit about having a standard value that is US or EU based.

Identifying a world average sounds a good alternative. We agree however on the current filling of "other columns" and nan values; so this PR looks good to go :)

davide-f (approved):

davide-f:
> Pleare remind the release_note once finalized with changes as this would be nice to track. Many thanks :)



Ah sorry @yerbol-akhmetov , please remind the release_note as this PR has data and modelling implications

yerbol-akhmetov:
> > Pleare remind the release_note once finalized with changes as this would be nice to track. Many thanks :)
> 
> Ah sorry @yerbol-akhmetov , please remind the release_note as this PR has data and modelling implications

Sorry just forgot to add. I will do it asap.

davide-f (approved):

davide-f:
Great :D Merging, many thanks!

PR #1469, opened 2025-05-31
Make costs path as consistent with existing references

yerbol-akhmetov:
Good day. This PR aims to make costs path consistent in Snakefile. This PR enables having cost `year` different than `planning_horizon`. Currently, we have `year` and `planning_horizon` being used interchangeably. If `retrieve_cost_data` is set to false, then sector run throws error because some rules use `resources + RDIR + costs_{planning_horizon}.csv` as cost input not `COSTS` that was defined. I have tested the whole workflow with this change and it worked fine. One thing to note is defining cost year for myopic runs, but I think we can tackle the issue step by step at different PR. What are your thoughts about PR? @davide-f @hazemakhalek 

## Checklist


Snakefile
```diff
@@ -405,13 +405,13 @@ if config["enable"].get("retrieve_cost_data", True):
...
-                + "costs_{year}.csv",
+                + f"costs_{config['costs']['year']}.csv",
                 keep_local=True,
             ),
         output:
-            "resources/" + RDIR + "costs_{year}.csv",
+            COSTS,
         log:
-            "logs/" + RDIR + "retrieve_cost_data_{year}.log",
+            "logs/" + RDIR + f"retrieve_cost_data_{config['costs']['year']}.log",
```
davide-f: I see the inconsistency year here and planning_horizon below; what about specifying planning_horizon as wildcard here and keeping planning_horizon below?

Snakefile
```diff
@@ -1089,7 +1089,7 @@ rule prepare_sector_network:
     input:
         network=RESDIR
         + "prenetworks/elec_s{simpl}_{clusters}_ec_l{ll}_{opts}_{sopts}_{planning_horizons}_{discountrate}_{demand}_presec.nc",
-        costs="resources/" + RDIR + "costs_{planning_horizons}.csv",
+        costs=COSTS,
```
davide-f: The side effect here is that in overnight, the costs is fixed to the cost file in the data or resource folder.
In myopic, it is no more going to change across time horizons unfortunately

davide-f (commented):
I've added few comments.

Beyond inconsistency of wildcard naming, have you encountered issues?

yerbol-akhmetov:
> I've added few comments.
> 
> Beyond inconsistency of wildcard naming, have you encountered issues?

Hi, @davide-f. Thanks for the comment. Yes, apart from inconsistency, the issue appears when you would like to simulate 2023 scenario as example. Suppose I would like to use cost year of 2020, as there is no cost assumptions for 2023 in technology-data. Therefore, you should place `planning_horizon: 2020`. However, in that case `efficiency_gains` and `growth_factors` will not be calculated accurately, because `no_years = planning_horizons - base_year`. If you select `base_year = 2021` and `planning_horizon = 2020`, then `no_years = -1`. While in fact we would like to simulate 2023 - 2021 = 2 years.

Snakefile
yerbol-akhmetov: > The side effect here is that in overnight, the costs is fixed to the cost file in the data or resource folder. In myopic, it is no more going to change across time horizons unfortunately

Yes, this is a concern that I have risen. Therefore, `planning_horizons` are left as wildcard for myopic rules. Alternative, might be providing list of cost years or dictionary for myopic, but maybe it can be handled in another PR, or in this.

Snakefile
yerbol-akhmetov: Hi, could you please clarify about inconsistency? I thought COSTS is either `COSTS = "resources/" + RDIR + f"costs_{config['costs']['year']}.csv"` if `retrieve_cost_data` is true, or `COSTS = data/costs.csv`. However, here config['costs']['year']} should be consistent with COSTS. If we place here planning_horizons, then we cannot use `planning_horizon` other than 2020, 2025, and so on. For `prepare_sector_networks` we can keep `planning_horizons`, but what would be the reason and shortcomings of COSTS?

davide-f: COSTS is either "data/costs.csv" or "resources/../costs_year" where year is from the costs section.
This means that this formulation won't allow downloading multiple cost years for a myopic optimization

davide-f (commented):
Great discussion today!

It seems it emerged:
- restore {planning_horizon} wildcard for the rules, instead of COSTS
- possibly rename wildcard {year} in retrieve_cost_data into {planning_horizon} for consistency
- a proposal in retrieve_cost_data to either round/ceil/floor the {planning_horizon} to the closest 5 or 0, so to allow any planning horizon.

@yerbol-akhmetov is this consistent with your view? any comment

yerbol-akhmetov:
> Great discussion today!
> 
> It seems it emerged:
> 
> * restore {planning_horizon} wildcard for the rules, instead of COSTS
> * possibly rename wildcard {year} in retrieve_cost_data into {planning_horizon} for consistency
> * a proposal in retrieve_cost_data to either round/ceil/floor the {planning_horizon} to the closest 5 or 0, so to allow any planning horizon.
> 
> @yerbol-akhmetov is this consistent with your view? any comment

Hi, @davide-f. I have thoughts about how we could go further with this PR. I thought about dictionary for mapping `planning_horizon` to `cost_year`. While playing with it I come across to your comment that sector rules use {planning_horizon} wildcard, while power model used COSTS from historical perspective. But this is still seems inconsistent. In `add_electricity` rule `COSTS` is used. So `elec.nc` is formed with only one cost year, even though you try to simulate two planning horizons with overnight method. Eg:
```
    planning_horizons:
    - 2020
    - 2025
```
In that case, we will still had costs from `costs: year` not planning horizons for `elec.nc`. So our costs for added generators (.e.g onwind, ror, and etc) will be not right for some horizons. Even though for other links costs are added in `prepare_sector_network`, our generators for both horizons will have the same costs. What I would like to say that simulating multiple horizons is not totally right with overnight, or myopic right now. Only having horizon wildcard for `elec.nc` might be ultimate solution but too much computation, or readding generators back in `prepare_sector_network` with proper costs.

Generally, my plan was having something like shown below, where cost year is taken from horizon. But for initial power rules we have not horizon as wildcard. I would think about the method in more detail.
```
costs:
  cost_year_for_horizon:  # For planning horizon X, use cost year Y in a form of X: Y
    2020: 2020  # Planning horizon 2020 uses cost data from 2020
    2025: 2025  # Planning horizon 2025 uses cost data from 2025
    2030: 2030
    2035: 2035
    2040: 2040
    2045: 2045
    2050: 2050
```

yerbol-akhmetov:
Hi, @davide-f. Hope you are doing well. I have revised the cost year implementation to decouple it from `planning_horizon`. In particular, what it does is we use `horizon_to_cost_year` dictionary to have mapping of planning_horizon to cost year. I have set all available cost years by default (for 2020, 2025, and till 2050). With such dictionary, it is possible to provide `2023: 2020`, which make it possible to use 2020 cost year for `planning_horizon` of 2023.

In short, here is rules:
1) Power model: the first planning horizon is used to get cost year from dictionary. We can select any planning_horizon unless you provide dictionary with cost year:
```
costs:
  horizon_to_cost_year: # For planning horizon X, use cost year Y in a form of X: Y
    2023: 2020
```
2) Sector model: cost year is selected based on `snakemake.wildcards.planning_horizons` from `horizon_to_cost_year` dictionary. Eg. 
```
costs:
  horizon_to_cost_year: # For planning horizon X, use cost year Y in a form of X: Y
    2023: 2020
    2040: 2040
```
3) Planning horizon that you select needs to be present in `horizon_to_cost_year`.

I have tested the implementation on all possible occasions:
1) Power model, 1 planning horizon (aligned with cost year) (e.g. 2030), `retrieve_cost_data=True`. 
2) Sector model, 2 planning horizons (aligned with cost year) (e.g. 2030, 2040), `retrieve_cost_data=True`. . 
3) Power model, 1 planning horizon (NOT aligned with cost year) (e.g. 2023), `retrieve_cost_data=True`.
4) Sector model, 2 planning horizons (NOT aligned with cost year) (e.g. 2023, 2040), `retrieve_cost_data=True`.
5) Power model, `retrieve_cost_data = False`
6) Sector model, `retrieve_cost_data = False`. It was prepared until prepare_sector_networks, then error appeared. It was because of the `data/costs.csv` needs revision for sector model. In the upstream, using `data/costs.csv` is not possible.

Can you please review it?

GbotemiB:
Hello Yerbol, Looks like this work might have been forgotten. Do you need resources to continue this work?

feel free to let me know.

yerbol-akhmetov:
> Hello Yerbol, Looks like this work might have been forgotten. Do you need resources to continue this work?
> 
> feel free to let me know.

Hi, @GbotemiB. Thank you for the interest. The aims of the PR were partially covered by PR #1719 on migrating costs by @virio-andreyana. And now it opens a door to focus on other ideas in this PR (namely making a separation between planning_horizon and cost year). This way we can use any planning_horizon (e.g. 2023), while assigning any cost year to it (e.g. 2020 or 2025). The idea was to provide dictionaries mapping planning horizons to cost years. Do you want to investigate this PR?

PR #1492, opened 2025-06-19
Fix logic default currency conversion

danielelerede-oet:
@davide-f @ekatef 

This PR fixes the logic for the selection of the default exchange rate in #1319 . In the current state, the default exchange rate is selected if present in the config file (i.e. always). With this patch, it would be possible to only use it if no data are actually found.

The target is to get an average exchange rate for any currency and any year in `currency_year` from `technology-data`.

### Behavior by Case

| Case | Description | Behavior |
|------|-------------|----------|
|  **Partial or complete data available for the year** | At least one daily exchange rate is available for the given year | Returns the **average of available exchange rates**, ignoring missing days |
|  **No data for the entire year** | No exchange rate available at all for the year | Returns `default_exchange_rate` if provided; otherwise, raises `RuntimeError` |
|  **Future year requested** | Year > current year | Internally replaced with the current year (`today.year`) and follows above logic |
|  **Leap year handling** | Year has 366 days | Iterates over 366 days instead of 365 |

## Checklist


danielelerede-oet:
@ekatef @davide-f @yerbol-akhmetov I've received feedback about currency conversion asking if we can also give the possibility to choose between different configurations for the future development of exchange rates. at the moment, the average of the latest available year is taken as fixed exchange rate for future years, but a few people asked me if we can give the possibility to also use "customized" rates. What do you think?

davide-f:
> @ekatef @davide-f @yerbol-akhmetov I've received feedback about currency conversion asking if we can also give the possibility to choose between different configurations for the future development of exchange rates. at the moment, the average of the latest available year is taken as fixed exchange rate for future years, but a few people asked me if we can give the possibility to also use "customized" rates. What do you think?

To me is absolutely reasonable

scripts/_helpers.py
```diff
@@ -966,33 +966,74 @@ def annuity(n, r):
         return 1 / n
 
 
+# Simple cache to avoid repeated computations and logging for same (currency, output_currency, year)
+_conversion_cache = {}
+
+
 def get_yearly_currency_exchange_average(
```
davide-f: It is worth having _conversion_cache as argument of the function to avoid the hidden dependency and allow customization of entries potentially. The name of the variable may be revised in _currency_conversion_cache to make it more explicit maybe

davide-f: It is worth adding a docstring explaining the function

davide-f (commented):
I've added a comment on the current implementation.

The proposal look simple to me; more complex formulation using a class for cost conversion may be considered but not worth now.
Alternatively, the code may be revised into two functions like:
1. a first function that creates the conversion cache for the relevant currencies, being executed over the unique values of the entries
2. a second function that adopts the dictionary for the conversion

I'm dropping it as an idea to hear opinions but no pressures on development. what do you think of this second approach?

danielelerede-oet:
@davide-f I have implemented your suggestions. The tests here work, but I'm trying all the different combinations to understand if they actually work as intended  when adopting them

danielelerede-oet:
Ok @davide-f I tested it and it works as intended. If `currency_year` > latest available year for currency conversion (e.g. 2030), the fallback to either the latest available value (2025 as of today) or a custom value can be activated using the config file, and log messages are provided to confirm the behavior.

davide-f (approved):

PR #1535, opened 2025-08-01
Generalize usage of SEG option

davide-f:
# Closes # (if applicable).
In some senarios, using in the opt wildcard the time segmentation (SEG option) instead of the typical H option can lead to errors due to nan values being generated.
This PR fixes the issue.

It has been tested for power-only so far; ongoing testing aims to verify its functionality for sector-coupled runs

## Checklist


davide-f:
Tested PR with 365SEG option for opts and sopts.

Result in terms of snapshot clustering is the following:
image

@energyLS could you please revise the PR? especially the revision to add_export to adapt the rule for different clustering options for temporal clustering

davide-f:
It seems that sopts option was only used here in add_export. I could drop the wildcard and simplify the workflow

davide-f:
Hello @energyLS :) would you mind having a look when you have time?

With this rule, we may be able to drop the sopt option but I'd like to align with you on this

scripts/add_export.py
```diff
@@ -180,13 +180,14 @@ def create_export_profile():
             )
 
     # Resample to temporal resolution defined in wildcard "sopts" with pandas resample
-    sopts = snakemake.wildcards.sopts.split("-")
-    export_profile = export_profile.resample(sopts[0].casefold()).mean()
+    sel_export = export_profile[n.snapshots]
+    pu_profile_export = sel_export / (1e-6 + sel_export.sum())
+    export_profile = pu_profile_export * export_profile.sum() * len(n.snapshots) / 8760
```
energyLS, line 185: Is there a specific reason why the pandas resampling function is dropped here? How is it done in other scripts?

energyLS: Found the answer myself: Since we do not have the `sopts` anymore, there is no direct information about the resolution, that is why it is obtained via `n.snapshots` and the pandas resample function can not be applied that simply.

energyLS (approved):
Dear @davide-f thank you for the PR! The code make sense to me. I support the goal to drop the `sopts`, however there are some more things to consider before doing so:

- update documentation
- `sopts` used in `solve_network.py`
- `sopts` used in `prepare_sector_network.py` to handle the resampling, by using the function `average_every_nhours()`. Can we use this function in this PR as well?

To sum up, in the current status the PR is fine as it is and ready to merge. However, if you want to drop the `sopts` completely, please consider my points (but this could also be done in a different PR)

ekatef:
Merged 🎉 
Many thanks for the contribution @davide-f, and thanks a lot for reviewing @energyLS 💚 Great team work 😄

PR #1540, opened 2025-08-06
Extend isolated nodes function to isolated networks

davide-f:
# Closes # (if applicable).

## Checklist

- [x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [x] I tested my contribution locally and it seems to work fine.
- [x] Code and workflow changes are sufficiently documented.
- [ ] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`.
- [ ] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [ ] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [x] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes, including reference to the requested PR.

PR #1542, opened 2025-08-07
Add subregionalization in `cluster_networks`

virio-andreyana:

This pull request adds the option of applying subregion not only in `simplify_network` but also in `cluster_network` using a switch:

```yaml
subregion:
  enable:
    simplify_network: true # activate subregion in simplify_network
    cluster_network: true # activate subregion in cluster_network
```

## Checklist


virio-andreyana:
I wonder if this aligned with the clustering improvement that you wanted @davide-f. 

What I usually do for example is to calculate the areas of each subregion in a jupyter notebook and distribute the buses using `focus_weights`. Or do you want something to be done automatically?

jome1:
Hi @virio-andreyana,
thanks for your work!
PyPSA-Earth would tremendously benefit from a function, so I can cluster to custom regions. Not always I want to have administrative boundaries as model regions, but my own defined regions within the country. For instance, I want to merge certain administrative zones to have fewer model regions within the country I am modelling.
Is your solution doing this?
I dont really understand what "subregions" means/does in `cluster_networks` and `simplify_networks`.

virio-andreyana:
Hi @jome1,
Thank you for your interest,
The concept behind subregions is that buses temporarily adopt a custom country format and then revert back to the original national format. This ensures compatibility with workflows that depend on the standard country code. The goal is to keep it non-intrusive.

If you have a geojson file formatted similarly to resources/shapes/country_shapes.geojson, you can specify its path from the Snakefile level. You can activate this in `simplify_network` and `cluster_network` by setting them as true respectively.

You can find the pull request documentation on subregions here:
https://pypsa-earth--1542.org.readthedocs.build/en/1542/configuration.html#subregion

davide-f (commented):
Great @virio-andreyana :D

Regarding this PR, if I understand right, you are trying to cluster some nodes based on proximity.
This reminds me of issues about "isolated nodes/networks" that add buses to the network but not really demand and create noise.

On that regard we have [these options](https://github.com/pypsa-meets-earth/pypsa-earth/blob/88d3e4c2515e6c8d5c81a7fa148c6718d3e25739/config.default.yaml#L85-L87) that allow to handle a bit that and I was wondering if there may be some duplication.
For clarity:
- p_threshold_drop_isolated: drops isolated nodes/networks with average demand below e.g. 20MW / snapshot
- p_threshold_merge_isolated: allows to merge isolated networks into a single node by country [so it doesn't drop the demand]. the isolated networks are identified by the average demand being no higher than the set threshold
- s_threshold_fetch_isolated: allows to merge the isolated networks to the closest node of the "main transmission system" being the subnetwork with the highest demand by country.

In the case of asean regions I may expect this third option to be interesting for you and may be further optimized potentially.

May there be some similarities?

FYI: https://github.com/pypsa-meets-earth/pypsa-earth/pull/1540 extends the isolated nodes to "isolated networks". Currently the formulation of the options above applies on "isolated nodes" that are AC buses with no lines; however if a super tiny island has 2 nodes and one line among them, the first 2 options won't apply to it. with that PR we extend the functionality also on them

virio-andreyana:
Hi @davide-f  :wave: 

Although the code `nearest_shape()` make use of proximity, its main objective is only to rename the country of buses within a given shape (and then reverting it back to the original form). It's not a method of clustering but impacts the existing clustering workflow. For example:

- in rule `simplify_network`, the function `merge_isolated_nodes()` and especially `merge_into_network()`, their aggregation is country specific.  
- in rule `cluster_network`, the function `busmap_for_n_cluster()` group buses based on countries. 

By redefining the countries into subregions, these code is altered. I wouldn't say its a duplication, but more of an enhancement of existing codes that you mentioned above.

The case application of this is the islands of Indonesia, the current aggregation scheme, will lead to most buses clustered to the island of Java simply because it has the most load and population. Using this subregion feature, I can distribute the number of buses to larger yet less populated islands using `focus_weights`

davide-f (commented):
Just moving here the summary of the discussion on Discord with @virio-andreyana ; feel free to revise anything here.

We discussed that this revision unlocks the possibility of using focus_weights using the name coding of the subregions.

We agreed on improving this PR by adding better documentation to document this new new functionality.

config.default.yaml
```diff
@@ -118,8 +118,12 @@ build_shape_options:
   contended_flag: "set_by_country" # "set_by_country" assigns the contended areas to the countries according to the GADM database, "drop" drops these contended areas from the model
 
 subregion:
+  enable:
+    simplify_network: true # activate subregion in simplify_network
```
davide-f, line 124: should this be false by default?

config.default.yaml
```diff
@@ -118,8 +118,12 @@ build_shape_options:
...
+    cluster_network: false # activate subregion in cluster_network
   define_by_gadm: false # name of the subregion. Multiple countries can be part in the same subregion.
   path_custom_shapes: false # (optional) provide the specific absolute path of the custom file e.g. (...\data\custom_shapes.geojson)
+  tolerance: 1000000 # Buffer distance (in meters) for assigning a country/subregion shape to a bus (the default tolerance is 100 km)
```
davide-f: this tolerance is currently 1000 km while default should be 100 by text. Shall we have the value directly in km instead of meters? It may be easier to read

config.default.yaml
virio-andreyana: Subregion for `simplify_network` prior to this pull request had function without any `enable` switch. Just in case someone had already use this feature and don't know that now it's a switch, I tend to prefer keeping them as on. But what do you think of this opinion?

config.default.yaml
virio-andreyana: Ah, my mistake. It should be 100 km. And you're right. Converting it now to km

virio-andreyana:
I just realized that `focus_weights` is poorly documented. So I made this separate pull request #1565

config.default.yaml
davide-f: I see that if simplify flag is true, but define by gadm and path are false basically the feature is disabled, which may make the flag simplify a kind of non-harmful duplication.
It can go as is and revise later if needed.

Unless you have a different opinion or no comment, I'd merge today as-is

davide-f (approved):

PR #1545, opened 2025-08-11
Change build_natura_raster

Eric-Nitschke:
## Description

Change `build_natura_raster.py` to only consider the countries of interest instead of the whole cutout.

### Reason

Right now, `build_natura_raster.py` calculates an exclusion raster for the entire cutout, while performing three main steps:
- unifying all geometries provided by the WDPA files in the `data\landcover` folder into one Multipolygon
- preparing a transformation matrix for rasterio
- saving the unified geometry as a rasterized map, spanning the entire cutout

This requires a lot of resources and time. Furthermore it is unnecessary, since the exclusion raster is only required for the analyzed countries, as well as their offshore regions.
Specifically the first step can take multiple hours (even for a small number of countries, depending on your setup) and the third step requires roughly 40 GB of RAM when using the Africa cutout, which a lot of setups might not have.

To address the time issue, I propose the following changes:
- prepare a unified geometry of the `country_shapes` and `offshore_shapes` of the analyzed countries. I've added an additional buffer of 100km (relatively random number, which in my testing seemed to ensure that onshore and offshore regions were always connected).
- preselect only WDPA geometries, which are inside the bounding box of the country/offshore regions. This is a relatively quick operation.
- from the preselection only select those which intersect with the country/offshore regions. Only unify this selection.

Running the change version for Benin, Nigeria, and Algeria took about 5min rather than ~8h.

To address the RAM issue, I propose the following changes:
- change the bounds of the rasterio `out_shape` to the extend of the unified country/offshore regions.

As long as you are running a model with only a couple countries (which are not placed in opposite corners of the cutout), this should drastically reduce the RAM requirements. My example was feasible on a 16 GB RAM laptop (though I sadly didn't check how much RAM exactly it required).

## Note

I am not that well versed in rasterio and would appreciate if someone could double check, that I didn't break anything by changing the bounds of the `out_shape`.

## Checklist


scripts/build_natura_raster.py
```diff
@@ -53,13 +53,42 @@
...
+    # load country shapes
+    countries_gdf = gpd.read_file(country_shapes).to_crs(natura_crs)
+    countries = countries_gdf.geometry.union_all()
+
+    # load offshore shapes
+    offshore_gdf = gpd.read_file(offshore_shapes).to_crs(natura_crs)
+    offshore = offshore_gdf.geometry.union_all()
+
+    # combine countries and offshore regions into one merged geometry
+    buffer = 1e5  # 100 km, adjust should important regions be missed
```
davide-f: shall this be a parameter to the function or a global variable in the script?

Note: the buffer 99% of the time is likely to work, though I expect issues with countries bordering the +-180°C.
No need to overcomplicate, but it would be good to have better visibility over the value.

Note moreover tha the value is CRS dependent. if it has EPSG:4326 the unit is not km and result can be unexpected

davide-f (commented):
Thanks for the contribution! :D

Indeed the proposal is interesting and can allow to customize the shapes availabilities very precisely without the need to create a global natura file.

Beside the minor comment, as an only "major remark" it would be best to possibly preserve the previous behavior optionally.
Indeed, the global natura tiff can be quite useful: can be created once for all users, uploaded in the bundles to be downloaded and used.

This new version is breaking as it wouldn't allow to preserve the previous behavior that would be quite useful.
As a curiosity: how did you improve the default natura tiff? I'm curious if we could generalize it fur

Eric-Nitschke:
> Beside the minor comment, as an only "major remark" it would be best to possibly preserve the previous behavior optionally. Indeed, the global natura tiff can be quite useful: can be created once for all users, uploaded in the bundles to be downloaded and used.
> 
> This new version is breaking as it wouldn't allow to preserve the previous behavior that would be quite useful. As a curiosity: how did you improve the default natura tiff? I'm curious if we could generalize it fur

Thank you @davide-f for the review!

I agree that I would be good to maintain the current functionality to generate a cutout wide natura file. I think we could maintain this functionality (and even add an option for fully global file) by utilizing a shifting window approach and sequentially generating the rastered file (e.g. a shifting 20°x20° window). The idea would be to:
- prepare the outer bounds based on a flag in the config. Maybe `natura_size: countries # countries, cutout, or global`. Use the old version for the cutout section, the new one for countries and simply [-180,180,-90,90] for global.
- prepare the transformation matrix.
- load the shp files.
- split the bounds into different windows.
- select the geometries wich are inside the window boundaries with `.cx`.
- unify the selected geometries and rasterize them.
- repeat for all windows.

This should still drastically reduce the RAM requirements and prevent the time consuming massive unary union. It does not include the preselection based on whether the geometry is inside the country/offshore region. However I am somewhat unsure anyway if it is a time save on average or not. It would definitely save time for country selection where the countries are quite far apart, but for single countries or countries that are close together it will likely waste some time instead.

Eric-Nitschke:
@davide-f the version I've just pushed includes a shifting window approach and I was able to generate a global natura.tiff in about 3h on a laptop. I've included options to modify the natura raster generation in the config. Importantly you can change how large the shifting window is going to be, so you can utilize a larger shifting window if your hardware allows it.

Be aware that I've not fully tested the code yet and haven't fully compared the results to the previous natura.tiff.

One downside to this implementation is that it deals with the files one after the other, resulting in every byte theoretically being read and changed up to six times. However, I couldn't get any implementation to work that avoids this. Future changes could improve on this.

EDIT:
Apparently my local installation didn't catch the flipped rows and columns.

scripts/build_natura_raster.py
```diff
@@ -190,27 +177,38 @@ def unify_protected_shape_areas(inputs, natura_crs, out_logging):
     inputs = snakemake.input
     cutouts = inputs.cutouts
     shapefiles = get_fileshapes(inputs)
+    country_shapes = inputs.country_shapes
+    offshore_shapes = inputs.offshore_shapes
+    natura_size = snakemake.params.natura["natura_size"]
+    natura_resolution = snakemake.params.natura["natura_resolution"]
+    window_size = snakemake.params.natura["window_size"]
+
+    regions = get_relevant_regions(country_shapes, offshore_shapes, natura_crs)
```
davide-f: This function is only relevant for option countries; it is better calculated inside determine region for the appropriate case most likely.
If we run it for the globe, doing the unary_shape of all shapes can take a while and may not be used

scripts/build_natura_raster.py
```diff
@@ -220,7 +218,64 @@ def unify_protected_shape_areas(inputs, natura_crs, out_logging):
...
-        width=raster.shape[1],
-        height=raster.shape[0],
+        width=out_shape[1],
+        height=out_shape[0],
+        nodata=0,
     ) as dst:
-        dst.write(raster, indexes=1)
+        pass
+
+    with rio.open(snakemake.output[0], "r+") as dst:
```
davide-f, line 256: The functionality looks interesting! needs testing to ensure the previous behavior works.
You mentioend you tested the past approach for Africa; may you compare the past result with the new one? and share the files?
Have you done the comparison?

scripts/build_natura_raster.py
```diff
@@ -220,7 +218,64 @@ def unify_protected_shape_areas(inputs, natura_crs, out_logging):
...
+                        transform=dst.window_transform(window),
+                        fill=0,
+                        dtype="uint8",
+                    )
+
+                    data[mask == 1] = 1
+
+                    dst.write(data, 1, window=window)
+
+                logger.info(f"Done writing rows {row0}-{row1} out of {max_row}")
```
davide-f: Could we share a % value? it may be easier to read.

davide-f (commented):
Many thanks @Eric-Nitschke !
The proposal looks very interesting and see it is even more advanced that initially suggested.
As you mention it needs testing before merging.

Before considering merging, we'd need to ensure the window approach allows to successfully reproduce the previous behavior.
I'm currently trying to have a period off and unlikely I'll have bandwidth before September unfortunately.

The natura.tiff that is automatically downloaded from the bundles is [this CC0 file](https://doi.org/10.7910/DVN/XIV9BL); we have the option to create it from shapes, though for example the protectedplanet ones are not as permissive as CC0 or CC By, which is the reason we opted to move for CC0.
It would be great to test the functionality of the moving window for the region, for example Africa and check if this functionality works as the previous one without the sliding window. That would be awesome to unlock usability on normal computers :)

Eric-Nitschke:
I've now got a functioning version of the code (though it doesn't yet address your comments @davide-f). I've tested it for a selection of countries ("NG", "BJ", "DZ"), the Africa cutout, as well as global. The runtimes were ca. 10min, 20min, 5h30min respectively on my laptop.

The results look relatively similar to the previous dataset, though it is tough to directly compare since some of the data seems to have changed over the last years and the file in `data\natura.tiff` is in a different crs so I had to reproject it. 

This is a section from the old file (reprojected). The fill value here is a large negative number instead of 0, that's why the cbar looks different:
Band_data_base_reprojected

And this is the same section with new code (using "cutout"):
Band_data_cutout

davide-f:
looks promising! how many windows have been used for the test you are sharing?
retesting the case with smaller windows (at least 3-5?) to ensure the result is the same should be enough

Ah, as an additional point, when we add new config parameters to the yaml, there is the need to add the documentation by:
- editing (or creating) the csvs in [this folder](https://github.com/pypsa-meets-earth/pypsa-earth/tree/main/doc/configtables)
- editing the appropriate documentation file (file rst) in https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/doc. In this case https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/doc/configuration.rst

Eric-Nitschke:
Hey @davide-f,
thanks for letting me know about the documentation requirements!

The runtime test were performed with the default `window_size: 10000`, resulting in roughly 18x36=648 windows used by the shifting window (and a minimum RAM requirement of 100MB purely from the window size). Depending on your computer an other `window_size` might be faster.

I used the following code to produce the images above (x and y in meters):
```
ds["band_data"].sel(x=slice(0,1000000),y=slice(1000000,0),band=1).plot()
```
At a resolution of 100 m, this is the same size as one of the shifting windows. So far I've checked a few windows around this area at the same scale. They all looked promissing, but you're right that we probably should check multiple smaller areas all around the world.

scripts/build_natura_raster.py
Eric-Nitschke: I agree!
A progress bar would probably be even nicer, but I am hesitant on adding one, as I've never used one before,

scripts/build_natura_raster.py
Eric-Nitschke: I agree!

scripts/build_natura_raster.py
davide-f: Doing a simple progress bar can be done using:
```
from tqdm import tqdm
for i in tqdm(range(10000),disable=disable_progress):
    ...
```
https://tqdm.github.io/

or more manually using
https://github.com/pypsa-meets-earth/pypsa-asean/blob/7d348d760122c91197b88771be57d6dedc98c21e/scripts/_helpers.py#L439-L446

The latter allows to update for every inner for loop not just the external one: 
by using:
```
pbar = tqdm(total=N, disable=disable_progress)  # outside the forloop with N the total number of iterations

# and inside the forloop:
pbar.n += 1
pbar.refresh()
```

scripts/build_natura_raster.py
davide-f: this number is still hard-coded, could you please comment?
Is it related to the sensitivity you did with 10km? if I understand were you trying to suggest that the value is not extremely important to customize? though the comment says otherwise. Could you clarify?

davide-f (commented):
Hello @Eric-Nitschke , recovering this PR: I'd move it as ready for review to enable notifications and proper visuals.
I've given a round, we can finalize it quite soon :D

scripts/build_natura_raster.py
Eric-Nitschke: Hey @davide-f, sorry I've seemed to have missed the initial comment. I agree that the hard coded buffer isn't ideal. I'd probably move it to the config and pass it to the function as a parameter so you can adjust it easily, should you ever switch to a crs that is not in metres.
(Though as you said it will probably work just fine 99% of the time)

scripts/build_natura_raster.py
Eric-Nitschke: Seems easy enough! I'll implement the latter.

ekatef:
Hey @Eric-Nitschke, many thanks for the great contribution!

As promised, attaching a link to the databundle with `natura.tiff` generated three years ago:

https://github.com/pypsa-meets-earth/pypsa-earth/blob/19b1ce3a20bcf621788395d4b5832fcdc55ef55f/configs/bundle_config.yaml#L194-L202

Given the computational requirements, it has been quite an effort to generate it. The functionality you have developed will be tremendously helpful to have regular updates of `natura.tiff` versions 🙂

Eric-Nitschke:
I've extensively tested the difference between the old 2022 dataset (available in bundle provided by @ekatef) and the new global dataset, generated by my code. I've then also compared the resoluts from specific regions with the [WDPA Interactive Map](https://www.protectedplanet.net/en/thematic-areas/wdpa?tab=WDPA). The code I've used is as follows:
<details>
<summary>Code (can lead to RAM overflow)</summary>

```python
import rasterio as rio
import matplotlib.pyplot as plt
import matplotlib.colors as colors
import matplotlib.patches as mpatches
import xarray as xr
import gc

# adjust filepaths to your datastructure
run_name = ""
fp_old = r"natura.tiff" 
fp_new = rf"resources\{run_name}\natura.tiff"

cmap = colors.ListedColormap(["white", "#A3672C", "#4F88B9", "#733957"])
bounds = [0, 1, 2, 3, 4]
norm = colors.BoundaryNorm(bounds, cmap.N)

da_old = xr.open_dataarray(fp_old, engine="rasterio")
da_new = xr.open_dataarray(fp_new, engine="rasterio")

new_crs = da_new.rio.crs
new_res = da_new.rio.resolution()

da_old.encoding["_FillValue"] = 0

for x in range(-18, 18, 1):

    for y in range(-9,9, 1):

        xmin = x * 1e6
        xmax = (x + 1) * 1e6
        ymin = y * 1e6
        ymax = (y + 1) * 1e6

        conv = 111_000 # m/°

        xmin_old = xmin / conv * 0.8 if xmin > 0 else xmin / conv * 1.2
        xmax_old = xmax / conv * 1.2 if xmax > 0 else xmax / conv * 0.8
        ymin_old = ymin / conv * 0.8 if ymin > 0 else ymin / conv * 1.2
        ymax_old = ymax / conv * 1.2 if ymax > 0 else ymax / conv * 0.8

        x_slice = slice(xmin, xmax)
        y_slice = slice(ymax, ymin)
        x_slice_old = slice(xmin_old, xmax_old)
        y_slice_old = slice(ymax_old, ymin_old)

        da_old_pre = da_old.sel(x=x_slice_old, y=y_slice_old)
        da_new_sel = da_new.sel(x=x_slice, y=y_slice)
        da_old_sel = da_old_pre.rio.reproject_match(da_new_sel)

        da_new_sel = da_new_sel.fillna(0)
        da_old_sel = da_old_sel.fillna(0)

        combined = da_new_sel+2*da_old_sel

        fig, ax = plt.subplots()
        combined.plot(ax=ax,cmap=cmap, norm=norm, add_colorbar=False)

        handles = [
            mpatches.Patch(color="#A3672C", label="New only"),
            mpatches.Patch(color="#4F88B9", label="Old only"),
            mpatches.Patch(color="#733957", label="Both"),
        ]
        ax.legend(handles=handles, loc="center left", bbox_to_anchor=(1, 0.5))
        ax.set_title(None)
        ax.set_xlabel("x in meters")
        ax.set_ylabel("y in meters")

        fig.savefig(f"natura_comp_x{x}_y{y}")
        fig.clf()
        plt.close(fig)

        del da_old_pre, da_new_sel, da_old_sel, combined

        gc.collect()
```

</details>

There are a couple things I've noticed:
- the old dataset is in ESPG:4326 and the new one in `config["crs"]["area_crs"]` which is ESRI:54009
- the old dataset is in ~ 0.0417° * 0.0417° (8640 * 4320 raster) while the new one has a resolution of `config["natura"]["natura_resolution"]` which is 100m * 100m (360802 * 180402 raster). I took this resolution from the previous `build_natura_raster.py` resolution, though I am not sure, if it really necessary. Solving time wise, it was fine since my 16 GB RAM laptop could generate the global file in about 5h.
- in some areas the protected areas are nearly identical between the two datasets with only minor additions such as here in Brazil, Colombia, and Venezuela:
natura_comp_BR_CO_VE
natura_comp_BR

- some regions which largerly match, but where some larger areas are new, such as here in central/western Europe, where a lot of areas in the south east of Germany, and in northern Italy are new:
natura_comp_AT_CH_DE_FR_IT

- whats especially visible in the example above, is that the old version doesnt seem to include larger offshore areas. This is true all around the world, such as here on the eastern coast of Australia:
natura_comp_AU

- there are also multiple large onshore regions which are not marked as protected in the old version, but should be, such as here in Japan and Korea (first image) and Russia (Irkutsk Oblast and Krasnoyarsk Krai, second image):
natura_comp_JP_KR
natura_comp_RU

- rarely, there are also regions which are marked as protected in the old version, but aren't protected anymore, such as here in India:
natura_comp_CN_IN_NP
natura_comp_IN


However, everywhere I checked the new dataset matches the [WDPA Interactive Map](https://www.protectedplanet.net/en/thematic-areas/wdpa?tab=WDPA). As a result I believe that:
- this PR works as intended and can be merged
- the default natura.tiff is vastly outdated and should be updated. I'll create an issue for this.

ekatef:
Hey @Eric-Nitschke, that is amazing! 
~Could you please share a link to the updated `natura.tiff` to update the datakits for retrieve scripts?~

Update: it's unfortunately not possible to re-host data obtained using WDPA data due to licensing restrictions:

> No commercial use
> no sub-licensing or redistribution on WDPA/OECM database data

Though, I think it also means that it's precious to have a code to generate the custom `natura.tiff` locally. 

I'm not touching the code since you are working with @davide-f on that 🙂

ekatef:
@davide-f when you would have time, it would be great to have your review of the revised code. 

As a very technical comment, not sure if you have had a chance to notice it as github seems to be a bit specifis. In case, the answer on your comment can be if you'd click on the commented part of the code (which opens [this link](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1545/files/242e757f15159d2e3f41f0bdd3e02fe299adc7b3#diff-444250aaa09fc861c1ec64471eddb69469e0c39d26b819ac5ef6ea87c5eae0a1))

davide-f:
Yeah, I'll do in the weekend

ekatef:
> Yeah, I'll do in the weekend

Thank you so much. Sorry, didn't mean to put pressure on you

davide-f (approved):
The PR is good to go! Many thanks @Eric-Nitschke !!!
I'll just remove the config.natura.yaml as it is unused; build_natura_tiff is used in the base config.tutorial case.
As CI passes (at least linux), the PR is squash merged :D

Issue #1505, opened 2025-07-08
Clustering in Egypt results in high transmission capacities


jome1:
## Checklist

- [x] I am using the current [`main`](https://github.com/pypsa-meets-earth/pypsa-earth/tree/main) branch or the latest [release](https://github.com/pypsa-meets-earth/pypsa-earth/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-earth` environment](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

The OSM data on power system elements is quite accurate for Egypt. However, when I am clustering the transmission lines (in this case according to GADM1 zones) I get huge transmission capacities which are way to big. 
See the picture below (just a quick and dirty plot):
 
<img width="634" height="636" alt="Image" src="https://github.com/user-attachments/assets/ae1305b6-461c-4561-9a75-1fe14ffb8c7e" />

I am using the config.yaml which I put in [this repo](https://github.com/jome1/share-files.git). I turned off the simplification of shapes, because this led to "holes" in the past (see [Issue 1051](https://github.com/pypsa-meets-earth/pypsa-earth/issues/1051))
Once the databundle is downloaded (turned off in the config), I created the clustered network with `snakemake -j1 networks/test_gadm/elec_s_7_ec_lcopt_Co2L-3h`

ekatef:
Hello @jome1, and many thanks for reporting!

Some overestimation has been noticed previously for the transmission capacities in a few regions for `elec.nc`, but we haven't been aware that there may be also issues linked with clustering. Does the issue happen only also with Voronoi clustering (`alternative_clustering: false`)?

jome1:
This is what I receive when using voronoi clustering (not really sure why I got 8 regions now instead of 7). There are still transmission capacities of 30-60 GW:

<img width="634" height="636" alt="Image" src="https://github.com/user-attachments/assets/9bd9a182-c5dc-47cc-b597-9e5597389db4" />

ekatef:
> This is what I receive when using voronoi clustering (not really sure why I got 8 regions now instead of 7). There are still transmission capacities of 30-60 GW:

Thank you so much! It's very helpful to have as detailed feedback as possible to understand what is going wrong and find a way to fix that.

jome1:
[...]

PR #1552, opened 2025-08-14
fix/improvement clean_osm_data.py

jome1:
**problem**: transmission capacities between regions are way to high for the case of Egypt (but this would probably apply to many other countries as well). see #1505

**reason**: While the raw OSM data is often pretty good, the script clean_osm_data.py introduces erros by not considering the circuits value in the raw data or incorrectly calculating the number of circuits based on the number of cables in the raw data. In the case of Egypt, this lead to too many circuits on many power lines in the cleaned OSM data. 

**fix**: read length of cables and circuits from raw data not just when they are stored as strings, but also when they are stored as floats; add "50.0" as possible frequency in dict cables_req (see #1505)

before            |  after fix
:-------------------------:|:-------------------------:
image |  image


jome1:
sorry, I dont know how to get the actual PR template

scripts/clean_osm_data.py
```diff
@@ -557,12 +565,7 @@ def _parse_float(x, ret_def=0.0):
             return ret_def
 
     # cables requirement for circuits calculation
-    cables_req = {
-        "50": 3,
-        "60": 3,
-        "16.7": 2,
-        "0": 2,
-    }
+    cables_req = {"50": 3, "60": 3, "16.7": 2, "0": 2, "50.0": 3}
```
ljansen-iee: I suggest to use the cleanup function "clean_frequency" instead and extend it by adding `"50.0": "50", `to the repl_freq dict in the clean_frequency function. This would give the same result but separate the cleanup of freq and lookup of cables_req, right? What do you think?

scripts/clean_osm_data.py
```diff
@@ -541,11 +541,19 @@ def fill_circuits(df):
     def _get_circuits_status(df):
         len_f = df["tag_frequency"].map(len)
         len_c = df["circuits"].map(
-            lambda x: x.count(";") + 1 if isinstance(x, str) else np.nan
+            lambda x: (
+                x.count(";") + 1
+                if isinstance(x, str)
+                else 1 if isinstance(x, float) and not pd.isna(x) else np.nan
```
ljansen-iee: I don't fully understand the 1 if float case, but this also looks to me like circuits requires more cleaning in clean_circuits.

scripts/clean_osm_data.py
```diff
@@ -541,11 +541,19 @@ def fill_circuits(df):
...
+            )
         )
         isna_c = df["circuits"].isna()
         len_cab = df["cables"].map(
-            lambda x: x.count(";") + 1 if isinstance(x, str) else np.nan
+            lambda x: (
+                x.count(";") + 1
+                if isinstance(x, str)
+                else 1 if isinstance(x, float) and not pd.isna(x) else np.nan
```
ljansen-iee: same suggestion here: extend clean_cables instead? :)

ljansen-iee (changes requested):
Hi @jome1, thank you for the PR and for identifying this issue.
I think we should extend the cleaning functions (clean_cables, …) to address it.
I’ve left some comments on this, but i guess @davide-f @ekatef  should also take a look. 🙂

scripts/clean_osm_data.py
ljansen-iee, line 1: Slightly off-topic: 
@davide-f, I just realised that repl_cables is not used in clean_cables. [clean_osm_data.py#L424-L448](https://github.com/pypsa-meets-earth/pypsa-earth/blob/a24a99cc3eb16f356ad758ed4ebb70f036c529e6/scripts/clean_osm_data.py#L424-L448)

If it's not on purpose, then the code could be: 
```
    df["cables"] = (
        df["cables"]
        .replace(repl_cables)
        .map(lambda x: x.replace(" ", "") if isinstance(x, str) else x)
    )
```

scripts/clean_osm_data.py
jome1: if the circuits value is stored as a float in the raw OSM data, then this is recognized and I assumed that always only one float value can be stored (length = 1), whereas if the circuits value was a string, it could contain multiple number seperated by semicolon.
We could convert a float into a string in clean_circuits?

scripts/clean_osm_data.py
jome1: see above

scripts/clean_osm_data.py
jome1: that makes a lot of sense! I didnt consider the cleanup functions at all when I was trying to find the error.

jome1: I just commited the change to put `"50.0": "50",` into "clean_frequency", ran a test and it worked

scripts/clean_osm_data.py
davide-f: interesting as in my test cases I never experienced the column being a float; the behavior may also change across different pandas versions (we have changed some too...).
also isinstance( float) may work for 1.0 but does not catch int and may not catch values like np-based types pandas usually rely on.
The code assumed the column to be either string or nan/null.

The 1 in case of numeric value makes sense as it is equivalent to no ";" and hence one circuit only, but I'm concerned about the type-dependent case. it may be safer to ensure the columns are string typed or nans

scripts/clean_osm_data.py
davide-f: not sure why it wasn't considered; I've seen @jome1 included it; this comment may be resolved

davide-f (commented):
@jome1 really really nice catch!
I think this PR has high priority and I'm looking forward to testing for further regions.
We can work together also at improving the heuristics on the parsing of the lines further. In transparency, I'm having some vacation next week and coming back in around a week if that works for you.
I see you are working a lot with OSM data and it would be great to have practical examples of lines that are improperly converted as the one you found.

Regarding the PR and activities, also in agreement with urgency, other maintainers ensure continuity in case

scripts/clean_osm_data.py
jome1: In this PR it is still missing the .replace() function. I can add it so the code is like @ljansen-iee suggested.

scripts/clean_osm_data.py
jome1: What do you recommend? 

There is the comment of "note: no string conversion here! it is performed later on" within the [clean_circuits function](https://github.com/pypsa-meets-earth/pypsa-earth/blob/16bf40b8b5396a59df968e352561fa9bd0a7b126/scripts/clean_osm_data.py#L388-L416)
```
    # note: no string conversion here! it is performed later on
    df["circuits"] = (
        df["circuits"]
        .replace(repl_circuits)
        .map(lambda x: x.replace(" ", "") if isinstance(x, str) else x)
    )
```

But within `fill_circtuits`, the [string conversion](https://github.com/pypsa-meets-earth/pypsa-earth/blob/16bf40b8b5396a59df968e352561fa9bd0a7b126/scripts/clean_osm_data.py#L639) happens after getting the length of circuits value in [_get_circuits_status](https://github.com/pypsa-meets-earth/pypsa-earth/blob/16bf40b8b5396a59df968e352561fa9bd0a7b126/scripts/clean_osm_data.py#L541)


I tested to convert cables and circuits columns into strings in `clean_cables` and `clean_circuits`, but then some lines got dropped/where deleted and not there anymore in all_clean_lines.geojson

jome1:
Hi @davide-f ,
thanks for reviewing!
I definitely want the clean_osm issue to be resolved because that would help me a lot with modelling Egypt. 
But my OSM data experience is also limited. Let me know what exactly you need as examples. 
I think all countries must face issue with the OSM cleaning part, because it seems to me that the string and float handling of cables and circuits is a structural problem in the code. Maybe it was different in previous version, but could it be the case that geopandas is turning some values into numbers automatically? Maybe values are stored as strings on OSM, but clean_osm_data.py is parsing the raw geojson files and not directly from OSM.

scripts/clean_osm_data.py
ljansen-iee: Hi, I created a PR from my branch to your branch @jome1 and added this line. :)

ljansen-iee:
I agree that this is an important PR. I have tested geopandas versions: There is no difference between geopandas 1.0 and 1.1. But I also suspect that it is due to a change in versions or raw data.

I’ve made some direct code suggestions to your branch @jome1, to improve numeric data type handling and introduce warnings for non-numeric values in the clean_circuits, and clean_cables functions:

- Added a conditional conversion of numbers (across different dtypes) to strings while preserving NaN and other string entries. As you noted, we should not convert all entries to strings because of the complex filtering logic later on. These conditional conversions in the clean_x functions might work — could you please test them? I only did a brief check and they seemed to work.
- Added a logger warning if any non-numeric and non-NaN values remain.

It would be great if you would take a look and also test them.

scripts/clean_osm_data.py
ljansen-iee: I suggest to add these lines to only convert numeric-like values to string. I’ve made a direct code suggestions to your branch @jome1. It would be greate if you can you test if this prevents the that lines get dropped? 

```
# note: no string conversion for all entries in clean_circuits! it is performed later on
df["circuits"] = (
    df["circuits"]
    .replace(repl_circuits)
    .map(lambda x: x.replace(" ", "") if isinstance(x, str) else x)
)

# Convert numbers in different dtypes to string while preserving NaN or other strings.
is_numeric = ~pd.to_numeric(df["circuits"], errors="coerce").isna()
df["circuits"] = df["circuits"].where(~is_numeric, df["circuits"].astype(str))

    # Report non-numeric and non-NaN values, which should be added to repl_circuits.
if df.loc[~is_numeric, "circuits"].notna().any():
    logger.warning(
        "Non-numeric and non-NaN values found in circuits column, consider replacement:",
        df.loc[~is_numeric, "circuits"].unique())
```

ekatef:
> Hi @davide-f , thanks for reviewing! I definitely want the clean_osm issue to be resolved because that would help me a lot with modelling Egypt. But my OSM data experience is also limited. Let me know what exactly you need as examples. I think all countries must face issue with the OSM cleaning part, because it seems to me that the string and float handling of cables and circuits is a structural problem in the code. Maybe it was different in previous version, but could it be the case that geopandas is turning some values into numbers automatically? Maybe values are stored as strings on OSM, but clean_osm_data.py is parsing the raw geojson files and not directly from OSM.

Hey @jome1, hey @ljansen-iee , many thanks for investigating the network part! That is indeed a part which is very high in the priority list. I'll try to provide some answers until Davide is back 🙂 

The relevant examples are those where the current approach is not working as expected. In particular, it's very helpful to have the results of your investigation placed in [this comment](https://github.com/pypsa-meets-earth/pypsa-earth/issues/1505#issuecomment-3179263130) for which the current method needs some adjustments.

Just to clarify, the current approach can be illustrated with the following scheme (also described in the docstrings of `clean_osm_data`):
image

The code relays on some relationships assumed for frequency and a number of circuits/cables (those described in [the docstring](https://github.com/pypsa-meets-earth/pypsa-earth/blob/16bf40b8b5396a59df968e352561fa9bd0a7b126/scripts/clean_osm_data.py#L516-L538) ) but validity of those assumptions may differ from country to country, and there may be also some special cases in data availability and formats.

I completely agree that it makes absolute sense to fix types of the OSM-extracted data columns, as you have just done. But there is also a more general question: what are we missing in the current implementation for every country of the world, and how can we improve the underlaying heuristics? Having evidence on that is crucial to increase the accuracy of the method. Once we'll understand which issues can be expected from the data we'll be able to tackle them in a more effective way.

For this particular PR, have you had an opportunity to test it? Does it resolve the issues experienced for Egypt?

jome1:
Thanks for your help @ekatef! 

Thanks a lot @ljansen-iee for your code suggestions! Very smart approach! I tested it with Egypt and Benin and it works! For Egypt I am getting the same results as in the [first post](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1552#issue-3323466169) of this thread. There are still some high transmission capacities in the north of Egypt between regions, but those are not as high as before (but I also dont have the exact real world data to compare). But clean and raw OSM data seem to me going hand in hand now. 


Which script and method are used to get from circuits and voltage of a line to its capacity? 


I consider the issue regarding the faulty reading of strings and numeric values as solved :)

I tried to explain the code behaviour in #1505. But in my view, this is resolved now. 
Looking forward for your opinion and tests once you are back @davide-f!

In case you have more ideas/questions on parsing the power lines, let me know :)

ekatef:
> Thanks for your help @ekatef!
> 
> Thanks a lot @ljansen-iee for your code suggestions! Very smart approach! I tested it with Egypt and Benin and it works! For Egypt I am getting the same results as in the [first post](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1552#issue-3323466169) of this thread. There are still some high transmission capacities in the north of Egypt between regions, but those are not as high as before (but I also dont have the exact real world data to compare). But clean and raw OSM data seem to me going hand in hand now.
> 
> Which script and method are used to get from circuits and voltage of a line to its capacity?
> 
> I consider the issue regarding the faulty reading of strings and numeric values as solved :)
> 
> I tried to explain the code behaviour in #1505. But in my view, this is resolved now. Looking forward for your opinion and tests once you are back @davide-f!
> 
> In case you have more ideas/questions on parsing the power lines, let me know :)

Hey @jome1, happy to hear that the discussion has been helpful to resolve some of the issues. A fantastic team work with @ljansen-iee 🙂 

Do you mind to share the updated pictures for Egypt? Some colleagues may have regional expertise and could help to assess accuracy of the resulted transmission capacities.

jome1:
That would be great to have an experts judgement on the capacities!

It is the same as in the first post ("after fix"):
image 

I am also having an interactive plot on my computer to plot the capacities of all single lines before clustering, but that I cannot really share it in a comment here.

ekatef:
> That would be great to have an experts judgement on the capacities!
> 
> It is the same as in the first post ("after fix"): image
> 
> I am also having an interactive plot on my computer to plot the capacities of all single lines before clustering, but that I cannot really share it in a comment here.

Many thanks for the clarification and posting! I'll ping the colleagues once they'll be back to the workspace

ekatef:
> Thanks for your help @ekatef!
> 
> Thanks a lot @ljansen-iee for your code suggestions! Very smart approach! I tested it with Egypt and Benin and it works! For Egypt I am getting the same results as in the [first post](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1552#issue-3323466169) of this thread. There are still some high transmission capacities in the north of Egypt between regions, but those are not as high as before (but I also dont have the exact real world data to compare). But clean and raw OSM data seem to me going hand in hand now.
> 
> Which script and method are used to get from circuits and voltage of a line to its capacity?
> 
> I consider the issue regarding the faulty reading of strings and numeric values as solved :)
> 
> I tried to explain the code behaviour in #1505. But in my view, this is resolved now. Looking forward for your opinion and tests once you are back @davide-f!
> 
> In case you have more ideas/questions on parsing the power lines, let me know :)

Hey @jome1, have just recognised that I have missed to answer your question on capacity evaluation. To translate the lines parameters into its' transmission capacity we assume some particular set of line type defined in this part of the config
https://github.com/pypsa-meets-earth/pypsa-earth/blob/ebb803335c45348da6692fde2cddfa2c7c098629/config.default.yaml#L210-L219

The transmission capacities of those standard types are being used in https://github.com/pypsa-meets-earth/pypsa-earth/blob/ebb803335c45348da6692fde2cddfa2c7c098629/scripts/base_network.py#L313 of `base_network` script to set the transmission capacity of the lines.

Does it answer your question?

jome1:
thanks for the clarification! Makes total sense!

ekatef:
@hazemakhalek could you please to have a look at the resulted transmission capacities of the network? An updated picture available in [this](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1552#issuecomment-3211424248) comment.

scripts/clean_osm_data.py
```diff
@@ -557,12 +582,7 @@ def _parse_float(x, ret_def=0.0):
...
     # cables requirement for circuits calculation
-    cables_req = {
-        "50": 3,
-        "60": 3,
-        "16.7": 2,
-        "0": 2,
-    }
+    cables_req = {"50": 3, "60": 3, "16.7": 2, "0": 2}
 
     def _basic_cables(f_val, cables_req=cables_req, def_circ=2):
```
davide-f: Since we are here, could you replace the default value into 3 as most of the lines do have 3 cables?

```suggestion
    def _basic_cables(f_val, cables_req=cables_req, def_circ=3):
```

scripts/clean_osm_data.py
```diff
@@ -342,6 +342,7 @@ def clean_frequency(df, default_frequency="50"):
         "16.67": "16.7",
         "50;50;16.716.7": "50;50;16.7;16.7",
         "50;16.7?": "50;16.7",
+        "50.0": "50",
         # "24 kHz": "24000",
```
davide-f: May it make sense to add also 60.0?
```suggestion
        "50.0": "50",
        "60.0": "60",
        # "24 kHz": "24000",
```

davide-f (commented):
I've been testing this for selected countries: it improves countries such as Egypt and I don't see why it affects others.

A release_note is recommended and added few minor improvements if interested :)
Much needed and valuable contribution!

If you could kindly add the release_note, the PR is merged or if you need support please share it

Many thanks!

ljansen-iee:
Hi all,

One reference that could be interesting: see [EA Energy Analysis (2022)](https://www.ea-energianalyse.dk/wp-content/uploads/2022/09/wind_power_intergration_in_egypt.pdf), Figure 3.5. In that study, Net Transfer Capacities (NTCs) between Egyptian transmission zones were estimated using actual flow data from the transmission company. However, note that the reference years are 2014–2015, and the metric differs.

It could be interesting to compare these values with the capacities used in a PyPSA-Earth model instance for 2025. However, I do not consider this a required analysis for this PR, which is already a clear improvement for me.

jome1:
thanks for pointing to the Ea publication @ljansen-iee !

We are currently using the NTCs from [this report](https://usercontent.one/wp/www.eaedu.dk/wp-content/uploads/2020/03/Integration-of-wind-power-in-the-Egyptian-power-system-2016.pdf) (see. page 58f).
I will ask our Egyptian partners during our next workshop. But they will only make general comments because they are not allowed to share the exact data.

thanks for the small improvements @davide-f!
I added them and the release_notes as well. I hope everything works out with the checks. It is the first time that I have a PR.

@ljansen-iee thanks again for the smart code improvements!

thanks to everyone for this great team effort! 

@ljansen-iee I think you have to accept the review. Please let me know if anything is missing.

ljansen-iee (approved):

ljansen-iee:
Hello, I have approved it, but I think you need to ask someone (.e.g @davide-f or @ekatef ) with write access. I don't have it yet :)

davide-f (approved):

davide-f:
Many thanks! ready to fly :D

Many thanks for reviewing the PR @ljansen-iee ! We really encourage contributions and revisions to grow the model stronger and fast! :D
We recently granted write permissions to long-standing contributors and indeed that is definitely possible :)

Issue #1518, opened 2025-07-24
Fix CO2 limits after the merge of power and sector models


danielelerede-oet:
## Checklist

- [x] I am using the current [`main`](https://github.com/pypsa-meets-earth/pypsa-earth/tree/main) branch or the latest [release](https://github.com/pypsa-meets-earth/pypsa-earth/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-earth` environment](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

Using the current, default CO2 limit field from the config file may be dangerous as that value is tailored on the electricity sector, and anyway quite low for large/largely emitting countries and users may not notice/know that the value has to be changed (especially if they're running the sector model). 

Here's the case of Nigeria when using the default CO2 limit value:

| electricity generation (TWh)     | Power | Sector |
|------------------|----------|----------|
| CCGT             | 18.44    | 0.13     |
| solar            | 15.96    | 36.81    |
| hydro            | 5.22     | 5.23     |
| coal             | 2.05     | 0.00     |
| onwind           | 0.00     | 0.00     |
| biomass          | –        | 183.52   |
| solar rooftop    | –        | 3.09     |

My proposal to overcome this issue is:
1) have a collection of CO2 limit values for e.g., 2020 for all the countries in the world for both the electricity generation / total energy-related emissions, so that the `countries` field is already sufficient to all the correct values
2)  Implement a differentiation between electricity-related CO2 limit and total emissions from energy uses, reducing the users' effort to select the correct value of `CO2limit`

FYI @ekatef @davide-f

ekatef:
@danielelerede-oet many thanks for opening the issue! A great point to address. Completely agree that the current design is not particularly user-friendly in the emissions part.

If I get your idea properly for point 1, you suggest an automated way to extract the emission level for each specific country from an emission database. To me, it sounds like a very convenient approach and I wonder if it may be possible to build it on top the infrastructure we still have in the model.

Currently, we have an option to calculate the emission limit automatically with the following options:
https://github.com/pypsa-meets-earth/pypsa-earth/blob/640f38ae3e26310d6360219766ad2cf4a2785263/config.default.yaml#L161-L162

[...]

PR #1553, opened 2025-08-15
Add `co2_budget` for emission targets in multiple planning horizon years

virio-andreyana:

When performing **myopic runs**, using the `Co2L` wildcards becomes impractical due to the presence of multiple planning horizon years. To address this, we utilize an existing feature in pypsa-eur: `co2_budget`

We've aimed to make this feature as flexible as possible. Users can choose from several options:

* Base the annual CO₂ budget on the existing `co2base` or `co2limit` values,
* Define a custom value and scale subsequent years relative to it,
* Or assign an **absolute value** for each specific year.

The `overwrite` option serves as a safeguard, informing users when an existing CO₂ limit (set earlier during `prepare_network`) will be overwritten.

```yaml
co2_budget:
  enable: false
  overwrite: true
  compare: co2limit # choose from: [co2limit, co2base, absolute, {float}]
  year:
    2020: 1.0
    2025: 0.85
    2030: 0.70
    2035: 0.55
    2040: 0.40
    2045: 0.25
    2050: 0.1
```

## Checklist

### DIFF
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -153,8 +153,8 @@ load_options:
 
 co2_budget:
   enable: false
-  overwrite: true
-  compare: co2limit # choose from: [co2limit, co2base, absolute, {float}]
+  override_co2opt: true
+  co2base_value: co2limit # choose from: [co2limit, co2base, absolute, {float}]
   year:
     2020: 1.0
     2025: 0.85
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -151,6 +151,19 @@ load_options:
   prediction_year: 2030 # Load scenarios available with different prediction year (GDP, population)
   scale: 1 # scales all load time-series, i.e. 2 = doubles load
 
+co2_budget:
+  enable: false
+  overwrite: true
+  compare: co2limit # choose from: [co2limit, co2base, absolute, {float}]
+  year:
+    2020: 1.0
+    2025: 0.85
+    2030: 0.70
+    2035: 0.55
+    2040: 0.40
+    2045: 0.25
+    2050: 0.1
+
 electricity:
   base_voltage: 380.
   voltages: [132., 220., 300., 380., 500., 750.]
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -2949,7 +2949,9 @@ def add_co2_budget(n, co2_budget, investment_year, elec_opts):
         raise ValueError(f"{compare} is not a valid option for co2_budget: compare")
 
     Nyears = n.snapshot_weightings.objective.sum() / 8760.0
-    logger.info(f"Annual emissions for {investment_year} set to {annual_emissions / 1e6:.2f} MtCO₂-eq/year")
+    logger.info(
+        f"Annual emissions for {investment_year} set to {annual_emissions / 1e6:.2f} MtCO₂-eq/year"
+    )
 
     add_co2limit(n, annual_emissions, Nyears)
 
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -28,6 +28,7 @@ from _helpers import (
     three_2_two_digits_country,
     two_2_three_digits_country,
 )
+from prepare_network import add_co2limit
 from prepare_transport_data import prepare_transport_data
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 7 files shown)


config.default.yaml
```diff
@@ -151,6 +151,19 @@ load_options:
   prediction_year: 2030 # Load scenarios available with different prediction year (GDP, population)
   scale: 1 # scales all load time-series, i.e. 2 = doubles load
 
+co2_budget:
+  enable: false
+  overwrite: true
+  compare: co2limit # choose from: [co2limit, co2base, absolute, {float}]
```
ekatef: I love the concept with `overwrite` flag. There is a feeling though that we may want to revise the naming taking the perspective of model user. [It's a completely normal step when working on interfaces since you obviously have a developers perspective when you are deep in the code. The trick is making sure that the configuration parameters are clear also for a user who didn't look into the code]

I see the following issues with the current naming:
- for `overwrite`: it's not obvious what exactly is being overwritten
- `compare` may give an impression that it's some kind of boolean flag while it's used rather to determine a method

Do you have any ideas on possible revision? As a possible next iteration I could suggest to revise `overwrite` -> `adjust_co2constraint` / `adjust_co2limit` / `override_co2`, `compare` -> `co2adjustment_method`, `co2limit_method`, `co2override_method`

config.default.yaml
```diff
@@ -151,6 +151,19 @@ load_options:
   prediction_year: 2030 # Load scenarios available with different prediction year (GDP, population)
   scale: 1 # scales all load time-series, i.e. 2 = doubles load
 
+co2_budget:
```
ekatef, line 154: Could you please document those parameters adding a config table to [the existing ones](https://github.com/pypsa-meets-earth/pypsa-earth/tree/main/doc/configtables)?

scripts/prepare_sector_network.py
```diff
@@ -2917,27 +2918,42 @@ def add_electricity_distribution_grid(n, costs):
...
-#     sectors = emission_sectors_from_opts(opts)
-
-#     # convert Mt to tCO2
-#     co2_totals = 1e6 * pd.read_csv(snakemake.input.co2_totals_name, index_col=0)
-
-#     co2_limit = co2_totals.loc[countries, sectors].sum().sum()
+def add_co2_budget(n, co2_budget, investment_year, elec_opts):
+    # Check if CO2Limit already exists
+    if "CO2Limit" in n.global_constraints.index:
+        if co2_budget["overwrite"]:
```
ekatef: I wonder if it would increase readability if nested `if`-s would be replaced by `and` condition? According to [pythonic zen](https://en.wikipedia.org/wiki/Zen_of_Python) `Flat is better than nested`

scripts/prepare_sector_network.py
```diff
@@ -2917,27 +2918,42 @@ def add_electricity_distribution_grid(n, costs):
...
+            logger.info("CO2Limit already exists, value will be overwritten.")
+            n.global_constraints.drop(index="CO2Limit", inplace=True)
+        else:
+            logger.warning("CO2Limit already exists, value will not be overwritten.")
```
ekatef: It feels a bit more natural to swap the info and the warning. If the constrain is being overwritten, we issue a warning while if nothing is to be done an info should be enough. What do you think?

scripts/prepare_sector_network.py
```diff
@@ -2917,27 +2918,42 @@ def add_electricity_distribution_grid(n, costs):
...
+            return
+
+    # Get base year emission factor
+    try:
+        year_emission = co2_budget["year"][investment_year]
+    except KeyError:
+        raise KeyError(
+            f"Investment year '{investment_year}' not found in co2_budget: year"
+        )
```
ekatef: Not sure about the need for `try` block as defining `investment_year` parameter is a fully controllable factor. Can we replace `try-except` with a simple `if` block and simply define a default `investment_year` if it's not available?

scripts/prepare_sector_network.py
```diff
@@ -2917,27 +2918,42 @@ def add_electricity_distribution_grid(n, costs):
...
+
+    compare = co2_budget["compare"]
+    if compare == "co2limit":
+        annual_emissions = year_emission * elec_opts["co2limit"]
```
ekatef: `year_emission` seems to be a factor used to modify a value of CO2 emission. Could it be a good idea to revise the variable name to make it more clear?

scripts/prepare_sector_network.py
```diff
@@ -2917,27 +2918,42 @@ def add_electricity_distribution_grid(n, costs):
...
+        if co2_budget["overwrite"]:
+            logger.info("CO2Limit already exists, value will be overwritten.")
+            n.global_constraints.drop(index="CO2Limit", inplace=True)
+        else:
+            logger.warning("CO2Limit already exists, value will not be overwritten.")
+            return
+
+    # Get base year emission factor
+    try:
+        year_emission = co2_budget["year"][investment_year]
```
ekatef: What would happen if `investment_year` is not contained in `co2_budget:year` collection?

ekatef (commented):
Hey @virio-andreyana, a cool idea and very interesting implementation! Have added some comments and questions. Could you please have a look?

config.default.yaml
virio-andreyana: Thanks for the suggestion Katia

- I think for `overwrite` -> `override_co2opts`. This is what is meant by this options
- For `compare` -> I would say `co2base_value`. The main point is selecting which value to have all of the year factor to be compared with.

What do you think? If you agree, I'll add documentations in the next commit

scripts/prepare_sector_network.py
virio-andreyana: It could be a factor (for `co2limit`, `co2base`, `{float}`) or the year emission if `absolute` is selected. This is to accommodate if certain countries have well determined emission targets. Maybe l'll rename it to `factor` since that is the answer most of the time

scripts/prepare_sector_network.py
virio-andreyana: maybe if the `investment_year` is not in the list, the value will be set back to the currently named `compare` value. I can try solving this without using `try-except`

config.default.yaml
ekatef: I think it's a great idea 🙂 Resolving this

scripts/prepare_sector_network.py
ekatef: I think you have found a very elegant solution. Thanks for the revision!

scripts/prepare_sector_network.py
ekatef: I think now it's neat and clear

ekatef (approved):
Hey @virio-andreyana many thanks for the revisions! I think it looks perfect now.

ekatef:
On my side, happy to merge once CI will be completed.

@danielelerede-oet as discussed, the PR may be the first step to fix the issue #1518 you raised. @yerbol-akhmetov It may be also relevant to make more convenient using `myopic` option.

PR #1565, opened 2025-09-03
Documenting and moving `focus_weights` to `cluster_options`

virio-andreyana:

``focus_weights`` is undocumented because it functions as a standalone configuration. When not explicitly used, it is simply omitted from ``config.default.yaml``. Similar to the approach taken in PyPSA-Eur, we need to relocate ``focus_weights`` to a more appropriate section.

> The configuration setting for ``focus_weights`` has been moved from ``focus_weights:`` to ``cluster_options: focus_weights:``. Backwards compatibility to old config files is maintained.

## Checklist

### DIFF
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -104,7 +104,7 @@ cluster_options:
       ramp_limit_up: max
       ramp_limit_down: max
       efficiency: mean
-  focus_weights: false # When a value specified, set the share of nodes allocated to each country
+  focus_weights: null # When a value specified, set the share of nodes allocated to each country
     # country: share
 
 build_shape_options:
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -104,7 +104,7 @@ cluster_options:
       ramp_limit_up: max
       ramp_limit_down: max
       efficiency: mean
-  focus_weights:      # When a value specified, set the share of nodes allocated to each country
+  focus_weights: false # When a value specified, set the share of nodes allocated to each country
     # country: share
 
 build_shape_options:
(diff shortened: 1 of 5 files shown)


virio-andreyana:
Ready to merge @ekatef

config.default.yaml
```diff
@@ -104,6 +104,8 @@ cluster_options:
       ramp_limit_up: max
       ramp_limit_down: max
       efficiency: mean
+  focus_weights: # When specified, set the share of nodes allocated to each country
```
ekatef: Can we specify `false` value in the config? It's not effecting functionality just may be more readable and more consistent with other parameters

ekatef (commented):
Hey @virio-andreyana, looks a great catch and very nice implementation. There is only a very minor comment which we may want to accommodate. Could you please have a look?

Otherwise, agree that it's ready to be merged 😄

config.default.yaml
ekatef:
```suggestion
  focus_weights: false # When a value specified, set the share of nodes allocated to each country
```

ekatef: Hey @virio-andreyana, have added a suggestion. If you are alright with applying it happy to proceed with merging 🙂

virio-andreyana: Done! Thank's for the suggestion

ekatef: Thank you! 😄

ekatef (approved):

virio-andreyana:
Oh no. the pre-commit wouldn't let me use null in yaml. But False is not None. Guess I have to modify the content to achieve this @ekatef

ekatef:
Many thanks @virio-andreyana! A great contribution and a nice way to treat defaults for `focus_weights`.

PR #1586, opened 2025-09-26
Add support for custom OSM data

GbotemiB:
# Closes # (if applicable).
This PR adds the capability to use custom OpenStreetMap (OSM) data instead of downloading it automatically from the OSM servers. This feature is useful for users who want to use pre-processed or specific OSM datasets.

### Usage

To use custom OSM data:
1. Set `custom_data.osm_data: true` in your configuration file
2. Place your custom OSM files (PBF and power files) in the `data/custom/osm` folder
3. Run the workflow - it will use your custom data instead of downloading

### Configuration Files
- **config.default.yaml**: Added `osm_data: false` option under `custom_data` section
- **config.tutorial.yaml**: Added `osm_data: false` option under `custom_data` section

### Scripts
- **scripts/download_osm_data.py**: 
  - Enhanced to check for custom OSM data configuration
  - Added logic to copy custom OSM data from `data/custom/osm` folder when enabled
  - Skips OSM download when custom data is used

### Benefits

- Allows users to work with specific OSM datasets
- Enables offline workflows when OSM download is not possible
- Supports reproducible research with fixed OSM data snapshots
- Useful for testing with smaller, controlled datasets

## Checklist


ekatef:
Many thanks @GbotemiB! Currently, I'm testing this functionality and should be able to provide some meaningful feedback soon 🙂

Snakefile
```diff
@@ -168,6 +168,9 @@ if config["enable"].get("download_osm_data", True):
     rule download_osm_data:
         params:
             countries=config["countries"],
+            store_path_resources="resources/" + RDIR + "osm/raw",
+            store_path_data="data/osm",
+            custom_data_path="data/custom/osm",
```
ekatef, line 173: I wonder if it may be a good idea to place custom data in a dedicated folder for a run? Theoretically, `data` folder itself must be used for globally-relevant datasets only (e.g. landcover). 

I know that we are currently violating that for quite a few data groups. But that is a point which we may need to revise in course of the data management improvements and avoid making it more complicated for now. What do you think?

scripts/download_osm_data.py
```diff
@@ -99,11 +99,19 @@ def convert_iso_to_geofk(
...
+    custom_data = snakemake.config.get("custom_data", {}).get("osm_data")
+    custom_data_path = Path(snakemake.params.get("custom_data_path", "data/custom/osm"))
+
+    # allow for custom data usage
+    if custom_data:
+        logger.info(
+            "Custom OSM data usage is activated. Skipping the download of OSM data."
+        )
+        os.makedirs(store_path_data, exist_ok=True)
+        shutil.copytree(custom_data_path, store_path_data, dirs_exist_ok=True)
```
ekatef, line 114: When testing, I have found that I need to amend `pbf` directory to `store_path_data`. It looks like that is what `earth-osm` is doing when downloading data, though you may have a more detailed understanding of the logic

ekatef (commented):
Hello @GbotemiB many thanks for developing this functionality! That makes life much easier when investigating an effect of different editions of OSM datasets. I hope that we may be able in future to use different kinds of enhanced OSM datasets and we may need to give some thoughts to the design. 

For now, I think the goal of this functionality is to be able paste versions of OSM data downloaded manually from geofabrik data server for a country of interest . I have tested the implementation with the raw datasets extracted from [https://download.geofabrik.de/south-america/](https://download.geofabrik.de/south-america/) for Bolivia and have found that there are some conditions which data in `custom/osm` folder must satisfy to be used by the workflow:
1) the names must be {country}-latest.osm.pbf and {country}-latest.osm.pbf.md5
2) md5 hash must match to the file content [haven't been able to find out from where `earth-osm` is downloading the reference md5 file; you have a better overview here thanks to your recent contributions into `earth-osm`]
3) a timestamp of the hash file must be later as compared to the pbf file timestamp.

If any of those conditions isn't satisfied, `earth-osm` just downloads the latest data from geofabrik server and they are being used instead of the custom ones. A work-around solution is to download a pbf file manually and tweak it in a way it matches expectations of `earth-osm` and can be actually used. It may be a perfectly feasible short-term solution, while your contribution to `earth-osm` can give a long-term one.

On the operative level, it would be good to make sure that we can keep several versions of OSM datasets on the same time. But that is a rather narrow question which is more relevant for a more focused investigation of OSM role in the energy modelling. I'd suggest to move the further discussion for the current development stage to `pypsa-earth-osm` repo. What do you think?

Snakefile
GbotemiB: Yeah, you are right. I will consider a different approach on how to store the data.

scripts/download_osm_data.py
GbotemiB: This is correct. The pbf and power folders are created if it doesnt exist

GbotemiB:
> Hello @GbotemiB many thanks for developing this functionality! That makes life much easier when investigating an effect of different editions of OSM datasets. I hope that we may be able in future to use different kinds of enhanced OSM datasets and we may need to give some thoughts to the design.
> 
> For now, I think the goal of this functionality is to be able paste versions of OSM data downloaded manually from geofabrik data server for a country of interest . I have tested the implementation with the raw datasets extracted from https://download.geofabrik.de/south-america/ for Bolivia and have found that there are some conditions which data in `custom/osm` folder must satisfy to be used by the workflow:
> 
> 1. the names must be {country}-latest.osm.pbf and {country}-latest.osm.pbf.md5
> 2. md5 hash must match to the file content [haven't been able to find out from where `earth-osm` is downloading the reference md5 file; you have a better overview here thanks to your recent contributions into `earth-osm`]
> 3. a timestamp of the hash file must be later as compared to the pbf file timestamp.
> 
> If any of those conditions isn't satisfied, `earth-osm` just downloads the latest data from geofabrik server and they are being used instead of the custom ones. A work-around solution is to download a pbf file manually and tweak it in a way it matches expectations of `earth-osm` and can be actually used. It may be a perfectly feasible short-term solution, while your contribution to `earth-osm` can give a long-term one.
> 
> On the operative level, it would be good to make sure that we can keep several versions of OSM datasets on the same time. But that is a rather narrow question which is more relevant for a more focused investigation of OSM role in the energy modelling. I'd suggest to move the further discussion for the current development stage to `pypsa-earth-osm` repo. What do you think?

Hi @ekatef, thank you for the review. I have also noticed some of the points you rasied. I am taking a closer look at them. 

I agree we can further the discussion in pypsa-earth-osm.

ekatef:
Hey @GbotemiB, thanks for [the fix](https://github.com/pypsa-meets-earth/pypsa-earth-osm/pull/10) in `pypsa-earth-osm `. I think it covers the issue. Would you mind to close this PR?

PR #1610, opened 2025-10-16
CCL constraint with existing capacities

yerbol-akhmetov:
Good day. This PR aims first to fix the CCL constraint, as it is not properly functioning due to new linopy version. In addition, the PR also aims to take into account existing capacities. It will be possible to select either to account or not existing capacities in config.

The current implementation is not ideal, because separate constraints are added for each carrier. Previous implementation was better, but I was not able to reproduce and adapt it with new version. Maybe I need to spend more time on it. Initially, I thought also to include links (conventional generation units) to constraints, but it become lengthy and more elegant way needs to be thought off.

Currently, the PR works fine. I have tested it with NG and BJ. Maximum limits of 300MW and 100MW were set to NG and BJ for solar. Maximum limit for onwind for NG was set to 2000MW, but it was not reached. Minimum limit for solar for NG was set to 100MW. The results were satisfying the constraints. But maybe I will tweak limits more to get minimum also be visible.

image

image

## Checklist


yerbol-akhmetov:
I think it is important to `add data/agg_p_nom_minmax.csv` file as an example. We can also include the CCL constraint to test to keep track of its functionality. @ekatef @davide-f

yerbol-akhmetov:
Here is another result of the simulation with CCL constraint with min max constraints:
image
`agg_p_nom_minmax.csv` that is used to set constraints:
image
We can see that constraints are followed. What do you think @ekatef @davide-f ?

yerbol-akhmetov:
I wanted to add CCL constraint to CI test. Therefore, I have added this option into `test/config.sector.yaml`.
Before CCL constraints, we have the following expansion for `config.sector.yaml`:
image
First of all, the expansion seems quite large. For testing CCL, we can set maximum limits for solar as 150GW and onwind 200GW for BJ. For BJ, it would be good to set minimum constraint for onwind as 100GW to make sure minimum CCL also works. For NG, we can set onwind maximum limit as 900GW and 250GW for solar.
After applying CCL constraints:
image
So, the constraints for test works. I have added that `data/agg_p_nom_minmax.csv` and `CCL` option to opts. Do you think I should change parameter in mock snakemake is `solve_network.py`? @davide-f

config.default.yaml
```diff
@@ -182,7 +182,9 @@ electricity:
   voltages: [132., 220., 300., 380., 500., 750.]
   co2limit: 7.75e+7 # European default, 0.05 * 3.1e9*0.5, needs to be adjusted for Africa
   co2base: 1.487e+9 # European default, adjustment to Africa necessary
-  agg_p_nom_limits: data/agg_p_nom_minmax.csv
+  agg_p_nom_limits:
+    file: data/agg_p_nom_minmax.csv
+    include_existing: false
```
ekatef: I wonder if it could make sense to add a comment explaining how the constraint will work for the sector-coupled model? My feeling is that it's perfectly fine to leave the current implementation focused on `Generator` component only [as wind and solar are likely to be the most demanded option in any case] we just need to make it evident for a user

doc/configtables/electricity.csv
```diff
@@ -5,7 +5,9 @@ co2limit,:math:`t_{CO_2-eq}/a`, float, "Cap on system total annual carbon dioxid
 co2base,:math:`t_{CO_2-eq}/a`, float, "Reference value of system total annual carbon dioxide equivalent emissions. Used only if relative emission reduction target is specified in ``{opts}`` wildcard."
 automatic_emission, bool, "{True, False}", "True: Emissions are obtained from automatic emission extraction procedure. False: Emissions are obtained manually"
 automatic_emission_base_year,, integer, "CO2 emissions of year 1990 from EDGAR category 1A1a (Public electricity and heat production)."
-agg_p_nom_limits, file, path, "Reference to ``.csv`` file specifying per carrier generator nominal capacity constraints for individual countries if ``'CCL'`` is in ``{opts}`` wildcard. Default is ``data/agg_p_nom_minmax.csv``."
+agg_p_nom_limits,,, "Configure per carrier generator nominal capacity constraints for individual countries if ``'CCL'`` is in ``{opts}`` wildcard."
```
ekatef: I wonder if we can improve a bit clarity of this fragment keeping, also for first-time users who is trying to understand the essence of the constraint? Sketching a suggestion bellow; feel free to revise obviously

```suggestion
agg_p_nom_limits,,, "Configure per carrier generator capacity constraints limiting minimum and maximum values of the expanded capacity `p_nom_opt` for individual countries. Is enabled if ``'CCL'`` is presents in ``{opts}`` wildcard."
```

scripts/solve_network.py
```diff
@@ -223,33 +229,62 @@ def add_CCL_constraints(n, config):
...
             .rename("country")
         )
-        ext_carrier_per_country = capacity_variable.loc[
-            country_grouper.index
-        ].groupby_sum(country_grouper)
-        lhs.append(ext_carrier_per_country)
-    lhs = merge(lhs, dim=pd.Index(ext_carriers, name="carrier"))
+        lhs = (
+            capacity_variable.loc[country_grouper.index].groupby(country_grouper).sum()
+        )
```
ekatef: Just to be sure: if country isn't included into `agg_p_nom_minmax.csv`, it isn't a part of `country_grouper` and will not be included into the constraint formulation, right?

test/config.sector.yaml
```diff
@@ -18,7 +18,7 @@ scenario:
   simpl: [""]
   ll: ["copt"]
   clusters: [4]
-  opts: [Co2L-24h]
+  opts: [CCL-Co2L-24h]
```
ekatef, line 21: I think it's a great implementation to keep CO2 constraint and the time aggregation wildcard together (since those two are most frequently used often seen in a combination). Could it be a good idea to revise this order also in the docstring of `add_CCL_constraints` for consistency? (this line https://github.com/pypsa-meets-earth/pypsa-earth/blob/4bfb13da0f24a5fc98f541babe00dfd72f7ba3f8/scripts/solve_network.py#L208C16-L208C20)

ekatef, line 21: May it be a good idea to add a test-specific configuration of `agg_p_nom_minmax.scv` (e.g. adding it to the test folder derectly) to make the constraint actually working in the test case but avoid adding limitations to the default setup which can be unexpected for a user?

ekatef (commented):
Hello @yerbol-akhmetov, very happy to hear that you have managed to get the implementation work! Have added some rather minor comments trying to support integration of the improved functionality.

An additional technical comment, currently the testing workflow doesn't find the reference objective for the `sector` test:
image
The reason is a changed name of a testing file (which is intentional as `CCL` wildcard had to be added to the file name). To fix this, the file name should be updated for `sector` test of `obj_ref.csv` file ([this line](https://github.com/pypsa-meets-earth/pypsa-earth/blob/4bfb13da0f24a5fc98f541babe00dfd72f7ba3f8/test/utils/obj_ref.csv#L19). I guess, there will be also need to update the the objective value for this case as it should change thanks to enabling `CCL` constraint.

Great work and looking forward to see this PR merged!

config.default.yaml
yerbol-akhmetov: Hello, @ekatef, thank you for the comment. Agree, it is important to add the explanation of how it works.

yerbol-akhmetov:
> Hello @yerbol-akhmetov, very happy to hear that you have managed to get the implementation work! Have added some rather minor comments trying to support integration of the improved functionality.
> 
> An additional technical comment, currently the testing workflow doesn't find the reference objective for the `sector` test: image The reason is a changed name of a testing file (which is intentional as `CCL` wildcard had to be added to the file name). To fix this, the file name should be updated for `sector` test of `obj_ref.csv` file ([this line](https://github.com/pypsa-meets-earth/pypsa-earth/blob/4bfb13da0f24a5fc98f541babe00dfd72f7ba3f8/test/utils/obj_ref.csv#L19). I guess, there will be also need to update the the objective value for this case as it should change thanks to enabling `CCL` constraint.
> 
> Great work and looking forward to see this PR merged!

Thank you, @ekatef. I have updated filename and objective value. Now they are matching:
image

scripts/solve_network.py
yerbol-akhmetov: Hi, @ekatef. Thank you for the comment. I have revised the implementation. Now, it is without for-loop, similar to PyPSA-Eur. I have tested it locally, we got the same results.
image

If country is not provided in the `agg_p_nom_minmax.csv`, the constraint for that country will not be applied. I have tested it also locally by removing `BJ` from `agg_p_nom_minmax.csv`. The workflow worked fine and only constraints for `NG` worked. The results are shown below:
image

test/config.sector.yaml
yerbol-akhmetov: Thank you, I have revised the order of parameters in docstring

yerbol-akhmetov: Thank you for the hint. I have moved `agg_p_nom_minmax.csv` from `data/` to `test/utils/` folder and updated the `config.sector.yaml`.

config.default.yaml
```diff
@@ -182,7 +182,9 @@ electricity:
   voltages: [132., 220., 300., 380., 500., 750.]
   co2limit: 7.75e+7 # European default, 0.05 * 3.1e9*0.5, needs to be adjusted for Africa
   co2base: 1.487e+9 # European default, adjustment to Africa necessary
-  agg_p_nom_limits: data/agg_p_nom_minmax.csv
+  agg_p_nom_limits: # enables to set country-wise maximum and minimum generation capacities for generators (e.g. renewables, nuclear, and geothermal)
+    file: data/agg_p_nom_minmax.csv # path to csv file containing country-wise generation capacity limits
```
ekatef, line 186: I think we need to add this file as well to give an example and make CI happy

REUSE.toml
```diff
@@ -25,6 +25,11 @@ precedence = "aggregate"
 SPDX-FileCopyrightText = "The PyPSA-Earth and PyPSA-Eur Authors"
 SPDX-License-Identifier = "CC-BY-4.0"
 
+[[annotations]]
+path = "test/utils/agg_p_nom_minmax.csv"
+precedence = "aggregate"
+SPDX-FileCopyrightText = "The PyPSA-Earth and PyPSA-Eur Authors"
+SPDX-License-Identifier = "CC-BY-4.0"
```
ekatef, line 32: Thanks for keeping REUSE in sight. Just putting a reminder here that if we'll decide to add `agg_p_nom_minmax` also to `data` would be good to list it here as well

scripts/solve_network.py
```diff
@@ -223,34 +229,69 @@ def add_CCL_constraints(n, config):
...
+            f"Considered existing capacities in CCL constraints for carrier {c}."
+        )
+
+    # Convert limits to xarray for masking
+    min_values = xr.DataArray(min_values.dropna()).rename(dim_0="group")
+    max_values = xr.DataArray(max_values.dropna()).rename(dim_0="group")
+
+    # Valid constraints
+    valid_min_index = min_values.indexes["group"].intersection(lhs.indexes["group"])
+    valid_max_index = max_values.indexes["group"].intersection(lhs.indexes["group"])
```
ekatef, line 281: An elegant implementation!

test/config.sector.yaml
ekatef: Thank you! I think it's a great solution to have `test/utils/agg_p_nom_minmax.csv` file set-up in way that it's actually tested by CI. It appears though that another `agg_p_nom_minmax.csv` is needed also in `data` to satisfy Snakemake expectations but this one can be simply empty. What do you think?

ekatef (commented):
Many thanks @yerbol-akhmetov! Great that you have been able to resolve this tricky issue with a shadow folder, and the revised implementation looks very neat.

My feeling is that the PR is basically ready to merge. The only point left is adding a "fake" `agg_p_nom_minmax.csv` to `data` folder to ensure the file is always there to make sure Snakemake is able to locate the file also for the cases which are using `agg_p_nom_limits : file` from `config.default.yaml` ["fake" meaning that it doesn't add CCL constraints which user may not intended to have]. CI is currently complaining because of this and an empty `agg_p_nom_minmax.csv` would resolve that. Does it make sense for you?

yerbol-akhmetov:
> at the PR is basically ready to merge. The only point left is adding a "fake" `agg_p_nom_minmax.csv` to `data` folder to ensure the file is always there to make sure Snakemake is able

Hi, @ekatef. Thank you for pointing it out. This is what I was thinking as a first option to make sure the CI works and also provide example structure of content. I was not yet added that, because I wanted to propose it. Thank you for suggestion.

config.default.yaml
yerbol-akhmetov: Yes, thank you for the comment.

yerbol-akhmetov:
Hi, @ekatef. Thank you for the comment. I have applied it and CI passed successfully.

REUSE.toml
ekatef, line 32:
```suggestion
[[annotations]]
path = "test/utils/agg_p_nom_minmax.csv"
precedence = "aggregate"
SPDX-FileCopyrightText = "The PyPSA-Earth and PyPSA-Eur Authors"
SPDX-License-Identifier = "CC-BY-4.0"

[[annotations]]
path = "data/agg_p_nom_minmax.csv"
precedence = "aggregate"
SPDX-FileCopyrightText = "The PyPSA-Earth and PyPSA-Eur Authors"
SPDX-License-Identifier = "CC-BY-4.0"
```

ekatef (approved):

ekatef:
> Hi, @ekatef. Thank you for the comment. I have applied it and CI passed successfully.

Perfect! 🙂 I think the only minor point left is adding also `data/agg_p_nom_minmax.csv` to REUSE specification and we can merge.

yerbol-akhmetov:
> > Hi, @ekatef. Thank you for the comment. I have applied it and CI passed successfully.
> 
> Perfect! 🙂 I think the only minor point left is adding also `data/agg_p_nom_minmax.csv` to REUSE specification and we can merge.

Thank you, @ekatef. I have applied your change on REUSE. I have not previously added it explicitly, because `REUSE` was already including `data/**` folder under CC-BY-4.0. Having it explicitly stated is also good to track the file.

davide-f (approved):
Looks great to me too! Great job katia and yerbol! 

Merging :)

PR #1622, opened 2025-10-31
Attach wind and solar generators using real positions from powerplants.csv

### LINKED ISSUE #753: Use powerplantmatching for wind and solar

Currently the build_powerplants rule filters out wind and solar. Now that we are adding GEM data to powerplantmatching it would be worth removing that filter. This is the line responsible: 

https://github.com/pypsa-meets-earth/pypsa-earth/blob/f7ddf6c4175b19c5661fe95dee2e96d884ad354b/scripts/build_powerplants.py#L304

This could be addressed in #750 

I am not sure how the IRENA scaling is currently working so this might have to be changed as well? @davide-f what do you think? Maybe worth adding a config parameter so that people can choose whether they want to use IRENA scaling or ppmatching.

danielelerede-oet:
# Closes #753 

This pull request modifies the renewable generator creation process to preserve the real geographic positions of existing wind and solar plants provided in powerplants.csv, instead of redistributing capacities according to population. @davide-f I'm not really sure this is everything we need to target this point, do you have any other insights?

## Checklist


ekatef:
Thank you very much @danielelerede-oet! That is great functionality which would be helpful for many applications. Sorry for delaying the review: we are currently about to re-structure the way we are working both on technical and organisational levels to be more effective in joining efforts

@virio-andreyana thanks a lot for being there to help! It would be great to have your review for this PR

scripts/add_electricity.py
```diff
@@ -365,83 +365,133 @@ def attach_wind_and_solar(
...
-                    + costs.at[tech + "-station", "capital_cost"]
+                    + costs.at[carrier + "-station", "capital_cost"]
                     + connection_cost
                 )
-                logger.info(
-                    "Added connection cost of {:0.0f}-{:0.0f} {}/MW/a to {}".format(
-                        connection_cost.min(),
-                        connection_cost.max(),
-                        snakemake.params.costs["output_currency"],
-                        tech,
```
virio-andreyana, line 505: Is this logger no longer relevant?

scripts/add_electricity.py
```diff
@@ -365,83 +365,133 @@ def attach_wind_and_solar(
...
-                df.technology == "Offshore", "offwind-ac"
-            )
+    df = ppl.rename(columns={"country": "Country"}).copy()
+
+    # Force offshore to offwind-ac
+    df.loc[df.technology == "Offshore", "carrier"] = "offwind-ac"
+    df.loc[df.technology == "Onshore", "carrier"] = "onwind"
 
-        df["carrier"] = df["carrier"].mask(df.technology == "Onshore", "onwind")
+    irena_targets = snakemake.params.electricity.get("irena_targets", {})
```
virio-andreyana: Looks like `irena_tagets` is a new config, should we add it to `config.default.yaml`?

scripts/add_electricity.py
```diff
@@ -945,10 +995,9 @@ def add_nice_carrier_names(n, config):
     attach_existing_batteries(n, costs, ppl)
 
     if snakemake.params.electricity.get("estimate_renewable_capacities"):
-        estimate_renewable_capacities_irena(
-            n,
-            snakemake.params.electricity["estimate_renewable_capacities"],
-            snakemake.params.countries,
+        logger.info(
+            "Skipping legacy IRENA capacity estimation for wind and solar: "
+            "handled directly in attach_wind_and_solar."
```
virio-andreyana: Should we really remove legacy options? Maybe some countries don't have robust renewable locations. Some renewable can also be very decentralized that its not stored in powerplants.csv

virio-andreyana (commented):
Thanks Daniele for this PR, I've been wondering as well why we didn't make use of the renewable options in `powerplants.csv`. Here is my comment:

scripts/add_electricity.py
danielelerede-oet: Good catch Virio, but I think it's better if we don't add `irena_targets` to the config entries.

I changed this to reuse the existing `electricity.estimate_renewable_capacities`
settings and the same IRENASTAT access already used in the legacy
`estimate_renewable_capacities_irena` implementation. So the IRENA targets are
now computed internally in `attach_wind_and_solar`, and no new key is needed in
`config.default.yaml`.

scripts/add_electricity.py
danielelerede-oet: It is, I should have deleted it by mistake. I restored it, thanks for noticing!

scripts/add_electricity.py
danielelerede-oet: Good point, but we are not removing that logic, just moving it upstream. Previously,  `estimate_renewable_capacities_irena` was assigning capacities only to extendable generators, which could ignore existing plants or lead to inconsistencies.
Now the same idea (using IRENA to fill gaps where powerplants data is incomplete) is applied directly in `attach_wind_and_solar`, i.e. at the stage where existing capacities are built.
So:
- `powerplants.csv` is still used where available (to consider the spatial detail)
- IRENA is used to fill missing national totals (especially relevant for e.g., decentralized renewables)

This should actually better handle the case you mention, since we preserve spatial information where we have it, and only redistribute the missing share.

If needed, we could still keep the legacy function as a fallback behind a config flag, but functionally the behavior is now already covered in a more consistent way. What do you think?

scripts/add_electricity.py
```diff
@@ -429,159 +429,150 @@ def attach_wind_and_solar(
...
-        Dictionary of extendable carriers for different component types.
-    line_length_factor : float
-        Factor to adjust line lengths for connection cost calculations.
+    Existing capacities are taken from powerplants.csv and spatialized to buses.
+    National capacity gaps with respect to IRENA targets are filled and
+    redistributed uniformly across buses within each country.
 
-    Returns
-    -------
-    None
```
virio-andreyana, line 459: If possible, keeping and updating the docstring

scripts/add_electricity.py
virio-andreyana: Hmm, originally I was thinking that that those function could be used just in case we have global data on solar rooftops or any other distributive technologies that we've haven't think about. But then again, each distributive technologies has its own heuristics (solar rooftop based on the availability of roofs, micro hydro based on water stream and population etc)

So I agree, we can remove them, and build better ones once the data is available :)

virio-andreyana (commented):
Thanks Daniele, here is my second comment! If the CI test results is within boundaries, then it looks good to me.

scripts/add_electricity.py
danielelerede-oet: Sure Virio, it was a placeholder and I forgot to update it

virio-andreyana:
Looks good to me @ekatef!

danielelerede-oet:
> Looks good to me @ekatef!

I've also re-tested it on Australia and works as intended!

Current upstream:
image

Including this PR (the effect on batteries is due to including a custom_powerplants.csv file):
image

scripts/add_electricity.py
```diff
@@ -425,17 +425,92 @@ def aggregate_inflow_by_group(
...
+    carrier: str,
+    estimate_renewable_capacities_config: dict,
+    countries: list,
+) -> pd.Series:
+    """
+    Return IRENA installed capacity targets for a given carrier as a Series
+    indexed by country (MW).
+
+    The data source and configuration are the same as in
+    estimate_renewable_capacities_irena.
```
ekatef: Are there any reasons to keep `estimate_renewable_capacities_irena` function, both in the code and in the docstring here? To my understanding, `get_irena_targets_for_carrier` is a full reimplementation of `estimate_renewable_capacities_irena` and it would be good to keep only `get_irena_targets_for_carrier` which is actually used.  @danielelerede-oet what do you think?

I guess the major reason for copy-pasting of `estimate_renewable_capacities_irena` into `get_irena_targets_for_carrier` has been the need to minimise merge conflcts, right?

scripts/add_electricity.py
ekatef: Great! I think it's more elegant to have IRENA-based estimation integrated into attach_wind_and_solar. Although, I agree with @virio-andreyana that the message raises some questions and can be potentially confusing. Could we probably remove it?

ekatef (commented):
@danielelerede-oet Thanks a lot for the great feature and in-depth investigation! @virio-andreyana having your review has really helped a lot!

Agree that the PR is basically ready to be merged, I'm just thinking that we could also remove the legacy code to keep the implementation lighted. @danielelerede-oet could you please check if the specific comments do make sense for you?

scripts/add_electricity.py
danielelerede-oet: Good point, I agree.
that's correct, `get_irena_targets_for_carrier` currently duplicates the IRENASTAT access logic, and keeping both functions in parallel is not ideal. The duplication was useful to introduce the new behaviour with minimal changes to the previous code path, but now that the gap-filling logic is handled directly in `attach_wind_and_solar`, it makes sense to consolidate this and remove the legacy function.

I will also update the docstring/comment so it does not refer back to
`estimate_renewable_capacities_irena`.

danielelerede-oet:
Thanks for the comments @ekatef . Apart from them, I have removed the legacy `estimate_renewable_capacities_irena` path. IRENA-based handling for wind and solar is now implemented only in `attach_wind_and_solar`, using the existing `estimate_renewable_capacities` config to read IRENASTAT targets and fill remaining country-level gaps after assigning capacities from `powerplants.csv`.

scripts/add_electricity.py
ekatef: Perfect, thank you!

There is also a question of retaining history and contributions: when an old code is copy-pasted and modified only after that, all the previous history is lost. In this case, copy-paste is justified by the need to deal with merge conflicts, so I think it's fine.

ekatef (approved):
Thanks a lot @danielelerede-oet! Great functionality and neat implementation 😄 

As a minor technical comment, I see that rst release notes have been some changes which don't seem relevant. However, rst notes themselves must be removed as we have transitioned to markdown format. Have opened #1797 to track and fix that.

Agree that the PR is ready to be merged once CI passes!

Issue #1208, opened 2024-11-27
Add sector-coupled configuration to docs


energyLS:
## Describe the feature you'd like to see

We have a nice [documentation of the electricity-only PyPSA-Earth](https://pypsa-earth.readthedocs.io/en/latest/configuration.html). However, this is also necessary for the sector-coupled part. Potentially rely on https://papers.ssrn.com/sol3/papers.cfm?abstract_id=4743242 for information.

This can be done here: https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/doc/configuration.rst

PR #1623, opened 2025-10-31
Account for existing batteries in powerplants file and improve logging on installed capacity

danielelerede-oet:
This PR ensures that battery storage from powerplants.csv is taken into account both in the power-only and sector-coupled networks, and improves logging in `add_electricity` and `prepare_sector_network`, while avoiding duplication of log messages in `add_electricity` and `simplify_network`.

## Checklist


scripts/add_electricity.py
```diff
@@ -111,6 +111,7 @@
 idx = pd.IndexSlice
 
 logger = create_logger(__name__)
+logger.propagate = False
```
ekatef: Is it an intentional change?

scripts/add_electricity.py
```diff
@@ -666,6 +667,42 @@ def attach_hydro(n, costs, ppl):
...
+    if batteries.empty:
+        logger.info("No existing batteries found in powerplants.csv.")
+        return
+
+    _add_missing_carriers_from_costs(n, costs, ["battery"])
+
+    # Remove duplicates and reset index like in attach_hydro
+    batteries = batteries.reset_index(drop=True).rename(index=lambda s: f"{s} battery")
+
+    max_hours = snakemake.params.electricity["max_hours"]["battery"]
```
ekatef, line 683: By any chance, aren't `max_hours` readily available in the data extracted by powerplantmatching? Could it be an option to calculate them using some simple empirics by the data we have on hand for the real batteries?

scripts/prepare_sector_network.py
```diff
@@ -1291,8 +1291,26 @@ def add_aviation(n, cost, energy_totals, airports_fn):
     )
 
 
-def add_storage(n, costs):
-    "function to add the different types of storage systems"
+def fetch_existing_battery_capacity_from_elec(n_elec) -> dict:
```
ekatef: We don't use type hinting right now. Could be potentially a good idea but need to be implemented in a consistent way

scripts/prepare_sector_network.py
```diff
@@ -1291,8 +1291,26 @@ def add_aviation(n, cost, energy_totals, airports_fn):
     )
 
 
-def add_storage(n, costs):
-    "function to add the different types of storage systems"
+def fetch_existing_battery_capacity_from_elec(n_elec) -> dict:
+    """
+    Read existing battery capacity from the electricity network and
+    return a dict {bus_name: p_nom_MW} aggregated per AC bus.
+    """
```
ekatef, line 1303: Thank you for remembering to add a docstring!

scripts/prepare_sector_network.py
```diff
@@ -1291,8 +1291,26 @@ def add_aviation(n, cost, energy_totals, airports_fn):
...
+    if bat.empty:
+        return {}
+
+    # Aggregate capacity per bus (MW)
+    return bat.groupby("bus")["p_nom"].sum().to_dict()
+
+
+def add_storage(n, costs, existing_battery_capacity: dict | None = None):
+    """Function to add the different types of storage systems, including
+    existing battery capacities from powerplants.csv."""
```
ekatef: Just to double-check: adding the existing batteries also here doesn't duplicate `add_existing_batteries` implemented in `add_electricity`, right?

scripts/prepare_sector_network.py
```diff
@@ -1306,6 +1324,104 @@ def add_storage(n, costs):
...
+    if existing_battery_capacity is None:
+        existing_battery_capacity = {}
+
+    total = float(sum(existing_battery_capacity.values()))
+    nodes_with_cap = len([p for p in existing_battery_capacity.values() if p > 0])
+    logger.info(
+        f"Battery carry-over from elec.nc: {total/1e3:.2f} GW across {nodes_with_cap} nodes."
+    )
+
+    try:
```
ekatef: What are particular reasons of having `try-catch` block here?

Generally, using of try-catch blocks is discouraged unless there is a strong need to address some special cases which can happen as a part of a normal workflow [e.g. some url source won't be available, or data field doesn't present in the dataset]. In those cases a type of the exception must be specified as otherwise the errors can be masked

scripts/prepare_sector_network.py
```diff
@@ -1306,6 +1324,104 @@ def add_storage(n, costs):
...
+        # Ensure 'bus' and 'DateOut' are numeric
+        if "bus" in df_powerplants.columns:
+            df_powerplants["bus"] = (
+                df_powerplants["bus"]
+                .astype(str)
+                .str.replace(r"[^0-9]", "", regex=True)
+                .replace("", "0")
+                .astype(int)
+            )
```
ekatef: Not sure I get where `df_powerplants` is defined... Could you please double check nothing is missed?

Also, will this regex magic work properly both for Voronoi and alternative clustering? I wonder if filtering by a carrier could be a more stable solution [not sure however I get the current implementation properly so feel free to disregard the comment]

scripts/prepare_sector_network.py
```diff
@@ -1306,6 +1324,104 @@ def add_storage(n, costs):
...
+        )
+        df_batteries = df_powerplants[
+            (df_powerplants.Fueltype == "battery")
+            & (df_powerplants.DateOut >= baseyear)
+        ].copy()
+
+        if not df_batteries.empty:
+            # assign clustered bus
+            busmap_s = pd.read_csv(snakemake.input.busmap_s, index_col=0).squeeze()
+            busmap = pd.read_csv(snakemake.input.busmap, index_col=0).squeeze()
```
ekatef: Not sure there are `busmap_s` and `busmap` inputs for `prepare_sector_network`. Am I missing something?

scripts/prepare_sector_network.py
```diff
@@ -1306,6 +1324,104 @@ def add_storage(n, costs):
...
+
+            inv_busmap = {}
+            for k, v in busmap.items():
+                inv_busmap[v] = inv_busmap.get(v, []) + [k]
+
+            clustermaps = busmap_s.map(busmap)
+            clustermaps.index = clustermaps.index.astype(int)
+
+            df_batteries["cluster_bus"] = df_batteries.bus.map(clustermaps)
```
ekatef: Just as a double-check: could you please explain briefly for which reason the buses had to be treated in a different way as compared with `add_generation`? [I suspect your implementation may be more correct and `add_electricity` requires in this case a TODO to be re-implemented in the same way]

scripts/prepare_sector_network.py
```diff
@@ -1306,6 +1324,104 @@ def add_storage(n, costs):
...
+            total = sum(existing_battery_capacity.values())
+            nodes = len(existing_battery_capacity)
+            logger.info(
+                f"Found existing battery capacity from powerplants.csv: "
+                f"{total:.1f} MW in {nodes} nodes"
+            )
+
+    except Exception as e:
+        logger.warning(f"Could not load existing battery capacities: {e}")
+        existing_battery_capacity = {}
```
ekatef: If `except-try` is needed a type of the exception must be specified to address the arising issues in a more targeted way

scripts/prepare_sector_network.py
```diff
@@ -1306,6 +1324,104 @@ def add_storage(n, costs):
...
+
+    e_nom_mins = []
+    p_nom_mins_charger = []
+    p_nom_mins_discharger = []
+
+    for node in spatial.nodes:
+        p = existing_battery_capacity.get(node, 0)
+        e_nom_mins.append(p * battery_duration)
+        p_nom_mins_charger.append(p)
+        p_nom_mins_discharger.append(p)
```
ekatef: I wonder if a list comprehension could be a more elegant solution here? [though, the current implementation may be more readable, so feel free to disregard]

scripts/simplify_network.py
```diff
@@ -113,6 +113,7 @@
 sys.settrace
 
 logger = create_logger(__name__)
+logger.propagate = False
```
ekatef: Is it an intentional change? May it be needed also in other scripts (could be a good idea for a further PR)?

ekatef (commented):
Many thanks @danielelerede-oet! Looks a great advancement and a neat implementation.

I have had a number of questions, in particular related to the sector part. Could you please go through them to make sure we aren't missing anything?

scripts/add_electricity.py
danielelerede-oet: Yes, to avoid duplicated logging. Honestly is not a big deal for me, but I think it makes things more elegant. Could be addressed in another PR and extended to the entire code BTW.

scripts/add_electricity.py
danielelerede-oet: As far as I know, they shouldn't be available from powerplantmatching

scripts/prepare_sector_network.py
danielelerede-oet: Makes sense. I'm removing it for the time being :)

scripts/prepare_sector_network.py
danielelerede-oet: In `prepare_sector_network`, `add_storage` rebuilds the storage formulation for the sector model and uses the carried-over capacities solely as lower bounds. The electricity-stage battery components are not active in the sector solve, so nothing should be counted twice.

scripts/prepare_sector_network.py
danielelerede-oet: This should mask issues related to malformed or missing powerplant metadata, but it's maybe too careful to have it here. I will remove it

scripts/prepare_sector_network.py
danielelerede-oet: Thanks for flagging this. I removed the `df_powerplants` logic from `add_storage` to avoid duplicating functionality already handled in `add_electricity` and to keep the sector model strictly based on carry-over from `elec.nc`. Existing battery capacities are now sourced exclusively from the electricity-stage network, which makes the data flow explicit and avoids hidden dependencies on `powerplants.csv`.

Regarding the regex: I’m confident it is safe in the current implementation because battery buses and links are created  from `spatial.nodes`, independent of whether Voronoi or alternative clustering is used. That said, filtering by `carrier == "battery"` would indeed be more robust and could be a reasonable follow-up refactor; I kept the current approach here to minimize the scope of this PR.

scripts/prepare_sector_network.py
danielelerede-oet: Those inputs were only needed for the now-removed powerplants.csv battery mapping. With battery capacities taken exclusively from `elec.nc`, `busmap_s` / `busmap` are no longer required here.

scripts/prepare_sector_network.py
danielelerede-oet: Yes, the difference comes from where in the pipeline this is applied. In `prepare_sector_network` the network has already been clustered, while `powerplants.csv` still refers to pre-clustering buses, so an explicit remapping via busmap / busmap_s was required to avoid assigning capacities to the wrong nodes. In `add_generation` this mapping effectively happens earlier and is therefore implicit.

So it’s not a conceptual difference, just a consequence of the execution stage. Aligning the two implementations could make sense, but I think it would require touching the data flow more broadly.

Just to clarify: this logic has been removed in the current version of the PR, so there is no longer any special bus handling here.

scripts/prepare_sector_network.py
danielelerede-oet: Removed now :)

scripts/prepare_sector_network.py
danielelerede-oet: Good point. A list comprehension would indeed be more compact, but I went with the explicit loop mainly for readability. Given the small size of `spatial.nodes`, performance is not a concern here, and I felt the explicit form makes it easier to verify correctness and extend later if additional logic is needed. Happy to switch to a list comprehension if we prefer a more compact style.

scripts/simplify_network.py
danielelerede-oet: Answered above too, I can remove it and address it comprehensively in another PR

danielelerede-oet:
Hi @ekatef thanks for the review! I have addressed your changes but I definitely need to re-test if everything's working as intended with a powerplants file containing batteries :)

danielelerede-oet:
@ekatef I've checked that the batteries are correctly fetched from the powerplants file if present both in the electricity and sector networks. Anything in particular you'd like me to check?

ekatef:
> @ekatef I've checked that the batteries are correctly fetched from the powerplants file if present both in the electricity and sector networks. Anything in particular you'd like me to check?

Hey @danielelerede-oet, happy New Year 😄 I hope you have had an opportunity to relax during the festive time.

The testing being successful is great news! Not sure any additional tests should be needed 🙂 It's always a good idea to check the change in the objective function but it looks alright at a quick glance.

I'll try to go back to the PR as fast as I can (in the middle of the next week if trying to be realistic), and looking forward to see this work completed!

danielelerede-oet:
> > @ekatef I've checked that the batteries are correctly fetched from the powerplants file if present both in the electricity and sector networks. Anything in particular you'd like me to check?
> 
> Hey @danielelerede-oet, happy New Year 😄 I hope you have had an opportunity to relax during the festive time.
> 
> The testing being successful is great news! Not sure any additional tests should be needed 🙂 It's always a good idea to check the change in the objective function but it looks alright at a quick glance.
> 
> I'll try to go back to the PR as fast as I can (in the middle of the next week if trying to be realistic), and looking forward to see this work completed!

@ekatef yes I've had both time to relax and to finally work on this PR :laughing: I've tested the sector model for the default configuration (NG, BJ) adding a 1GW plant in NG and the objective function decreases very slightly (by 0.0000403%, `6780187922.450884` main vs `6780185193.653492` existing batteries) wrt the `main` branch due to dispatch choices as the battery plant comes for free. In the electricity network, instead, the objective function is 1512090080.4144554 for `main` and 1512091305.7489874 for `existing batteries`.

scripts/prepare_sector_network.py
ekatef: Tanks for checking! If you feel an explicit loop is more readable let's keep it there.

scripts/prepare_sector_network.py
```diff
@@ -1312,6 +1330,31 @@ def add_storage(n, costs):
...
+    total = float(sum(existing_battery_capacity.values()))
+    nodes_with_cap = len([p for p in existing_battery_capacity.values() if p > 0])
+    logger.info(
+        f"Battery carry-over from elec.nc: {total/1e3:.2f} GW across {nodes_with_cap} nodes."
+    )
+
+    # Use configured max_hours for battery duration
+    elec_config = snakemake.config["electricity"]
+    max_hours = elec_config["max_hours"]
+    battery_duration = max_hours["battery"]
```
ekatef: Let's use `params` also here for consistency? I mean something like this (have copy-pasted from your implementation in `add_electricity` ):

```suggestion
    battery_duration = snakemake.params.electricity["max_hours"]["battery"]
```

scripts/prepare_sector_network.py
```diff
@@ -1312,6 +1330,31 @@ def add_storage(n, costs):
...
+
+    e_nom_mins = []
+    p_nom_mins_charger = []
+    p_nom_mins_discharger = []
+
+    for node in spatial.nodes:
+        p = existing_battery_capacity.get(node, 0)
+        e_nom_mins.append(p * battery_duration)
+        p_nom_mins_charger.append(p)
+        p_nom_mins_discharger.append(p)
```
ekatef: Not sure I get how `e_nom_mins `, `p_nom_mins_charger ` and `p_nom_mins_discharger ` are used. Could you please double-check?

scripts/simplify_network.py
ekatef: No problem to keep it here while agree that a more comprehensive logging improvement deserves a dedicated PR

ekatef (commented):
Hello @danielelerede-oet, thanks a lot for polishing the PR! I think we are very close to finalise. Could you please check the remaining points? They are quite technical ones while the major functionality is there and neatly implemented.

ekatef:
> > > @ekatef I've checked that the batteries are correctly fetched from the powerplants file if present both in the electricity and sector networks. Anything in particular you'd like me to check?
> > 
> > 
> > Hey @danielelerede-oet, happy New Year 😄 I hope you have had an opportunity to relax during the festive time.
> > The testing being successful is great news! Not sure any additional tests should be needed 🙂 It's always a good idea to check the change in the objective function but it looks alright at a quick glance.
> > I'll try to go back to the PR as fast as I can (in the middle of the next week if trying to be realistic), and looking forward to see this work completed!
> 
> @ekatef yes I've had both time to relax and to finally work on this PR 😆 I've tested the sector model for the default configuration (NG, BJ) adding a 1GW plant in NG and the objective function decreases very slightly (by 0.0000403%, `6780187922.450884` main vs `6780185193.653492` existing batteries) wrt the `main` branch due to dispatch choices as the battery plant comes for free. In the electricity network, instead, the objective function is 1512090080.4144554 for `main` and 1512091305.7489874 for `existing batteries`.

Hey @danielelerede-oet, that is a pleasure to get back to reviewing your PR! Interesting, thanks for checking. I'd say `0.0000403%` might be even too low 😄 Have you checked that the existing batteries have been in fact fetched? [This comment](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1623#discussion_r2680411019) could be potentially related to that.

One more point: I see that you add batteries as StorageUnits to the power-only model while our default config attaches extendable batteries as Stores https://github.com/pypsa-meets-earth/pypsa-earth/blob/fdf7f4ae6f9abe1cd4b60c9f9296643351c6452a/config.default.yaml#L205-L211
Have you used `StorageUnits` in `add_electricity` for simplicity or have there been also other reasons? [I think it's perfectly well as it is now just trying to understand if there are any points which deserves to me mentioned in the tutorials or tracked further in issues]

scripts/prepare_sector_network.py
danielelerede-oet: Yes, agreed, good point. Using `snakemake.params.electricity` here is more consistent with add_electricity. I’ll switch to the suggested form.

scripts/prepare_sector_network.py
danielelerede-oet: You’re right, these variables are currently not used anywhere downstream. They were leftovers from an earlier version, so I’ll drop them to avoid confusion.

danielelerede-oet:
> > > > @ekatef I've checked that the batteries are correctly fetched from the powerplants file if present both in the electricity and sector networks. Anything in particular you'd like me to check?
> > > 
> > > 
> > > Hey @danielelerede-oet, happy New Year 😄 I hope you have had an opportunity to relax during the festive time.
> > > The testing being successful is great news! Not sure any additional tests should be needed 🙂 It's always a good idea to check the change in the objective function but it looks alright at a quick glance.
> > > I'll try to go back to the PR as fast as I can (in the middle of the next week if trying to be realistic), and looking forward to see this work completed!
> > 
> > 
> > @ekatef yes I've had both time to relax and to finally work on this PR 😆 I've tested the sector model for the default configuration (NG, BJ) adding a 1GW plant in NG and the objective function decreases very slightly (by 0.0000403%, `6780187922.450884` main vs `6780185193.653492` existing batteries) wrt the `main` branch due to dispatch choices as the battery plant comes for free. In the electricity network, instead, the objective function is 1512090080.4144554 for `main` and 1512091305.7489874 for `existing batteries`.
> 
> Hey @danielelerede-oet, that is a pleasure to get back to reviewing your PR! Interesting, thanks for checking. I'd say `0.0000403%` might be even too low 😄 Have you checked that the existing batteries have been in fact fetched? [This comment](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1623#discussion_r2680411019) could be potentially related to that.
> 
> One more point: I see that you add batteries as StorageUnits to the power-only model while our default config attaches extendable batteries as Stores
> 
> https://github.com/pypsa-meets-earth/pypsa-earth/blob/fdf7f4ae6f9abe1cd4b60c9f9296643351c6452a/config.default.yaml#L205-L211
> 
> 
> Have you used `StorageUnits` in `add_electricity` for simplicity or have there been also other reasons? [I think it's perfectly well as it is now just trying to understand if there are any points which deserves to me mentioned in the tutorials or tracked further in issues]

@ekatef Yes, this turned out to be the key issue. In my initial tests the deviation was indeed misleadingly small because the existing batteries were not correctly fetched into the sector-coupled networks. After adjusting the setup and re-running the model with a more representative existing battery capacity (100 GW instead of 0.1 GW), it became clear that this needed to be fixed.

Regarding the representation: existing batteries are added as StorageUnits in `add_electricity` mainly for consistency with other existing storage assets (e.g. hydro, PHS) and to keep the power-only formulation simple, where power and energy capacities are fixed and coupled. In contrast, extendable batteries are represented via the Store + charger/discharger Link formulation, where separating power and energy investments is intentional and useful.

There was no additional modelling assumption behind this choice. However, I agree that the difference in representation between existing and extendable batteries is worth documenting, and possibly revisiting in the future if we want to fully unify the formulations.

With the fixes in place, existing battery capacity is now correctly retained when moving from the electricity-only to the sector-coupled network. I’m posting below the relevant results for both networks to illustrate this.

I'm posting the relevant results from the test with the latest improvements below (both related to electricity-only and sector networks).

image

ekatef (approved):
Many thanks for cross-checking @danielelerede-oet! I think it's excellent and ready to be merged. Thank you for the great contribution! 🎉

ekatef:
> @ekatef Yes, this turned out to be the key issue. In my initial tests the deviation was indeed misleadingly small because the existing batteries were not correctly fetched into the sector-coupled networks. After adjusting the setup and re-running the model with a more representative existing battery capacity (100 GW instead of 0.1 GW), it became clear that this needed to be fixed.
> 
> Regarding the representation: existing batteries are added as StorageUnits in `add_electricity` mainly for consistency with other existing storage assets (e.g. hydro, PHS) and to keep the power-only formulation simple, where power and energy capacities are fixed and coupled. In contrast, extendable batteries are represented via the Store + charger/discharger Link formulation, where separating power and energy investments is intentional and useful.
> 
> There was no additional modelling assumption behind this choice. However, I agree that the difference in representation between existing and extendable batteries is worth documenting, and possibly revisiting in the future if we want to fully unify the formulations.
> 
> With the fixes in place, existing battery capacity is now correctly retained when moving from the electricity-only to the sector-coupled network. I’m posting below the relevant results for both networks to illustrate this.
> 
> I'm posting the relevant results from the test with the latest improvements below (both related to electricity-only and sector networks).
> 
> image

Hey @danielelerede-oet congratulations with getting it done 😄 A nice investigation, and a great demo for the functionality implemented! 

Thanks for the explanation of your idea behind the different representation of extendable and non-extendable storage. I think it makes perfect sense. For documentation, I'd suggest to track the need for documenting this part as sub-issue of #1208 and proceed with merging this PR. Would it be alright for you, or do you prefer to amend a note on methodology to the docs as a part of this PR?

danielelerede-oet:
@ekatef thank you! I have opened #1713 to track the documentation status for the new implementation of existing batteries. Any suggestion about where I could also amend a note on methodology to the docs as a part of this PR regarding this feature?

ekatef:
> @ekatef thank you! I have opened #1713 to track the documentation status for the new implementation of existing batteries. Any suggestion about where I could also amend a note on methodology to the docs as a part of this PR regarding this feature?

Perfect, thank you so much @danielelerede-oet! Honestly, I'm not quite sure what is the best place to insert a methodology note on the batteries implementation. It looks we don't really touch details of the modelling methodology. So, I completely agree with you that it's worth opening an issue to track possible improvements on that.

The implementation is completed and I think it's a really nice one! Merging 😄

PR #1494, opened 2025-06-30
Sector documentation

yerbol-akhmetov:
## Changes proposed in this Pull Request
Good day. This is the draft PR to improve documentation of the sector part. In particular, it is focused on improving `Configuration` page of the documentation.

## Checklist

- [x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [x] I tested my contribution locally and it seems to work fine.
- [ ] Code and workflow changes are sufficiently documented.
- [ ] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`.
- [ ] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [ ] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [ ] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes, including reference to the requested PR.

PR #1629, opened 2025-11-07
Use Global Buildings datasets to estimate solar-rooftop potentials

virio-andreyana:

The current implementation of solar rooftop potential uses a simplified approach based on the following equation:
**potential = population × 20 m²/person × 0.1 kW/m²**

This follows the default assumption from PyPSA-Eur. However, since solar rooftop could play a more significant role in the Global South, more refined assumptions are needed.

This pull request is inspired by [Hideaki Obane’s paper](https://eneken.ieej.or.jp/data/12710.pdf). While the paper uses the [Google Open Buildings](https://sites.research.google/gr/open-buildings/) dataset, this implementation relies on the [Microsoft Global Buildings](https://github.com/microsoft/GlobalMLBuildingFootprints) dataset. The key difference is that Open Buildings provides more recent data, whereas Global Buildings offers broader, worldwide coverage.

This is the workflow:
1. We first download all the buildings shapes in a given country and saved in the data folder.
2. The rooftop area of each building is then calculated. Only a portion of each rooftop can be allocated for PV installation. Smaller buildings are less likely to host PV systems, while larger buildings can dedicate a greater share of their rooftop to PVs. (Optional: you can also filtered out small buildings)
```yaml
    install_ratio:
      0: 0
      10: 0.3 # Damayanti et al Residential rooftop solar: Technical and market potential in 34 provinces in Indonesia (IESR)
      100: 0.36 # Damayanti et al
      200: 0.41
      450: 0.49 # Gagnon et al Roftop solar photovoltaic technical potential in the United States: A detailed assessment (NREL)
      2300: 0.66 # Gagnon et al
```
3. The useful area is then clustered to be used in the `prepare_sector_networks`

Since this workflow requires more time to run, we still provide the option to use the original population-based assumptions as an approximation. Even better, these settings are now more transparent, as they are defined directly in the configuration file.

The configuration remains backward compatible. In this version, `solar_rooftop` is defined as a dictionary, whereas in previous versions it was a simple `true/false` flag. Both formats are supported, but the new dictionary format is required to enable this workflow.

Finally, the `download_global_buildings` rule can be used not only to estimate solar rooftop potential, but also to assess building spaces for battery storage, heating/cooling demand, and other applications. We just need more literature to implement this.

I've placed the CI test of this feature in `test/config.sector.yaml`

## Checklist


davide-f:
Hello virio and great idea!

I'm reporting here some comments emerged during the afternoon meeting yesterday.

Katia mentioned the ram issues you have been struggling with. In the following we mention some options for possibly supporting:
- it would be interesting to get inspired from build_natura_raster for the analysis. For a normal user, we could create a default tiff file with smaller resolution, similar to build_natura_raster. The rule also adopts procedures to reduce rime requirements by executing operations iteratively each time on a smaller fraction of the dataset. More can follow if interested
- Pierre also mentioned the possible use of land classification from copernicus to estimate the value. Land use would overestimate the rooftop area though

Would like to hear your opinion and happy to follow from there. Great work!

virio-andreyana:
> Hello virio and great idea!
> 
> I'm reporting here some comments emerged during the afternoon meeting yesterday.
> 
> Katia mentioned the ram issues you have been struggling with. In the following we mention some options for possibly supporting:
> 
>     * it would be interesting to get inspired from build_natura_raster for the analysis. For a normal user, we could create a default tiff file with smaller resolution, similar to build_natura_raster. The rule also adopts procedures to reduce rime requirements by executing operations iteratively each time on a smaller fraction of the dataset. More can follow if interested
> 
>     * Pierre also mentioned the possible use of land classification from copernicus to estimate the value. Land use would overestimate the rooftop area though
> 
> 
> Would like to hear your opinion and happy to follow from there. Great work!

Thanks @davide-f for joining the discussion! After some input we did some improvement:

* First, we split the workflow to `download_global_buildings` and `cluster_global_buildings`. The former only have to be executed once, while the latter have to be run for each clustering.
* In `download_global_buildings`, we don't save the entire building information. Instead, we just save the center coordinate and the surface area of the building. This save the memory storage by 80%. As a context for Nigeria. What was once 10 GB are now 1.8 GB.
* In case a feature in the future requires the height, volume or the surface area for the wall itself (heat/cooling dissipation, demand distribution), this will need to be processed in  `download_global_buildings` as well.

Inspite of that, the time required to run the CI doubles from one hour to 2 hours. It takes 30 min to download, and another 30 to process in `cluster_global_buildings`. 
* At least for the CI, we can make a mock example (100 buildings instead of 100.000) and save them somewhere either in the same package as the tutorial bundle.
* Or a simpler solution is to disable it in the CI.

What do you think Davide?

ekatef:
Hello @virio-andreyana, great work and looking forward to have this feature implemented. Adding a brief summary of our discussion today to keep @davide-f in the loop.

It would be a great idea to create a smaller "tutorial" buildings raster for testing purposes. However, that would require some additional work due to the need to update the tutorial datakit. The way how did we do it previously:
- creating a tutorial buildings raster
- uploading it to your google drive [or any other storage]
- sharing a link to the file
- someone from maintainers will need to download the buildings raster, include it into the tutorial datakit and upload to PyPSA-meets-Earth google drive and Zenodo
- updating links to the tutorial datakit in [bundle_config.yaml](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/configs/bundle_config.yaml)

Personally, I feel that adding a tutorial bundle is worth a dedicated PR, while for the current one it may be enough to do some local testing. 

@virio-andreyana @davide-f what is your opinion on that?

virio-andreyana:
Alright, I've created a simplified model for NG and BJ and store it in gDrive according to @ekatef suggestion. it looks like whats left is some documentations. The sector documentation is still ongoing https://github.com/pypsa-meets-earth/pypsa-earth/pull/1494. Where should I place the citation for [Hideaki Obane’s paper](https://eneken.ieej.or.jp/data/12710.pdf) because they inspire this work?

virio-andreyana:
This code is pretty mature. By using parquet instead of CSVs, the saving and retrieving time is much faster, while being 50-60% smaller in memory size. 

Docstrings are added in the code for better documentation

@ekatef I'd say it's ready to merged

config.default.yaml
```diff
@@ -23,6 +23,7 @@ enable:
   retrieve_databundle_sector: true
   retrieve_cost_data: true # true: retrieves cost data from technology data and saves in resources/costs.csv, false: uses cost data in data/costs.csv
   download_osm_data: true # If 'true', OpenStreetMap data will be downloaded for the above given countries
+  download_global_buildings: false # If 'true', GlobalMLBuildingFootprints data will be downloaded for the above given countries
```
ekatef, line 26: I wonder what would happen if we leave `download_global_buildings: false` while changing `use_building_size` to `true`? My understanding is that it'll lead to:
1) an attempt to use the existing tutorial data if `NG` and/or `BJ` are being modelled
2) a complain that Snakemake can't find the input files

If this understanding is correct, the second edge case could be probably addressed by having a check for that and a clear error message. Not sure it makes sense to add this in this PR but may be worth tracking in a dedicated issue

config.default.yaml
```diff
@@ -657,7 +658,20 @@ sector:
   efficiency_heat_gas_to_elec: 0.9
 
   electricity_distribution_grid: true # adds low voltage buses and shifts AC loads, BEVs, heat pumps, and resistive heaters, micro CHPs to low voltage buses if technologies are present
-  solar_rooftop: true # adds distribution side customer rooftop PV (only work if electricity_distribution_grid: true)
+  solar_rooftop: # adds distribution side customer rooftop PV (only work if electricity_distribution_grid: true)
+    enable: true
+    kW_per_m2: 0.1
+    m2_per_person: 20
+    use_building_size: false
+    install_ratio:
```
ekatef, line 671: Not sure `install_ratio` is crystal clear. Could it make sense to add a clarification comment for that? E.g. we can directly use a definition from Hideaki Obane’s paper:

```suggestion
    # proportion of rooftop suitable for PV installation
    install_ratio:
```

configs/bundle_config.yaml
```diff
@@ -46,6 +46,18 @@ databundles:
...
+    countries: [NG, BJ]
+    tutorial: true
+    category: global_buildings
+    destination: "data"
+    urls:
+      gdrive: https://drive.google.com/file/d/1nnseF1A740Gp0IaryeCQ8IRcEF91-aUe/view?usp=drive_link
+    output:
+    - data/global_buildings/NG_global_buildings_raw.parquet  # needed in cluster_global_buildings
+    - data/global_buildings/BJ_global_buildings_raw.parquet  # needed in cluster_global_buildings
+
```
ekatef, line 60: That is really cool to have it in the databundle

config.default.yaml
```diff
@@ -657,7 +658,20 @@ sector:
...
+      0: 0
+      10: 0.3
+      100: 0.36
+      200: 0.41
+      450: 0.49
+      2300: 0.66
```
ekatef, line 677: Completely agree that a reference could be helpful here for the paper used as a data source

config.default.yaml
```diff
@@ -657,7 +658,20 @@ sector:
...
+    tolerance: 100
```
ekatef, line 679: It could be worth explaining what exactly `tolerance` mean as not all users may be comfortable to check the code. What if we'd add a definition following the doc string from `calculate_solar_rooftop_area` e.g.:
```suggestion
     # maximum distance [km] to allocate a building within the nearest bus shapes
    tolerance: 100
```

scripts/download_global_buildings.py
```diff
@@ -0,0 +1,162 @@
...
+    if not os.path.exists(global_buildings_path) or update is True:
+
+        country_name = {
+            "CongoDRC": "Democratic Republic of the Congo",
+            "RepublicoftheCongo": "Republic of the Congo",
+            "FYROMakedonija": "North Macedonia",
+            "KingdomofSaudiArabia": "Saudi Arabia",
+            "SultanateofOman": "Oman",
+            "IsleofMan": "Isle of Man",
+        }
```
ekatef, line 58: Are those regions only ones which need a special treatment for naming? I think it's perfectly fine to cover only a part of regional-specific naming exceptions we just need to make sure an issue is opened to facilitate future improvements

scripts/prepare_sector_network.py
```diff
@@ -2851,15 +2851,53 @@ def add_electricity_distribution_grid(n, costs):
     n.links.loc[mchp, "bus1"] += " low voltage"
 
     if options.get("solar_rooftop", False):
-        logger.info("Adding solar rooftop technology")
+        if isinstance(options["solar_rooftop"], dict):
```
ekatef, line 2854: Generally, using if + a logical AND is more clear as compared with the nested if's. But I think that is alright for the current implementation thanks to the clear naming and neat implementation otherwise

ekatef (commented):
Hello @virio-andreyana, that is magnificent! Happy to see that you have managed to decrease the computational time and many thanks for adding the datasets to the tutorial databundles.

Have added a few comments to the code which mainly relates to explanations of new configuration parameters. There are also a couple of points related to edge cases but I think they are more suitable to be tracked as issues and tackled in further PRs.

ekatef:
Checking results of the objective function, I see that the objective has remained basically the same which means that the current upstream assumptions are consistent with results or a refined approach for NG-BJ region used in a tutorial run. [NB [this run](https://github.com/pypsa-meets-earth/pypsa-earth/actions/runs/20182299635) for the PR before merging the latest changes]

ekatef:
> Alright, I've created a simplified model for NG and BJ and store it in gDrive according to @ekatef suggestion. it looks like whats left is some documentations. The sector documentation is still ongoing #1494. Where should I place the citation for [Hideaki Obane’s paper](https://eneken.ieej.or.jp/data/12710.pdf) because they inspire this work?

Many thanks for the databundles. As a coordination question, have you used PyPSA-meets-Earth google drive?

For the citation, I completely agree that we must cite properly and see that you have added a reference to the doc string. For now, I'd say it's enough. What else can be probably worth is adding a graphic citation of their assumptions behind `installation_ratio` values:

image

ekatef (approved):
I think the PR is ready to be merged while it'd be good to track the remaining questions in a dedicated issue. Many thanks for the amazing contribution @virio-andreyana!

ekatef:
Merged 😄 
A follow-up issue opened #1693

PR #1630, opened 2025-11-09
Bugfix: Avoid creating duplicate conventional generators by setting `keep_existing_capacities` to `false`

virio-andreyana:
This issue is flagged by @sophiabrl

There is an inaccuracy in modeling myopic scenarios: the capacities of conventional power plants keep increasing even though those carriers are set as non-extendable. After tracing the issue, the underlying cause appears to be the following:

1. In the `rule prepare_sector_network`, the parameter `keep_existing_capacities` is set to `true` by default (unlike in PyPSA-Eur, where it is `false`). This causes conventional capacities calculated in `prepare_sector` to be re-added.
2. In the function `add_generation()` from the same rule, all power plants listed under `sector: conventional_generation` are added. In PyPSA-Eur, this originally only included gas power plants, but was extended here to cover other fossil plants (possibly to reflect interest in expanding fossil capacity or adding CCs). Importantly, these power plants lack a `build_year`.
3. Since `build_year` is missing, the `rule add_brownfield` triggers the function `add_build_year_to_new_assets()`, which assigns these fossil capacities the `build_year` of the current `planning_horizon`. 
4. In the following planning horizon, the same rule detects these assets and re-adds them as existing capacities.
5. Consequently, **the capacity of conventional power plants multiplies with each new planning horizon**.

In most cases, the function `add_power_capacities_installed_before_baseyear()` within the `rule add_existing_baseyear` already adds all conventional capacities along with their respective `build_year`s. Therefore, `add_generation()` is primarily relevant only for extendable carriers.

This pull request introduces the following changes:

* Set `keep_existing_capacities` to `false` by default. This option should only be enabled in **non-myopic** sector models where the user intends to include all conventional capacities regardless of their construction year. (Alternatively, to make it compatible with myopic settings, we would need to assign `build_year`s to all capacities in `prepare_sector_networks` and preceding rules, which would be quite complex.)
* Add `CCGT: gas` to `sector: conventional_generation` to align with PyPSA-Eur.
* Remove newly added links that have zero capacity and are not extendable in  `add_generation()`

## Checklist

### DIFF
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -747,11 +747,12 @@ sector:
 
   conventional_generation: # generator : carrier
     OCGT: gas
+    CCGT: gas
     oil: oil
     coal: coal
     lignite: lignite
     biomass: biomass
-  keep_existing_capacities: true
+  keep_existing_capacities: false
 
 solving:
   options:
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -39,6 +39,8 @@ This part of documentation collects descriptive release notes to capture the mai
 
 **Minor Changes and bug-fixing**
 
+* Avoid creating duplicate conventional generators by setting `keep_existing_capacities` to `false`. Convert CCGT as links in `prepare_sector_network` `PR #1630 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1630>`__
+
 * Refine load shedding capacity calculation to use bus-specific maximum loads instead of fixed large values, improving solver performance and numeric stability. Rename load shedding generator carrier to 'load shedding' (now) instead of just 'load' (previously) `PR #1581 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1581>`__
 
 * The configuration setting for ``focus_weights`` has been moved from ``focus_weights:`` to ``cluster_options: focus_weights:``. Backwards compatibility to old config files is maintained `PR #1565 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1565>`__
--- a/scripts/add_brownfield.py
+++ b/scripts/add_brownfield.py
@@ -73,6 +73,12 @@ def add_brownfield(n, n_p, year):
         c.df[f"{attr}_nom"] = c.df[f"{attr}_nom_opt"]
         c.df[f"{attr}_nom_extendable"] = False
 
+        # remove assets if name already exist in the new network
+        n_p.mremove(
+            c.name,
+            c.df.index.intersection(getattr(n,c.list_name).index)
+        )
+
         n.import_components_from_dataframe(c.df, c.name)
 
         # copy time-dependent
--- a/scripts/add_brownfield.py
+++ b/scripts/add_brownfield.py
@@ -74,10 +74,7 @@ def add_brownfield(n, n_p, year):
         c.df[f"{attr}_nom_extendable"] = False
 
         # remove assets if name already exist in the new network
-        n_p.mremove(
-            c.name,
-            c.df.index.intersection(getattr(n,c.list_name).index)
-        )
+        n_p.mremove(c.name, c.df.index.intersection(getattr(n, c.list_name).index))
 
         n.import_components_from_dataframe(c.df, c.name)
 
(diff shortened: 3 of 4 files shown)


ekatef:
Hey @virio-andreyana, many thanks for the investigation!

@yerbol-akhmetov I remember you have been working a lot to improve the implementation for myopic optimisation. Would you have time to review the PR?

yerbol-akhmetov:
> Hey @virio-andreyana, many thanks for the investigation!
> 
> @yerbol-akhmetov I remember you have been working a lot to improve the implementation for myopic optimisation. Would you have time to review the PR?

Hi, @virio-andreyana, @ekatef. Thanks for the PR. I think here revising `prepare_sector_network` is dealing with the result, while the cause of such bug is myopic implementation. The functions there needs revision. Maybe we need to prioritize it. We have managed to make myopic work well in `efuels-supply-potentials` repo and `keep_existing_capacities` was not causing duplication. Otherwise, setting it false make capacity related powerplant information not transferred for conventional generation units when modelled as link. We can merge this PR as a fast ad-hoc solution to a problem. However, I suggest turning it true, after revision of myopic implementation.

That `keep_existing_capacities` is only available in PyPSA-Earth. Though we initially tested in PyPSA-Eur for EEE project, but did not opened a PR to upstream.

virio-andreyana:
> Hi, @virio-andreyana, @ekatef. Thanks for the PR. I think here revising `prepare_sector_network` is dealing with the result, while the cause of such bug is myopic implementation. The functions there needs revision. Maybe we need to prioritize it. We have managed to make myopic work well in `efuels-supply-potentials` repo and `keep_existing_capacities` was not causing duplication. Otherwise, setting it false make capacity related powerplant information not transferred for conventional generation units when modelled as link. We can merge this PR as a fast ad-hoc solution to a problem. However, I suggest turning it true, after revision of myopic implementation.
> 
> That `keep_existing_capacities` is only available in PyPSA-Earth. Though we initially tested in PyPSA-Eur for EEE project, but did not opened a PR to upstream.

Thanks, @yerbol-akhmetov, for joining the discussion! I didn’t notice any significant differences in the implementation of the `get_capacities_from_elec()` or `add_generation()` functions across PyPSA-Earth, PyPSA-Eur, or efuels-supply-potentials.

You are right though when it comes to transferring the information of the capacity, however, they didn't save the information of `build_year` which in theory and in my practice seems to be the root of the issue I’m currently facing.

There also appears to be some overlap between the efforts of `add_generation()` in `prepare_sector_network.py` and `add_power_capacities_installed_before_baseyear()` in `add_existing_baseyear.py`. I’m curious to know what changes or additional information you’ve preserved that prevented you from encountering this problem.

yerbol-akhmetov (approved):
Hi, @virio-andreyana. Thank you for the PR. Implementation seems neat. I have tested it locally and doubling of conventional capacities are prevented. I think we can merge it. What do you think, @ekatef?

ekatef:
> Hi, @virio-andreyana. Thank you for the PR. Implementation seems neat. I have tested it locally and doubling of conventional capacities are prevented. I think we can merge it. What do you think, @ekatef?

Hey @yerbol-akhmetov, many thanks for taking care about the PR. I think you have the best overview on the development strategy for myopic and trust your opinion. Personally, I'm very much pro merging this PR once CI passes, according to the development plan we have discussed with you. Just waiting until #1653 will be in main to introduce it also there and get CI back 🚀

@virio-andreyana sorry for taking a while in handling your PR. As mentioned, there have been some issues with CI and some coordinated team work has been needed data management. But we are getting there and looking forward to accommodate your contribution.

virio-andreyana:
No worries @ekatef and @yerbol-akhmetov. The dual strategy of pushing this bugfix early in PyPSA-ASEAN, and have a complete check when implementing this in PyPSA-Earth takes the best of both world. Looking forward for it to be merged!

yerbol-akhmetov:
Thank you @virio-andreyana for the PR and @ekatef for support. PR is merged.

ekatef:
Happy to see the PR completed 💚 
Amazing team work @virio-andreyana @yerbol-akhmetov!

PR #1650, opened 2025-11-26
Adjust default HiGHS parameters

ekatef:

Parameters used to solve a model with HiGHS are adjusted in improve convergence.

## Checklist

- [ ] Code and workflow changes are sufficiently documented including ammending docstrings for meaningful functions.


davide-f (commented):
Thanks @ekatef !

Could you please clarify the source of the proposal?

FYI: adjusting some parameters improves convergence for small instances but worsen performances on large ones.

For example highs default primal_feasibility_tolerance is 1e-7 (higher than the proposed one) and default ipm_optimality_tolerance is 1e-8
https://ergo-code.github.io/HiGHS/dev/options/definitions

in gurobi equivalent default tolerances are around 1e-6

https://docs.gurobi.com/projects/optimizer/en/current/reference/parameters.html OptimalityTol (equivalent of dual feasibility tol) and FeasibilityTol (equivalent of primal_feasibility_tolerance)

Many thanks!

ekatef:
> Thanks @ekatef !
> 
> Could you please clarify the source of the proposal?
> 
> FYI: adjusting some parameters improves convergence for small instances but worsen performances on large ones.
> 
> For example highs default primal_feasibility_tolerance is 1e-7 (higher than the proposed one) and default ipm_optimality_tolerance is 1e-8 https://ergo-code.github.io/HiGHS/dev/options/definitions
> 
> in gurobi equivalent default tolerances are around 1e-6
> 
> https://docs.gurobi.com/projects/optimizer/en/current/reference/parameters.html OptimalityTol (equivalent of dual feasibility tol) and FeasibilityTol (equivalent of primal_feasibility_tolerance)
> 
> Many thanks!

Hey @davide-f, many thanks for looking into that. The PR resulted from some convergence issues I have observed when modelling Colombian power system, especially for the scenarios for which shedding played a considerable role.

With the current HiGHS settings, I have faced convergence issues during a post-solve phase with the following warning: `WARNING: LP solver claims optimality, but with WARNING: 2.57e-05 relative P-D objective error (tolerance = 1e-07) WARNING: Model status changed from "Optimal" to "Unknown" since relative violation of tolerances is 257`. Agree however that the particular value might be problem-specific. It looks though that our current defaults are significantly lower as compared with HiGHS defaults:
-  `primal_feasibility_tolerance`: 1e-5 vs 1e-07
- `dual_feasibility_tolerance`: 1e-5 vs 1e-07
- `ipm_optimality_tolerance`: 1e-4 vs 1e-08

For `small_matrix_value` and `large_matrix_value ` I have just returned HiGHS defaults, while excluding `parallel` seems to be my mistake (fixed sorry for that!).

Do you think it'll be safer to adjust tolerances to match the default values? I'd be also happy to have some lines on how to debug such solver-related convergence issues. Could you have any ideas on that?

davide-f:
> > Thanks @ekatef !
> > Could you please clarify the source of the proposal?
> > FYI: adjusting some parameters improves convergence for small instances but worsen performances on large ones.
> > For example highs default primal_feasibility_tolerance is 1e-7 (higher than the proposed one) and default ipm_optimality_tolerance is 1e-8 https://ergo-code.github.io/HiGHS/dev/options/definitions
> > in gurobi equivalent default tolerances are around 1e-6
> > https://docs.gurobi.com/projects/optimizer/en/current/reference/parameters.html OptimalityTol (equivalent of dual feasibility tol) and FeasibilityTol (equivalent of primal_feasibility_tolerance)
> > Many thanks!
> 
> Hey @davide-f, many thanks for looking into that. The PR resulted from some convergence issues I have observed when modelling Colombian power system, especially for the scenarios for which shedding played a considerable role.
> 
> With the current HiGHS settings, I have faced convergence issues during a post-solve phase with the following warning: `WARNING: LP solver claims optimality, but with WARNING: 2.57e-05 relative P-D objective error (tolerance = 1e-07) WARNING: Model status changed from "Optimal" to "Unknown" since relative violation of tolerances is 257`. Agree however that the particular value might be problem-specific. It looks though that our current defaults are significantly lower as compared with HiGHS defaults:
> 
> * `primal_feasibility_tolerance`: 1e-5 vs 1e-07
> * `dual_feasibility_tolerance`: 1e-5 vs 1e-07
> * `ipm_optimality_tolerance`: 1e-4 vs 1e-08
> 
> For `small_matrix_value` and `large_matrix_value ` I have just returned HiGHS defaults, while excluding `parallel` seems to be my mistake (fixed sorry for that!).
> 
> Do you think it'll be safer to adjust tolerances to match the default values? I'd be also happy to have some lines on how to debug such solver-related convergence issues. Could you have any ideas on that?

If we fall back to default values, I'd suggest to drop the entry instead.
Having tolerances tigther than the default would look risky, especially with 1e-8 when we have very large numbers in our model, it makes it more likely to fall into numerical troubles.

How large was the instance you tested? was it power-only or sector-coupled?

To me 1e-6 may look more reasonable (same as gurobi), though it would be better testing it

small_matrix_value of 1e-6 may look reasonable as generally we have very large numbers in the matrix, likewise, we don't have them so large.
large_matrix_value of 1e9 may look risky as indeed we have big numbers

ekatef:
> If we fall back to default values, I'd suggest to drop the entry instead. Having tolerances tigther than the default would look risky, especially with 1e-8 when we have very large numbers in our model, it makes it more likely to fall into numerical troubles.
> 
> How large was the instance you tested? was it power-only or sector-coupled?
> 
> To me 1e-6 may look more reasonable (same as gurobi), though it would be better testing it
> 
> small_matrix_value of 1e-6 may look reasonable as generally we have very large numbers in the matrix, likewise, we don't have them so large. large_matrix_value of 1e9 may look risky as indeed we have big numbers

My modelling instance was a small one: power-only model with 10 or 18 nodes. I'd expect though that HiGHS is more likely to be used on smaller models, at least before it's new version will be available.

What if we would use this PR to add some hints on HiGHS parameters which work well on smaller models? I can comment out most of the suggested changes and add a comment which explains that they are worth trying if convergence issues appear especially if a model is small? [Btw, I wonder if it may be realistic to sketch some approach which allows for automated fitting of solver parameters to improve and facilitate convergence? Out of question for this PR obviously I'm just curious to understand principal feasibility / potential utility]

For the actionable parameters change, only `large_matrix_value` is safe to modify changing from the current `1e9` to `1e15`, right? Or do you think shifting to `1e-6` also makes sense for some tolerances?

davide-f:
> > If we fall back to default values, I'd suggest to drop the entry instead. Having tolerances tigther than the default would look risky, especially with 1e-8 when we have very large numbers in our model, it makes it more likely to fall into numerical troubles.
> > How large was the instance you tested? was it power-only or sector-coupled?
> > To me 1e-6 may look more reasonable (same as gurobi), though it would be better testing it
> > small_matrix_value of 1e-6 may look reasonable as generally we have very large numbers in the matrix, likewise, we don't have them so large. large_matrix_value of 1e9 may look risky as indeed we have big numbers
> 
> My modelling instance was a small one: power-only model with 10 or 18 nodes. I'd expect though that HiGHS is more likely to be used on smaller models, at least before it's new version will be available.
> 
> What if we would use this PR to add some hints on HiGHS parameters which work well on smaller models? I can comment out most of the suggested changes and add a comment which explains that they are worth trying if convergence issues appear especially if a model is small? [Btw, I wonder if it may be realistic to sketch some approach which allows for automated fitting of solver parameters to improve and facilitate convergence? Out of question for this PR obviously I'm just curious to understand principal feasibility / potential utility]
> 
> For the actionable parameters change, only `large_matrix_value` is safe to modify changing from the current `1e9` to `1e15`, right? Or do you think shifting to `1e-6` also makes sense for some tolerances?

I'd be more inclined to have tolerances 1e-6 that is in the between between current values and defaults
Moving large values to 1e15 should be safe

ekatef:
> I'd be more inclined to have tolerances 1e-6 that is in the between between current values and defaults Moving large values to 1e15 should be safe

I trust you intuition. Thanks for the hint! Updated

`run_crossover` has been changed to `on` which is consistent with the defaults though I don't have a strong opinion on that

davide-f (approved):
This is ready to merge for me.
Optionally, we could have highs as default solver for the tutorial. What do you think?

ekatef:
> This is ready to merge for me. Optionally, we could have highs as default solver for the tutorial. What do you think?

Sounds a great idea. Let's test :)

davide-f (approved):

PR #1678, opened 2025-12-10
Bugfix: fix missing baseyear capacity, differentiate CHP and PP biomass for bioenergy

virio-andreyana:

After that pull request is merged, I've did some validation analysis and it seems we have a minor bug that filters out gas and biomass in `add_power_capacities_installed_before_baseyear(n, grouping_years, costs, baseyear)`

Cause of missing capacity:
```python
if generator not in vars(spatial).keys():
```
Solution:
```python
if carrier[generator] not in vars(spatial).keys():
```

While on this section, since we track most of the biomass capacity from [Global Bioenergy Power Tracker](https://globalenergymonitor.org/projects/global-bioenergy-power-tracker/) Most of the capacity doesn't seem to be CHP plants. Especially those located in hot climate. I was wondering if we can revert this back to `biomass` because that was the case if the capacity is added in `rule prepare_sector_network`

What's your opinion on the biomass parts?

## Checklist


scripts/add_existing_baseyear.py
```diff
@@ -363,44 +363,44 @@ def add_power_capacities_installed_before_baseyear(n, grouping_years, costs, bas
...
+                #         carrier=generator,
+                #         p_nom=new_capacity / costs.at[key, "efficiency"],
+                #         capital_cost=costs.at[key, "fixed"]
+                #         * costs.at[key, "efficiency"],
+                #         marginal_cost=costs.at[key, "VOM"],
+                #         efficiency=costs.at[key, "efficiency"],
+                #         build_year=grouping_year,
+                #         efficiency2=costs.at[key, "efficiency-heat"],
+                #         lifetime=lifetime_assets.loc[new_capacity.index],
+                #     )
```
davide-f: Was this intended? this alters the previous formulation.
Were you proposing to change only the previous line?

davide-f (commented):
Hello @virio-andreyana :D
Many thanks for raising the issue! Just a minor comment

scripts/add_existing_baseyear.py
virio-andreyana: Yes, this is the part where we set the carrier `biomass` as default instead of `urban central solid biomass CHP`, hence this section is no longer necessary and can be hashed.

virio-andreyana:
Any chance we have a conclusion to the matter of using biomass instead of central solid biomass CHP as the default setting @davide-f?

For @ekatef maybe your heating and cooling workflow might be impacted by this changes, do you have suggestions to this PR?

davide-f (commented):
Hello Virio!
Agree with you that we shall come to a conclusion.
I summarize what I see and we can finalize also with katia.

This PR aims to simplify the biomass and avoid dropping these carriers.
In particular, the proposal is assume that biomass no longer provides CHP, while before that was the case.

As questions:
- with this PR, how can we preserve CHPs? after this PR, CHPs are no longer available with this carrier. If complex and not relevant, we can consider it for the future and add an issue on that. Opinion by @ekatef that is working on the heating part would be ideal
- currently, the biomass generators would have co2 emissions. I'd advise to make sure to exclude co2 emissions for biomass

virio-andreyana:
Thanks @davide-f for the follow up. Here is my answer to the two questions:
- Users are still able to add "central solid biomass CHP" in `prepare_sector_network` if they set both as true as shown below. I think this PR meets the users expectation better because if feels strange that CHP is automatically added even though one of these options are disabled.
```yaml
sector:
  enable:
    biomass: true
  chp: true
```
- Some users might disagree wherever biomass is considered CO2 neutral or not. The best compromise I think is to allow the users to overwrite CO2 intensity like so:
```yaml
costs:
  'CO2 intensity':
    biomass: 0
```
(Though I think this could be its own PR to align biomass emission) What do you think?

virio-andreyana:
Turns out that only `solid biomass` has `CO2 intensity`. `biomass` doesn't. So the current implementation already avoids biomass having emission.

ekatef:
Hello @virio-andreyana and @davide-f!

Have investigated the biomass part a bit, and I do agree with Virio that biomass isn't being handled in a perfectly consistent way. I think it's worth tackling this point in a systematic way as otherwise we can get some inconsistencies, such as losing some groups of technologies, as Davide has mentioned.

To my understanding, the current way to represent biomass-fuelled plants is as follows. 
1. We add biomass-fed power generators in `add_electricity`
2. In the sector-coupled version, we remove biomass-fuelled generators along with other conventional generators with `remove_elec_base_techs( )` and reimplement them with link representation:
- biomass-fuelled power plants must be added by `add_generation( )` together with other conventional generators
 - additionally, there is `add_biomass` function which adds `central solid biomass CHP` and `urban central solid biomass CHP CC` plants as a greenfield capacity
3. For myopic implementation, biomass is currently treated only as `urban central solid biomass CHP`

To me, it looks this PR is related to the previous fix #1630 for the myopic implementation and we may need to consider this PR as an interim solution, while #1698 will provide a more stable one. As a reminder, the idea of #1630 was to avoid re-adding installed capacities at each stage of a myopic simulation by setting `keep_existing_capacities: false`. So, I'm not sure that it'll work to delegate adding biomass-CHPs to `prepare_sector_network`. Am I missing something?

As for heating/cooling PR, I'm happy to adapt it to any changes elaborated in this PR, no problem for this side.

virio-andreyana:
@ekatef I think your representation of biomass-fueled generators are correct! And yes, #1698 will be the long term solution, while #1630 in combination with this PR is the interim fix. I didn't see adding biomass-CHPs only in `prepare_sector_network` as a problem.

I've noticed an issue in `add_biomass` related to the types of carriers biomass is being used (`solid biomass`, `biomass`, `biomass EOP`) and how this impacts the marginal cost of each carrier. But I think it should be its own PR (after this one is done)

ekatef:
> @ekatef I think your representation of biomass-fueled generators are correct! And yes, #1698 will be the long term solution, while #1630 in combination with this PR is the interim fix. I didn't see adding biomass-CHPs only in `prepare_sector_network` as a problem.

Hey @virio-andreyana, thanks for the response! I think we are converging here 🙂 

Great to hear you see the current implementation as a robust interim solution. Just to make sure I understand how it's working with `biomass CHP` in myopic implementation after changes from this PR are applied listing steps of the current workflow [please feel free to correct me if needed]:
1) `prepare_sector_network` adds extendable capacity of `biomass CHP` for a run on each time horizon [but neglecting with the existing installed capacity]
2) for the first time horizon, `add_existing_baseyear` adds the installed capacities for power plants which exist in the power system, considering biomass-fuelled plants as electricity-only
3) for all the consecutive planning years, `add_brownfield` transfers the installed capacities from the previous time horizon

So, the only data loss we can have due to the changes in this PR is cogeneration capabilities of `biomass CHP`s which still exist in the power system. If checking the global statistics, there are two points:
1) biomass CHPs are currently mainly used in Europe (source: [GLOBAL BIOENERGY STATISTICS 2022](https://www.worldbioenergy.org/uploads/221223%20WBA%20GBS%202022.pdf) by World Bioenergy Association, page 3)
> Europe is the world leader in producing heat from biomass in power
plants with a share of 88% globally
2) In Europe, biomass CHP is primary used in Nordic countries, and its' share is not crazy (source: [IEA Bioenergy: Countries’ Report – update 2024](https://www.ieabioenergy.com/wp-content/uploads/2025/01/CountriesReport2024_final.pdf))
> Levels of bioelectricity have been relatively stable (or only slightly growing) in the past decade and typically represent 2-5% of power production. There are some exceptions where higher levels (>15%) are reached, e.g. Denmark, Sweden and Finland; in these countries, bioelectricity is mainly produced through CHP installations where the heat is also used, so this evolution is also part of a strategy to displace fossil fuels in industry and the district heating sector.

Given that, I'd say that it could be probably acceptable to neglect with cogeneration capabilities of biomass CHPs for now, but open a dedicated issue on that to make sure we keep track of it. What do you think @virio-andreyana @davide-f?

ekatef:
> I've noticed an issue in `add_biomass` related to the types of carriers biomass is being used (`solid biomass`, `biomass`, `biomass EOP`) and how this impacts the marginal cost of each carrier. But I think it should be its own PR (after this one is done)

Agree that harmonising the way we are dealing with biomass in general could have a higher priority for now, and @virio-andreyana it would be great to have your review on that. I guess your experience with PyPSA-ASEAN would be highly relevant for that.

virio-andreyana:
As informed by Yerbol, It turns out that the `powerplants.csv` already contains a column that differentiate between powerplants (PP) and CHP. I restore most of the change and if there is any Bioenergy that also is a CHP, then it will consider as central solid biomass CHP

The changes are now minor and accommodate both needs. @ekatef @davide-f If the workflow runs successfully, I think we can merge it.

ekatef (approved):
Thank you @virio-andreyana!! It has been a long way but I think the result is completely worth it. A minimalistic and very elegant solution. Agree that the PR is ready to be merged.

ekatef:
As discussed with @yerbol-akhmetov, the PR is also relevant for advancing myopic implementation and will need to be accounted for in #1698 and #1734.

Merging and looking forward for further developments! Thanks a lot @virio-andreyana

Issue #1669, opened 2025-12-03
Improve representation of hydro generation


ekatef:
## Describe the feature you'd like to see

The issue is opened to coordinate developments for hydro representation.

Issue #1668, opened 2025-12-03
Improve data management


ekatef:
## Describe the feature you'd like to see

Retrieval of the databundles is one of the major steps of the initial modelling setup and increasing its' robustness in essential to improve user experience. Additionally, instabilities of external databases is one of the most frequent reasons of CI breaking.

It would be good to revise the current logic behind data management starting from creating a unified list of url links used by the code and addressing other related points which have been raised before.

willu47:
I suggest adding #1740 as a sub-issue too.

ekatef:
> I suggest adding [#1740](https://github.com/pypsa-meets-earth/pypsa-earth/issues/1740) as a sub-issue too.

Thank you @willu47 and completely agree. Done!

Issue #518, opened 2022-11-15
Revise retrieve_databundle_light


davide-f:
## Describe the feature you'd like to see

As discussed with @pz-max and @Hazem-IEG , the rule retrieve_databundle_light shall be revised to decompose it into several snakemake rules, one for every input data.
Examples are provided at [pypsa-eur](https://github.com/PyPSA/pypsa-eur/blob/ce5317b06d690b82c427b1e4ee0a00254acdbe20/Snakefile#L93)

davide-f:
Hydrobasins can now be easily downloaded by continent here https://www.hydrosheds.org/products/hydrobasins

willu47:
Over in open-energy-transition/pypsa-zambia/issues/9, we've been considering how to refactor the retrieval of data bundles. I'm copying the discussion so far here for better visibility and discussion with the community:

I'm investigating `scripts/retrieve_databundle_light.py` (links to https://github.com/pypsa-meets-earth/pypsa-earth/issues/518)

This script contains the logic for downloading, unzipping, and moving the online resources listed in the `config_databundles.yaml` file to the local working directory.

## Problem statement

The problem at the moment is that all of this logic is under a single rule in Snakemake. This causes the rule to fail and restart if there's a problem with any one of the [multiple] large downloads, leading to repeated queries of the online sources.  A side effect of this is that repeated downloads across all forks of PyPSA-Earth, as occurs during test runs, result in very high utilisation of the file hosts (the data volume for Zenodo tutorial files is about 20TB!). The new `Storage` feature in Snakemake will help avoid this, especially the snakemake_cached_http_plugin created for PyPSA-Eur. However, a sequential refactor of the script is required to integrate it.

## Description of the current situation

The `scripts/retrieve_databundle_light.py` contains the following logic:

1. functions to choose which databundles to retrieve, based on the regions selected in the main config. The purpose of this is to download only the relevant data for the selected region.  At the moment, this doesn't seem particularly well used, as the databundles are highly aggregated.  There's also a mixture of ISO2 and ISO3-letter country codes, as well as non-ISO descriptions, e.g. "Africa" and "Earth".

        get_best_bundles_by_category(country_list: list, category: str, config_bundles: dict, tutorial: bool, config_enable: dict)
        get_best_bundles(countries: list, config_bundles: dict, tutorial: bool, config_enable: dict, include_categories: list = [], exclude_categories: list = [])
[...]

PR #1681, opened 2025-12-11
Add IRENA statistics for hydro generation normalization

virio-andreyana:
# Related to https://github.com/pypsa-meets-earth/pypsa-earth/issues/1669

Hydropower capacity continues to expand across the Global South, so relying on outdated 2013 data would lead to an underestimation of its potential. In ASEAN for example, the results is 70 TWh where in reality hydropower produce 220 TWh in 2023 at least.

Unlike `powerplantmatching.data.IRENASTAT()`, which only provides capacity figures, this IRENA dataset also includes generation data. This historical hydropower generation can be used for normalization, and the dataset is available up to 2023.

There are several options for where we could integrate the IRENA data:

1. Store it directly in the `data` folder.
2. Download and store it during `build_renewable_profiles`.
3. Download it within `retrieve_databundle_light` (though this currently has issues).

## Checklist


virio-andreyana:
Nigeria hydropower generation
- IEA 2013: 5.273 TWh
- IRENA 2013: 6.127 TWh
- IRENA 2023: 9.426 TWh

Higher hydro generation leads to lower system cost

virio-andreyana:
Mention to @energyLS. The failing workflows are unrelated to the changes. It works locally

ekatef:
Hello @virio-andreyana, totally support the idea of the PR and sorry we put it on hold for a while. We had to introduce some changes to data management for keeping CI testing functional due to the changes in policies of some data storages. 

I see that your PR may experience similar troubles with `https://www.irena.org/-/media/Files/IRENA/Agency/Publication/2025/Jul/IRENA_Statistics_Extract_2025H2.xlsx`. It's accessible for manual download but leads to 403 error when trying to extract it with a script. A solution we found is re-hosting the datafiles in [(temporary_storage)](https://github.com/pypsa-meets-earth/temporary_storage) repo. I wonder if it could be feasible also with IRENA data. Are you by any chance aware if they license allows for re-hosting?

virio-andreyana:
Hi @ekatef, at least according to IRENAs [Terms and Conditions](https://www.irena.org/terms-and-conditions) section 4, it is allowed to reproduce their data for non-commercial use. Compared to other sources, IRENA seems to allow it.

Let me know if the temporary storage is ready to use. And what will be the permanent solution?

ekatef:
> Hi @ekatef, at least according to IRENAs [Terms and Conditions](https://www.irena.org/terms-and-conditions) section 4, it is allowed to reproduce their data for non-commercial use. Compared to other sources, IRENA seems to allow it.
> 

Hey @virio-andreyana, that is great news about re-hosting. Thanks for checking! Agree that it looks like we can re-host it, just need to amend a note notifying that no commercial usage is possible. We have also other datasets which restrict commercial applications so it's anyway point which is worth checking.

Feel free to add the data file to [temporary_storage](https://github.com/pypsa-meets-earth/temporary_storage) repo and mention it's attribution and license in README.

ekatef:
> Let me know if the temporary storage is ready to use. And what will be the permanent solution?

A permanent solution is to improve data management in a systematic way for which we have a list of tasks linked as sub-issues to #1668. 

For IRENA file, I'd say the following points are relevant:
1) add it to an inventory #1711 for data sources the workflow is currently using [needed to make sure we keep an overview]
2) prepare the databundles to be hosted in a more stable manner
re-host the source file to some more suitable place than github together (@davide-f has still done some quite comprehensive work re-arranging and cleaning databundles)
3) re-host IRENA files together with other databundles to some data server. An original idea was to use Zenodo but we have discussed also using data buckets hosted at some commercial provider
4*) ideally, we can consider revision of the current data workflow when moving data extraction into a dedicated script from which only those parts are triggered which are needed for a particular modelling implementation. To ensure flexibility, we can also split the current databundles into parts as #518 suggests.

@virio-andreyana let me know please if you have any comments or suggestions on that. I recognise my answer may be extensively lengthly but data management is one of the current priorities and it would be good to make sure we are coordinated in tackling it.

virio-andreyana:
I agree with all @ekatef suggestion here. How should we move forward from here? Should we wait for https://github.com/pypsa-meets-earth/pypsa-earth/pull/1711 to be merge first? or I can revert back to the IEA default and have this merged? (the second options requires a minor PR to change the yaml from IEA=>IRENA)

configs/bundle_config.yaml
```diff
@@ -185,6 +185,16 @@ databundles:
     output:
     - data/hydrobasins/hybas_world.shp
 
+  # global data for irena generation, capacity, heat generation and more
```
ekatef: Could you please add also an indication on a specific script where this source is used? It'll be very handy for implementing a "standartised" retrieval for this particular dataset

scripts/build_renewable_profiles.py
```diff
@@ -257,6 +257,58 @@ def get_eia_annual_hydro_generation(fn, countries):
     return df
 
 
+# Temporary function, to be migrate to retrieve databundle once working
+def extract_IRENA_energy_stats(file_path="data/IRENA_Statistics_Extract_2025H2.xlsx"):
+    import requests
```
ekatef: Are there specific reasons to put import inside the function? It facilitates an overview when all the imports are grouped in the beginning of the file, and that is the reason PEP8 standard [recommends](https://peps.python.org/pep-0008/#imports) `Imports are always put at the top of the file`

scripts/build_renewable_profiles.py
```diff
@@ -257,6 +257,58 @@ def get_eia_annual_hydro_generation(fn, countries):
...
+    df = pd.read_excel(fn, sheet_name="Country")
+
+    countries_iso3 = {coco.convert(names=name, to="ISO3"): name for name in countries}
+
+    df = (
+        df.loc[
+            lambda df: df.Technology.isin(["Renewable hydropower"])
+            & ~df["Producer Type"].isin(["All types"])
+            & df["ISO3 code"].isin(countries_iso3.keys())
+        ]
```
ekatef: Isn't using `lambda` a bit of overshoot there? I guess indexing by a boolean would do the job being more conscience. While `query` would be probably the shortest option.

scripts/build_renewable_profiles.py
```diff
@@ -257,6 +257,58 @@ def get_eia_annual_hydro_generation(fn, countries):
...
+        .assign(countries=lambda df: df["ISO3 code"].map(countries_iso3))
```
ekatef: If the idea is to replace ISO2 with ISO3 using `countries_iso3` dictionary, `replace(countries_iso3)` could be a simpler implementation.

scripts/build_renewable_profiles.py
```diff
@@ -257,6 +257,58 @@ def get_eia_annual_hydro_generation(fn, countries):
...
+            logger.warning("No internet connection and file not found locally.")
+            raise FileNotFoundError(
+                f"File {file_path} not found and cannot download from the internet."
+            )
+
+    return file_path
+
+
+def get_irena_annual_hydro_generation(fn, countries):
+
```
ekatef, line 267: It would be good to add a short docstring, including a brief mention of `df` format. I see that the function outputs the dataframe in wide format with the columns being ISO3 codes. It's worth noticing since it differents from long format used for most datasets (when country codes are placed in a dedicated column, e.g. [hydro_capacities.csv](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/data/hydro_capacities.csv).

config.default.yaml
```diff
@@ -369,8 +369,8 @@ renewable:
     clip_min_inflow: 1.0
     extendable: true
     normalization:
-      method: eia # 'hydro_capacities' to rescale country hydro production by using hydro_capacities, 'eia' to rescale by eia data, false for no rescaling
-      year: 2013 # (optional) year of statistics used to rescale the runoff time series. When not provided, the cutout weather year is used
+      method: irena # 'hydro_capacities' to rescale country hydro production by using hydro_capacities, 'eia' or 'irena' to rescale by eia or irena data, false for no rescaling
+      year: 2023 # (optional) year of statistics used to rescale the runoff time series. When not provided, the cutout weather year is used
```
ekatef, line 375: Let's keep here `2013`. That is crucial that hydro normalisation year corresponds to the cutout year since runoff strongly depends on the year.

ekatef (commented):
Hello @virio-andreyana, looks great contribution! Sorry it took a while to review. Have added a few technical comments. Could you please have a look?

ekatef:
> I agree with all @ekatef suggestion here. How should we move forward from here? Should we wait for #1711 to be merge first? or I can revert back to the IEA default and have this merged? (the second options requires a minor PR to change the yaml from IEA=>IRENA)

Hey @virio-andreyana, great that we are aligned! For the general improvement on the data management, good to be coordinated but I don't think we must care too much about it right now. #1711 should go after this PR, and I'll use it to harmonise retrieval data on fly across the scripts.

So, I think we must proceed with completing and merging this PR focusing on cross-checking the technical points I mentioned in the review. 

For the data management, the only obligatory point we must fix is CI which is currently failing due to some troubled in retrieving source data. My suggestion is to re-host IRENA data in our [temporary_storage](https://github.com/pypsa-meets-earth/temporary_storage) repo and replacing the original link to a re-hosted one, but happy to opt also for an alternative solution if you see it.

scripts/build_renewable_profiles.py
virio-andreyana: Noted! The resulting output for countries are in ISO2. Maybe I should rename some variables

config.default.yaml
virio-andreyana: Hmm, although runoffs strongly depends on the weather year, the total generation of most developing countries has been increasing simply because there are more hydros being built. I'm still in favor of using 2023 data because that year is available for most countries, but you can make the final call if 2013 is still better.

ekatef:
Adding some examples on year-to-year variation of hydro production

## Weather-dominated variations

image

Decreased generation in 2022 is linked with a pronounced [drought event](<https://publications.jrc.ec.europa.eu/repository/handle/JRC128974?utm_source=chatgpt.com>), while installed capacity was gradually increasing in the years before and after this drought episode:

image

## Policy-driven changes

image

When the installed hydro capacity is increasing rapidly, you can still notice large drought events (such as [2009-2010](https://www.sciencedirect.com/science/article/abs/pii/S027737912300015X?utm_source=chatgpt.com) drought, or [2023](https://www.yicaiglobal.com/news/20230620-03-hydropower-generation-in-chinas-sichuan-yunnan-continues-to-drop-in-may-amid-droughts?utm_source=chatgpt.com)). But their effect in noticeable only as rather small oscillations around the increasing trend determined by dynamics of the installed capacity

image

For hydro normalisation, it's more appropriate to use the weather year for weather-dominated variations (countries where the installed capacity is changing slowly), and technology-year for policy-determined changes (countries where installed capacity is changing quickly).

@virio-andreyana does it make sense for you?

config.default.yaml
ekatef: Mmm... Agree that both installed capacity and the available hydro potential play a role here. Depending on a particular country, different choices would be "correct" (have added a some details in this comment). Given the fact that PyPSA-Earth is widely used for Global South, agree with leaving 2023 here 🙂 

Probably, adding a comment may be helpful, something like:
```suggestion
     # use the weather year for weather-dominated variations (installed capacity is changing slowly)
     # the technology-year for policy-determined changes (installed capacity is changing quickly)
      year: 2023 # (optional) year of statistics used to rescale the runoff time series. When not provided, the cutout weather year is used
```

Snakefile
```diff
@@ -487,21 +487,31 @@ rule build_demand_profiles:
...
+HYDRO_PROFILES = {
+    "hydro_capacities": "data/hydro_capacities.csv",
+    "eia_hydro_generation": "data/eia_hydro_annual_generation.csv",
+    "irena_stats": "data/IRENA_Statistics_Extract_2025H2.xlsx",
+    "powerplants": "resources/" + RDIR + "powerplants.csv",
+}
+
+
+def inputs_hydro(w):
+    return HYDRO_PROFILES if w.technology == "hydro" else {}
```
ekatef, line 499: Quite a neat solution

ekatef (approved):
Hey @virio-andreyana, I very much like the changes! For retrieval, my feeling is that we can your design to unify dealing with data download along the workflow.

Overall, a very nice implementation, and ready to be merged. Thanks a lot for the amazing work!

ekatef:
I think we can merge once CI passes unless there are any further comments.

ekatef:
Ouch, it appears we have missed the release note... Thanks a lot for remembering about it @virio-andreyana!

Merging as agreed 🚀  

Shall also post an update in pypsa-earth Discord to flag the advancements and the strategy needed to ensure backward compatibility for the existing modelling workflows.

ekatef:
Merged 😄 
Thank you so much @virio-andreyana!!

PR #1684, opened 2025-12-13
Add heuristics to determine missing hydro technologies

virio-andreyana:

Following up from this pull request https://github.com/pypsa-meets-earth/pypsa-earth/pull/1681, it appears that the total sum of inflows is not fully accounted for for the hydro generations. Further investigation shows that hydropower generation potential is reduced whenever a technology is classified as run-of-river, since its maximum dispatch is capped by installed capacity.

Moreover, the current workaround for missing hydropower technologies is to define them as run-of-river, which further amplifies this issue.

This pull request introduces heuristics to infer whether a missing hydropower technology should be treated as reservoir-based or run-of-river. Specifically, if the energy-to-capacity ratio exceeds a threshold (e.g., 1.0), the plant is classified as reservoir-based; otherwise, it is considered run-of-river.

This is the new configuration
```yaml
renewable:
   hydro:
      hydro_min_inflow_pu: 1.0
```

## Checklist


virio-andreyana:
Mention to @energyLS . The failing workflows are unrelated to the changes. It works locally

scripts/add_electricity.py
```diff
@@ -567,6 +563,25 @@ def attach_hydro(n, costs, ppl):
                     .to_pandas()
                 )
 
+    # Heuristics for missing hydro technologies
+    if not tbd.empty:
+        inflow_pu_limit = inflow_t[tbd.index].divide(tbd["p_nom"], axis=1).sum() / 8760
```
ekatef: Not sure I get for which reason do we need `8760` here: shouldn't it be a number of hours calculated by `n.snapshots`? 

Also, it could be worth adding a comment on units of measurements since. They aren't completely clear from the context but are important for maintaining this part

ekatef (commented):
Hey @virio-andreyana, thanks for a neat solution! Completely agree it's a very handy option, and I expect it to be in quite a wide use. Have added a couple of comments related to methodology and it's representation. 

Apart of them, I wonder what is the specific connections between `hydro_min_inflow_pu` and `hydro_max_hours_default`. To my understanding, both are essentially the maximum number of hours for which energy can be stored by a reservoir hydro power plant, while `hydro_min_inflow_pu` is a default for the lowest value, while `hydro_max_hours_default` is a default for the average one. Is it correct, or am I missing something?

scripts/add_electricity.py
virio-andreyana: You're right! This looks more complicated than it needs to. 8760 only assumes if the inflow profile is has the time series of the complete weather year. Averaging it out is sufficient.

The units here are:
- `inflow_t` = MWh
- `tbd["p_nom"]` = MW
- `inflow_pu_limit` = Average MWh / MW = (MWh/h) / MW = no unit

the timesteps in `build_renewable_profile` which generates the hydro `inflow` has 1 hour resolution, independent from the PyPSA network. So using `mean()` should be safe

```suggestion
        inflow_pu_limit = inflow_t[tbd.index].mean() / tbd["p_nom"]  # Average MWh/MW
```

virio-andreyana:
Thanks for reviewing @ekatef 

- `hydro_max_hours_default` only relates to the `max_hour` pumped hydro storage if the countries are not listed in `data/hydro_capacities.csv`
- whereas `hydro_min_inflow_pu` relates to the missing hydro technologies where values higher than it becomes `hydro` and lower than it becomes `ror`.
- PHS is rare worldwide and I assumed that GEM categorize pumped storage well.
- So, there is no direct relation between `hydro_max_hours_default` and `hydro_min_inflow_pu` 

The unintended effects of setting `hydro_min_inflow_pu` at a value of 1.0 for example increases the number of hydro reservoirs, which subtracts the storage potential of PHS from `hydro_capacities.csv` and hence reduce the average PHS `max_hours` of each country. (only if `hydro_max_hours: "energy_capacity_totals_by_country"`)

But come to think of it `hydro_capacities.csv` doesn't really have a clear source. It's already 3 years old but it's the only one that has the hydro storage capacity called `E_store[TWh]` and `p_nom_discharge[GW]` of most countries, neither IEA nor IRENA have those data implemented in PyPSA-Earth. I don't have any solution in mind, and it should be its own PR

scripts/add_electricity.py
ekatef: Great, thank you for the revision and detailed explanations @virio-andreyana!
I think the code look good now. 

For the units part, I believe `[Average MWh]` is  rather `[MWh / 1]` as you divide by a number of hours when averaging, not on the time itself. So, inflow_pu_limit should have dimension of time `Average MWh / MW = (MWh/1) / MW = h` which also makes sense as that is energy-to-capacity ratio which has dimension of time. Given that, I wonder if we could improve naming here to make physical sense of `inflow_pu_limit` more clear. Currently, `_pu_` feels as an indication that the value is dimensionless which is not quite correct. What if we'd revise `inflow_pu_limit` into something like `en2cap_limit`?

I recognise that naming is a tricky part but it's also something which effects code readability quite strongly, so it may be worth to spend some thoughts here.

ekatef:
> Thanks for reviewing @ekatef
> 
> * `hydro_max_hours_default` only relates to the `max_hour` pumped hydro storage if the countries are not listed in `data/hydro_capacities.csv`
> * whereas `hydro_min_inflow_pu` relates to the missing hydro technologies where values higher than it becomes `hydro` and lower than it becomes `ror`.
> * PHS is rare worldwide and I assumed that GEM categorize pumped storage well.
> * So, there is no direct relation between `hydro_max_hours_default` and `hydro_min_inflow_pu`
> 
> The unintended effects of setting `hydro_min_inflow_pu` at a value of 1.0 for example increases the number of hydro reservoirs, which subtracts the storage potential of PHS from `hydro_capacities.csv` and hence reduce the average PHS `max_hours` of each country. (only if `hydro_max_hours: "energy_capacity_totals_by_country"`)
> 
> But come to think of it `hydro_capacities.csv` doesn't really have a clear source. It's already 3 years old but it's the only one that has the hydro storage capacity called `E_store[TWh]` and `p_nom_discharge[GW]` of most countries, neither IEA nor IRENA have those data implemented in PyPSA-Earth. I don't have any solution in mind, and it should be its own PR

Many thanks for the detailed explanation @virio-andreyana. I think you have given a perfect overview on the parameters from the functional perspective, and totally agree that we shouldn't try to solve all problems with a single PR. I'm rather trying to deepen my understanding of the physical part. The current situation is we classify all hydro as run-of-river by default should be certainly improved, and your implementation is certainly improving that. I'd like just make sure that we are well aware of the physical basis of the solution as it would facilitate a lot usage and maintenance of the solution.

As we discussed in [those comments](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1684#discussion_r2852200815) above, `hydro_min_inflow_pu` is the energy-to-power ratio which is measured in time units and has physical meaning of the time duration (in hours, minutes or seconds) that the system can operate while delivering its rated output (taken from the definition [here](https://gridpiq.pnnl.gov/v2-beta/doc/technologies/es/es-energy-power/)). Having `hydro_min_inflow_pu: 1` by default means that we assume that if a hydro power plan can provide the full rated power during 1 hour [since we measure time in hours in the model], this plant is reservoir one. 

You are completely right that that `hydro_max_hours_default` and  `hydro_min_inflow_pu` are different but I don't think they are completely independent. So, we need to make sure that the values we assume for them are consistent. My understanding is that having `1 hour` for the average energy-to-capacity ratio with `6 hours` as the maximum value since averaging normally is smoothing peaks. So, it's good as it's right now, we just could add a few more comments documenting physical assumptions.

doc/configtables/hydro.csv
```diff
@@ -14,3 +14,4 @@ normalization,--,dict,"When specified, it describes how to normalize hydro time
 --method,--,str,"Data source used to rescale the hydro runoff; option 'hydro_capacities' to use the provided 'data/hydro_capacities.csv', 'eia' or 'irena' for using the eia or irena file"
 --year,--,year,"(optional) Specify the desired year to be used for normalization, the default value corresponds to the cutout weather year"
 multiplier,--,float,"Multiplier factor of the rescaling process (default 1.0)"
+hydro_min_inflow_pu,--,float,"Minimum energy-to-capacity ratio used to classify hydropower plants with missing technology information: values above this threshold are treated as reservoir plants, while smaller values are classified as run-of-river."
```
ekatef: I think we must correct units of measurments here 

```suggestion
hydro_min_inflow_pu,hours,float,"Minimum energy-to-capacity ratio used to classify hydropower plants with missing technology information: values above this threshold are treated as reservoir plants, while smaller values are classified as run-of-river."
```

config.default.yaml
```diff
@@ -374,6 +374,7 @@ renewable:
      # the technology-year for policy-determined changes (installed capacity is changing quickly)
       year: 2023 # (optional) year of statistics used to rescale the runoff time series. When not provided, the cutout weather year is used
     multiplier: 1.1 # multiplier applied after the normalization of the hydro production; default 1.0
+    hydro_min_inflow_pu: 1.0
```
ekatef: I think it would be great to add a line explaining the meaning of the parameter. Something like:

```suggestion
    hydro_min_inflow_pu: 1.0 # [hour] A threshold of energy to power ratio to classify hydro powerplants as reservoirs in case type is missed
```

ekatef (commented):
Hey @virio-andreyana, I think the only point left is fixing units of measurements for the threshold. Could you please go through the suggestions in the review? After that, the PR is ready to be merged.

ekatef (approved):
Perfect, thanks a lot for the adjustments @virio-andreyana!

PR #1638, opened 2025-11-17
MkDocs Migration

GbotemiB:
# Closes # (if applicable).

## Changes proposed in this Pull Request
This PR completes the migration from RST to MkDocs Material theme for PyPSA-Earth documentation.

## Checklist

- [x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [x] I tested my contribution locally and it seems to work fine.
- [ ] Code and workflow changes are sufficiently documented including ammending docstrings for meaningful functions.
- [ ] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`.
- [ ] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [ ] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [ ] Archives of the uploaded data do not have an enclosing folder and archive names correspond to the conventions of `configs/bundle_config.yaml`.
- [ ] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes, including reference to the requested PR.

PR #1596, opened 2025-10-02
Restore fetch isolated networks

davide-f:
# Closes #131 

This PR re-enables the use of fetch_isolated networks

## Checklist

- [x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [ ] I tested my contribution locally and it seems to work fine.
- [ ] Code and workflow changes are sufficiently documented.
- [ ] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`.
- [ ] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [ ] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [ ] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes, including reference to the requested PR.

PR #1698, opened 2026-01-02
Revise myopic implementation (Part 1)

yerbol-akhmetov:
Good day. This PR aims to revise implementation of myopic runs as well as addition of components.

## Checklist


ekatef:
Hey @yerbol-akhmetov, just flagging that #1705 is fixing reference objective values in CI summary table. I hope it can facilitate for checking of the objective function.

scripts/cluster_network.py
```diff
@@ -604,6 +620,105 @@ def cluster_regions(busmaps, inputs, output):
...
 
 
+def restore_base_carrier_names(n: pypsa.Network) -> None:
+    """
+    Restore carrier names from carrier_gy format (e.g., "solar-2020") to base carrier (e.g., "solar").
+
+    This is called after all aggregation operations to clean up carrier names while
+    preserving build year information in component names/indices.
+
+    Generator indices keep build year information (e.g., "US0 1 solar-2025"), but carrier becomes base ("solar").
```
ekatef: Could it be a good idea to move this function to _helpers module alongside with `get_base_carrier`?

scripts/cluster_network.py
```diff
@@ -604,6 +620,105 @@ def cluster_regions(busmaps, inputs, output):
...
+    n.generators["carrier"] = n.generators["carrier"].apply(get_base_carrier)
+
+    # Restore base carrier names for storage units
+    n.storage_units["carrier"] = n.storage_units["carrier"].apply(get_base_carrier)
+
+    # Remove year-tagged carriers
+    year_tagged_carriers = []
+    for carrier in n.carriers.index:
+        parts = carrier.rsplit("-", 1)
+        if len(parts) == 2 and parts[1].isdigit() and len(parts[1]) == 4:
```
ekatef: There is a fragment in `get_base_carrier` which must be aligned with this part. Could it be a good idea to wrap it up into a small dedicated function and call it from both `get_base_carrier` and `restore_base_carrier_names`?

scripts/add_electricity.py
```diff
@@ -360,15 +446,122 @@ def update_transmission_costs(n, costs, length_factor=1.0, simple_hvdc_costs=Fal
     )
 
 
+def get_grouping_year(build_year, grouping_years):
```
ekatef: I think we need to add a missed type hint for the function output

scripts/add_electricity.py
```diff
@@ -360,15 +446,122 @@ def update_transmission_costs(n, costs, length_factor=1.0, simple_hvdc_costs=Fal
...
+    """
+    Map build_year to the nearest grouping year (rounded up).
```
ekatef: A neat-picking comment: I think you mean mapping to the bins rather than rounding up, right? That is what np.digitize is doing which also makes perfect sense

scripts/add_electricity.py
```diff
@@ -360,15 +446,122 @@ def update_transmission_costs(n, costs, length_factor=1.0, simple_hvdc_costs=Fal
...
+    indices = np.digitize(build_year, grouping_years, right=True)
+    return np.take(grouping_years, indices)
+
+
+def aggregate_ppl_by_bus_carrier_year(ppl: pd.DataFrame) -> pd.DataFrame:
+    """
+    Aggregate power plants by (bus, carrier, grouping_year).
+
+    Creates a new carrier name with grouping year suffix (e.g., "CCGT-2020")
+    and aggregates capacity and other attributes.
```
ekatef, line 320: Those docstrings are awesome!

scripts/add_electricity.py
```diff
@@ -360,15 +446,122 @@ def update_transmission_costs(n, costs, length_factor=1.0, simple_hvdc_costs=Fal
...
-    line_length_factor,
+    n: pypsa.Network,
+    costs: pd.DataFrame,
+    ppl: pd.DataFrame,
+    input_files: dict,  # snakemake input
+    technologies: set,
+    extendable_carriers: dict,
+    line_length_factor: float,
 ):
+    """
```
ekatef, line 437: I think a type hint should still be added even if the function returns None to be consistent. What do you think?

scripts/build_powerplants.py
```diff
@@ -353,6 +353,9 @@ def replace_natural_gas_technology(df: pd.DataFrame):
         ppl_query
     )  # add carriers from own powerplant files
 
+    # define unique index
+    ppl = ppl.reset_index(drop=True)
+
```
ekatef, line 358: Just for my understanding: could you please clarify specific reasons we need a unique index? To avoid issues by further grouping?

scripts/simplify_network.py
```diff
@@ -297,7 +323,13 @@ def replace_components(n, c, df, pnl):
...
 
     for one_port in aggregate_one_ports:
-        df, pnl = aggregateoneport(n, busmap, component=one_port)
+        one_port_strategy = one_port_strategies.get(one_port, dict())
+        df, pnl = aggregateoneport(
+            n,
+            busmap,
+            component=one_port,
+            custom_strategies=one_port_strategy,
+        )
```
ekatef, line 330: Could you please give some hints on the reason this aggregation was needed? To give some context, `aggregateoneport` was creating some weird troubles for specific versions of PyPSA in the past. 

If you remember fetching isolated networks which stopped working, that was linked exactly with changed behaviour of `aggregateoneport`. That is the reason we have commented it out when restoring this fetch-network functionality in #1596: https://github.com/pypsa-meets-earth/pypsa-earth/blob/575c030591aa3a28ee9dcfd2dd89d89e2eec4cc9/scripts/simplify_network.py#L873

I think you may be perfectly right in the getting it back in this PR just sharing some context which can be potentially relevant

scripts/cluster_network.py
```diff
@@ -587,6 +590,19 @@ def clustering_for_n_clusters(
...
+        if not df.empty:
+            # Check if all values are either 0 or NaN
+            all_zero_or_nan = ((df == 0) | df.isna()).all().all()
+            if all_zero_or_nan:
+                # Make it empty with correct index but no columns
+                clustering.network.storage_units_t[attr] = pd.DataFrame(index=df.index)
+                logger.info(
+                    f"Cleared zero/NaN-filled storage_units_t.{attr} after clustering"
+                )
+
```
ekatef, line 609: Could you please remind for which reason has it been needed? I remember you mentioned the need to deal with NaNs in time-series for storage units but don't recall all the details

ekatef (commented):
Hey @yerbol-akhmetov, that is a pleasure to review your PR. Completely agree that having properly implemented type hinting and docstrings make the code much easier to work with. Have added a few comments and questions. Additionally, I see the objective function has noticeably changed for the monter-carlo test. Do you have any ideas on what could be the reason?

scripts/cluster_network.py
yerbol-akhmetov: Hi, @ekatef. Thank you for the suggestion. I have moved it to the `_helpers.py`.

scripts/cluster_network.py
yerbol-akhmetov: Hi, @ekatef. Thanks for the comment. Up until recently, there was an issue in `get_clustering_from_busmap` of PyPSA itself. In particular, after using this function, we get non-empty timeseries for storage units. It is because particular line shown below. It was previously adding timeseries even having empty DataFrame. Function `_import_series_from_df` was filling DataFrame with default values of attributes. For example, you will get `storage_units_t.p` filled with zero after using `get_clustering_from_busmap` function, while it should be just empty DataFrame. The function works fine for generators and links, but not for storage units. Therefore I have created a [PR](<https://github.com/PyPSA/PyPSA/pull/1528>) to PyPSA to make it consistent. The PR was merged and the fix is available after [PyPSA v1.1.0](https://github.com/PyPSA/PyPSA/releases/tag/v1.1.0). IN myopic PR, I have used fix to make timeseries empty. It is ad-hoc solution. We will remove this part of the code, when we update to PyPSA v1.1.0.

Image

scripts/cluster_network.py
yerbol-akhmetov: Thanks for noting it, @ekatef. I have put that repetitive part to a separate function.

scripts/build_powerplants.py
yerbol-akhmetov: Yes, exactly. `custom_powerplants` might have repetitive index. That will cause error during clustering. This line will help to prevent it.

scripts/simplify_network.py
yerbol-akhmetov: Hi, @ekatef. Thanks for the comment. I see your concern. I have not added aggregation here, it was already there. I have just made one port components aggregation strategy being read from the config. `aggregateoneport` will anyway be needed here, because otherwise the loads and storage units will not be clustered/aggregated. As you have noted, potential issue happens in `merge_into_network` function, which I have not touched. While this part of the code was existing in the current workflow.

scripts/add_electricity.py
yerbol-akhmetov: Hi, @ekatef. Thank you for noting it. I have added it.

scripts/add_electricity.py
yerbol-akhmetov: Thank you, I have fixed the docstring.

scripts/add_electricity.py
yerbol-akhmetov: Thank you, appreciate it.

scripts/add_electricity.py
yerbol-akhmetov: Thank you for noting it. I have fixed in several places too.

scripts/cluster_network.py
ekatef: Thank you! Looks quite neat I think 🙂

scripts/build_powerplants.py
ekatef: Great, many thanks for explaining!

scripts/cluster_network.py
ekatef: I think it's brilliant @yerbol-akhmetov! 
Thanks for providing the details. Could you please add a TODO to this fragment explaining that it won't be needed after PyPSA v1.1.0? Otherwise, looks great

scripts/simplify_network.py
ekatef: Thank you! I'm sure your implementation is correct just double-checking we don't miss anything important here. My hypothesis is that the previous issues could have been linked with the issue time-series for StorageUnits you have fixed, and it should be thus resolved with PyPSA v1.1.+

ekatef (commented):
Thank you @yerbol-akhmetov, I think the implementation looks perfect from the technical side. The difference in the function values is most likely linked with the changes introduced recently with merging some bug-fixes into the main (we'll update the reference values for the final CI run before merging).

To cross-check maintainability and overall functionality, I'll outline my understanding of the major design ideas, and please correct the points which are not captured precise enough.
1. In `add_electricity`, we split all the existing generators and storages by a year they were built and modify carriers adding an indication of the corresponding year bin (e.g. `CCGT-2015`)
2. In `elec.nc`, all the carriers of the existing generators have now also a year indicating their build year
3. Functionality of `simplify_network` has remained exactly the same [while the code quality has been improved]
4. When clustering in `cluster_network`, the year tag is dropped for all the carriers, and generators had to be additionally aggregated by non-tagged carrier for each bus

Could you please check wether it's correct understanding and update if needed?

yerbol-akhmetov:
> Thank you @yerbol-akhmetov, I think the implementation looks perfect from the technical side. The difference in the function values is most likely linked with the changes introduced recently with merging some bug-fixes into the main (we'll update the reference values for the final CI run before merging).
> 
> To cross-check maintainability and overall functionality, I'll outline my understanding of the major design ideas, and please correct the points which are not captured precise enough.
> 
> 1. In `add_electricity`, we split all the existing generators and storages by a year they were built and modify carriers adding an indication of the corresponding year bin (e.g. `CCGT-2015`)
> 2. In `elec.nc`, all the carriers of the existing generators have now also a year indicating their build year
> 3. Functionality of `simplify_network` has remained exactly the same [while the code quality has been improved]
> 4. When clustering in `cluster_network`, the year tag is dropped for all the carriers, and generators had to be additionally aggregated by non-tagged carrier for each bus
> 
> Could you please check wether it's correct understanding and update if needed?

Hi, @ekatef. Thanks for the summary. 
1. Yes, exactly. We already provided `build_year` in `add_electricity` for components. Now they have carrier with grouping year tag.
2. Yes, existing generators and storage units now have carrier with grouping year (`build_year`) tag. It is not exactly build year, but grouping year based on `build_year`. It is necessary for further aggregation/clustering happening in `simplify_network` and `cluster_network`. Because innate functions there aggregate nodes by carrier. Therefore, we need to have separate carrier names for existing generators built in different year. So we can differentiate then and decommission according to the horizon.
3. Yes, there is minor change in `simplify_network`. Other than that functionality is the same. Because we already differentiated generators by carrier and grouping years.
4. Yes, but let me clarify a bit. We have year tag in name of the component in index (eg. `US0 0 biomass-1990`) and in the carrier name (`biomass-1990`). Year tag in carrier names is dropped for both `overnight` and `myopic` optimization (eg. `biomass-1990` -> `biomass`). We no longer need it because it played its role in aggregation and clustering, later we just want to have carrier name as currently (simply `biomass` or `coal`). 
Year tag in component name in index is dropped only when we simulate with `overnight` foresight (eg. `US0 0 biomass-1990` -> `US0 0 biomass`). It is done to get exactly what we have currently in overnight simulation. We do not differentiate between build years, we just have carrier. There might be improvement in the future to be able to filter powerplants not only based on `powerplant_filter` in config, but using `planning_horizon`. But it is better to have in another PR. 
For myopic optimization, year tag in component name (e.g. `US0 0 biomass-1990`) remains in index of the components, because we need it to distinguish then and decommission when needed.

yerbol-akhmetov:
Thanks, @ekatef. I will add TODO regarding changes needed when we switch to PyPSA v1.1.0.

scripts/cluster_network.py
ekatef: Thanks for adding the comment! It'll be very helpful in maintainance

ekatef (commented):
Thanks a lot @yerbol-akhmetov for the detailed explanation. It would be a precious part of the documentation per se as myopic optimisation is generally not quite well understood. [But it's something to be added after #1638 is merged to avoid creating conflicts]

I think the implementation looks great, and the only point left is checking that we understand the effect on the objective function. I have an impression that the objective value has changes quite significantly for a few Monte-Carlo runs. It may be also fine but it would be good to understand a particular reason for that. Do you have any ideas what may be a possible explanation for that?

As a technical comment, the objective function has been changed recently as a result of merging some bug-fixes which makes comparison not especially convenient. Have opened #1749 to update the objective values in the upstream.

ekatef:
On the methodological level, it would be perfect to have a look of @davide-f. Davide, if you could have resources to support, the methodology is described in [this comment](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1698), while the major challenge is to understand the reason of objective changes which we observe for a few Monte-Carlo runs.

yerbol-akhmetov:
Hi, @ekatef. Although not visible by CI, I have noticed the issue with biomass in `add_existing_baseyear`. I have noticed it when simulating sector model for US with biomass entries. Here, full dataframe is assigned to subset of biomass. As a result, we get empty df for biomass. Thus, it is dropped. I will fix it and push changes here.

image

ekatef:
> Hi, @ekatef. Although not visible by CI, I have noticed the issue with biomass in `add_existing_baseyear`. I have noticed it when simulating sector model for US with biomass entries. Here, full dataframe is assigned to subset of biomass. As a result, we get empty df for biomass. Thus, it is dropped. I will fix it and push changes here.
> 
> image

Mmmm... I think it deserves a dedicated bugfix. Would you mind to open a focused PR on that?

yerbol-akhmetov:
> > Hi, @ekatef. Although not visible by CI, I have noticed the issue with biomass in `add_existing_baseyear`. I have noticed it when simulating sector model for US with biomass entries. Here, full dataframe is assigned to subset of biomass. As a result, we get empty df for biomass. Thus, it is dropped. I will fix it and push changes here.
> > image
> 
> Mmmm... I think it deserves a dedicated bugfix. Would you mind to open a focused PR on that?

I was investigating the issue, and it was going deeper in `add_existing_baseyear` as biomass and biomass CHP is split. I do not think we need to fix it. Because if we merge both myopic PRs simultaneously then the problem will not appear. Because in Revise myopic Part 2 we remove re-addition of power capacities altogether. So our work will be in vain.

ekatef:
> > > Hi, @ekatef. Although not visible by CI, I have noticed the issue with biomass in `add_existing_baseyear`. I have noticed it when simulating sector model for US with biomass entries. Here, full dataframe is assigned to subset of biomass. As a result, we get empty df for biomass. Thus, it is dropped. I will fix it and push changes here.
> > > image
> > 
> > 
> > Mmmm... I think it deserves a dedicated bugfix. Would you mind to open a focused PR on that?
> 
> I was investigating the issue, and it was going deeper in `add_existing_baseyear` as biomass and biomass CHP is split. I do not think we need to fix it. Because if we merge both myopic PRs simultaneously then the problem will not appear. Because in Revise myopic Part 2 we remove re-addition of power capacities altogether. So our work will be in vain.

Great! My major concern was to avoid increasing scope for this PR. I think it's totally fine to take `won't-fix` strategy which is also consistent with the approach we discussed #1678 (where actually the split was introduced).

So, let's focus on completing this and #1722.

ekatef:
@yerbol-akhmetov have checked the optimised networks which resulted from CI runs ([results-ubuntu](https://github.com/pypsa-meets-earth/pypsa-earth/actions/runs/23180553686/artifacts/5959175675) for CI run here vs [results-ubuntu](https://github.com/pypsa-meets-earth/pypsa-earth/actions/runs/23290083935/artifacts/6004439265) in the main), and it looks like clustering has changed in myopic implementation

When plotting the network for `tutorial` test case, I obtain the following pictures:
### main
image

### myopic
image

As a result, there are some differences in the installed capacity of lines and optimisation outputs are different for generation capacities:

```
(
    n_main.statistics.installed_capacity() - n_myopic.statistics.installed_capacity()
)
```
image

```
(
    (n_main.statistics.optimal_capacity() - n_myopic.statistics.optimal_capacity()) /
    n_main.statistics.optimal_capacity()
)
```
image

Do you have any ideas on what may be the issue with clustering?

yerbol-akhmetov:
Thanks for analysis @ekatef. I think the issue happens in `cluster_network`. Because after simplification we got exactly the same networks for `main` and `revise_myopic`. I have just showed the plot blow. Nodes and lines: `red` - main, `blue` - `revise_myopic`.

image

ekatef (approved):
Brilliant! The objective value is maintained precisely. Fantastic work @yerbol-akhmetov 🎉 

No more comments from my side, only couple of follow-up points.

1. It would be just good to craft an educational section regarding the idea of myopic optimisation in general which would fit good to continue Alessandro's intro to the optimisation. Not urgent though, just good to have in the roadmap.

2. I think it could be also good to have Davide's opinion on the implementation to understand the bigger picture better. In particular, if there may be any edge cases, or points to account for when applying the developed concept.

davide-f (commented):
Great work and thanks for my insights.

I am notifying this revision will likely be disrupting for some users, especially for validation.
I understand the elec network will contain {tech}-{year} carriers as well, thus when doing n.statistics() they will be all displayed.
I believe some notebooks in the documentation won't be reproducible, e.g. [this one](https://github.com/pypsa-meets-earth/documentation/blob/main/notebooks/viz/regional_transm_system_viz.ipynb). It is not critical though good to have that in mind.
Had you capacity, I'd suggest to test if [this](https://github.com/pypsa-meets-earth/documentation/blob/main/notebooks/network_analysis.ipynb) works as well as its [sec version](https://github.com/pypsa-meets-earth/documentation/blob/main/notebooks/network_analysis_sec.ipynb); if you don't have capacity, it is no problem.

I understand the carrier-X definition is needed as a fix in cluster network, and simplify_network where clustering is applied as well.

To share context also for future readers, what were the downsides of the other options you thought of? like making the splitting by build_year optional, limit the carrier-X usage to cluster_network or possibly fix the issue that makes clustering not feasible with multiple objects with different name and same carrier
You have worked a lot here and we can build a better architecture also for future PRs

Many thanks!

yerbol-akhmetov:
Hi, @davide-f. Thanks for the comments. I am investigating the effect of myopic implementation on visualization and validation scripts. Let's go one by one. 

## Regional transmission system visualization
The [notebook](<https://github.com/pypsa-meets-earth/documentation/blob/main/notebooks/viz/regional_transm_system_viz.ipynb>) provides a map of regional installed generation and line capacity. Fore revised myopic implementation, if we apply the notebook as it is with `elec.nc` including year-tagged carriers, then we get the following image (for `test/config.myopic.yaml`). The problem here is vivid, we have all carriers separate.

image

 In short, the plotting script needs some adjustments. We can import use/import `restore_base_carrier_names` function from pypsa-earth for that. This way, we just apply it before plotting. Thus, year tag from carrier is removed if exists. Below, I have plotted the same network with restoring the carrier names. The plot is similar to what we get if we used to have carriers without year-tag.

image

yerbol-akhmetov:
## Network analysis
The [notebook](<https://github.com/pypsa-meets-earth/documentation/blob/main/notebooks/network_analysis.ipynb>) that provides analysis on network works just fine with new myopic implementation. It is because `elec.nc` is used there only to plot *Onshore wind and Solar PV potential density*. In particular, `p_nom_max` is taken for `onwind` and `solar`. In new myopic implementation, extendable renewable generators have carrier name without year-tag (i.e. `solar` and `onwind`).  And their `p_nom_max` refers to newly installable potentials. 

In all other plots, network from `postnetworks/` is used. Those plots are fine because year-tag from carrier is removed in `cluster_network` and we do not have them in sector networks. I have tested the network and below I provided plots for networks produced using `test/config.myopic.yaml`.

image
image

This is one of the plots from this notebook:
image

yerbol-akhmetov:
## Network analysis sector
The [notebook](<https://github.com/pypsa-meets-earth/documentation/blob/main/notebooks/network_analysis_sec.ipynb>) works fine with new myopic implementation. Because we no longer have year tags in networks of sector model. I have tested it and it worked just fine.

Some example plots for `test/config.myopic.yaml`:

image

image

yerbol-akhmetov:
> Great work and thanks for my insights.
> 
> I am notifying this revision will likely be disrupting for some users, especially for validation. I understand the elec network will contain {tech}-{year} carriers as well, thus when doing n.statistics() they will be all displayed. I believe some notebooks in the documentation won't be reproducible, e.g. [this one](https://github.com/pypsa-meets-earth/documentation/blob/main/notebooks/viz/regional_transm_system_viz.ipynb). It is not critical though good to have that in mind. Had you capacity, I'd suggest to test if [this](https://github.com/pypsa-meets-earth/documentation/blob/main/notebooks/network_analysis.ipynb) works as well as its [sec version](https://github.com/pypsa-meets-earth/documentation/blob/main/notebooks/network_analysis_sec.ipynb); if you don't have capacity, it is no problem.
> 
> I understand the carrier-X definition is needed as a fix in cluster network, and simplify_network where clustering is applied as well.
> 
> To share context also for future readers, what were the downsides of the other options you thought of? like making the splitting by build_year optional, limit the carrier-X usage to cluster_network or possibly fix the issue that makes clustering not feasible with multiple objects with different name and same carrier You have worked a lot here and we can build a better architecture also for future PRs
> 
> Many thanks!

Thanks for the comment, @davide-f. Yes, having year-tag in carrier name is necessary for clustering. Because clustering functions currently use carrier and bus to aggregate the units. I do not see benefit of making year-tag optional. It is necessary for myopic optimization. We could add those `if` conditions to prevent year-tag for overnight optimization, but I think it will be too involved and too hand-made. The current approach seems more general.

In `add_electricity`, we add year tag to carrier name and component name (index). We could probably add year tag to carrier name in the beginning of `simplify_network` (by reading from component name, or build_year). This way we will have `elec.nc` more clean. However, we will still need year tag in carrier name in `simplify_network` and `cluster_network`. I think the current approach is more general and consistent (although very dramatic).

Another solution might be to adapt clustering function to allow having just carrier name. But it seemed for intrusive.

yerbol-akhmetov:
Generally, I think the changes to notebooks can be minimal. What do you think @davide-f ?

davide-f (commented):
Thanks a lot for the investigation, Yerbol. I fully agree that these changes in the notebook are doable.

I am sharing a possible alternative and leave it to you to judge whether this fits in this PR, a future PR, or is not needed.

My main concern is that changing the output of add_electricity by encoding carrier+year in the carrier field makes life harder for users. In particular, it reduces the usability of the intermediate network files for electric-only applications, and makes n.statistics() on intermediate outputs difficult to handle because the carrier is no longer preserved in its original form.

Therefore, I am wondering whether we could preserve the proper carrier in the output files while still keeping the current functionality. One possible approach would be the following:

In add_electricity, keep the current generator naming convention (bus + carrier + year), but store the actual carrier in the carrier field rather than carrier+year.

Then, at the beginning of simplify_network and cluster_network, introduce a direct temporary conversion from carrier to carrier+year only for the purpose of bypassing the clustering issue, and then convert it back from carrier+year to carrier at the end.

This would preserve the current functionality while also keeping the carriers consistent in the elec, simplify, and cluster outputs.

I would be interested to hear your view on the complexity of this approach, and whether you think it is possibly relevant for this PR or a future one

yerbol-akhmetov:
> Thanks a lot for the investigation, Yerbol. I fully agree that these changes in the notebook are doable.
> 
> I am sharing a possible alternative and leave it to you to judge whether this fits in this PR, a future PR, or is not needed.
> 
> My main concern is that changing the output of add_electricity by encoding carrier+year in the carrier field makes life harder for users. In particular, it reduces the usability of the intermediate network files for electric-only applications, and makes n.statistics() on intermediate outputs difficult to handle because the carrier is no longer preserved in its original form.
> 
> Therefore, I am wondering whether we could preserve the proper carrier in the output files while still keeping the current functionality. One possible approach would be the following:
> 
> In add_electricity, keep the current generator naming convention (bus + carrier + year), but store the actual carrier in the carrier field rather than carrier+year.
> 
> Then, at the beginning of simplify_network and cluster_network, introduce a direct temporary conversion from carrier to carrier+year only for the purpose of bypassing the clustering issue, and then convert it back from carrier+year to carrier at the end.
> 
> This would preserve the current functionality while also keeping the carriers consistent in the elec, simplify, and cluster outputs.
> 
> I would be interested to hear your view on the complexity of this approach, and whether you think it is possibly relevant for this PR or a future one

Hi, @davide-f. Yes, this was one of the options that I have looked. I think it is doable. I have not implemented it because it required extra steps of conversions. I can try that. I will let you know the results.

yerbol-akhmetov:
Hi, @davide-f. I have implemented your suggestion of having temporary `carrier -> carrier+year` inside of scripts, while returning to carrier names at the end of scripts. It worked well (CI passed and functionality retained). I will modify the #1722 accordingly. What do you think, @davide-f?

## After `add_electricity`:
image

## After `simplify_network`:
image

## After `cluster_network`:
image

image

scripts/add_electricity.py
```diff
@@ -422,6 +798,44 @@ def attach_hydro(n, costs, ppl, hydro_min_inflow_pu=1):
         to_be_hydro = tbd[mask_reservoir]
         to_be_ror = tbd[mask_ror]
 
+        # Aggregate to_be_ror and to_be_hydro by (bus, carrier, grouping_year)
+        to_be_hydro.loc[:, "carrier"] = "hydro"
+        to_be_ror.loc[:, "carrier"] = "ror"
```
davide-f, line 803: I've seen this now: attention that this operation is executed on views, rather than concrete dataframes.
The behavior has also changed across versions. If to_be_hydro are intended to be different objects, I recommend using .copy() at the end of the tbd[...] to avoid issues.

davide-f (commented):
Fantastic @yerbol-akhmetov !!! I like it!

I've just seen a potential pandas issue, had you the option to fix it. Then, ready to go for me

scripts/add_electricity.py
yerbol-akhmetov: Hi, @davide-f. Thank you for the comment. I have applied the suggestion and added `.copy()`.

davide-f (approved):
Many thanks @yerbol-akhmetov ! All good on my side.
@ekatef feel free to finalize as you prefer

ekatef:
> Many thanks @yerbol-akhmetov ! All good on my side. @ekatef feel free to finalize as you prefer

Hey @davide-f thanks a lot for reviewing! Agree that the PR is ready to be merged and very much looking forward to have this functionality accommodated into the model!

As discussed with @yerbol-akhmetov, the next steps could be:
1) coordinate on possible merge conflicts;
2) open a PR on adding an educational entry for myopic;
3) proceed with merging.

I'm happy to take 1 and 2; @davide-f would you be available to review the plan, and do the actual merge? You have the best overview on the current state of the work in the repo and a more fresh look on the PR itself.

yerbol-akhmetov:
I am closing this PR, because changes proposed in this PR are already included in already merged PR #1722

PR #1722, opened 2026-02-11
Revise myopic implementation (Part 2)

yerbol-akhmetov:
Good day. This PR is a part 2 of PR #1698 on revising myopic implementation. It is built on top of that PR.

## Checklist


scripts/add_brownfield.py
```diff
@@ -32,53 +48,34 @@ def add_brownfield(n, n_p, year):
...
-                threshold
-                * c.df.efficiency[chp_heat.str.replace("heat", "electric")].values
-                * c.df.p_nom_ratio[chp_heat.str.replace("heat", "electric")].values
-                / c.df.efficiency[chp_heat].values
-            )
-            n_p.mremove(
-                c.name,
-                chp_heat[c.df.loc[chp_heat, f"{attr}_nom_opt"] < threshold_chp_heat],
-            )
-
```
yerbol-akhmetov, line 63: It is legacy code and in no longer in use. There is no ` electric` and ` heat` parts for CHP. I do not see a point to have different threshold for CHP or adjust it. By lower part of the code we already drop all extendable components that have optimized `p_nom/s_nom` less than threshold.

scripts/add_brownfield.py
yerbol-akhmetov, line 109: The gas pipeline/network is currently not supported by the model. This part related to H2 pipeline repurposed needs to be revised in another PR.

yerbol-akhmetov:
After merge of #1698 and #1722, few things need to be addressed. I have noticed them recently.
TODOs:
1. Revise `estimate_renewable_capacities_irena` in `add_electricity`. Currently, it just sets `p_nom` for renewables from IRENA for buses. Ideally, it should not set `p_nom`, but add generators with `build_year` and `lifetime`. So they will have proper year-tag (approximately). It will make processing in myopic scripts easier. What do you think, @ekatef? Should it be fixed within this PR?
2. Adapt `test/config.myopic.yaml` to run two horizons. So myopic scripts (`add_existing_baseyear` and `add_brownfield`) are tested.
3. Split of biomass and biomass CHP early in `add_electricity`.
4. Implement myopic running script for electricity model.

scripts/add_existing_baseyear.py
```diff
@@ -33,15 +32,22 @@
 spatial = SimpleNamespace()
 
 
-def add_build_year_to_new_assets(n, baseyear):
+def add_build_year_to_new_assets(n: pypsa.Network, baseyear: int):
```
ekatef, line 36: A type indication is missed for type hinting

scripts/prepare_sector_network.py
```diff
@@ -90,26 +81,25 @@ def add_carrier_buses(n, carrier, nodes=None):
...
+        Dictionary containing existing capacities, efficiencies, and nodes for conventional generation technologies.
 
     Returns:
-        _type_: _description_
-    """ """"""
-
-    logger.info("adding electricity generation")
-
-    # Not required, because nodes are already defined in "nodes"
-    # nodes = pop_layout.index
```
ekatef, line 111: That is just a regular clean-up, right? So, no far-reaching consequences of this change are expected

scripts/prepare_sector_network.py
```diff
@@ -885,6 +891,8 @@ def define_spatial(nodes, options):
         spatial.oil.nodes = ["Earth oil"]
         spatial.oil.locations = ["Earth"]
 
+    spatial.oil.df = pd.DataFrame(vars(spatial.oil), index=nodes)
+
```
ekatef, line 818: Does it relate to the PR, or is it bug-fixing which is independent from the myopic functionality?

scripts/prepare_sector_network.py
```diff
@@ -1336,30 +1353,78 @@ def add_storage(n, costs, existing_battery_capacity=None):
...
+        logger.info("No existing batteries found in electricity-only network")
 
     # Use configured max_hours for battery duration
-    battery_duration = snakemake.params.electricity["max_hours"]["battery"]
+    default_max_hours = snakemake.params.electricity["max_hours"]["battery"]
+
+    if not existing_batteries.empty:
+        # Convert each battery StorageUnit to a Store and Link in the sector network, carrying over capacities
+        max_hours = existing_batteries.get("max_hours", default_max_hours)
+        e_nom = existing_batteries["p_nom"] * max_hours
```
ekatef, line 1294: To my understanding, the major change for batteries is adding the existing capacities. While fetching existing is simpler after revision of carriers grouping and can be done directly in add_storage, so `fetch_existing_battery_capacity_from_elec` is not needed anymore, right?

scripts/prepare_sector_network.py
```diff
@@ -2943,18 +3006,19 @@ def add_electricity_distribution_grid(n, costs):
...
-            solar,
+            solar_rooftop_index,
             suffix=" rooftop",
-            bus=n.generators.loc[solar, "bus"] + " low voltage",
+            bus=n.generators.loc[solar_rooftop_index, "bus"] + " low voltage",
             carrier="solar rooftop",
             p_nom_extendable=True,
-            p_nom_max=potential.loc[solar],
-            marginal_cost=n.generators.loc[solar, "marginal_cost"],
+            p_nom_max=potential.loc[solar_rooftop_index],
```
ekatef, line 2940: Agree that it's safer to subset potential with it's own index instead of using `solar`. I wonder though what are implications for naming and if any additional changes are needed to fix potential side-effects of the changed names?

scripts/prepare_sector_network.py
```diff
@@ -3100,59 +3164,112 @@ def add_rail_transport(n, costs, nodal_energy_totals_fn):
     )
 
 
-def get_capacities_from_elec(n, carriers, component):
+def convert_conventional_generators_to_links(
+    n: pypsa.Network, costs: pd.DataFrame
+) -> None:
```
ekatef, line 3092: I see that we replace `get_capacities_from_elec` which previously only calculated capacities with `convert_conventional_generators_to_links` which is now also adding links. A if there is no duplication with `add_generation`?

ekatef (approved):
Hello @yerbol-akhmetov, looks a great and comprehensive improvement. Have added a few questions which are focused on cross-checking that the implementation is self-explanatory and can be maintained easily. My general impressing is that your have done in-depth analysis and created quite an elegant and robust design. We just must make sure that we accommodate the implementation properly including any adjustments which may be needed in the infrastructure.

As discussed, it would be very helpful if you could share outputs of the functionality testing you did to demonstrate how the approach is working. I think it could be also an awesome addition to the educational chapter of our documentation (and also a potential `Case Study` section in a dissemination output).

ekatef:
@davide-f I don't have any blocking comments for this PR but we have agreed with @yerbol-akhmetov that it would be good to have your opinion.

@yerbol-akhmetov I think it'll be useful if you could address my understanding questions to facilitate capturing the main design ideas by Davide. What do you think?

scripts/add_existing_baseyear.py
yerbol-akhmetov: Thank you for noting it. It seems like I have forgot to push the commit.

scripts/prepare_sector_network.py
yerbol-akhmetov: Yes, it is cleaning up unused code. It was already commented and `nodes` is no longer used by itself. We use `spatial.nodes`.

scripts/prepare_sector_network.py
yerbol-akhmetov: Thanks for clarification. Yes, exactly. If we already know existing capacities, then we can add in `add_storage` directly.

scripts/prepare_sector_network.py
yerbol-akhmetov: It is a bug-fix which was necessary for this PR. I do not remember exactly, but it was causing error. I will post the place where it causes error.

scripts/prepare_sector_network.py
yerbol-akhmetov: Thanks for the comment. Here we could not use `solar` for solar rooftop, because we no longer have number of solar generators exactly equal to nodes. We might have more solar generators (because there might be existing generators, and there will be year-tag in name (e.g. `solar-2010`). Previously existing capacities were set to generators, while the number of them was equal to nodes exactly. Now it is not, because we want to keep `build_year` information. Therefore, solar rooftop needs to have its own index.

scripts/prepare_sector_network.py
yerbol-akhmetov: Thanks for the comment. `add_generation` is no longer in use. I have forgot to delete it. Now `convert_conventional_generators_to_links` does what `add_generation` did. I have named to bring the meaning of the function to name. Because in fact, `add_generation` was not adding generation, but converting from generators to links user specified conventional generators.

yerbol-akhmetov (commented):
Thanks for the comments, @ekatef. I hope I managed to address them.

scripts/prepare_sector_network.py
yerbol-akhmetov: We use df in fuel carrier for oil. For all other conventional carriers we have df, but not for oil.

image
image

yerbol-akhmetov:
Let’s consider how myopic implementation works taking `test/config.myopic.yaml` scenario as example. 
1. For proper myopic implementation, it is necessary to know commissioning and decommissioning date of powerplants. Therefore, before attaching the powerplants to the network, it is important to handle missing `datein` and `dateout` parameters. It is implemented using `fill_datein_dateout` function in `add_electricity ` rule. In particular, missing `datein` parameters are filled based on average build year per carrier, while missing `dateout` parameters are estimated using `datein` and lifetime from cost assumptions (or `fill_values`, if not provided in costs). Then, powerplants are aggregated based on build years using `grouping_year` list. Finally, `grouping_year` column is added to powerplant dataframe. 
image

2. In `add_electricity` rule, all powerplants are attached to network with `build_year` and `lifetime`. The `build_year` is set from `grouping_year` column in powerplants dataframe. In addition, components obtain the year tag in carrier name and component names (e.g. carrier `CCGT-2010` and component name `NG0 2 CCGT-2010`). The year tag in carrier name is necessary for proper aggregation in `simplify_network` and `cluster_network` rules. Extendable components are attached without any year tag as previously. Thus, existing capacities become separate from extendable ones, making it easier to decommission subset of generation units (i.e. wind, solar, hydro, and  conventional powerplants and batteries).
image

3. `estimate_renewable_capacities_irena` function in `add_electricity` is modified to take into account existing capacities previously defined. Some renewable capacities can be attached from custom powerplants dataset. In that case, only remaining capacities are distributed. 
4. In `cluster_network` rule, original base carrier names are restored and year tag is removed from carrier names (e.g. `CCGT-2010` -> `CCGT`) after clustering. In addition, for overnight simulation the generation units are aggregated per carriers per buses, thus we get one generator per carrier per bus as previously (i.e. `NG0 2 CCGT-2010` - > `NG0 2 CCGT`).
5. In `prepare_sector_network` rule, a new function, named `convert_conventional_generators_to_links` was created to replace `add_generation`, `get_capacities_from_elec`, `remove_elec_base_techs`. There is no longer need for `fetch_existing_battery_capacity_from_elec` function, because existing battery capacities are handled within `add_storage` function.
6. In `add_existing_baseyear` rule, functions `add_existing_renewables` and `add_power_capacities_installed_before_baseyear` are  removed  because existing capacities are already attached in `add_electricity` rule.
7. In `add_brownfield` rule, `p_nom` and `p_nom_min` of extendable renewable generators that was set from IRENA stats are reset to 0, because they are already accounted in the base year.

davide-f (commented):
Wrong PR, I'm moving this comment in the other PR #1698 , apologies

virio-andreyana:
Great PR @yerbol-akhmetov! Since this PR touches many parts of the workflow, it would be great for this PR to be merged before any other additions that I have in mind. Do you have a timeline of when this is merged?

ekatef:
> Great PR @yerbol-akhmetov! Since this PR touches many parts of the workflow, it would be great for this PR to be merged before any other additions that I have in mind. Do you have a timeline of when this is merged?

Hey @virio-andreyana, thanks for coordination! Agree that we must merge the myopic contribution as soon as possible since it's ready and polished. My guess is that it could be realistic to aim for merging in 2-3 days.

davide-f (commented):
Hello!

Many thanks Yerbol!
Minor comment: would you mind adding the release-note?
The release note is also missing into part 1; we could merge this to include also part 1 automatically, unless you see particular benefits in having them separate

yerbol-akhmetov:
> Hello!
> 
> Many thanks Yerbol! Minor comment: would you mind adding the release-note? The release note is also missing into part 1; we could merge this to include also part 1 automatically, unless you see particular benefits in having them separate

Thank you, I have added the release notes.

davide-f (approved):
Many thanks!

ekatef:
@yerbol-akhmetov when doing the final check, I see that the objective has slightly changed which is linked with same clustering issues we saw previously:

image

I think it could be worth to check weightings as I see that has been part recently updated. 

To facilitate exploration, sharing a couple of technical details: it appears the objective changes have happened between `Mar 30, 11:06 AM GMT+2` and `Mar 30, 2:19 PM GMT+2` ([this](https://github.com/pypsa-meets-earth/pypsa-earth/actions/runs/23736808837) and [that](https://github.com/pypsa-meets-earth/pypsa-earth/actions/runs/23744340887) CI runs; sorry can't restore particular commit hashes due to some issues with git history). 

Could you please cross-check? It has been a long way, and the implementation is close to be perfect. Just a final polish is needed.

yerbol-akhmetov:
> @yerbol-akhmetov when doing the final check, I see that the objective has slightly changed which is linked with same clustering issues we saw previously:
> 
> image
> I think it could be worth to check weightings as I see that has been part recently updated.
> 
> To facilitate exploration, sharing a couple of technical details: it appears the objective changes have happened between `Mar 30, 11:06 AM GMT+2` and `Mar 30, 2:19 PM GMT+2` ([this](https://github.com/pypsa-meets-earth/pypsa-earth/actions/runs/23736808837) and [that](https://github.com/pypsa-meets-earth/pypsa-earth/actions/runs/23744340887) CI runs; sorry can't restore particular commit hashes due to some issues with git history).
> 
> Could you please cross-check? It has been a long way, and the implementation is close to be perfect. Just a final polish is needed.

Thanks for noting it, @ekatef. I have mistakenly reverted `weighting_for_country`, brought it back.

ekatef:
Excellent @yerbol-akhmetov! Everything looks great now. Merging 🙂

ekatef:
@yerbol-akhmetov congratulations with getting that done!! Amazing work and awesome functionality 🎉 🎉 🎉 

Thank you @davide-f for the great support with reviewing 🙂

PR #1719, opened 2026-02-09
Migrate all cost data processing to `process_cost_data.py`

virio-andreyana:
# Overwrites [#1572](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1572) and [#1577](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1577)

## Changes proposed in this Pull Request

This pull request is inspired by [PyPSA-Eur #1752](https://github.com/PyPSA/pypsa-eur/pull/1752) which centralizes the cost processing into a single script to reduce the duplication across the workflow. Compared to the pull request [#1572](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1572), we try to avoid any changes in the cost by splitting the outputs as follow:

| Category | Processing Function | Output File |
| --- | --- | --- |
| **Electricity Costs** | `load_costs()` | `_elec.csv` |
| **Sector Costs** | `prepare_costs()` | `_sec.csv` |

All of this are conducted in one script: `process_cost_data.py`. A minor improvement in the code is that if `enable: retrieve_cost_data` is false, then only `data/costs.csv` is used regardless of planning_horizon.

This feature can lead to this PR:
- merging `load_costs()` and `prepare_costs()` (if we want to open that again)
- add novel storage technologies as mentioned in [#1577](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1577)
- custom cost csv as mentioned in [PyPSA-Eur #1752](https://github.com/PyPSA/pypsa-eur/pull/1752)

## Checklist

- [x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [x] I tested my contribution locally and it seems to work fine.
- [x] Code and workflow changes are sufficiently documented, including updates to docstrings for meaningful functions.
- [x] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`.
- [x] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [x] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [x] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [x] Archives of the uploaded data do not have an enclosing folder and archive names correspond to the conventions of `configs/bundle_config.yaml`.
[...]

PR #1577, opened 2025-09-23
Extend storage options to include li-ion, vanadium, lfp, lair, pair and iron-air

virio-andreyana:
# Requires https://github.com/pypsa-meets-earth/pypsa-earth/pull/1572 to be merged beforehand

## Changes proposed in this Pull Request

This features aims to make any storage technologies implementable both in the forms of stores and links, or as storage units. It now works both in the electricity only and sector-coupled networks.

- With it, novel storage technologies such as li-ion, vanadium, lfp, lair, pair, iron-air can be included.
- The original design rename the term batteries to li-ion for more clarity, but this would lead to backward incompatibility, so we just let both terms be valid.
- Technology mapping is added in `_helpers.py` and coloring as well

## Checklist

- [x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [x] I tested my contribution locally and it seems to work fine.
- [x] Code and workflow changes are sufficiently documented.
- [ ] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [x] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [ ] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes, including reference to the requested PR.

PR #1731, opened 2026-02-18
Extend storage options to include li-ion, vanadium, lfp, lair, pair and iron-air (re-attempt)

virio-andreyana:
# Requires https://github.com/pypsa-meets-earth/pypsa-earth/pull/1719 to be merged beforehand

**This is a re-attempt of https://github.com/pypsa-meets-earth/pypsa-earth/pull/1577**

This features aims to make any storage technologies implementable both in the forms of stores and links, or as storage units. It now works both in the electricity only and sector-coupled networks.

- With it, novel storage technologies such as li-ion, vanadium, lfp, lair, pair, iron-air can be included.
- The original design rename the term batteries to li-ion for more clarity, but this would lead to backward incompatibility, so we just let both terms be valid.
- Technology mapping is added in `_helpers.py` and coloring as well

## Checklist


config.default.yaml
```diff
@@ -207,13 +207,19 @@ electricity:
...
+    "iron-air": 100
     H2: 168
 
   extendable_carriers:
       # Note: For landlocked countries (e.g., Jordan, Austria, Switzerland), remove 'offwind-ac' and 'offwind-dc'
           # from the Generator list below, as these offshore wind technologies require coastline access.
     Generator: [solar, onwind, offwind-ac, offwind-dc, OCGT]
-    StorageUnit: [] # battery, H2
+    StorageUnit: [] # battery, H2, li-ion, vanadium, lfp, lair, pair, iron-air
     Store: [battery, H2]
```
davide-f: Please, clarify what technologies can be supported as Stores and what as StorageUnit.

While modelling-wise it is possible to model storages as "Stores" or "StorageUnits" with some equivalences, the numeric values would not represent the same assets, while being functional.

Example. If we have a lithium battery with Max C-Rate of 2, that can be represented as Store and links (as a bicharger). Then the optimal capacity of the store will be in MWh.
If we use a StorageUnit, instead of a Store, the optimal capacity will be in MW. If we artificially set max_hour=1, then the equivalence between MW and MWh will be achieved, however the maximum charging and discharging power of the storageunit won't be as with the store implementation. This may be wanted if the user knows what he is doing though absolutely not trivial

scripts/process_cost_data.py
```diff
@@ -294,8 +294,91 @@ def convert_row(x):
...
+                costs.loc[charger] if charger else None,
+                costs.loc[discharger] if discharger else None,
+                max_hours=v,
+                costs_name=costs_name,
+            )
+        else:
+            missing_store += [k]
+
+    if missing_store:
+        logger.warning(
```
davide-f: This should be an error, otherwise we'll likely have issues later in the code

davide-f (commented):
Hello virio and many thanks to propose this very interesting feature!

Beside these minor comments, it would be interesting to include the feature into the tests, by enabling the multiple options in a run, ideally sector-coupled, to ensure it is compatible with the whole workflow.

Many thanks!

config.default.yaml
virio-andreyana: For the first statement, All of the storage technologies are listed in `config.default.yaml` `storage_techs`. All of them can be modeled as "Store" but only those with their `max_hours` defined can be used as storage units. I don't have the precise number for all of them.

You're right on the second statement. I will have to extend the documentation on the configuration to make sure people consider this.

virio-andreyana:
I've also have to fix this issue that is also valid for this PR https://github.com/PyPSA/pypsa-eur/pull/2157

scripts/process_cost_data.py
virio-andreyana: Indeed, this is an issue if the user use this as a "StorageUnit". The best mitigation to this issue is to modify `get_available_storage_carriers()` to not allow them to be implemented.

I have a follow up question for @davide-f on this. Do you think its better if the workflow:
1. Skip the technology that is not implemented even though its written in `extendable_carrier` by the user
2. Stop the workflow and raise an exception if the user select a technology that is not implemented

virio-andreyana: I've turned this into an error. This issue is resolved

config.default.yaml
virio-andreyana: Extensive documentation on this characteristics is now included. See: https://pypsa-earth--1731.org.readthedocs.build/en/1731/user-guide/configuration/#storage_tech

scripts/process_cost_data.py
```diff
@@ -294,8 +294,92 @@ def convert_row(x):
     return cost_dataframe
 
 
+def calculate_cost_for_storage_units(
```
davide-f, line 297: If a technology is modelled as a store or as a storage_unit, does the entry in costs have different name?

The risk is having confusion across runs and values with same name and different units.

config.default.yaml
```diff
@@ -212,13 +212,19 @@ electricity:
...
```
davide-f: Should we add here the same comment as for StorageUnit? I understand the same technologies are supported here too

davide-f (commented):
Hello @virio-andreyana !

Adding a few minor comments, many thanks for the revision after the myopic implementation!

scripts/process_cost_data.py
virio-andreyana: For Store, the cost is retrieved from the individual charger, store and discharger. Whereas for StorageUnit, the cost is unified based on the name of that carrier. The only true difference in parameter is related to the max_hours.

The configuration: `storage_techs:` is the way in which we made this information more transparent.

davide-f: Indeed, so, in the csv file when it is implemented as storageunit, the value would be different than when using store, with the difference being related to the max_hours. Are we aligned?

From what I see, this ensures the model to be simple; avoiding this ambiguity would require ad-hoc solution that may not be really needed. Moreover, our current model already has this ambiguity so this PR is not changing this behaviour.
Are we seeing it in the same way?

virio-andreyana: We are aligned on this, but what are your proposed solution? Or is it something that we just need to keep in mind?

davide-f: > We are aligned on this, but what are your proposed solution? Or is it something that we just need to keep in mind?

I understand this PR is not worsening that issue as we had it already. As such, let's not tackle it now and let's leave it to the future. My main goal is to align on this and identify the best path forward.
A possible solution is assigning a different name depending of whether they are stores or storage unit, but it is another problem, not for this PR

davide-f (commented):
Thanks @virio-andreyana !
Do you mind resolving the conflicts? Good to go for me after that.
If you need support, please let us know

doc/configtables/electricity.csv
```diff
@@ -15,8 +15,8 @@ max_hours,,,
...
--- StorageUnit,--,"Any subset of {battery, H2}",Adds extendable storage units (battery and/or hydrogen) at every node/bus after clustering without capacity limits and with zero initial capacity. It is used in the add_extra_components rule.
--- Store,--,"Any subset of {battery,H2}",Adds extendable storage units (battery and/or hydrogen) at every node/bus after clustering without capacity limits and with zero initial capacity.
+-- StorageUnit,--,"Any subset of {battery, H2, li-ion, vanadium, lfp, lair, pair, iron-air}","Adds extendable storage units at every node/bus after clustering without capacity limits and with zero initial capacity. Supported technologies include battery, H2, li-ion, vanadium, lfp, lair, pair, and iron-air."
+-- Store,--,"Any subset of {battery, H2, li-ion, vanadium, lfp, lair, pair, iron-air}","Adds extendable storage at every node/bus after clustering without capacity limits and with zero initial capacity. Supported technologies include battery, H2, li-ion, vanadium, lfp, lair, pair, and iron-air."
```
yerbol-akhmetov: Thanks for the PR, @virio-andreyana. Can we add the meaning of each storage technology? Because abbreviation might not be always clear. So users at least can check them in documentation.

doc/configtables/storage_techs.csv
```diff
@@ -0,0 +1,8 @@
+Parameter,Unit,Values,Description
+{carrier},  ,  ,Name of the storage carrier.
+-- store,  ,str,"Storage name as listed in costs.csv, used to determine their storage capacity cost and lifetime."
+-- bicharger,  ,str,"Charger and discharger name as listed in costs.csv, used to determine their charging/discharging costs, efficiencies, and lifetimes."
+{carrier},  ,  ,Name of the storage carrier.
+-- store,  ,str,"Storage name as listed in costs.csv, used to determine their storage capacity cost and lifetime."
```
yerbol-akhmetov, line 6: I was curious if we need such repetitions. Although we have different pairs (store+bicharger/store+charger+discharger), I think we can have only one definition for store. What do you think?

scripts/add_electricity.py
```diff
@@ -1144,29 +1152,90 @@ def attach_existing_batteries(
...
+        # Add Stores for existing battery energy capacity
+        n.madd(
+            "Store",
+            batteries_grouped.index,
+            bus=batteries_grouped["bus"] + " battery",
+            e_cyclic=True,
+            e_nom_extendable=False,
+            e_nom=batteries_grouped["p_nom"] * max_hours,
+            carrier="battery",
+            capital_cost=costs.at["battery storage", "capital_cost"],
```
yerbol-akhmetov: I was curious, why we do not use `costs.at[config["storage_techs"]["battery"]["store"]], "capital_costs"]`. Some thing like this, as we already define store and links for different storage technologies in config.

yerbol-akhmetov (commented):
Thanks, @virio-andreyana, for the PR. I have left few comments.

scripts/add_electricity.py
virio-andreyana: That is a valid point. We can do that.

doc/configtables/electricity.csv
virio-andreyana: You can find more detail on what's the difference between stores and storage units in configuration.md. What do you think of those clarifications?

doc/configtables/storage_techs.csv
virio-andreyana: I'm ambivalent on this. I just want to make sure that users know that regardless of pairs, store is always required in those pairs.

doc/configtables/electricity.csv
yerbol-akhmetov: Hi, @virio-andreyana. I think I have not explained myself well. Since this csv is shown in the `configuration.md` and is kind of ultimate documentation, can't we provide what each of them means? So user gets what each abbreviation means without going deep into code and configuration. Eg. lair - is Liquid air store, vanadium - vanadium redox flow store and so on.

doc/configtables/storage_techs.csv
yerbol-akhmetov: We can keep it as it is, or we can just add sentence to `configuration.md` about necessity of store (or how it is modelled).

config.default.yaml
```diff
@@ -191,15 +191,21 @@ electricity:
   # ----------------- Electricity generation -------------------
   extendable_carriers:
     Generator: [solar, onwind, offwind-ac, offwind-dc, OCGT] # Remove 'offwind-ac' and 'offwind-dc' for landlocked countries (e.g., Jordan, Austria, Switzerland)
-    StorageUnit: [] # battery, H2
-    Store: [battery, H2]
+    StorageUnit: [] # battery, H2, li-ion, vanadium, lfp, lair, pair, iron-air
```
yerbol-akhmetov: Maybe we can say in the comment here, please see full list and details in configuration.md giving path.

doc/configtables/electricity.csv
yerbol-akhmetov: What about other storage technologies provided in `storage_techs` (e.g. `znbr`, `gravity` and etc)? Can't we use them now?If we can why not to provide details about them in configuration too?

virio-andreyana: Well, that's a lot of information, and it's really dynamic as well depending on what data is available in PyPSA Technology Data. The middle ground would be a csv describing the main abbreviation of each storage technolgies. It will be a lot, are you sure you want to add those to `configuration.md`?

config.default.yaml
virio-andreyana: Good point!

doc/configtables/electricity.csv
yerbol-akhmetov: We can describe each abbreviation and then provide link to `technology-data` (our version) and say users to check them there.

davide-f (approved):
Good to go on my side with thanks!

yerbol-akhmetov:
Thank you, @virio-andreyana for the quite valid and valuable PR. Thanks, @davide-f for the great discussions and suggestions.

yerbol-akhmetov:
Tagged probably related issues #1972, #1973 that can be handled in the future

Issue #1740, opened 2026-03-03
Use cached http snakemake plugin


willu47:
The snakemake plugin developed for PyPSA-Eur would also be useful for dealing with retrieving files for PyPSA-Earth.

This was originally developed for PyPSA-Eur in [PR#1913](https://github.com/PyPSA/pypsa-eur/pull/1913).

The [plugin](https://github.com/PyPSA/snakemake-storage-plugin-cached-http) is available on bioconda or can be installed via [pip](https://pypi.org/project/snakemake-storage-plugin-cached-http/). 

The repo readme describes configuration

    # Optional: Configure cached HTTP storage with custom settings
    # This extends the existing storage configuration (e.g., for HTTP)
    storage cached_http:
        provider="cached-http",
        cache="~/.cache/snakemake-pypsa-eur",  # Default location
        max_concurrent_downloads=3,  # Download max 3 files at once

and usage

    rule download_zenodo:
        input:
            storage("https://zenodo.org/records/3520874/files/natura.tiff"),
        output:
            "resources/natura.tiff"
        shell:
            "cp {input} {output}"
    
    rule download_pypsa:
        input:
            storage("https://data.pypsa.org/workflows/eur/eez/v12_20231025/World_EEZ_v12_20231025_LR.zip"),
        output:
            "resources/eez.zip"
        shell:
            "cp {input} {output}"
    
    rule download_gcs:
        input:
            storage("https://storage.googleapis.com/open-tyndp-data-store/CBA_projects.zip"),
        output:
            "resources/cba_projects.zip"
        shell:
            "cp {input} {output}"

Issue #1508, opened 2025-07-10
Implement data management and versioning system


ekatef:
As suggested by @euronion, it would be great to share versioned data on zenodo as an intermediate layer of the modeling workflow. [An link](https://github.com/PyPSA/pypsa-eur/pull/1675) to an implementation of a similar system for PyPSA-Eur.

As discussed during today's [weekly meeting](https://docs.google.com/document/d/1WDkF3fl0wOCBgvNdQ4TghrZa4WsM8I8S/edit?usp=sharing&ouid=102857250950370967808&rtpof=true&sd=true), the following points are worth attention:

- Checking licensing is necessary (in particular, hydrobasins and gadm restric re-hosting)
- We may need a design decision on sharing inputs [extracted by the workflow from some sources, such as land usage, bathymetric or population datasets] and/or output data obtained by the workflow itself [e.g. topology of power grids]
- Synergies with other initiatives, in particular with [PyPSA-Earth-Status](https://github.com/pypsa-meets-earth/pypsa-earth-status)

ekatef:
This feature can be very relevant to increase stability of the model, and improve user experience. On the usability point, #1507 discussion can help to fit the implementation into a broader context.

PR #1312, opened 2025-01-20
Update Snakemake and adapt for dag

Eric-Nitschke:
This pull request addresses issue #1307.
Since it updates Snakemake to version 8, this fully breaks the Snakefile and is thus currently not operational. Additionally, switching to a newer Snakemake version requires a newer Python version as well, which is incompatible with the current matplotlib specifications.
Thus, this PR requires further investigation and more changes.

# Closes #1307 .

## Changes proposed in this Pull Request
Major changes:
- update Snakemake in environment.yaml to version 8

Minor changes:
- move the input file logging from build_demand_profiles.py to retrieve_databundle_light.py to prevent issues with the creation of the rule graph

## Checklist

- [x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [ ] I tested my contribution locally and it seems to work fine.
- [ ] Code and workflow changes are sufficiently documented.
- [ ] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`.
- [ ] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [ ] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [ ] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes, including reference to the requested PR.

PR #1676, opened 2025-12-09
Update environment to support latest PyPSA versions

davide-f:
# Closes # (if applicable).

## Checklist

- [x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [ ] I tested my contribution locally and it seems to work fine.
- [ ] Code and workflow changes are sufficiently documented including ammending docstrings for meaningful functions.
- [ ] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`.
- [ ] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [ ] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [ ] Archives of the uploaded data do not have an enclosing folder and archive names correspond to the conventions of `configs/bundle_config.yaml`.
- [x] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes, including reference to the requested PR.

PR #1741, opened 2026-03-04
Explore data versioning approach

willu47:
Addresses #1740 and #1508 

- Adds the snakemake cached http plugin.
- The plugin uses the `storage` feature found in snakemake 9, so this is updated in the environment.yaml file.

This prepares the ground for refactoring of the `retrieve` scripts which is explored in commit 72d1b53e412f8f0bab9acb8826d21eb012a91d88 and implements a demo of the PyPSA-Eur approach of data versioning for the hydrobasins datasets. Checkout the branch and then run `snakemake -j8 create_hydrobasins_world` to see it in action.

- Similar to #1312 (updates snakemake to v8)

## Checklist


Snakefile
```diff
@@ -24,12 +22,17 @@ from _helpers import (
...
 from pathlib import Path
 
 
-HTTP = HTTPRemoteProvider()
+# This extends the existing storage configuration (e.g., for HTTP)
+storage cached_http:
+    provider="cached-http",
+    cache="~/.cache/snakemake-pypsa-earth",  # Default location
+    max_concurrent_downloads=3,  # Download max 3 files at once
+
```
ekatef, line 35: I see CI is complaining about this fragment but don't get what is the problem as  PyPSA-Eur seems to use [exactly the same](https://github.com/PyPSA/pypsa-eur/blob/21bc4faa0e8720cd5e1d9d4bbd4e78451a774d47/rules/retrieve.smk#L21) syntax. @willu47 do you have any idea what is going on there? May it be caused by using the old Snakemake version in testing? (there is the need to update the conda lock files since they determine particular dependencies versions used in CI tests)

ekatef (dismissed):
Thanks a lot for looking into that @willu47. Have added a technical comment to the code. Could you please have a looks?

I also see that pre-commit is [complaining](https://results.pre-commit.ci/run/github/343194588/1772617363.wIEOFT0lQpKbiKurIDiSFA) which can be connected with some specific settings of OET github organisation. My guess is the most straightforward way to fix it could be run pre-commit locally `run pre-commit --all` and commit the automatically suggested changes.

ekatef:
Thank you @willu47, I think it's awesome. Have shared technical comments in the review, adding more in-depth ones here for better visibility as they require some coordination.

1. Updating Snakemake to 9+ is known to be a bit of challenge. The first exploration was done in #1312 and it was found to be a bit more tricky than expected. The current status on dependencies updates is tracked in #1617 but is focused mainly on PyPSA. 

As a follow-up of #1617 @davide-f is testing Snakemake 9+ in [this branch](https://github.com/davide-f/pypsa-earth/tree/refs/heads/test_env_34) but some nasty bug was discovered. the testing results of are available [here](https://github.com/davide-f/pypsa-earth/actions/runs/22457009080/job/65040795382)). As discussed with @lkstrp, the behaviour looks are some Snakemake-related troubles which are most likely linked with the fact that retrive_databundle is implemented in `not-quite-snakemake` way due to the need to deal with cutouts.

2. For a proper Zenodo release, you are totally right that it's something quite urgent. Thank you for providing a quick solution for that. I think we may need to revise the meta-data, in particularly authorship as creating databundles was teamwork [happy to compose a list of people who have contributed]. Also, it would be good to coordinate with @davide-f who have been working on re-structuring databundles. 

I wonder if we can replace the sandbox links with the release created by @willu47 as the first step, and continue to work on re-structuring after? That would relieve pressure and allow to focus on designing data management revision, as it doesn't feel nice when we know that the most convenient data source can fail basically every moment.

We discussed previously that we must be 100% sure that Zenodo release is alright from licensing perspective. For now, I guess we there on ~95%, and I'll try to confirm remaining 5% tomorrow.

Generally, I think we need to have those technical issue finally resolved, and this PR is a step in this direction.

Snakefile
willu47: Yes, I need to update the lock files to get the CI to work.

ekatef:
Hey @willu47, checking CI logs I see we are facing here same issues @davide-f has reported for testing Snakemake 9+ ([a sample output](https://github.com/davide-f/pypsa-earth/actions/runs/22457009080/job/65040795382)).

The workflow is complaining on that the cutout has no intersection with an area of the country considered by the model in the landlock test. As far as I remember our discussion on that, the most likely explanation has been this the list of data bundles to download is not treated properly by `retrieve_databundle_light` rule https://github.com/pypsa-meets-earth/pypsa-earth/blob/575c030591aa3a28ee9dcfd2dd89d89e2eec4cc9/Snakefile#L149-L151

Presumably, it has something to do with not quite correct behaviour of Snakemake itself which wasn't well tested for a use case when we transfer file names as parameters instead of inputs. While from our side the major reason for the implemented solution is the need to deal with cutouts as discussed in [SPIKE issue](https://github.com/open-energy-transition/pypsa-zambia/issues/9).

@davide-f please feel free to correct and comment on that.

willu47:
Thanks @ekatef. Yes, I just came to the same conclusion.

I can revert commit f359bbd05625c5c35 in this PR to point back to Zenodo, and create a new PR with only that change.

Then, in this PR, we focus on updating `snakemake>9<10`, which will enable further work on the retrieval script.

ekatef:
> f359bbd



> Thanks @ekatef. Yes, I just came to the same conclusion.
> 
> I can revert commit [f359bbd](https://github.com/pypsa-meets-earth/pypsa-earth/commit/f359bbd05625c5c355c7ee01bde36360178c35ea) in this PR to point back to Zenodo, and create a new PR with only that change.
> 
> Then, in this PR, we focus on updating `snakemake>9<10`, which will enable further work on the retrieval script.

Agree, it makes perfect sense to keep PRs focused. Thank you for keeping this in mind. Great, looking forward for a Zenodo update PR.

For updating snakemake version, I think it could be a good idea to synchronise snakemake update with updating PyPSA version. #1676 provides an overview of the changes needed to accommodate different versions of PyPSA without touching Snakemake which could be probably provide some useful hints also for this PR.

willu47:
All tests passed locally after merging #1746

ekatef:
> All tests passed locally after merging #1746

That is a precious insight. The fact the current design is compatible also with newer Snakemake versions and caching is an important input for a discussion in #518 on revision of data management.

For the merging strategy, we agreed previously that we'll do first incremental update of PyPSA version with the releases in-between, and upgrade Snakemake version after. Work on Snakemake update has progressed more than I have been aware of (sorry for that!), and the current implementation is functional ([this](https://github.com/davide-f/pypsa-earth/pull/41) internal PR). However, we didn't consider caching plugin which sounds a much needed improvement.

willu47:
Yes, I see, this duplicates a lot of the work also completed in https://github.com/davide-f/pypsa-earth/pull/41 of @davide-f 

I'm no closer to understanding why the tests pass on my local machine (`make test`) but the CI fails...

ekatef:
> Yes, I see, this duplicates a lot of the work also completed in [davide-f#41](https://github.com/davide-f/pypsa-earth/pull/41) of @davide-f
> 
> I'm no closer to understanding why the tests pass on my local machine (`make test`) but the CI fails...

I think it's only a partial duplication, and feel free to build on the insights @davide-f has obtained to save a bit of resources. It would be perfect to understand better implications of using the caching plugin. As mentioned, there are concerns of possible inflation of disk storage, especially when downloading continental-scale data. What is your experience in this regard?

ekatef:
I think [the implementation](https://github.com/open-energy-transition/pypsa-zambia/pull/74) is a great example how the approach developed in this PR can work with the current environment. 

@davide-f if you have resources it would be great to have your review of the PR.

doc/release_notes.rst
```diff
@@ -45,6 +45,10 @@ This part of documentation collects descriptive release notes to capture the mai
 
 * Fix stochastic bundle names causing rerun of workflow every time `PR #1746 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1746`__
 
+* Update snakemake, replace HTTPRemote with Storage, and use snakemake_cached_http_plugin `PR #1741 <https://github.com/pypsa-meets-earth/pypsa-earth/pull/1741`__
+
+* Add hydrobasins to rule input for build_renewable_profiles `PR #xxxx <https://github.com/pypsa-meets-earth/pypsa-earth/pull/XXXX`__
+
```
davide-f, line 51: We moved to .md documentation; there is the need to port this into doc/release_notes

also installation.rst is to be dropped. I'd advise against keeping references to conda lock in the documentation as the user won't generate the files using conda-lock

scripts/_helpers.py
```diff
@@ -447,22 +447,31 @@ def get_aggregation_strategies(aggregation_strategies):
...
+    **wildcards,
 ):
     """
-    This function is expected to be executed from the "scripts"-directory of "
+    This function is expected to be executed from the 'scripts'-directory of '
     the snakemake project. It returns a snakemake.script.Snakemake object,
     based on the Snakefile.
 
-    If a rule has wildcards, you have to specify them in **wildcards**.
+    If a rule has wildcards, you have to specify them in **wildcards.
```
davide-f, line 461:
```suggestion
    This function is expected to be executed from the 'scripts'-directory of
   the snakemake project. It returns a snakemake.script.Snakemake object,
   based on the Snakefile.

    If a rule has wildcards, you have to specify them in **wildcards**.
```

davide-f (commented):
I really like the architecture. Great draft for when we switch to snakemake 9+.
In a planning perspective, we can integrate this architecture first in a compact way, and start integrate into this structure the various data sources that are currently downloaded. E.g. [emissions file](https://github.com/pypsa-meets-earth/pypsa-earth/blob/a2c5f7b33bb12c4a6f5e06868606ce1dc2d5718c/scripts/prepare_network.py#L82), [gadm data](https://github.com/pypsa-meets-earth/pypsa-earth/blob/a2c5f7b33bb12c4a6f5e06868606ce1dc2d5718c/scripts/build_shapes.py#L71), [steel database](https://github.com/pypsa-meets-earth/pypsa-earth/blob/a2c5f7b33bb12c4a6f5e06868606ce1dc2d5718c/scripts/build_industrial_database.py#L67), [ports](https://github.com/pypsa-meets-earth/pypsa-earth/blob/a2c5f7b33bb12c4a6f5e06868606ce1dc2d5718c/scripts/prepare_ports.py#L21), [airports](https://github.com/pypsa-meets-earth/pypsa-earth/blob/a2c5f7b33bb12c4a6f5e06868606ce1dc2d5718c/scripts/prepare_airports.py#L19), among others

To enhance data availability, it would be nice to support backup sources e.g. try link A first, and B if it doesn't work. This improves reliability.
Currently, the retrievedatabundle supports that functionality.

Moreover, it would be interesting to include an option to enable the retrieval of bundles from the local hard drive. Some users cannot access internet while running the tool. However, they could upload the files in a common folder in the cloud.
It would be interesting to enable a flag that switches the retrieval of each bundle, if requested to the retrieval from a global local folder.
We could also provide a simple rule that prepares this bundle

I'm moving these comments also in issue #518

PR #1746, opened 2026-03-05
Fix #1745 by sorting unordered list of bundle names

willu47:
# Closes #1745

## Changes proposed in this Pull Request

Sort the list of bundle names returned from the helper function to avoid running the entire workflow (including downloading of bundles) on every rerun.

## Checklist

- [x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [x] I tested my contribution locally and it seems to work fine.
- [x] Code and workflow changes are sufficiently documented, including updates to docstrings for meaningful functions.
- [x] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`.
- [x] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [x] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [x] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [x] Archives of the uploaded data do not have an enclosing folder and archive names correspond to the conventions of `configs/bundle_config.yaml`.
- [x] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes, including reference to the requested PR.

PR #1748, opened 2026-03-08
Add options to retrieve only missing databundles

virio-andreyana:
# Supports #1746

This PR replaces the unused `retrieve_databundle_sector` with `keep_existing_databundle`. What it does is that it checks first if the output file listed in the databundle have already been downloaded beforehand. If so, then users can safely chose not to download them again.

The challenge of using this method is that we have to make sure that all filenames are unique, which is not always the case when it comes to tutorial files, most notably the cutouts. This PR must be seen in the wider context of streamlining the data management.

## Checklist


willu47:
See the discussion in #518 for ideas on refactoring the retrieval of databundles. Ideally, we should use the built-in Snakemake tools to check for the existence of files. For example, if snakemake rules were written for each databundle, this PR would be redundant.

virio-andreyana:
Thanks for reviewing @willu47. I understand where you're coming from. In PyPSA-Eur, `retrieve.smk` contains all the rules to download each databundle. These are I think the challenges of implementing them in PyPSA-Earth:

`tutorial`:  
- The tutorials databundle in PyPSA-Earth contains simplified files to save memory of new users and speed up the CI. The output name of tutorial files are similar to the full version. 
-  PyPSA-Eur use of tutorials are depreciated. 
-  Can we make sure that snakemake notice the rules have change even though the output remains the same?

`countries`: 
- PyPSA-Earth data retrieval need to be flexible enough to retrieve some countries instead of the entire datasets (see cutouts).
- PyPSA-Eur's datasets are consistent regardless of countries selected. 
- There is an order of prioritization when it comes to country-specific files, the more accurate the files, the better. How to prevent multiple rules having the same output? 

I believe these problems can be resolved eventually. But I'm not sure how soon. Once https://github.com/pypsa-meets-earth/pypsa-earth/pull/1741 is merged, we can investigate further :)

willu47:
Hi @virio-andreyana 

For Snakemake, if the file names are identical but contain different information (e.g., tutorial versus full data), it is important to ensure that Snakemake can distinguish between them. This avoids weird bugs where a workflow might be inadvertently run with the incorrect (e.g. tutorial) data when it needs the full data. Snakemake will be able to do this if it has knowledge of the inputs. For example:

```python

if config['tutorial'] is True:

rule retrieve_tutorial
    input: https://a.link.to/tutorial_data/a_tutorial.csv
    output: data.csv

if config['tutorial'] is False:

rule retrieve_full_data
    input: https://link.to/full_data/a.csv
    output: data.csv
```

For countries, I think the decision needs to be made whether to use country as a wildcard. This would then allow for country-specific downloads, followed by an aggregation rule. Another pseudocode example:

```python

COUNTRIES = config['countries']

rule retrieve_country:

    input: https://my.data/{country}/a.csv
    output: data/{country}/a.csv

rule collate_country_data:

    input: expand(`data/{country}/a.csv`, country=COUNTRIES)
    output: data.csv
```

ekatef:
Hello!
@virio-andreyana thank for pushing data management forward.

@willu47 totally agree that it's worth to keep a bigger picture in sight, and it's a very good moment for revisions of the general data management strategy. @virio-andreyana we have moved a more detailed discussion to #518 to avoid hijacking your PR. Feel free to continue the high-level discussion there.

For the essence of the PR, I think it's a step in a right direction. Unintentional overwriting the existing data bundles is one of the most annoying current issues and anything but good design. Also, saving some resources during initial data retrieval would be a great outcome.

virio-andreyana:
Hi @willu47 and @ekatef. I've now read through #518 and yes it is ambitious indeed. I'll close this PR and support the development that are more aligned with what Will mentioned. Cheers!

doc/tutorial_sector.rst
```diff
@@ -45,14 +45,14 @@ Scenarios can be defined like for electricity-only studies, but with additional
...
-    It is important to set the following flags ``retrieve_databundle`` and ``retrieve_databundle_sector``
+    It is important to set the following flags ``retrieve_databundle`` and ``keep_existing_databundle``
     to ``false`` after the first run to prevent unnecessary re-downloads, as the files only need to be downloaded once.
 
 .. code:: yaml
 
     enable:
       retrieve_databundle: true
-      retrieve_databundle_sector: true
+      keep_existing_databundle: true
```
ekatef: A great catch!

doc/tutorial_sector.rst
```diff
@@ -349,14 +349,14 @@ in the ``test/config.test_myopic.yaml`` file.
 
 .. note::
 
-    It is important to set the following flags ``retrieve_databundle`` and ``retrieve_databundle_sector``
+    It is important to set the following flags ``retrieve_databundle`` and ``keep_existing_databundle``
```
ekatef: Here as well would be good to cross-check a recommendations for `keep_existing_databundle`

doc/tutorial_sector.rst
```diff
@@ -45,14 +45,14 @@ Scenarios can be defined like for electricity-only studies, but with additional
...
```
ekatef: My understanding is `keep_existing_databundle` must be `true` for regular modelling runs (when all the data needed are still extracted), right? If that is the case, the recommended value will be opposite to the one for `retrieve_databundle` which must be `false` normally

doc/tutorial_sector.rst
ekatef: Same as above: would be great to cross-check the recommended values

scripts/retrieve_databundle_light.py
```diff
@@ -778,6 +778,15 @@ def get_best_bundles(
     return sorted(bundles_to_download)
 
 
+def bundle_exists(config, bundle):
+    """
+    Check whether all files for a given data bundle already exist locally.
+    """
+    files = datafiles_retrivedatabundle(config, [bundle])
+    exists = all(os.path.exists(f) for f in files)
+    return exists
```
ekatef, line 826: That is the point where it would help to have type hinting and more detailed docstrings in the existing code base. Personally, I need some thinking to recognise what exactly checks `datafiles_retrivedatabundle`: are those only destination names, or also details of the bundle itself?

test/config.landlock.yaml
```diff
@@ -8,6 +8,9 @@ countries: ["BW"]
 run:
   name: "landlock"
 
+enable:
+  keep_existing_databundle: false # to retrieve cutouts for BW
```
ekatef, line 12: It would be good to harmonise across test configs. Not sure I completely understand the reason `keep_existing_databundle` is kept `false` for the landlock but `true` for `config.sector`

ekatef (commented):
Thank you @virio-andreyana! Looks a great idea and neat implementation. Have scanned through the code and added a few comments and questions. Could you please have a look?

Additionally, I think it could be helpful to formalise outputs of our discussion on the retrieval strategy on comparing design when everything is downloaded in a retrieval script (or scripts) vs downloading data on fly in the middle of the workflow execution. Currently, we are mixing both approaches which is not great. I agree with you that some solution is needed on the level of the project.

Issue #1715, opened 2026-02-04
Address pandas 2.3 deprecations


ekatef:
## Describe the feature you'd like to see

pandas 3.0 has been released recently meaning that all deprecations of pandas 2.3 will become errors.

We have two following points to be addressed.
1. [Dedicated string data type by default](https://pandas.pydata.org/docs/dev/whatsnew/v3.0.0.html#dedicated-string-data-type-by-default)
This deprecation leads to the warning:
> clean_osm_data.py:620: FutureWarning: Setting an item of incompatible dtype is deprecated and will raise an error in a future version of pandas. Value '['1.0' '1.0' '1.0' '1.0' '1.0' '1.0' '2.0' '2.0' '2.0' '1.0' '2.0' '1.0'

2. [Chained assignment has stopped working](https://pandas.pydata.org/docs/dev/whatsnew/v3.0.0.html#consistent-copy-view-behaviour-with-copy-on-write)
Which results in a warning in the solving script:
> solve_network.py:661: FutureWarning: A value is trying to be set on a copy of a DataFrame or Series through chained assignment using an inplace method.The behavior will change in pandas 3.0. This inplace method will never work because the intermediate object on which we are setting values always behaves as a copy.

PR #1779, opened 2026-04-03
bugfix in clean osm data

SermishaNarayana:
# Closes #

Issue arises as below  in clean_osm_data due to pandas treating a series as str instead of float. 
Screenshot 2026-04-03 105710

Fix introduced by type-casting to float64.

Addresses one of issues mentioned in #1715 

## Checklist

### DIFF
--- a/scripts/clean_osm_data.py
+++ b/scripts/clean_osm_data.py
@@ -613,10 +613,10 @@ def fill_circuits(df):
                 for (vc, vf) in zip(row["cables"], row["tag_frequency"])
             ]
         )
-
+    
     df.loc[df_match_by_cables.index, "circuits"] = df_match_by_cables.apply(
         _filter_cables, axis=1
-    )
+    ).astype("float64")
 
     # length of cables elements is larger than frequency; the last cable data are merged to match
     df_merge_by_cables = df[to_fill_merge][["tag_frequency", "cables"]].copy()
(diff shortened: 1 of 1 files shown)


scripts/clean_osm_data.py
```diff
@@ -616,7 +616,7 @@ def _filter_cables(row):
 
     df.loc[df_match_by_cables.index, "circuits"] = df_match_by_cables.apply(
         _filter_cables, axis=1
-    )
+    ).astype("float64")
```
ekatef: I wonder if it could be the case that some values in `"circuits"` can't be transformed into float. Have you been able to trace a particular point where having `"curcuits"` as string leads to issues?

davide-f (changes requested):
Hello @SermishaNarayana and thanks for reporting! Having a robust network representation is key!

Could you share whether you are using pandas 3 or below 3?
The workflow is not compatible with pandas 3+ and it may be an issue induced by it.
It would be great to fix issues while ensuring preparation for pandas 3+

Please check if your dataset has always data with no semicolumn like "3;3" in columns "circuits" or "cables".
The "circuits"/"cables" columns are intended to be of string type as they may contain entries like "1;1" and in some cases they are numeric. Forcing them to be float-only will lose information in case of entries with multiple circuits.

Indeed,  single OSM entry maps two circuits separated by semicolumns, e.g. "3;3" for cables or "1;1" for circuits. For those cases, the astype(float64) doesn't work. Indeed, circuits/cables/frequency/voltage columns are intended to be in string format to treat those cases.
However, in special cases they become float, which leads to complications with pandas 3+ as it no longers enable a trivial change in the column type

The robust solution is to ensure columns with string type for circuits,cables,frequency,voltage, had you capacity, and equivalent ones.
Had you capacity to tackle this, it would be awesome

PR #1711, opened 2026-01-28
Streamline data management

ekatef:
Relates to #1668 

## Changes proposed in this Pull Request

The PR aims to increase robustness of the workflow by using an inventory of urls utilised to fetch data on-fly. Having an overview of the remote sources being in use facilitates introducing of changes and updates, and is the first step to data versioning.

## Checklist

- [x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [ ] I tested my contribution locally and it seems to work fine.
- [ ] Code and workflow changes are sufficiently documented, including updates to docstrings for meaningful functions.
- [ ] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`.
- [ ] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [ ] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [ ] Archives of the uploaded data do not have an enclosing folder and archive names correspond to the conventions of `configs/bundle_config.yaml`.
- [ ] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes, including reference to the requested PR.

PR #1783, opened 2026-04-10
Add ammonia industry explicitly

### LINKED ISSUE #102: Upstream modeling of ammonia production

Bring to upstream a detailed implementation of ammonia production done for e-fuels project [this PR](https://github.com/open-energy-transition/efuels-supply-potentials/pull/50). As discussed with @yerbol-akhmetov there is a global database which makes generalisation relatively straightforward

yerbol-akhmetov:
Good day. This PR aims to add ammonia industry explicitly into the model. Global ammonia production dataset is taken from latest release of [USGS](<https://www.usgs.gov/centers/national-minerals-information-center/nitrogen-statistics-and-information>) by 2023. USGS dataset includes also ammonia production capacities by plants for the USA. Using the provided location of the plants and [US cities dataset](<https://simplemaps.com/data/us-cities>), the production capacities are geo-located. For European countries, PyPSA-Eur uses [ICIS data](<https://www.icis.com/explore/resources/news/2023/01/18/10846094/insight-poor-demand-high-costs-stifle-europe-industry-despite-falling-gas-prices/>) for the plant-level production capacity. We use the same source for distributing European production capacities.

Some locations provided in USGS dataset for US cities cannot be found in US cities dataset. In some cases, the name of location slightly different, while in some cases the location is not present. In first situations, the name is fixed, while in the second situations, the location name is replaced to the closest existing location name. 

P.S. An issue #1804 was added to differentiate ammonia based on colors as suggested by @Eddy-JV. That issue can be handled as a separate PR.

## Checklist


data/industry/ammonia_plants.csv
davide-f, line 1: is the pedix "_eu" necessary?

scripts/build_ammonia_production.py
```diff
@@ -0,0 +1,327 @@
...
+    if pd.isnull(location):
+        return None
+
+    parts = location.split(", ")
+    if len(parts) > 1:
+        return parts[1][:2].upper()
+    return None
+
+
+def download_ammonia_production_data(ammonia_sources: dict) -> bytes:
```
davide-f: Download functions are better treated with separate retrieve functions, which will help the data management.

scripts/build_ammonia_production.py
```diff
@@ -0,0 +1,327 @@
...
+
+    # Load parameters
+    ammonia_sources = snakemake.params.ammonia_sources
+    us_cities_source = snakemake.params.us_cities_source
+
+    # Load inputs
+    eu_ammonia_plants_path = snakemake.input.eu_ammonia_plants
+
+    # Download ammonia production dataset
+    content = download_ammonia_production_data(ammonia_sources)
```
davide-f: I understand we have different source of data that are processed.
I'd suggest to make this more clear, as this can extend to other countries and we'd like to have the infrastructure to handle it.

I'd suggest to move the "download" to a different rule that has the sole scope of downloading the data. Then, the output will be come an input to this rule.
Then, I'd advise to have a clear distinction between the treatment of each source independently.
In this main, I assume what functions are US-only, what are EU-only, what are common etc, which can lead to complications.

Moreoever, it is advisable to provide instructions on how a user can include their own data for a different section.
Using csvs is quite confortable and we could advise to fill additional rows in the ammonia_plants_eu.csv. If that is the chosen option, it is best to drop the "_eu" pedix to avoid confusion that they must be eu-specific

scripts/build_industrial_database.py
```diff
@@ -483,6 +484,57 @@ def create_paper_df():
...
+        Path to the ammonia plants CSV file.
+
+    Returns
+    -------
+    pd.DataFrame
+        A DataFrame containing ammonia plant information with columns:
+        ['country', 'y', 'x', 'location', 'technology', 'capacity', 'unit', 'quality', 'ID'].
+    """
+    # Load ammonia plants data
+    df_ammonia = pd.read_csv(ammonia_plants_file)
```
davide-f: When reading/saving csvs, it is best to use read/to_csv_nafix from helpers to increase robustness with "NA" being namibia instead of nan. This applies also in other sections. A PR by anton is covering that in upstream, and it is best to preserve this for code additions
```suggestion
    df_ammonia = read_csv_nafix(ammonia_plants_file)
```

scripts/build_ammonia_production.py
```diff
@@ -0,0 +1,327 @@
...
+
+    # Extract EU plant data
+    eu_ammonia_plants = extract_eu_plant_data(eu_ammonia_plants_path)
+
+    # Combine ammonia plants data
+    ammonia_plants = combine_ammonia_plants(us_ammonia_plants, eu_ammonia_plants)
+
+    # Save the results
+    ammonia_totals.to_csv(snakemake.output.ammonia_production)
+    ammonia_plants.to_csv(snakemake.output.ammonia_plants, index=False)
```
davide-f: When exporting to csv, it is recommended to use to_csv_nafix to prevent issues with "NA" values being interpreted as nans

scripts/build_ammonia_production.py
```diff
@@ -0,0 +1,327 @@
...
+        columns={"Capacity2": "capacity", "lat": "y", "lng": "x"}
+    )
+
+    us_plants.index.name = "ktonNH3/a"
+    us_plants["country"] = "US"
+
+    return us_plants
+
+
+def load_us_cities_data(us_cities_source: dict) -> pd.DataFrame:
```
davide-f: Also this function "dowloads" data. Also this function shall have its dedicated rule that downloads the data only and then it is used here as an input

scripts/build_ammonia_production.py
davide-f: The additional data_sources file is not strictly necessary depending on the implementation that you'll do.
If we can avoid it for now, better, though I'm not against keeping it for the moment until we agree on the data management #1668 / #1508

scripts/build_industry_demand.py
```diff
@@ -289,12 +291,59 @@ def match_technology(df):
...
+                industry_base_totals.loc[(country, "ammonia"), "ammonia"] = nh3_mwh
+
+                # Subtract gas and electricity feedstock from chemical and petrochemical (clip at 0)
+                for carrier, value in [("gas", gas_mwh), ("electricity", elec_mwh)]:
+                    current = industry_base_totals.loc[
+                        (country, carrier), "chemical and petrochemical"
+                    ]
+                    industry_base_totals.loc[
+                        (country, carrier), "chemical and petrochemical"
+                    ] = max(0.0, current - value)
```
davide-f, line 344: if value is higher than current, energy totals are not preserved. Best to raise a logger.warning in that case

Snakefile
```diff
@@ -2087,6 +2095,26 @@ rule build_industry_demand:  #default data
...
 
+rule build_ammonia_production:
+    params:
+        ammonia_sources=config["data_sources"]["ammonia"],
+        us_cities_source=config["data_sources"]["us_cities"],
+    input:
+        eu_ammonia_plants="data/industry/ammonia_plants_eu.csv",
+    output:
+        ammonia_production="data/industry/ammonia_production.csv",
+        ammonia_plants="data/industry/ammonia_plants.csv",
```
davide-f: outputs should be in resources. In data we host "only" raw data

davide-f (commented):
Many thanks @yerbol-akhmetov for the addition! Added a few comments

@danielelerede-oet , in past discussions did I understand right that you'll be needing ammonia too?
If so and had you the availability, it would be interesting to have your opinion on how we can make ammonia section easier to customize for other regions.
If you are not expected to use this feature in the foreseeable feature, feel free to ignore the comment.
Many thanks!

data/industry/ammonia_plants.csv
yerbol-akhmetov: Hi, @davide-f. Thanks for the comment. Totally, not necessary. I have added later. Initially, I have kept it without `_eu` prefix. I have even thought that we can use that file to fill ammonia capacity data for other regions. So it can be our ammonia dataset, which is filled in the future. (then I thought should that be in separate repo or not). This is what I wanted to discuss too.

scripts/build_ammonia_production.py
yerbol-akhmetov: > Download functions are better treated with separate retrieve functions, which will help the data management.

Thank you for the comment, @davide-f. This is another point that I have discussed. My thought was to create a separate rule to download the dataset. We have discussed with @ekatef, if we should have separate rules for each dataset downloaded (ammonia, steel, and so on). Or we have a single rule `retrieve_dataset`, which will have wildcards or so with dataset name. I totally agree that having separate download function makes scripts more readable and workflow more clear. The only question should I go with `retrieve_ammonia_dataset` or general `retrieve_dataset`?

yerbol-akhmetov: > The additional data_sources file is not strictly necessary depending on the implementation that you'll do. If we can avoid it for now, better, though I'm not against keeping it for the moment until we agree on the data management #1668 / #1508

Thank you for the comment. This is another point we discussed with @ekatef. If we go with a single general rule for downloading datasets, like `retrieve_dataset` then having such `data_sources` might be helpful. I just did not want to put URLs inside of the scripts. This `data_sources` is just proposal, which can be easily changed/discarded. I will put the url inside of the script for now. Do you think it is ok, @davide-f?

scripts/build_ammonia_production.py
yerbol-akhmetov: Hi, @davide-f. Thank you for the feedback and clear instructions. I will apply them.

Snakefile
yerbol-akhmetov: Thank you for the comment, @davide-f. I have put the outputs to data, because we have `industrial_database.csv` in `data` folder. Ammonia output is used for `industrial_database.csv` creation. If we will have it in resources, then `industrial_database` will be rerun each time for new run name. Also, I think we can have ammonia outputs generation only once as for other industries. What do you think?

yerbol-akhmetov: Another option is to move `industrial_database.csv` to resources too. But it means that industrial database will be rerun for each run name. However, it seems unnecessary increase of simulation time. As a suggestion, I have put ammonia outputs and industrial database output into `resources` directly without `RDIR`.

yerbol-akhmetov:
Hello, @davide-f. I have applied all your suggestions. Download and processing functions are separated to individual rules. Could you please review it?

ekatef:
Hello @yerbol-akhmetov @davide-f, thank you for working on that! As mentioned, I see it more effective if you'd focus for this PR on a data man approach you are most comfortable with. Just a few quick high-level comments.
1. I think the original idea by @yerbol-akhmetov of using yaml as a format for data inventory. In contract to csv, it feels more human-readble, and I think we must use it in cases when manual data input is expected from user (and #1711 must be revised accordingly)
2. Just to be sure: could you please make double-check all the csv committed to the repo are licensed in a way which allows for re-distribution?

yerbol-akhmetov:
> Hello @yerbol-akhmetov @davide-f, thank you for working on that! As mentioned, I see it more effective if you'd focus for this PR on a data man approach you are most comfortable with. Just a few quick high-level comments.
> 
> 1. I think the original idea by @yerbol-akhmetov of using yaml as a format for data inventory. In contract to csv, it feels more human-readble, and I think we must use it in cases when manual data input is expected from user (and [Streamline data management #1711](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1711) must be revised accordingly)
> 2. Just to be sure: could you please make double-check all the csv committed to the repo are licensed in a way which allows for re-distribution?

Thank you for the comments, @ekatef. 
1. Here, I think both ways work, but I really liked the idea of @davide-f to tackle everything within one PR of streamline managaement.
2. I have added only one csv (`data/industry/ammonia_plants.csv`). I have added the license information for it. Other datasets, like USGS or US cities we are not hosting.

yerbol-akhmetov:
Hello, @davide-f, @ekatef. I have applied all requested changes. I would be glad if you could provide feedback,  please.

ekatef:
Hey @yerbol-akhmetov, thank you for checking the license and completely agree that revision of the data management must be done in other PRs. I just like your initial idea of using yaml to store meta data, and have noted it for further work. But it's a different story.

For the implementation itself, I don't want to mess-up review work you are doing with @davide-f but happy to assist if needed.

Snakefile
davide-f: I like your proposal. It is in resources folder and independent of RDIR. Good for me, thanks!

data/industry/ammonia_plants.csv
davide-f: Thanks for the revision!

scripts/build_ammonia_production.py
davide-f: Thanks! Good for me

scripts/build_industry_demand.py
davide-f: we don't have yet the warning correct?

davide-f (approved):
Hello!
This PR is basically ready to go for me and I'm already approving it.
Yerbol and Katia feel free to finalize as you best see fit.

I'd only have a non-blocking comment mentioned above about a warning if the max function is executed. From what I see, that may induce a mismatch of values with energy totals and it would deserve a warning.

Many thanks!

scripts/build_industry_demand.py
yerbol-akhmetov: Hi, @davide-f. The warning was added. However, I have forgot to mention it as a response to comment. Thanks for noting, Here is the warning message in the script.

https://github.com/yerbol-akhmetov/pypsa-earth/blob/1b3f32bbf41ec1b020718dd88e64f8b14e9ea304/scripts/build_industry_demand.py#L336-L344

yerbol-akhmetov:
> Hello! This PR is basically ready to go for me and I'm already approving it. Yerbol and Katia feel free to finalize as you best see fit.
> 
> I'd only have a non-blocking comment mentioned above about a warning if the max function is executed. From what I see, that may induce a mismatch of values with energy totals and it would deserve a warning.
> 
> Many thanks!

Hi, @davide-f. Thank you for the review. The warning message was added to the script. However, I have forgotten to mention it.
image

ekatef:
Hello!
The PR looks good to me, and I think we can merge it. A nice feature and neat implementation. Well done @yerbol-akhmetov!

ekatef:
Have merged as agreed
Great team work @yerbol-akhmetov @davide-f!

Issue #1203, opened 2024-11-20
Implement unit testing for extra-functionality constraints in the solving script


ekatef:
## Describe the feature you'd like to see

We have a number of functions in `solve_network` script which implements diverse extra-constraints for the model, while they testing could have been limited up to the data. Examples: `add_chp_constraints` or `add_EQ_constraints`

Given the domain significance of these functionality, unit testing can work perfectly there to ensure stability against regression. The implementation approach can contain of the following steps:
- refactoring of the solving scripts
- debug and introducing fixes for all the functions in question
- formulation requirements of UT for each function
- implementation of UT themselves.

ekatef:
FYI @GbotemiB 
May you be interested to formulate the requirements for unit testing? I remember that you have still be testing `add_chp_constraints` which could give great insights for this task. 

As a comment, that is crucial to understand what exactly do we aim to check to maintain the balance between simplicity, ability to capture bugs and stability to false-positive.

GbotemiB:
Hello @ekatef, I am dropping a list of functions that we might need to write unit tests for.

Constraints Checklist
- add_land_use_constraint
- add_CCL_constraints
- add_EQ_constraints
- add_BAU_constraints
- add_SAFE_constraints
- add_operational_reserve_margin_constraint
- update_capacity_constraint
- add_battery_constraints
- add_RES_constraints
- H2_export_yearly_constraint
- monthly_constraints
- add_chp_constraints
- add_co2_sequestration_limit
- set_h2_colors

darsh-mishr:
Hi all, sharing a small unit test for `add_chp_constraints` that came up while fixing #1750 (PR #1801). Posting here so it can be tracked with the broader testing effort instead of bundling it into the bug-fix PR.

The test covers the no-links case: if `n.links` is empty, the function should return without raising.

```python
def _network_without_links():
    n = pypsa.Network()
    n.set_snapshots(pd.date_range("2013-01-01", periods=3, freq="h"))
    n.add("Bus", "bus0")
    n.add("Generator", "gen0", bus="bus0", p_nom=100, marginal_cost=10)
    n.add("Load", "load0", bus="bus0", p_set=50)
    return n


def test_add_chp_constraints_returns_when_no_links():
    """Regression test for #1750.

    add_chp_constraints must not raise when the network has no links.
    Operational models without DC interconnects, batteries, or pumped
    storage produce an empty n.links table; pandas .str accessor on the
[...]

PR #1801, opened 2026-04-25
Fix crash in add_chp_constraints for networks without links

### LINKED ISSUE #1750: Improve robustness of solving scripts for operational models
## Checklist

## Describe the Bug

Currently, we silently assume in `solve_network` that there are always some links in the model. However, operational-focused models can have no links at all (if there are no brownfield or expandable batteries, and no pumped storage, or DC links). In this case, `n.links.index.str` fails with an error message which isn't particularly clear.

## Error Message

```bash
add_chp_constraints(n)
  File "pypsa-earth/.snakemake/shadow/tmpdlp9rbag/.snakemake/scripts/tmpsjechpod.solve_network.py", line 819, in add_chp_constraints
    n.links.index.str.contains("urban central")
    ^^^^^^^^^^^^^^^^^
  File "miniforge3/envs/pypsa-earth-aug/lib/python3.11/site-packages/pandas/core/accessor.py", line 224, in __get__
    accessor_obj = self._accessor(obj)
                   ^^^^^^^^^^^^^^^^^^^
  File "miniforge3/envs/pypsa-earth-aug/lib/python3.11/site-packages/pandas/core/strings/accessor.py", line 194, in __init__
    self._inferred_dtype = self._validate(data)
                           ^^^^^^^^^^^^^^^^^^^^
  File "miniforge3/envs/pypsa-earth-aug/lib/python3.11/site-packages/pandas/core/strings/accessor.py", line 248, in _validate
    raise AttributeError("Can only use .str accessor with string values!")
AttributeError: Can only use .str accessor with string values!
```

darmis007:
## Summary

Fixes #1750.

`add_chp_constraints` calls `n.links.index.str.contains(...)` unconditionally, which raises `AttributeError: Can only use .str accessor with string values!` when `n.links` is empty (the empty Index has no string dtype to infer from).

This trips operational and dispatch-validation models for regions without DC interconnects, batteries, or pumped storage — `n.links` is legitimately empty in those cases, but the constraint setup runs anyway because `sector.chp` defaults to `true` in `config.default.yaml`.

The function would have produced no constraints on an empty `n.links` regardless: every downstream block is gated by `if not <subset>.empty`, all of which are false on an empty network. The fix is an explicit early return that short-circuits before the `.str` accessor.

## Changes

- `scripts/solve_network.py` — early return in `add_chp_constraints` when `n.links` is empty (3 lines).
- `test/test_solve_network.py` — new pytest regression test asserting the function returns cleanly on an empty-links network.
- `doc/release-notes.md` — note under *Minor Changes and bug-fixing*.

## Verification

- `pytest test/test_solve_network.py` passes locally.
- Confirmed the test catches the original bug: replaying the pre-fix function against an empty-links network reproduces the exact `AttributeError` from the issue.
- `pre-commit run` passes on all changed files (isort, black, codespell, REUSE).

## Prior art

The same one-line guard was applied independently in [open-energy-transition/pypsa-zambia#12](https://github.com/open-energy-transition/pypsa-zambia/pull/12) (merged 2026-03-13) for a Zambia dispatch-validation run. That fix was bundled with downstream-only changes and was not upstreamed; this PR brings it to the upstream repo with a regression test.

## Test plan

- [x] `pytest test/test_solve_network.py` passes
- [x] Pre-fix repro confirms the test would have caught the bug
- [x] `pre-commit run --files` clean on all changes
- [ ] CI on the PR

### DIFF
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -30,6 +30,8 @@ This part of documentation collects descriptive release notes to capture the mai
 
 **Minor Changes and bug-fixing**
 
+* Avoid crash in `add_chp_constraints` when the network has no links, supporting operational models without brownfield/expandable batteries, pumped storage, or DC links [#1750](https://github.com/pypsa-meets-earth/pypsa-earth/issues/1750)
+
 * Avoid crash in add_electricity when no extendable generators are configured [PR #1794](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1794)
 
 * Include country-specific nuclear `p_max_pu` values based on IAEA dataset [PR #1718](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1718)
--- a/scripts/solve_network.py
+++ b/scripts/solve_network.py
@@ -826,6 +826,9 @@ def hydrogen_temporal_constraint(n, n_ref, time_period):
 
 
 def add_chp_constraints(n):
+    if n.links.empty:
+        return
+
     electric_bool = (
         n.links.index.str.contains("urban central")
         & n.links.index.str.contains("CHP")


davide-f (changes requested):
Many thanks!

This fix is straightforward, no comments about it.

The test unit is very interesting for the future when test units are included.
To merge the fix fast, I propose to remove the test_solve_network from this PR and post the proposed solution in this issue
https://github.com/pypsa-meets-earth/pypsa-earth/issues/1203
so we can track that.

If you have other ideas, please share.

Many thanks!

darsh-mishr:
Hi @davide-f, thanks for the review!

I have removed `scripts/test_solve_network.py` from this PR (force-pushed), so it now only contains the fix and the release note entry. The test has been posted as a comment on #1203 so it can be tracked alongside the broader unit testing effort.

Let me know if anything else is needed before merge.

davide-f (approved):
Many thanks! Congratulations for becoming contributor to the initiative! :D

Merging :)

PR #1816, opened 2026-05-05
fix(solve_network): rhs UnboundLocalError and back-pressure leak in add_chp_constraints

### LINKED ISSUE #1814: add_chp_constraints: UnboundLocalError on greenfield CHP, leaked RHS on brownfield mix
## Description

In `scripts/solve_network.py`, `add_chp_constraints` mishandles the local
variable `rhs`. The variable is assigned inside one `if` block but read
inside a different `if` block. Because the two blocks are gated by
different conditions, two distinct broken scenarios exist:

- **Scenario A: crash on greenfield CHP.** When all CHP links are
  extendable (no fixed CHP), `rhs` is never assigned, but it is still
  read at the back-pressure step. Function raises `UnboundLocalError`.

- **Scenario B: silent wrong constraint on brownfield mix.** When a
  network has both fixed and extendable CHP, `rhs` is assigned in the
  fixed-CHP block (set to the fixed plants' nominal capacity) and then
  reused in the back-pressure block, where the structurally correct RHS
  is `0`. The back-pressure inequality is silently relaxed by the
  fixed-CHP capacity.

The relevant lines:

```python
# top_iso_fuel_line for fixed
if not electric_fix.empty:
    rhs = n.links.p_nom[electric_fix]                                 # assigned here
    n.model.add_constraints(lhs <= rhs, name="chplink-top_iso_fuel_line_fix")

# back-pressure
if not electric.empty:
    lhs = (...)
    n.model.add_constraints(lhs <= rhs, name="chplink-backpressure")  # read here
```

Behavior across the four CHP scenarios:

| `electric_fix.empty` | `electric.empty` | `rhs` at back-pressure step       | Outcome                                               |
| -------------------- | ---------------- | --------------------------------- | ----------------------------------------------------- |
| True                 | True             | (never read)                      | OK                                                    |
| False                | False            | leaked from fix block (e.g., 100) | Wrong constraint: `lhs <= 100` instead of `lhs <= 0`  |
| **True**             | **False**        | undefined                         | **`UnboundLocalError`**                               |

Three of the four CHP-bearing scenarios are broken: one with a crash,
two with a silent wrong constraint.

### Scenario A: `UnboundLocalError`

```python
import sys; sys.path.insert(0, "scripts")
import pandas as pd
import pypsa
from solve_network import add_chp_constraints

n = pypsa.Network()
n.set_snapshots(pd.date_range("2013-01-01", periods=2, freq="h"))
n.add("Bus", "AC")
n.add("Generator", "g", bus="AC", p_nom=200, marginal_cost=10)
n.add("Load", "l", bus="AC", p_set=50)
n.add("Link", "urban central CHP electric",
      bus0="AC", bus1="AC", p_nom_extendable=True, efficiency=0.40)
n.add("Link", "urban central CHP heat",
      bus0="AC", bus1="AC", p_nom_extendable=True, efficiency=0.50)
n.links["p_nom_ratio"] = 1.0
n.links["c_b"] = 0.75

n.optimize.create_model()
add_chp_constraints(n)
```

### Scenario B: leaked RHS

```python
import sys; sys.path.insert(0, "scripts")
import pandas as pd
import pypsa
from solve_network import add_chp_constraints

n = pypsa.Network()
n.set_snapshots(pd.date_range("2013-01-01", periods=2, freq="h"))
n.add("Bus", "AC")
n.add("Generator", "g", bus="AC", p_nom=200, marginal_cost=10)
n.add("Load", "l", bus="AC", p_set=50)
# fixed CHP
n.add("Link", "urban central CHP electric fix",
      bus0="AC", bus1="AC", p_nom=100, p_nom_extendable=False, efficiency=0.40)
n.add("Link", "urban central CHP heat fix",
      bus0="AC", bus1="AC", p_nom=100, p_nom_extendable=False, efficiency=0.50)
# extendable CHP
n.add("Link", "urban central CHP electric ext",
      bus0="AC", bus1="AC", p_nom_extendable=True, efficiency=0.40)
n.add("Link", "urban central CHP heat ext",
      bus0="AC", bus1="AC", p_nom_extendable=True, efficiency=0.50)
n.links["p_nom_ratio"] = 1.0
n.links["c_b"] = 0.75

n.optimize.create_model()
add_chp_constraints(n)
print(n.model.constraints["chplink-backpressure"].rhs.values)
```

### Scenario A

The function returns without error and registers three constraints:

- `chplink-fix_p_nom_ratio`
- `chplink-top_iso_fuel_line_ext`
- `chplink-backpressure` with `rhs = 0` everywhere

### Scenario B

The function returns without error and registers four constraints
(the three above plus `chplink-top_iso_fuel_line_fix`).

`chplink-backpressure.rhs` should be `0` for every entry.

### Scenario A

```
  File ".../scripts/solve_network.py", line 885, in add_chp_constraints
    n.model.add_constraints(lhs <= rhs, name="chplink-backpressure")
                                   ^^^
UnboundLocalError: cannot access local variable 'rhs' where it is not
associated with a value
```

### Scenario B

`chplink-backpressure.rhs` contains `100.0` for the rows corresponding
to fixed CHP (leaked from the fix block) instead of the expected `0.0`:

```
[[100. 100.]
 [ -0.  -0.]
 [ -0.  -0.]]
```

## Suggested fix

Initialize `rhs = 0` at the top of the back-pressure block. `0` is the
structurally correct RHS for the back-pressure inequality (`lhs <= 0`),
and the explicit assignment also overrides any value leaked from the
preceding fix block:

```python
# back-pressure
if not electric.empty:
    rhs = 0
    lhs = (...)
    n.model.add_constraints(lhs <= rhs, name="chplink-backpressure")
```

One line. Resolves both scenarios.

darmis007:
## Summary
Resolves #1814 with a single line change to `add_chp_constraints` in
`scripts/solve_network.py`.

The local variable `rhs` was assigned only inside the fixed-CHP block
(`if not electric_fix.empty:`) but read inside the back-pressure block
(`if not electric.empty:`). The two blocks are gated by different
conditions, so two scenarios were broken:

- **Greenfield CHP** (only extendable CHP, no fixed CHP): function
  raised `UnboundLocalError` because `rhs` was never assigned.
- **Brownfield mix** (fixed + extendable CHP coexist): back-pressure
  RHS silently inherited the fixed plant's `p_nom` value (e.g. 100)
  from the preceding block, where the structurally correct RHS is `0`.

The fix initializes `rhs = 0` at the top of the back-pressure block.
This guarantees `rhs` is always defined when read, and any value left
over from the fix block is overridden by the correct value for the
back-pressure inequality `lhs <= 0`.

A unit-test regression for both scenarios will follow once the test
scaffolding in #1813 lands.

Closes #1814.

## Test plan
- [x] Reproducer from #1814 Scenario A: function returns without error
- [x] Reproducer from #1814 Scenario B: `chplink-backpressure.rhs` is `0` everywhere
- [x] `pre-commit run` passes
- [ ] CI green

### DIFF
--- a/scripts/solve_network.py
+++ b/scripts/solve_network.py
@@ -882,7 +882,7 @@ def add_chp_constraints(n):
             p.loc[:, heat] * (n.links.efficiency[heat] * n.links.c_b[electric].values)
             - p.loc[:, electric] * n.links.efficiency[electric]
         )
-        n.model.add_constraints(lhs <= rhs, name="chplink-backpressure")
+        n.model.add_constraints(lhs <= 0, name="chplink-backpressure")
 
 
 def add_co2_sequestration_limit(n, sns):
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -32,6 +32,8 @@ This part of documentation collects descriptive release notes to capture the mai
 
 **Minor Changes and bug-fixing**
 
+* Fix `UnboundLocalError` and back-pressure RHS leak in `add_chp_constraints` [#1814](https://github.com/pypsa-meets-earth/pypsa-earth/issues/1814)
+
 * Add config parameter `enable_electricity_connection_cost`, which adds electricity grid connection costs to fixed capital costs for solar and onwind generators in prepare_sector_network [PR #1810](https://github.com/pypsa-meets-earth/pypsa-earth/issues/1810)
 
 * Avoid crash in `add_chp_constraints` when the network has no links, supporting operational models without brownfield/expandable batteries, pumped storage, or DC links [PR #1750](https://github.com/pypsa-meets-earth/pypsa-earth/issues/1750)


scripts/solve_network.py
```diff
@@ -878,6 +878,7 @@ def add_chp_constraints(n):
 
     # back-pressure
     if not electric.empty:
+        rhs = 0
```
ekatef: As a very minor comment, I wonder if we could improve consistency with the extendable block above just hard-coding rhs directly into the constraint definition `n.model.add_constraints(lhs <= 0, name="chplink-backpressure")`?

ekatef (commented):
Hello @darmis007, and thank you so much for the detailed investigation! Completely agree with the fix. Have added a minor styling suggestion but it's rather subjective.

I think the PR is ready to merge once CI passes. Would you mind to resolve merge conflicts?

scripts/solve_network.py
darsh-mishr: Done in 63e5faa5 — inlined `0` directly in the back-pressure constraint, dropped the `rhs = 0` local. Matches the convention of `chplink-fix_p_nom_ratio` and `chplink-top_iso_fuel_line_ext` above. Merge conflict in `doc/release-notes.md` also resolved by the rebase.

ekatef: Thanks a lot!

ekatef (approved):
Thank you @darmis007! Looks great and ready to merge once CI successful.

darsh-mishr:
Hi @ekatef, the macOS check failed again at the same `prepare_ports` step (third time, same external download issue). Ubuntu and Windows pass, and the failure is unrelated to the diff. Would you be open to merging as-is? Happy to wait for another rerun if you'd prefer.

ekatef:
> Hi @ekatef, the macOS check failed again at the same `prepare_ports` step (third time, same external download issue). Ubuntu and Windows pass, and the failure is unrelated to the diff. Would you be open to merging as-is? Happy to wait for another rerun if you'd prefer.

For code-related PRs, our usual quality standard is to have CI successful for all the operation systems. Sometimes, there may be some system-related issues.

Have re-run CI, and it has passed so we can merge with all the respect to QA procedures 😄 

Congratulations with getting this done, and thank you so much for the contribution! 🎉

PR #654, opened 2023-03-30
Improve earth coverage

davide-f:
# Closes #556 

## Checklist

- [x] I tested my contribution locally and it seems to work fine.
- [x] Code and workflow changes are sufficiently documented.
- [x] Newly introduced dependencies are added to `envs/environment.yaml` and `envs/environment.docs.yaml`.
- [x] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [x] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [x] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [x] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes, including reference to the requested PR.

PR #1819, opened 2026-05-10
Reduce CI

### LINKED ISSUE #1800: Shorten testing CI

Our CI is currently taking about 1.5 hours which is hampering developments, and it could be handy to shorten it a bit.

Summarising the options we have discussed with @yerbol-akhmetov and @GbotemiB :
1. Decrease a number of runs in Monte-Carlo tests (e.g. make 2-3 simulations instead current 9)
2. Clean-up config files used in testing and make sure our integration testing doesn't have duplications

yerbol-akhmetov:
# Closes #1800 

Good day. This PR aims to reduce the execution time of the CI tests.

## Checklist


yerbol-akhmetov:
Execution times before and after the test updates:

image

It is important to note that `landlock` and `custom` test configs are executed with separate `RDIR`. I think it is possible to bring `custom` config under common `test` directory. We just need to make sure that output names will be different from previous tests. It might save some more additional time.

test/config.test.yaml
```diff
@@ -0,0 +1,9 @@
+# SPDX-FileCopyrightText:  PyPSA-Earth and PyPSA-Eur Authors
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+run:
+  name: test
+  sector_name: test
+  base_config: config.tutorial.yaml  # base configuration file
```
ekatef, line 8: Both `base_config` and `scenarios` folder have been implemented as a part of `run_all_scenarios` infrastructure in [#654](https://github.com/pypsa-meets-earth/pypsa-earth/pull/654) which we left aside from testing for now

To keep consistency, I'd suggest to place `config.test.yaml` into `test` folder just in the same way as other testing configs

scripts/_helpers.py
```diff
@@ -833,8 +833,11 @@ def filter_codes(c_list, iso_coding=True):
         # create a list with all countries
         full_codes_list.extend(codes_list)
 
-    # Removing duplicates and filter outputs by coding
-    full_codes_list = filter_codes(list(set(full_codes_list)), iso_coding=iso_coding)
+    # Removing duplicates while preserving order keeps Snakemake params stable
+    # across separate workflow invocations.
+    full_codes_list = filter_codes(
+        list(dict.fromkeys(full_codes_list)), iso_coding=iso_coding
```
ekatef: A good point with keeping the order stable. 

I wonder if it could make sense to sort alphabetically? Not sure it's easy to comprehend what exactly the original order is

ekatef:
Hello @yerbol-akhmetov, that is excellent!
A really insightful investigation, and the result is impressive.

For `configs/scenarios/config.NG.yaml` agree that it's better to avoid including it into testing. However, I see value in keeping it since it demonstrates how to use `run_all_scenarios` for a scenarios sweep. Indeed, some improvements may be needed to document usage of `run_all_scenarios` but it's a different story.

On point to cross-check is that each solved test network can be distinguished from the others to make sure we can extract the objective value from them. When checking the objective outputs, that seems to be satisfied but good to be checked

image

We'll also need to update the reference objectives, and there is #1806 which enables automation for that.

On the implementation side, I'm adding two comments directly to the code but they are minor. Really looking forward to see this PR completed 😄

ekatef (commented):
Sharing more specific comments in the code, as promised

yerbol-akhmetov:
Thanks for the feedback, @ekatef. For testing `run_all_scenarios`, I have placed `config.tutorial.test.yaml` from `tests/` to `configs/scenarios/` and renamed it `config.test.yaml`. It works fine. The only problem is that the following `monte-carlo` with re-run all rules after `run_all_scenarios`. With providing `solve_all_networks --configfile config.tutorial.yaml configs/scenarios/config.test.yaml`, the rules are not re-run in the following test. I think with `run_all_scenarios` somehow params differ. If we use, `run_all_scenarios`, we should add around 6 minutes to full execution time.

scripts/_helpers.py
yerbol-akhmetov: Thank you for the comment, @ekatef. Having ordering can be quite helpful to be consistnt. I will implement it.

test/config.test.yaml
yerbol-akhmetov: Hi, @ekatef. Thank you for the suggestion. Can you please clarify it a bit, I did not fully get it? I mean the part about `base_config`? Do you say we can omit it?

scripts/_helpers.py
gustavo-dias: Hi yerbol, I see you changed from set() to dict.fromkeys() to enforce uniqueness. Perhaps it would be good to mention the reason behind the change  (performance, preserve order, etc) in a line comment to avoid a future developer revert it back to set(). Something like: "Using dict.fromkeys() instead of set() because..."

yerbol-akhmetov: Thank you, @gustavo-dias. I have added the explanation as suggested.

test/config.test.yaml
yerbol-akhmetov: > Both `base_config` and `scenarios` folder have been implemented as a part of `run_all_scenarios` infrastructure in [#654](https://github.com/pypsa-meets-earth/pypsa-earth/pull/654) which we left aside from testing for now
> 
> To keep consistency, I'd suggest to place `config.test.yaml` into `test` folder just in the same way as other testing configs

Thank you, @ekatef. I have moved the config to `test/`.

ekatef:
> Thanks for the feedback, @ekatef. For testing `run_all_scenarios`, I have placed `config.tutorial.test.yaml` from `tests/` to `configs/scenarios/` and renamed it `config.test.yaml`. It works fine. The only problem is that the following `monte-carlo` with re-run all rules after `run_all_scenarios`. With providing `solve_all_networks --configfile config.tutorial.yaml configs/scenarios/config.test.yaml`, the rules are not re-run in the following test. I think with `run_all_scenarios` somehow params differ. If we use, `run_all_scenarios`, we should add around 6 minutes to full execution time.

Hey @yerbol-akhmetov thank you for exploring options for testing `run_all_scenarios`. As a technical comment, it's behaviour is a bit different from all the other rules and relays on a number of additional assumptions which allow for a scenario-management system. The logic is great and functionality can be very effective but I'm not sure how relevant it's for CI right now.

test/config.test.yaml
ekatef: > Hi, @ekatef. Thank you for the suggestion. Can you please clarify it a bit, I did not fully get it? I mean the part about `base_config`? Do you say we can omit it?

Yeah, I think we can leave `run_all_scenarios` outside of this PR to keep focus, and accordingly to remove `base_config` from testing configs. Does it make sense for you?

ekatef: > > Both `base_config` and `scenarios` folder have been implemented as a part of `run_all_scenarios` infrastructure in [#654](https://github.com/pypsa-meets-earth/pypsa-earth/pull/654) which we left aside from testing for now
> > To keep consistency, I'd suggest to place `config.test.yaml` into `test` folder just in the same way as other testing configs
> 
> Thank you, @ekatef. I have moved the config to `test/`.

Great, thanks for the revision!

configs/scenarios/config.NG.yaml
```diff
@@ -1,15 +0,0 @@
...
-run:
-  name: NG  # Name of the configuration (arbitrary string value)
-  shared_cutouts: true  # set to true to share the default cutout(s) across runs
-  base_config: config.tutorial.yaml  # base configuration file
-
-countries:
-- NG
-
-scenario:
-  clusters: [5]
```
ekatef, line 15: I'd leave `config.NG.yaml` as-is since it's needed as an example for `run_all_scenarios`

ekatef (commented):
Thanks a lot @yerbol-akhmetov. I think the PR is close to be completed, and really looking forward to see it merged. A reduction in the testing time is impressive and will help a lot to facilitate maintenance.

Although, I do agree that `run_all_scenarios` may need some additional alignment. Personally, I'd be pro keeping `run_all_scenarios` out of scope of this PR by 1) leaving `config.NG.yaml` unchanged, 2) removing `bas_config` from all the configs used for testing. Does it make sense for you?

yerbol-akhmetov:
> > Thanks for the feedback, @ekatef. For testing `run_all_scenarios`, I have placed `config.tutorial.test.yaml` from `tests/` to `configs/scenarios/` and renamed it `config.test.yaml`. It works fine. The only problem is that the following `monte-carlo` with re-run all rules after `run_all_scenarios`. With providing `solve_all_networks --configfile config.tutorial.yaml configs/scenarios/config.test.yaml`, the rules are not re-run in the following test. I think with `run_all_scenarios` somehow params differ. If we use, `run_all_scenarios`, we should add around 6 minutes to full execution time.
> 
> Hey @yerbol-akhmetov thank you for exploring options for testing `run_all_scenarios`. As a technical comment, it's behaviour is a bit different from all the other rules and relays on a number of additional assumptions which allow for a scenario-management system. The logic is great and functionality can be very effective but I'm not sure how relevant it's for CI right now.

Hi, @ekatef. Thank you for the comment. I totally agree with you. Therefore, I have not included it in the final version of CI.

yerbol-akhmetov:
> Thanks a lot @yerbol-akhmetov. I think the PR is close to be completed, and really looking forward to see it merged. A reduction in the testing time is impressive and will help a lot to facilitate maintenance.
> 
> Although, I do agree that `run_all_scenarios` may need some additional alignment. Personally, I'd be pro keeping `run_all_scenarios` out of scope of this PR by 1) leaving `config.NG.yaml` unchanged, 2) removing `bas_config` from all the configs used for testing. Does it make sense for you?

Thanks, @ekatef. I agree with not including `run_all_scenarios`. However, regarding the `config.NG.yaml`, it seemed to me as being there to test `run_all_scenarios` manually, right? It will take time to run this rule, because we have 5 clusters, so most of the rules will be re-executed. I was thinking if `config.test.yaml` which is basically tutorial config can replace it. It was that reason I put `config.test.yaml` to `configs/scenarios/config.test.yaml`, thus we will have it in scenarios and being run with CI. What do you think? So it will be possible to manually run `run_all_scenarios`. If we keep `config.NG.yaml` and run after myopic run, it will take maybe additional 7-8 minutes.

ekatef:
> > Thanks a lot @yerbol-akhmetov. I think the PR is close to be completed, and really looking forward to see it merged. A reduction in the testing time is impressive and will help a lot to facilitate maintenance.
> > Although, I do agree that `run_all_scenarios` may need some additional alignment. Personally, I'd be pro keeping `run_all_scenarios` out of scope of this PR by 1) leaving `config.NG.yaml` unchanged, 2) removing `bas_config` from all the configs used for testing. Does it make sense for you?
> 
> Thanks, @ekatef. I agree with not including `run_all_scenarios`. However, regarding the `config.NG.yaml`, it seemed to me as being there to test `run_all_scenarios` manually, right? It will take time to run this rule, because we have 5 clusters, so most of the rules will be re-executed. I was thinking if `config.test.yaml` which is basically tutorial config can replace it. It was that reason I put `config.test.yaml` to `configs/scenarios/config.test.yaml`, thus we will have it in scenarios and being run with CI. What do you think? So it will be possible to manually run `run_all_scenarios`. If we keep `config.NG.yaml` and run after myopic run, it will take maybe additional 7-8 minutes.

Great! I do appreciate your attention to the details, and agree that it could be awesome to have some system in how do we deal with testing.

For `configs/config.NG.yaml`, it is needed rather as an example than for testing, and I think we must make sure that it'll remain in some form in the repo since it's currently the only way to find how `run_all_scenarios` is working. Not sure I get the point about `configs/scenarios/config.test.yaml`: is it added to this PR?

ekatef:
@yerbol-akhmetov to give a bit more details about `run_all_scenarios`: @davide-f has developed `run_all_scenarios` to build and run the model for every country of the world which is a reason of having `NG` in the name. While not sure we have many opportunities to use it since then while the method definitely deserves it.

Currently, I'm testing `run_all_scenarios` for Zambian project and have found that it provides a very effective way to run a scenario sweep but we need to document it and eventually add some design polishes (in particular, current way `run_all_scenarios` is working based to re-writing `config.yaml` which can lead to some side effects). My understanding is that it may be a bit too early for including `run_all_scenarios ` into a testing chain but you may be absolutely right that it could be worth to have an opportunity to test it manually.

@davide-f could you have any comments on `run_all_scenarios`? It would be great to have your opinion on that since you have the most detailed understanding here.

yerbol-akhmetov:
> > > Thanks a lot @yerbol-akhmetov. I think the PR is close to be completed, and really looking forward to see it merged. A reduction in the testing time is impressive and will help a lot to facilitate maintenance.
> > > Although, I do agree that `run_all_scenarios` may need some additional alignment. Personally, I'd be pro keeping `run_all_scenarios` out of scope of this PR by 1) leaving `config.NG.yaml` unchanged, 2) removing `bas_config` from all the configs used for testing. Does it make sense for you?
> > 
> > 
> > Thanks, @ekatef. I agree with not including `run_all_scenarios`. However, regarding the `config.NG.yaml`, it seemed to me as being there to test `run_all_scenarios` manually, right? It will take time to run this rule, because we have 5 clusters, so most of the rules will be re-executed. I was thinking if `config.test.yaml` which is basically tutorial config can replace it. It was that reason I put `config.test.yaml` to `configs/scenarios/config.test.yaml`, thus we will have it in scenarios and being run with CI. What do you think? So it will be possible to manually run `run_all_scenarios`. If we keep `config.NG.yaml` and run after myopic run, it will take maybe additional 7-8 minutes.
> 
> Great! I do appreciate your attention to the details, and agree that it could be awesome to have some system in how do we deal with testing.
> 
> For `configs/config.NG.yaml`, it is needed rather as an example than for testing, and I think we must make sure that it'll remain in some form in the repo since it's currently the only way to find how `run_all_scenarios` is working. Not sure I get the point about `configs/scenarios/config.test.yaml`: is it added to this PR?

Hi, @ekatef. Thanks for the feedback. During the implementation phase I have placed test config into `config/scenarios/` folder: `configs/scenarios/config.test.yaml`. Then, you have suggested to put it to `test/` directory. I think we can keep `configs/scenarios/config.NG.yaml` there. But we can just not include it to CI test. Because it does not bring anything new without `run_all_scenarios`. 

Regarding `run_all_scenarios`, I do not think we need to test it in CI as you have said. One problem with that snakemake does not tracks params well with `run_all_scenarios`.

yerbol-akhmetov:
> @yerbol-akhmetov to give a bit more details about `run_all_scenarios`: @davide-f has developed `run_all_scenarios` to build and run the model for every country of the world which is a reason of having `NG` in the name. While not sure we have many opportunities to use it since then while the method definitely deserves it.
> 
> Currently, I'm testing `run_all_scenarios` for Zambian project and have found that it provides a very effective way to run a scenario sweep but we need to document it and eventually add some design polishes (in particular, current way `run_all_scenarios` is working based to re-writing `config.yaml` which can lead to some side effects). My understanding is that it may be a bit too early for including `run_all_scenarios ` into a testing chain but you may be absolutely right that it could be worth to have an opportunity to test it manually.
> 
> @davide-f could you have any comments on `run_all_scenarios`? It would be great to have your opinion on that since you have the most detailed understanding here.

Totally agree with you, @ekatef. It a great feature. But with the current state we cannot re-use rules if we mix `run_all_scenarios` with regular `solve_all_networks`. `snakemake` somehow does see params to be different.

ekatef:
> Hi, @ekatef. Thanks for the feedback. During the implementation phase I have placed test config into `config/scenarios/` folder: `configs/scenarios/config.test.yaml`. Then, you have suggested to put it to `test/` directory. I think we can keep `configs/scenarios/config.NG.yaml` there. But we can just not include it to CI test. Because it does not bring anything new without `run_all_scenarios`.
> 
> Regarding `run_all_scenarios`, I do not think we need to test it in CI as you have said. One problem with that snakemake does not tracks params well with `run_all_scenarios`.

Hey @yerbol-akhmetov, I confirm that it feels a good solution to put all the testing files into `test` directory. So, all good there. 

The only point I'm not quite comfortable with is removing `configs/scenarios/config.NG.yaml` since it doesn't relate to testing but is a way to give an example, in same way as `data/custom_powerplants.csv` and alike. So, I think it must remain regardless on advancement introduced in testing. Does it make sense?

On the actions side, I suggest to restore `configs/scenarios/config.NG.yaml`, add a release note and go ahead with merging.

yerbol-akhmetov:
Hi, @ekatef. Thank you for the comments. I have restored `configs/scenarios/config.NG.yaml` and added the release notes.

ekatef (approved):
Many thanks @yerbol-akhmetov!!

Issue #1077, opened 2024-08-07
check consistency of options


davide-f:
## Describe the feature you'd like to see

Option "Co2L" to set the co2 limits is case sensitive; "CO2L" does not work.
We could plan if we can avoid case sensitiveness to lead to issues

energyLS:
Dear @davide-f thank you for flagging this issue, I aim to "revitalise" this issue, since @doneachh and me came across this recently. Would you propose to change the regex for the Co2L wildcard to not beeing case senstive (probably a one liner in a few scripts) or do you have a larger solution in mind?

davide-f:
Agree, making it case insensitive is important.
I'll also suggest to harmonize the documentation and configs to have values all uppercase or all lowercase for simplicity

ekatef:
An additional point to track for consistency options: it would be good to track compatibility of year-related parameters in the config. Currently, we have quite many including an implicit assumptions like this one https://github.com/pypsa-meets-earth/pypsa-earth/blob/0ee924bad34316a09bea76b14f3c3ba9384e37c3/config.default.yaml#L213

PR #1845, opened 2026-05-28
Add validation of the actual configuration file

ekatef:
Relates to #1077

Sketching a way to cross-check consistency of the parameters in an actual configuration file. The aim is to focus on the most impactful points:

- [x] the requested load weather years present in a specified demand dataset (GEGIS or DemandCast)
- [ ] the weather year is the same for the snapshots, renewable potential (cutouts) and demand

## Checklist

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -23,6 +23,7 @@ from _helpers import (
     branch,  # Remove if Snakemake >= 8.3.0
 )
 from build_demand_profiles import get_load_paths_gegis
+from check_config_consistency import write_config_consistency_report
 from retrieve_databundle_light import (
     datafiles_retrivedatabundle,
     get_best_bundles_in_snakemake,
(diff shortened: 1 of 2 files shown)


davide-f:
This is quite interesting and I like starting with a simple and common use case!
I'd be prone on triggering errors when the value is not matching the expected value.
To scale, we could also use pydantic, which hopefully simplifies the verification.

What do you think?

ekatef:
> This is quite interesting and I like starting with a simple and common use case! I'd be prone on triggering errors when the value is not matching the expected value. To scale, we could also use pydantic, which hopefully simplifies the verification.
> 
> What do you think?

Thank you for checking that! Agree on the error part. For pydantic, we absolutely must implement it as discussed but keep for simple validations only to avoid over-loading specifications.

For demand however I see it more useful to leave some space for more complex validation logic as compared with a simple year-to-dataset correspondence. At the very least, we must account for the region. In particular, I have found that DemandCast covers the declared 1999-2024 not for all countries, e.g. data for 2024 are not available for Zambia. Additionally, it could be handy to apply other heuristic checks such as cross-checking peak/average ratio. That is out of scope of this PR but I think it's relevant to agree on the architecture.

Does it make sense for you?

davide-f:
> Does it make sense for you?

Fully agree

PR #1846, opened 2026-05-28
Add a check for existence of a cutout file

ekatef:

We have now functionality to retrieve a dedicated cutout which can be a very handy for a number of use-cases. One of example is using pre-compiled cutouts for various years (without the need to extract the whole databundle).

To retrieve a cutout, a name of the cutout must be specified. However, a respective databundle will basically ignore the name and search for the best cutout using only a name specification. That can lead to some inconsistency when the cutout is being checked.

The PR adds a check that an expected cutout file actually exists before trying to download it and issues a meaningful error message in case of a mismatch. 

## Checklist


ekatef:
@virio-andreyana could you please check if the check makes sense for you?

virio-andreyana:
I think it explains the problem better that what we have currently. I agree with this PR

ekatef:
> I think it explains the problem better that what we have currently. I agree with this PR

Thanks a lot for checking @virio-andreyana! Then, I'm adding final touches and requesting review. Would you have capacity to give it a proper analysis?

scripts/retrieve_databundle_light.py
```diff
@@ -1014,7 +1014,22 @@ def check_retrieved_cutout_match(snakemake: Snakemake) -> None:
...
+
+    if not os.path.isfile(cutout_output_fl):
+        raise FileNotFoundError(
+            f"A target cutout file '{cutout_output_fl}' was not extracted. "
+            f"The downloaded bundle is {', '.join(map(str, snakemake.params.bundles_to_download))}"
+            f"Expected bundle output(s): "
+            f"{', '.join(configured_bundle_outputs)}"
+        )
+
+    cutout = atlite.Cutout(cutout_output_fl)
```
davide-f, line 1077: Interesing point and it is a good idea to be more informative. I feel like the error is more subtle than it seems.
This script (and thus this function) runs twice: once for retrieve_cutout (for cutouts) and another for retrieve_databundle (all but cutouts). I don't see if cases that prevent execution for the latter.

The challenge is that in the non-cutout execution, the cutout file may not exist yet as the parallel retrieve_cutout may have not completed the download. Thus this function arrived at "atlite.Cutout(cutout_output_fl)" and could fail; in the next iteration, we had retrieve_cutout completed from the previous iteration and the error no longer appears..
For the algorithm of retrieve_databundle,also, only a single cutout should be downloaded.

I think this function should be revised to limit its application only for retrieve_cutout. The check can be done by checking about whether cutout bundles are downloaded.
May this justification align with the error you are experiencing?

ekatef: No, the error you describe is a different one. I haven't used parallel download. 
Do you see any side effect in the proposed solution?

davide-f: Ok for me: I understand that the download process may fail before this function is triggered. When that happens, snakemake.output[0] is not available, thus leading to a weird error.

Rather than raising an error here, I'd raise a logger.error/warning and return the function.
At the end of retrieve_databundle there is already a log that explains what bundles have not been downloaded and raising the error here prevents that log to appear. I'm proposing a revision

scripts/retrieve_databundle_light.py
```diff
@@ -1059,7 +1059,22 @@ def check_retrieved_cutout_match(snakemake: Snakemake) -> None:
...
+        for output in snakemake.config["databundles"][bundle].get("output", [])
+    ]
+
+    if not os.path.isfile(cutout_output_fl):
+        raise FileNotFoundError(
+            f"A target cutout file '{cutout_output_fl}' was not extracted. "
+            f"The downloaded bundle is {', '.join(map(str, snakemake.params.bundles_to_download))}"
+            f"Expected bundle output(s): "
+            f"{', '.join(configured_bundle_outputs)}"
+        )
```
davide-f, line 1075: I'd propose a lighter implementation: retrieve_databundle already logs the bundles whose download failed, here we re-show them. It feels redundant to me.

I'd suggest skipping this check here and refer to the existing log, see [here](https://github.com/pypsa-meets-earth/pypsa-earth/blob/deed23526bf6d335da383f53b68c3f32411db13b/scripts/retrieve_databundle_light.py#L1039-L1042).
Feel free to suggest improvements in the logging of missing bundles, if you see relevant.

```suggestion
    cutout_output_fl = snakemake.output[0]

    if not os.path.isfile(cutout_output_fl):
        logger.warning(
            f"Skip check of cutout coverage: the target cutout file '{cutout_output_fl}' was not properly downloaded due to the missing bundles explained above."
        )
        return
```

scripts/retrieve_databundle_light.py
davide-f: update: the original issue I was mentioning is not real: snakemake.input is non-null only for retrieve_cutout, though not explicit.
If retrieve_databundle will be used with snakemake.input for entries beyong the cutout, that will lead to unexpected errors.
It would be worth adding a note/TODO here:
https://github.com/pypsa-meets-earth/pypsa-earth/blob/deed23526bf6d335da383f53b68c3f32411db13b/scripts/retrieve_databundle_light.py#L1137

davide-f (commented):
Hello,
Shared minor comments, very close to merge with thanks!

Issue #1850, opened 2026-06-01
Advance and polish run_all_scenarios


ekatef:
## Describe the feature you'd like to see

To run a set of scenarios, `run_all_scenarios` provides great functionality to run set of scenarios in a clean and robust way. To make it more convenient to use, the following points could be addressed:

- [ ] Avoid over-writing config.yaml which is reserved to keep local config settings
- [ ] Improve documentation to provide docstrings, a brief description of functionality and a usage example 
- [ ] Fix duplication of blocks (an detailed description is given bellow)

ekatef:
For updating the config with `update` function, there may be challenges for some edge cases. Adding below a minimal example to reproduce.

### diff config
```python
run:
  name: benchmark_run
  base_config: config.default.yaml

atlite:
  nprocesses: 4
  cutouts:
    cutout-2023-era5:
      module: era5
```

### expected output for atlite block
```python
atlite:
  nprocesses: 4
  cutouts:
    cutout-2023-era5:
      module: era5
  default: cutout-2013-era5
```

### current output for atlite block
(the commented-out part we have in `config.default.yaml` is removed for clarity)
```python
atlite:
  nprocesses: 4
  cutouts:
    cutout-2013-era5:
      module: era5
      dx: 0.3 # cutout resolution
      dy: 0.3 # cutout resolution
    cutout-2023-era5:
      module: era5
  default: cutout-2013-era5
```

PR #1852, opened 2026-06-01
Advance run scenarios

ekatef:
Relates to #1850 and aims to fix a duplication issue in the first place.

The PR is advancing `update()` function which aims to update the base config by setting values according to the diff config. The revision adds functionality which is replacing a block under a key instead adding one more block under the same key.

Additionally, some minor refactoring changes are introduced, such as adding a docstring and naming revision. 

## Checklist


ekatef:
As discussed with @davide-f , the implementation must allow to modify only a few entries in the config. E.g. it should be possible to specify in the config diff only a few parameters for `electricity` block without the need to copy-paste the whole block.

Also, we need to find a way to avoid over-writing `config.yaml`.

ekatef:
Sharing feedback from @tacwebservices who has been working with `run_all_scenarios`. There has been the need to introduce some additionally changes into Snakemake as implemented by [this commit](https://github.com/open-energy-transition/pypsa-zambia/pull/243/changes/a069a44638ba509bbd22731b2ebd5644a8aeabd5)\

@tacwebservices I'm adding an explanation on that you have shared, feel free to correct me if needed. 

For `--unlock`, the reason to use it was to avoid conflicts between different Snakemake processes. When outer process `run_all_scenarios` starts an inner process running a single scenario, the inner one sees the lock already there and stops, thinking someone else is using the folder. So only the first scenario would finish and the rest would not start

ekatef:
For `--nolock`, it appears behaviour can differ depending on an operation system. I have experienced some weird issues for `build_powerplants` when having `--nolock` enabled.

The error looked as follows: `ERROR:_helpers:An error happened in module '~/miniforge3/envs/pypsa-earth-aug/lib/python3.11/tokenize.py', function '_tokenize': ('EOF in multi-line statement', (2, 0))`

scripts/build_test_configs.py
```diff
@@ -20,13 +20,29 @@
...
+                    base_dictionary[key_to_apply], value_to_apply
+                )
+            else:
+                base_dictionary[key_to_apply] = copy.deepcopy(value_to_apply)
         else:
-            d[k] = v
-    return d
+            base_dictionary[key_to_apply] = copy.deepcopy(value_to_apply)
+
+    return base_dictionary
```
davide-f, line 45: I've investigated this issue aiming to understand the needs. When looking at the pinned issue #1850 , I perceive the key issue is the inability to "delete" entries from the original dictionary.
We could explicitly capture that feature by enablind deleting options when the diff value is empty, such as "cutout-2013-era5: {}".
May this solve the issue? I'm attaching a code below that does that, though it requires polishing and local testing with run_all_scenarios. I've been testing with the notebook [test_update.ipynb](https://github.com/user-attachments/files/30900502/test_update.ipynb) attached to this message.
Many thanks!

Moreover, please remind the release_note

```suggestion
def update(base_dictionary, diff_dictionary):
    """
    A function to recursively update items in a dictionary which
    intended usage is unpdating the base config according to the diff
    config
    """
    for key_to_apply, value_to_apply in diff_dictionary.items():
        if isinstance(value_to_apply, collections.abc.Mapping) and not set(value_to_apply):
            # if the diff value is an empty dict, delete the key from the base dict
            del base_dictionary[key_to_apply]
        elif isinstance(value_to_apply, collections.abc.Mapping) and isinstance(
            base_dictionary.get(key_to_apply), collections.abc.Mapping
        ):
            # if the diff value is a non-empty dict and the base value is also a dict, update the base dict with the diff dict
            base_dictionary[key_to_apply] = update(
                base_dictionary[key_to_apply], value_to_apply
            )
        else:
            # if the diff value is not a dict, update the base dict with the diff value
            base_dictionary[key_to_apply] = copy.deepcopy(value_to_apply)

    return base_dictionary
```

PR #1856, opened 2026-06-06
fix: raise clear error when emission_extractor returns no data

### LINKED ISSUE #1289: Raise issue when emission_extractor fails

- when emission_extractor (prepare_network) fails -> raise an issue with clear message
-

wingerter:

`emission_extractor()` in `scripts/prepare_network.py` only logged a *warning*
when some country codes were missing from the global emission file, and then
returned the (possibly empty) result. When **no** country matched, the caller
computed `co2limit = emission_extractor(...).sum()`, which evaluates to `0` and
silently sets a wrong "automatic" CO2 limit instead of failing.

This PR raises a `ValueError` with a clear message (file, base year, requested
countries) when the extraction yields an empty result, so the failure surfaces
explicitly rather than producing a misleading limit. The existing warning for
*partially* missing countries is kept unchanged.

## Verification

The repository's tests are Snakemake/config integration tests, and this
function reads the global emission Excel file inside the `prepare_network`
rule, so it isn't covered by a standalone unit test. The change is a guard on
an already-computed value: an empty `emission_by_country` now raises instead of
flowing into `.sum() == 0`. The normal (non-empty) path is unchanged.

## Checklist

- [ ] I tested my contribution locally and it seems to work fine. _(Not run against the full Snakemake setup; see Verification above — the change only adds a guard on an empty result and leaves the normal path untouched.)_

### DIFF
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -38,6 +38,8 @@ This part of documentation collects descriptive release notes to capture the mai
 
 **Minor Changes and bug-fixing**
 
+* Add error when emission_extractor fails [PR #1856](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1856)
+
 * Generalize the hard-coded legacy name `africa_shape` [PR #1848](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1848)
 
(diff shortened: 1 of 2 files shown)


davide-f (approved):
Thanks @wingerter ! I've added the release note. PR ready to go!

PR #1859, opened 2026-06-08
Integrate `planning horizon` to electricity network

### LINKED ISSUE #1869: Integrate `planning_horizon` to electricity network

It would be good to add `planning_horizon` wildcard into electricity network.

yerbol-akhmetov:
# Closes #1869 

Good day. This PR integrates `planning_horizon` into electricity network, thus enabling use of horizon-specific costs. This feature will be quite useful for myopic implementation for electricity-network. In addition, it will benefit sector model by facilitating use of proper costs for electricity part of the model. 

# Implementation notes
`assign_costs` rule was added to write horizon-specific costs to electricity network components. The script updates capital costs, marginal costs, efficiencies, and lifetimes based on horizon-specific cost file for extendable components (i.e. new components from provided horizon). For existing units, only marginal costs are updated according to horizon-specific costs. 

## Checklist


scripts/assign_costs.py
```diff
@@ -0,0 +1,688 @@
...
+    """
+    if lines_or_links.empty:
+        return
+
+    if lines_or_links.loc[lines_or_links.carrier == "DC"].empty:
+        return
+
+    dc_b = lines_or_links.carrier == "DC"
+    if simple_hvdc_costs:
+        cap = (
```
ekatef: Previously, we had `costs` variable here. Thant may be too generic but was slightly more readable. Currently, `cap` can be misleading, especially in a quick scan (e.g. read as limit on something). May it be a good idea to combine advantages and use something like `capital_costs` or alike?

Snakefile
```diff
@@ -905,6 +907,33 @@ rule add_extra_components:
...
 
 
+rule assign_costs:
+    params:
+        electricity=config["electricity"],
+        length_factor=config["lines"]["length_factor"],
+        hydro_capital_cost=config["renewable"]["hydro"].get("hydro_capital_cost", False),
+    input:
+        network="networks/" + RDIR + "elec_s{simpl}_{clusters}_ec.nc",
+        tech_costs="resources/" + RDIR + "costs_{planning_horizons}_elec.csv",
```
ekatef: Here `elec` is taken from the earlier `process_cost_data` rule via `{scope}` wildcard. However, this connection is not quite evident from the current Snakemake, and some exploration is needed to undersand this connection. Can we add `scope` to [wildcard_constraints](https://github.com/pypsa-meets-earth/pypsa-earth/blob/5e936fa713fc26058e2fa2479b9969d981030ada/Snakefile#L78-L86), something like:
```
scope="elec|sec",
```

Snakefile
```diff
@@ -905,6 +907,33 @@ rule add_extra_components:
...
+    output:
+        network="networks/"
+        + RDIR
+        + "elec_s{simpl}_{clusters}_ec_{planning_horizons}.nc",
```
ekatef, line 920: As we have discussed, `assign_costs` rule overwrites costs calculations made in `add_electricity` to by actual costs for a planning horizon. That keeps computational and storage costs lower for myopic runs avoiding duplication of `elec.nc` which is a strong advantage, especially for bigger power systems.

One potential pitfall is `simplify_network` outputting per-bus connection costs https://github.com/pypsa-meets-earth/pypsa-earth/blob/5e936fa713fc26058e2fa2479b9969d981030ada/Snakefile#L750

In the current PR implementation these costs will remain not updated which can lead to inconsistencies. Do you have any ideas how to fix it? I guess the most straightforward way would be to move this output from `simplify_network` to `assign_costs` but I can be missing something. What is your view on the matter?

scripts/assign_costs.py
```diff
@@ -0,0 +1,688 @@
...
+
+The electricity build pipeline is split into two phases:
+
+1. **Structure** (built once per scenario, shared across all horizons):
+   ``add_electricity`` → ``simplify_network`` → ``cluster_network`` →
+   ``add_extra_components`` → ``elec_s{simpl}_{clusters}_ec.nc``
+
+2. **Costs** (assigned once per planning horizon, cheap):
+   ``assign_costs`` → ``elec_s{simpl}_{clusters}_ec_{planning_horizons}.nc``
+
```
ekatef: An excellent explanation for the logic of the myopic implementation. For the header docstring however I'd expect explanation more focused on the script itself. Do you think it could be possible to re-phrase with the goal to highlight more a place of  `assign_costs` itself?

Snakefile
```diff
@@ -1034,29 +1075,29 @@ if config["monte_carlo"]["options"].get("add_to_snakefile", False) == True:
...
             network="networks/"
             + RDIR
-            + "elec_s{simpl}_{clusters}_ec_l{ll}_{opts}_{unc}.nc",
+            + "elec_s{simpl}_{clusters}_ec_l{ll}_{opts}_{planning_horizons}_{unc}.nc",
             agg_p_nom_minmax=config["electricity"]["agg_p_nom_limits"]["file"],  # ensure the CSV with capacity constraints is copied into the shadow directory (needed on Windows, since shadowed scripts can’t access files outside `input`)
         output:
             "results/"
             + RDIR
-            + "networks/elec_s{simpl}_{clusters}_ec_l{ll}_{opts}_{unc}.nc",
+            + "networks/elec_s{simpl}_{clusters}_ec_l{ll}_{opts}_{planning_horizons}_{unc}.nc",
```
ekatef, line 1122: As a backward-compatibility comment, this replacement will break existing post-processing workflows which often relay on specific pattern of a solved network file. Do you see any reasonable ways to mitigate it? One opportunity could be to use `branch` and define the output name differently for overnight and myopic runs. But that will make the code much less clear so not sure it's a really way to go.

Another option is to accept that this functionality will be "slightly" breaking for post-processing and announce it properly when merging. What do you think?

scripts/assign_costs.py
```diff
@@ -0,0 +1,688 @@
...
+    costs = pd.read_csv(snakemake.input.tech_costs, index_col=0)
+
+    update_electricity_costs(
+        n,
+        costs,
+        renewable_carriers=snakemake.params.electricity["renewable_carriers"],
+        length_factor=snakemake.params.length_factor,
+        hydro_capital_cost=snakemake.params.hydro_capital_cost,
+    )
```
ekatef, line 771: I wonder what if we would bypass costs update if `overnight` run is requested? That would keep costs corresponding to costs["year"] keeping the approach fully consistent with the current conventions.

ekatef (commented):
Thank you @yerbol-akhmetov! A well-designed and elegantly implemented contribution. Have added a few comments.

To understand the full picture better, #1887 is planned as the next step with the aim to harmonise myopic implementation between the power-only and sector-coupled implementations. Do I get it right?

Additionally, I think we still need a documentation section. Do you mind me opening a PR to cover that?

yerbol-akhmetov:
> Do you mind me opening a PR to cover that?

Thanks for the comment, @ekatef. Yes, having documentation is important. we can have a separate PR for that. #1887 is a continuation of this PR. This current PR is standalone PR.

scripts/assign_costs.py
yerbol-akhmetov: Thank you, @ekatef. I will rephrase it.

scripts/assign_costs.py
yerbol-akhmetov: Thank you, @ekatef, it is a quite a valid comment. I will address it.

scripts/assign_costs.py
yerbol-akhmetov: Hi, @ekatef, thanks for the comment. Not sure if we need such difference. It would be good to have the workflow consistent across optimization methods. It would be then easier to maintain. If the user wants to run `overnight` simulation for two horizons (e.g. 2030 and 2040), using only `cost.year` will lead to using single cost for both horizons. If we use rewriting mechanism, then user can easily simulate several horizons for overnight simulation. Also, if user simulates only 1 horizon, there is no problem, because if user have `cost.year: 2030` and planning `horizon: 2030`, it will just overwrites itself. So it does not contradicts the current conventions. What do you think? Do you have maybe more detailed edge case to discuss?

Snakefile
yerbol-akhmetov: Thanks, I will add `scope` to `wildcard_constraints`, and make it clear in `assign_costs` rule.

Snakefile
yerbol-akhmetov: Thank you for the comment, @ekatef. That is a quite valid point. I have implemented horizon-specific displacement connection costs to `assign_costs`. Otherwise, it is just lost and overwritten. Now, in `simplify_network` the displacement length is calculated, which is used in `assign_cost` to get associated costs.

Snakefile
yerbol-akhmetov: Thanks for the comment, @ekatef. I agree that this changes the solved-network filename and can break custom post-processing that assumes the old pattern. In PyPSA-Earth repository rules (`make_summary`, `plots`, `solve_all_networks`) already expand over `{planning_horizons}`. I would prefer one consistent naming scheme over **overnight/myopic** branching. I have locally tested `make_all_summaries` and `plot_all_summaries`, and they worked fine. Can you please provide the details or example on possible post-workflow issue. We can add details on documentation.

Snakefile
ekatef: Perfect, thank you! Looks much more clear now.

scripts/simplify_network.py
```diff
@@ -210,113 +206,173 @@ def simplify_network_to_base_voltage(
...
-                * lines_length_factor
-                * (
-                    unterwater_fractions
-                    * costs.at[tech + "-connection-submarine", "capital_cost"]
-                    + (1.0 - unterwater_fractions)
-                    * costs.at[tech + "-connection-underground", "capital_cost"]
-                )
-            )
+            # length only — underwater share uses generator underwater_fraction later
+            connection_lengths_per_link[tech] = dc_lengths * lines_length_factor
```
ekatef, line 251: A value of the underwater fraction has been previously accounted in the costs. For the current transition to the displacement length, not sure I get how a contribution of the `underwate_fraction` is accounted for, while I see it's value is still used in assign_costs. My understanding is `underwater_fraction` is transferred from n.links, but topology of the links is changing due to simplification. I wonder if it could make sense to transfer also some equivalent change in the `underwater_fraction`, along with the change in length?

scripts/assign_costs.py
```diff
@@ -0,0 +1,923 @@
...
+                    length_factor=length_factor,
+                )
+
+                displacement_connection_cost = 0.0
+                # displacement_length already includes length_factor from simplify_network
+                if "displacement_length" in n.generators.columns:
+                    displacement_length = n.generators.loc[
+                        sub, "displacement_length"
+                    ].fillna(0.0)
+                    if displacement_length.gt(0).any():
```
ekatef, line 402: Not sure I get the meaning of `.gt(0)`. Could you please clarify a bit? That is an advanced bit of the code, and it could be good to make sure that it is as readable as possible.

scripts/assign_costs.py
```diff
@@ -0,0 +1,923 @@
...
+            sub = n.generators.index[is_ext]
+            has_geometry = (
+                "average_distance" in n.generators.columns
+                and "underwater_fraction" in n.generators.columns
+                and n.generators.loc[sub, "average_distance"].notna().all()
+            )
+
+            if supcarrier == "offwind" and has_geometry:
+                farm_connection_cost = _offwind_connection_cost(
+                    carrier,
```
ekatef, line 389: I wonder if it could make sense to wrap-up calculations for the offwind into a dedicated function to make `update_generator_costs` lighter?

Snakefile
ekatef: Thanks for the revision! I like the idea and its' implementation. The only point to double-check is how the underwater fraction values are being treated. Have added this comment also to simplify_network directly.

Snakefile
ekatef: Agree that having a consistent pattern has some advantages. For backward compatibility issues, I have meant the need for those adjustment which you have added to the post-processing rules. For an experienced modeller, they may be evident but a bit annoying if they would appear in the middle of work on some modelling pipeline. I expect though that the most widely used pattern is copy-pasting a name of the network though it should be probably fine to re-design the pattern of the output name.

scripts/assign_costs.py
ekatef: Agree that consistency makes things easier and easier to maintain in the long run. Just to make sure I understand you properly, let's recap the idea:
1. `costs.year` is overwritten by `scenario.planning_horizons`
2. `scenario.planning_horizons` can be a list also for `foresight: overnight` in which case it triggers rewriting mechanism.

Could you please confirm it is right understanding? If that is the case, then I think `costs.year` shouldn't be needed anymore, and we can remove it from the config enabling a safety backward-compatibility mechanism you have implemented.

ekatef (commented):
Hello @yerbol-akhmetov, I think we are almost there. Thanks a lot for all the revisions! 
The only two remaining significant comments relates to a possible removal of `costs.year` and treatment of `underwater_fraction`. Could you please have a look on the comments?

As mentioned, I see also some risks from having a column with custom attribute but it could be an acceptable risk as workarounds are likely to be more complex and prone to errors. So, I'd stick to completing the current implementation while in future we can keep eye on possible changes in the way PyPSA is treating custom attributes.

Issue #1583, opened 2025-09-26
Discount rate mismatch


danielelerede-oet:
## Checklist

- [x] I am using the current [`main`](https://github.com/pypsa-meets-earth/pypsa-earth/tree/main) branch or the latest [release](https://github.com/pypsa-meets-earth/pypsa-earth/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-earth` environment](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

At the moment, the discount rate used for the calculation of the annuity is the one assigned via `fill_values` as `discount rate`, while `config.default.yaml` also includes a `discountrate`, which is used as a wildcard in sector model results (I suppose only to distinguish runs with different discount rates). 

However, I see three critical points in the current approach:
1) this approach doesn't allow to provide custom discount rate values for specific technologies
2) it may generate confusion between the `discountrate` value and the value in `fill_values` as they're also different in `config.default.yaml`
3) is the wildcard necessary?

A proposal could be to remove the wildcard or, alternatively, to uniform all the calls to a discount rate to `discountrate`, so that the value under `costs` could be used as a generic discount rate and the possibility to customize it is preserved under `fill_values`.

cc @davide-f @yerbol-akhmetov @ekatef @hazemakhalek @energyLS @GbotemiB

ljansen-iee:
Hi, I agree that the current implementation confuses and also confused me a while ago. Apparently, the wildcard is only used in the customisation step override_respot, and was probably used in past experiments to attach different renewable costs (based on different discountrates). Maybe @Eddy-JV could comment on that. 

I'm afraid I don't understand your suggestion. Are you proposing to actually use the discountrate as fill_value[discount rate] in the prepare_cost call and preserve the option to define technology-specific discount rates via custom cost.csv data?

PR #1862, opened 2026-06-09
fix and rename discount rate wildcard

ljansen-iee:
# Closes # (if applicable)

See #1583. The issue could be closed after the merge of this PR

Summary: Fix critical silent inconsistency in discount rate handling: the `discount_rate` config parameter now functions as a true Snakemake wildcard, generating separate cost files and network pipelines for each value in the list. This enables proper sensitivity analysis and eliminates the bug where multiple discount rates in the config would silently use only the first value.

Breaking Changes:

- **Renamed `discountrate` → `discount_rate`**: Standardized parameter naming across the codebase to use snake_case for consistency (affects `config.default.yaml`, all scenario configs, Snakefile wildcards, and all Python scripts with mock_snakemake references).
- **Removed redundant `costs.fill_values.discount_rate` entry**: This fallback value was causing confusion and silent bugs (if one wanted to use the wildcard, see #1583). The discount rate is now sourced exclusively from the wildcard-parametrized `costs.discount_rate` list.

Implementation:

- **Parametrized `discount_rate` as a Snakemake wildcard**: Added `{discount_rate}` suffix to all cost file paths (10+ rules across electricity and sector workflows) to trigger proper wildcard expansion.
- **Value injection in `process_cost_data.py`** via snakemake.params.costs["fill_values"]["discount rate"] = float(snakemake.wildcards.discount_rate → uses the actual wildcard value as fillna value per rule execution.

Future Enhancements:

- Consider making the `discount_rate` wildcard optional (e.g., to allow single-value runs without requiring explicit list syntax and without extra string in filenames). However, i dont see an obvious way to do this at the moment. The wildcards opts and sopts contain optional configs, but process_cost_data is executed before prepare_network. 
- Automatic discout_rate/WACC per country → I will propose a follow-up PR for this.

@ekatef @doneachh @davide-f  : Do you have capacities to review the PR?

## Checklist


doneachh (approved):
@ljansen-iee Looks good to me! 

@ekatef In my opinion, this can be merged, as soon as the checklist entries are checked and CI is running. I have no write access, so the final approval has to be done by you :)

ljansen-iee:
I ran all test configs and they worked locally.

ljansen-iee:
@ekatef @davide-f @doneachh @energyLS
Hi, do you have capacity to review or comment on the proposed fix/feature?

My main concern is that, with this PR, the discount_rate wildcard already affects the electricity model. As a consequence, it also propagates into all subsequent file names, which makes the Snakemake file names rather long and potentially inconvenient.

One simple option would be to shorten the wildcard name from discount_rate to dr.

Alternatively, I could try to apply the wildcard later in the workflow, for example only for the sector-coupled model, and optionally as part of the sopts wildcard. This would require updating all capital_cost values for all technologies based on the new WACC/discount rate extracted from the wildcard string - if it is defined.

I would appreciate your view on which direction is preferable before I make further changes.

scripts/process_cost_data.py
```diff
@@ -552,6 +552,10 @@ def annuity_factor(v):
     n = pypsa.Network(snakemake.input.network)
     Nyears = n.snapshot_weightings.generators.sum() / 8760.0
 
+    snakemake.params.costs["fill_values"]["discount rate"] = float(
```
doneachh, line 555: I would rather prefer to just use directly the discount_rate and modify it in the whole script, instead of creating a virtual fill_values config setting.

doneachh (commented):
@ljansen-iee My opinion: its good, that its also affects the electricity model and I see no problem, that the file names in the snakemake file are getting longer. It doesn't affect the user, just the code. That being sad, I also prefer the option to leave the wildcard name how it is right now (discount_rate) as the code is easier understandable.

In my opinion, I would modify the script process_cost_data and directly use the config parameter discount_rate there instead of creating a virtual one with fill_values again. 

More over, it would be nice (as discussed) to implement a new config parameter "automatic_wacc", which automatically sets the wacc for the chosen region based on damodaran or your wacc calculator.

@ekatef @davide-f feel free to point out how you are feeling about this approach

ekatef:
Hello,
@ljansen-iee thank you very much for the contribution and sharing your experience, and thanks a lot @doneachh for the outstanding support.
 
@ljansen-iee apologies for being slow to answer. 

I think adding a wildcard for the discount rate is perfectly in line with the current logic of the sector-coupled part which is using wildcards quite intensively. However, we have identified the need to facilitate dealing with the model as one of the most pressing development priorities which implies the need to change the current design. That includes re-structuring of the configuration parameters and also reducing the number of the wildcards to simplify things.

Another point is alignment with PyPSA-Eur. We are about to start the work for harmonise both models. From the perspective of the wildcards, PyPSA-Eur is more minimalistic and is using only four wildcards, while we still have nine of them.

Finally, using a discount rate as a wildcard limits options to use different discount rates for different technologies as @danielelerede-oet has mentioned in #1583. 

So, I'm not sure enabling `discountrate` as a proper wildcard is consistent with this logic.

ekatef:
> @ljansen-iee My opinion: its good, that its also affects the electricity model and I see no problem, that the file names in the snakemake file are getting longer. It doesn't affect the user, just the code. That being sad, I also prefer the option to leave the wildcard name how it is right now (discount_rate) as the code is easier understandable.
> 
> In my opinion, I would modify the script process_cost_data and directly use the config parameter discount_rate there instead of creating a virtual one with fill_values again.
> 
> More over, it would be nice (as discussed) to implement a new config parameter "automatic_wacc", which automatically sets the wacc for the chosen region based on damodaran or your wacc calculator.
> 
> @ekatef @davide-f feel free to point out how you are feeling about this approach

Implementing an automated way to fetch regional-specific economical parameters is an amazing idea. To my understanding, an ideal approach would be combining a default country-specific value of the discount rate with an option to apply custom discount rates, also for particular technologies.

ljansen-iee:
> I think adding a wildcard for the discount rate is perfectly in line with the current logic of the sector-coupled part which is using wildcards quite intensively. However, we have identified the need to facilitate dealing with the model as one of the most pressing development priorities which implies the need to change the current design. That includes re-structuring of the configuration parameters and also reducing the number of the wildcards to simplify things.
> 
> Another point is alignment with PyPSA-Eur. We are about to start the work for harmonise both models. From the perspective of the wildcards, PyPSA-Eur is more minimalistic and is using only four wildcards, while we still have nine of them.


Hi @ekatef , thank you for your view!

Just to make sure I understand correctly: do you mean that we should implement a build_scenarios rule and the corresponding config-overwrite pattern in PyPSA-Eur / PyPSA-DE, following the latest, more minimalistic scenario approach?

So essentially, scenarios would be defined through layered YAML configs and scenario-specific overrides, and the workflow would automatically iterate over them without relying on an extensive set of scenario wildcards. 

Is that what you have in mind?

ljansen-iee:
An alternative would be to combine all sector wildcards into sopts and use prefixes and pattern matching to find the sections that affect the sensitivity parameter.

ekatef:
> > I think adding a wildcard for the discount rate is perfectly in line with the current logic of the sector-coupled part which is using wildcards quite intensively. However, we have identified the need to facilitate dealing with the model as one of the most pressing development priorities which implies the need to change the current design. That includes re-structuring of the configuration parameters and also reducing the number of the wildcards to simplify things.
> > Another point is alignment with PyPSA-Eur. We are about to start the work for harmonise both models. From the perspective of the wildcards, PyPSA-Eur is more minimalistic and is using only four wildcards, while we still have nine of them.
> 
> Hi @ekatef , thank you for your view!
> 
> Just to make sure I understand correctly: do you mean that we should implement a build_scenarios rule and the corresponding config-overwrite pattern in PyPSA-Eur / PyPSA-DE, following the latest, more minimalistic scenario approach?
> 
> So essentially, scenarios would be defined through layered YAML configs and scenario-specific overrides, and the workflow would automatically iterate over them without relying on an extensive set of scenario wildcards.
> 
> Is that what you have in mind?

Hey @ljansen-iee, we are touching here quite deep questions of the long-term development strategy, and it would be great to have your input on them given your experience. 

I see two different points we need to tackle:
1. Design must be improved for the model itself to make it lighter.
2. Scenario management must be enhanced to facilitate application of the model to different use cases.

For the first point, we have seen a lot of user feedback on the current modelling workflows being too heavy which seems to be currently the biggest challenge for adoption. A long-term solution is making a workflow more modular. In short-term, we aim to re-structure the configuration file making it more slim and more structured, and reducing a number of wildcards is a part of this work (tackled in #1962, #1957, #1879 and #1863).

For scenario management, we currently have [run_all_scenarios](https://github.com/pypsa-meets-earth/pypsa-earth/blob/d7cf8ccf507689566ad2418cae29dee5258c9cb0/Snakefile#L2409) rule which allows running a set of scenarios in one go. Additionally, bigger specific modelling projects normally have some dedicated scenario management tools similar to [build_scenarios](https://github.com/PyPSA/pypsa-de/blob/main/scripts/pypsa-de/build_scenarios.py) PyPSA-DE or [create_scenarios](https://github.com/open-energy-transition/open-tyndp/blob/master/config/create_scenarios.py) in Open-TYNDP. For this kind of work we don't have a dedicated strategy right now, though I completely agree that it would be good to find a way to elaborate some solutions to accommodate the existing best practices.

ekatef:
> An alternative would be to combine all sector wildcards into sopts and use prefixes and pattern matching to find the sections that affect the sensitivity parameter.

Sounds a very interesting idea. Not sure however if it'll be straightforward to implement for parameters which must be modified on earlier stages of the workflow. 

@yerbol-akhmetov I'm tagging you here since you have authored restructuring of the configuration file, and Lukas' idea can resonate with your experience on enhancement of the myopic implementation.

yerbol-akhmetov:
> > An alternative would be to combine all sector wildcards into sopts and use prefixes and pattern matching to find the sections that affect the sensitivity parameter.
> 
> Sounds a very interesting idea. Not sure however if it'll be straightforward to implement for parameters which must be modified on earlier stages of the workflow.
> 
> @yerbol-akhmetov I'm tagging you here since you have authored restructuring of the configuration file, and Lukas' idea can resonate with your experience on enhancement of the myopic implementation.

Hi, @ekatef, @ljansen-iee. I think setting those wildcards into sector opts might be good idea. I see no significant obstacles, because those sector wildcards used to be added close to the end of the model building.

Issue #1820, opened 2026-05-13
Overestimation of transmission capacity expansion (prepare_network.py)


carlosfv92:
## Checklist

- [ ] I am using the current [`main`](https://github.com/pypsa-meets-earth/pypsa-earth/tree/main) branch or the latest [release](https://github.com/pypsa-meets-earth/pypsa-earth/releases). Please indicate.
- [ ] I am running on an up-to-date [`pypsa-earth` environment](https://github.com/pypsa-meets-earth/pypsa-earth/blob/main/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

(I'm using an old branch of the model but I checked that this particular section of the code hasn't changed compared to the current version)

## Describe the Bug

Hey! I've been exploring some outputs linked to transmission lines for a case study I'm running and I found that the minimal capacity expansion in the model is always recalculated in the prepare_network script. 
This is fine when exploring future scenarios as one would expect for the lines to eventually expand, however, I found that this also happens even in cases where the system wouldn't really need to expand transmission lines (running a historical year with the existing capacity of the system). 
Following the code, the minimal expansion recalculation I found was made in this specific section of the code: 

...
def set_transmission_limit(n, ll_type, factor, costs, Nyears=1):
    links_dc_b = n.links.carrier == "DC" if not n.links.empty else pd.Series()

    _lines_s_nom = (
        np.sqrt(3)
        * n.lines.type.map(n.line_types.i_nom)
        * n.lines.num_parallel
        * n.lines.bus0.map(n.buses.v_nom)
    )
    lines_s_nom = n.lines.s_nom.where(n.lines.type == "", _lines_s_nom)
...

From what I understood, the model is actively ignoring stored s_nom (calculated based on the existing lines voltages and the characteristics of the closest line type defined in the config file). If the lines "types" don't match an existing line type, it recomputes the minimum capacity expansion based on v_nom from buses instead of v_nom from lines. 
I guess this makes sense as we want to match the line voltage to the substation, but in doing so, we force the line to expand even in cases where it is not necessary and where a previously calculated transmission capacity was available (even if they are not assigned a particular line type).

My fix to avoid this over expansion was to change the last line in the section showed to 

lines_s_nom = n.lines.s_nom.where(n.lines.s_nom > 0, _lines_s_nom)

[...]

PR #1933, opened 2026-07-09
Support country-specific mappings and custom transmission line types

danielelerede-oet:
## Description

This PR adds support for country-specific base voltages and transmission line-type mappings while preserving the existing PyPSA-Earth mappings as the default.

The use of country-specific line types is configurable through `lines.use_country_specific_types`.

When this option is enabled, country-specific mappings are used only if mappings are available for every configured country. Otherwise, the complete `default` mapping is used. Availability is evaluated separately for AC and DC mappings, so country-specific and default mappings are not mixed within the same category.

The country associated with a line is determined from `bus0`, consistently across base-network construction, network simplification, and augmented connections. The closest available voltage is selected from the chosen mapping.

## Configuration

Line-type mappings remain in the configuration under:

```yaml
lines:
  use_country_specific_types: false

  ac_types:
    default:
      # voltage-to-line-type mappings
    AU:
      # country-specific mappings
    US:
      # country-specific mappings

  dc_types:
    default:
      # voltage-to-line-type mappings
```

The currently existing mapping is stored under `default`. Optional country-specific mappings can be added using country codes.

* If `lines.use_country_specific_types` is `false`, the complete default mapping is used.
* If it is `true` and mappings exist for every configured country, the country-specific mappings are used.
* If at least one configured country is missing from a category, the complete default mapping is used for that category.
* AC and DC mapping availability is evaluated independently.
* The closest available voltage in the selected mapping is used.
* The country of each line is determined from the country of `bus0`.

For example, if every configured country has an AC mapping but the DC mappings are incomplete, country-specific mappings are used for AC lines while the complete default mapping is used for DC lines.

## Checklist


data/line_type_mapping_ac.csv
```diff
@@ -0,0 +1,9 @@
+voltage,default,AU
+132,243-AL1/39-ST1A 110.0,243-AL1/39-ST1A 110.0
+220,Al/St 240/40 2-bundle 220.0,
+275,Al/St 240/40 3-bundle 300.0,Al/St 240/40 3-bundle 300.0
+300,Al/St 240/40 3-bundle 300.0,
+330,Al/St 240/40 4-bundle 380.0,Al/St 240/40 4-bundle 380.0
+380,Al/St 240/40 4-bundle 380.0,
+500,Al/St 240/40 4-bundle 380.0,Al/St 560/50 4-bundle 750.0
+750,Al/St 560/50 4-bundle 750.0,
```
davide-f: Could we keep the assignment of line types in the config file and add the csv for the characteristics of custom line types that are not defined in pypsa?

The advantage is that for each country we can define only the relevant line types and avoid nan values. The higher the number of countries, the more complex the handling becomes with csv format

davide-f (commented):
Many thanks @danielelerede-oet !
Really like the proposal that enables custom line-types by country that is a long wanted feature! Adding a comment

data/line_type_mapping_ac.csv
danielelerede-oet: The point is that this should address a slightly different case. In this PR, the csv file doesn't define the electrical characteristics of custom line types. It is supposed to only map each country and voltage level to a line type which is already existing in PyPSA (where applicable, as in the case of Australia).

The idea of having custom line types assigned to countries as the US where the mapping with already existing PyPSA line types is not accurated would be complementary to this country-specific assignment mechanism rather than a replacement for it.

Keeping the assignment in the config would also preserve the current limitation, since the purpose of this PR is precisely to support different mappings by country without adding country-specific configuration sections. Let me know what you think :)

danielelerede-oet:
Hi @davide-f @ekatef I have extended the PR beyond the mapping you checked in initial review that only accounted for countries with line types already defined in PyPSA (i.e. Australia).

The PR now also covers the US case, where representative types are not available in the standard library. I added a central `data/custom_line_types.csv` registry containing US electrical parameters derived from the ACTIVSg82k synthetic network, together with a non-workflow script that makes the derivation reproducible.

`base_network.py` now loads these additional definitions into `n.line_types` before assigning types and calculating line capacities. I also updated `simplify_network.py` and `augmented_line_connections.py` so country-specific types are retained when voltage layers are consolidated or new HVAC connections are introduced.

The existing default mapping remains unchanged for countries without dedicated mappings.

config.default.yaml
```diff
@@ -243,15 +243,6 @@ electricity:
...
 lines:
-  ac_types:
-    132.: "243-AL1/39-ST1A 20.0"
-    220.: "Al/St 240/40 2-bundle 220.0"
-    300.: "Al/St 240/40 3-bundle 300.0"
-    380.: "Al/St 240/40 4-bundle 380.0"
-    500.: "Al/St 240/40 4-bundle 380.0"
-    750.: "Al/St 560/50 4-bundle 750.0"
-  dc_types:
-    500.: "HVDC XLPE 1000"
```
davide-f, line 269: I'd be in favor of keeping the voltage to line-type assignment in the config as it enables flexibility and clarity.
In the CSV it is hard to follow the lines and more inflexible the larger the set of countries.
We could use something like:
ac_types:
   default:
     ...
   US:
     ....
   AU:
     ...

scripts/base_network.py
```diff
@@ -262,86 +262,138 @@ def _load_transformers_from_osm(fp_osm_transformers, buses):
...
+
+    existing_types = line_types.index.intersection(n.line_types.index)
+    if not existing_types.empty:
+        raise ValueError(
+            "Custom line types already exist in the PyPSA register: "
+            f"{existing_types.tolist()}"
+        )
+
+    line_types = line_types[required_columns]
+    n.line_types = pd.concat([n.line_types, line_types])
```
davide-f: I'd advise testing the use of n.add("LineType", **line_types), had you resources, otherwise we keep for the future

scripts/simplify_network.py
```diff
@@ -108,23 +111,81 @@
 logger = create_logger(__name__)
 
 
-def simplify_network_to_base_voltage(n, linetype, base_voltage):
+def _load_linetypes_from_csv(path):
```
davide-f: Depending on the other comments, we may revisit this.
Ideally we should find a way to reduce the length of the code; prior comments are structural and affect code here.
I'd expect some functions to be possibly reused across scripts

davide-f (commented):
Amazing Daniele! I'm adding few comments; this is a great feature!

As a comment: have you verified that linetypes properly propagate across scripts especially after cluster_network, just to avoid surprises?
When linetype is defined, pypsa internally may recalculate values by using linetype info and it is advisable to ensure the information is preserved.

config.default.yaml
danielelerede-oet: I also like it more, I was only thinking about when something like this is applied to many countries. But we can work with the config only for the time being

scripts/base_network.py
danielelerede-oet: Done, I replaced the direct concatenation with `n.add("LineType", line_types.index, **line_types)`.

scripts/simplify_network.py
danielelerede-oet: Agreed. The country-specific line-type selection is now used in three scripts, so I moved it to _helpers and reused the shared function across them.

scripts/base_network.py
danielelerede-oet: Looking at the failing tests I found out that passing the full index and DataFrame columns therefore it to treat each Series as a scalar.

I changed the implementation to add each line type individually with `network.add()`, so that it would bot h work with the current PyPSA version and remains compatible with PyPSA >1.0

danielelerede-oet:
> Amazing Daniele! I'm adding few comments; this is a great feature!
> 
> As a comment: have you verified that linetypes properly propagate across scripts especially after cluster_network, just to avoid surprises? When linetype is defined, pypsa internally may recalculate values by using linetype info and it is advisable to ensure the information is preserved.

Yes @davide-f this should be ensured but I'm testing the model on the US to double-check

danielelerede-oet:
> Amazing Daniele! I'm adding few comments; this is a great feature!
> 
> As a comment: have you verified that linetypes properly propagate across scripts especially after cluster_network, just to avoid surprises? When linetype is defined, pypsa internally may recalculate values by using linetype info and it is advisable to ensure the information is preserved.

@davide-f, this comment identified an important point. I checked the networks generated across the different workflow steps and found that the voltage-specific US line types are correctly assigned in elec.nc, but simplify_network subsequently remaps every line to the type associated with the configured base voltage. With the current 380 kV base voltage, the US mapping has no corresponding entry and therefore falls back to the default Al/St 240/40 4-bundle 380.0 type.

The procedure preserves `s_nom` by recalculating `num_parallel`, and the dependent electrical parameters are populated before solving. However, the original US-specific impedance characteristics are not retained directly, since r, x, and b are recalculated from the fallback 380 kV type.

We could either retain the current behaviour or revise the simplification settings to use more representative regional base voltages and mapped line types: potentially `US_345kV` at 345 kV for the US, and the currently mapped `Al/St 240/40 4-bundle 380.0` type at a 330 kV base voltage for Australia.

A preliminary comparison on the same simplified network confirms that using a 345 kV base voltage with US_345kV:
- preserves `s_nom` exactly
- increases `num_parallel` by 41.7%
- decreases `r` by 28.6%
- increases `x` by 4.4%
- increases `b` by 77.6%. 

Therefore, the regional simplification mapping is not merely a naming change: it materially affects the electrical representation, even though the thermal capacities remain unchanged. I will implement the US-specific mapping and rerun the network to quantify its impact on flows, congestion, dispatch and the objective.

I'm now testing this in https://github.com/pypsa-meets-earth/pypsa-northamerica/pull/43 : `electricity.base_voltage` is now defined as a single mapping in `config.default.yaml`:

```yaml
base_voltage:
  default: 380.
  AU: 330.
  US: 345.
```

In `simplify_network_to_base_voltage()`, the simplification voltage is resolved from the country of each bus, falling back to `default` when no country-specific value is configured. Lines are consequently assigned the corresponding country-specific voltage and line type, while `num_parallel` is recalculated to preserve `s_nom`, as before.

For the current US-only test, all buses and lines are simplified at 345 kV and use `US_345kV` as mentioned above.

One crucial point remains open before using this implementation in multi-country networks: an AC interconnector could connect countries to which different simplification voltages are assigned. The current logic follows the country of `bus0`, consistently with the existing line-type selection, but this could leave a regular `Line` connecting buses with different nominal voltages. 

I have not added a fail-fast test because we should first decide whether interconnected countries should share a common regional base voltage or whether voltage transitions should be represented explicitly (y preference would be to use a common simplification voltage for countries belonging to the same interconnected network).

## Result of test in PyPSA-NorthAmerica ([PR #43](https://github.com/pypsa-meets-earth/pypsa-northamerica/pull/43))

The `US_345kV` line type definition is now preserved through both simplification and spatial clustering, all 219 clustered lines referencing it have a valid associated linetype and the optimization completes successfully with slight changes to the objective function (+0.3%). This confirms that the country-specific voltage and line-type mapping is applied correctly without changing the lines’ thermal capacities.

ekatef:
@danielelerede-oet @davide-f looks an absolutely amazing work! 

@danielelerede-oet that is a very impactful improvement you have implemented to preserve the line type definition.

We have experienced dropping of custom line parameters also in PyPSA-Zambia model, and can confirm that it's essential to get the linetype right. However, the fix we have applied has been an ad hoc solution. Tagging @tacwebservices who has worked a lot on examining regional representation the power grid and its' effects on modelling.

scripts/_helpers.py
```diff
@@ -2296,3 +2296,28 @@ def sanitize_locations(n: pypsa.Network) -> None:
...
+    """
+    Return the closest available line type for a voltage and country.
+    """
+    if "default" not in linetypes:
+        raise ValueError("Missing 'default' line type mapping.")
+
+    mapping = {
+        **linetypes["default"],
+        **linetypes.get(country, {}),
+    }
```
davide-f: I'd revise the implementation: when country-specific options exist, the defaults are not used entirely, otherwise we risk mixing default and non-default linetypes for countries for small voltage differences

davide-f (changes requested):
Thanks @danielelerede-oet ! Added a minor change.
As this PR adds a breaking change in the use of line_types, it is advisable to signal backcompatibility.
Yerbol has integrated an option to track migrations:
https://github.com/pypsa-meets-earth/pypsa-earth/blob/297062401cd67b60624eabc31c482c3296d137ea/scripts/_helpers.py#L117

Please add the config references there to support the migrations. From a functional perspective I have only a comment left.
Many thanks!

danielelerede-oet:
> Thanks @danielelerede-oet ! Added a minor change. As this PR adds a breaking change in the use of line_types, it is advisable to signal backcompatibility. Yerbol has integrated an option to track migrations:
> 
> https://github.com/pypsa-meets-earth/pypsa-earth/blob/297062401cd67b60624eabc31c482c3296d137ea/scripts/_helpers.py#L117
> 
> Please add the config references there to support the migrations. From a functional perspective I have only a comment left. Many thanks!

@davide-f I checked `CONFIG_MIGRATIONS`, but I don’t think this change belongs there because `lines.ac_types` and `lines.dc_types` keep the same paths, while only the internal structure changes.

Since `CONFIG_MIGRATIONS` introduced by Yerbol only handles path renames, I added a dedicated handler in `_helpers.py` to detect legacy top-level voltage mappings, moves them under `default`, preserves country-specific mappings, and emits a `FutureWarning`.

scripts/_helpers.py
danielelerede-oet: Yeah, it's addressed now. It was already in my mind. Now the default and country-specific mappings are no longer merged. Country-specific line types are only used when `lines.use_country_specific_types` (as suggested by @ekatef ) is enabled and mappings are available for every configured country. Otherwise, the complete default mapping is used. This is evaluated separately for AC and DC line types, so default and country-specific line types are never mixed within the same category.

scripts/_helpers.py
```diff
@@ -2379,3 +2415,33 @@ def sanitize_locations(n: pypsa.Network) -> None:
...
+    use_country_specific_types,
+):
+    """Return the closest line type from the selected mapping."""
+    if "default" not in linetypes:
+        raise ValueError("Missing 'default' line type mapping.")
+
+    mapping_name = country if use_country_specific_types else "default"
+
+    if mapping_name not in linetypes:
+        raise ValueError(f"Missing line type mapping for country '{mapping_name}'.")
```
davide-f: The fallback to default shall apply when country is not available. This line may fail pre-commit
```suggestion
    mapping_name = country if use_country_specific_types and (country in linetypes) else "default"
```

scripts/_helpers.py
```diff
@@ -2379,3 +2415,33 @@ def sanitize_locations(n: pypsa.Network) -> None:
...
+
+    mapping = linetypes[mapping_name]
+
+    if not mapping:
+        raise ValueError(
+            f"No line type mapping found for voltage {v_nom} kV "
+            f"in mapping '{mapping_name}'."
+        )
```
davide-f, line 2490:
```suggestion
        raise ValueError(
            f"Empty line type mapping found for line mapping '{mapping_name}'."
        )
```
v_nom is not yet used

scripts/augmented_line_connections.py
```diff
@@ -135,14 +147,57 @@ def haversine(p):
...
+        False,
+    )
+
+    use_country_specific_ac_types = use_country_specific_types and all(
+        country in ac_linetypes for country in countries
+    )
+
+    use_country_specific_dc_types = use_country_specific_types and all(
+        country in dc_linetypes for country in countries
+    )
```
davide-f: The "all" condition is too strong. when flag is on, the country-specific setting is used for those countries that have it set, for all the others the default applies.
Otherwise it is complicated to use for many countries

danielelerede-oet:
Thanks @davide-f I applied your latest comments too

ekatef:
@carlosfv92 this PR may resolve the issues with the transmission capacity you reported in #1820

scripts/simplify_network.py
```diff
@@ -108,23 +111,72 @@
...
     """
 
     logger.info(f"Mapping all network lines onto a single {int(base_voltage)}kV layer")
     n.buses["v_nom"] = base_voltage
-    n.lines["type"] = linetype
+
+    line_countries = n.lines["bus0"].map(n.buses["country"])
+
+    ac_lines = n.lines["carrier"] == "AC"
+    dc_lines = n.lines["carrier"] == "DC"
```
ekatef: I wonder if it could be a good idea to revise naming to make it evident that `ac_lines` and `dc_lines` are not a subset of `n.lines` but a series of booleans?

scripts/non_workflow/build_us_line_types.py
```diff
@@ -0,0 +1,406 @@
+# SPDX-FileCopyrightText: PyPSA-Earth and PyPSA-Eur Authors
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+"""
+Derives representative US transmission line types from the ACTIVSg82k network.
```
ekatef: Could it be possible to add a reference to the source and/or some explanation on `ACTIVSg82k`?

scripts/non_workflow/build_us_line_types.py
```diff
@@ -0,0 +1,406 @@
...
+import pandas as pd
+
+S_BASE_MVA = 100.0
+MILE_TO_KM = 1.609344
+FREQUENCY_HZ = 60.0
+OMEGA = 2 * np.pi * FREQUENCY_HZ
+
+COUNTRY = "US"
+RAW_PATH = Path("data/SyntheticUSA.RAW")
+OUTPUT_PATH = Path("data/custom_line_types.csv")
```
ekatef, line 58: Do you have any hints on what might be a strategy to extract `SyntheticUSA.RAW` data? I recognise that management of the raw data are out of scope for this PR while it's crucial to understand which inputs are needed to reproduce a data building workflow (including externalisation of the grid processing).

scripts/non_workflow/build_us_line_types.py
```diff
@@ -0,0 +1,406 @@
...
+        Representative line types indexed by stable type names.
+    """
+    valid = filter_valid_branches(branches)
+
+    line_types = (
+        valid.groupby("v_nom")
+        .agg(
+            r_per_length=("r_per_length", "median"),
+            x_per_length=("x_per_length", "median"),
+            c_per_length=("c_per_length", "median"),
```
ekatef, line 317: I guess median would work best in case when variations of `{parameter}_per_length` for the same `v_nom` are caused by some precision issues, like measurement errors.  

I wonder if it may be possible when same voltage level corresponds to different linetypes? In this case, using median will lead to mixing parameters for different linetypes, and it would be potentially more accurate to take the most frequently used type of lines.

The best way to deal with that obviously depends on the region, and I trust your judgment on what works best for US. Just thinking about possible ways to generalise.

scripts/simplify_network.py
```diff
@@ -108,23 +111,72 @@
...
+
+    Returns
+    -------
+    tuple
+        Simplified network and transformer bus mapping.
     """
 
     logger.info(f"Mapping all network lines onto a single {int(base_voltage)}kV layer")
     n.buses["v_nom"] = base_voltage
-    n.lines["type"] = linetype
```
ekatef, line 125: Do we have any advantages from having `base_voltage` configurable? Given it's effect on handling custom linetypes, I'm thinking if it could be a good idea to define the base voltage automatically as the highest voltage of the country-specific linetypes if they are enabled?

ekatef (commented):
Hello @danielelerede-oet, a great and much needed feature! Have added a few comments mainly aiming to understand what is the best way to generalise the implementation. I don't have strong opinion on any of them, so please feel free to complete this PR according to what you have discussed with Davide.

I think that it would be perfect to finalise and merge as soon as we can. As the next steps, I see the following points: 
- to integrate the custom line approach into PyPSA-ZM;
- transfer lessons learnt into GridBuilder.
For both of them, it would be perfect to have your opinion.

ekatef:
@danielelerede-oet one more point: we have recognised when experimenting with linetypes in PyPSA-ZM that it is essential to have some standardised testing procedure to check the effect of modifications. One of the tests is checking `type` attribute of `n.lines`

In [this comment](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1933#issuecomment-5145598890) you are also mentioning about having some quantitative screening for the network representation:
> preserves s_nom exactly
> increases num_parallel by 41.7%
> decreases r by 28.6%
> increases x by 4.4%
> increases b by 77.6%.

From you experience, what is the best way to do such a screening? I normally use maps, such as those attached.
image

image

Is it similar to what you are doing? And what is the method you apply to reduce the distribution to a single number (which is actually much more convenient for a quick screening)?

danielelerede-oet:
Hi @ekatef 
Your maps are indeed useful for inspecting the spatial distribution and identifying local outliers. My screening was more limited and intended as a quick global consistency check, especially as I expected substantial changes to happen with respect to the default behaviour, but I can definitely highlight the differences in the US case through graphical tests. 

In my tests, for each attribute I reduced the network to its aggregate value and calculated:

`relative change = 100 \times (sum_after / sum_before − 1)`

I applied this independently to `s_nom `, `num_parallel`, `r`, `x`, and `b`. Then I checked that `s_nom` was preserved by the transformation.

For this particular change, I would additionally compare the `type` attribute of `n.lines`, since line-type assignment is the behaviour being modified.

scripts/non_workflow/build_us_line_types.py
danielelerede-oet: Sure @ekatef , I added a reference and explained how to use it to derive custom line types

ekatef:
> Hi @ekatef Your maps are indeed useful for inspecting the spatial distribution and identifying local outliers. My screening was more limited and intended as a quick global consistency check, especially as I expected substantial changes to happen with respect to the default behaviour, but I can definitely highlight the differences in the US case through graphical tests.
> 
> In my tests, for each attribute I reduced the network to its aggregate value and calculated:
> 
> `relative change = 100 \times (sum_after / sum_before − 1)`
> 
> I applied this independently to `s_nom `, `num_parallel`, `r`, `x`, and `b`. Then I checked that `s_nom` was preserved by the transformation.
> 
> For this particular change, I would additionally compare the `type` attribute of `n.lines`, since line-type assignment is the behaviour being modified.

Hello @danielelerede-oet, great, thank you for the detailed explanation! As the aggregate value, you mean the sum across all the network branches, correct?

For `type` attribute, completely agree. I'd also add `v_nom` to the testing scope for the work which is aiming to enhance representation of the transmission network.

For maps, I have sketched some options in this notebook https://github.com/open-energy-transition/zambia-dev-notebooks/blob/linetypes_exploration/linetypes_exploration.ipynb

As a disclaimer: all the scanning discussion relate to the follow-ups of this PR, whenever in PyPSA-ZM, or as a part of GridBuilder project. I'll open a dedicated issue in GridBuilder to continue a discussion there.

scripts/non_workflow/build_us_line_types.py
danielelerede-oet: @ekatef  as from the updated  docstring, I think we don't have many choices but to download and extract the dataset from the Texas A&M Electric Grid Test Case Repository and then place the file at the input path expected by the script. For longer-term data management, the download page states that the dataset is provided free for commercial and non-commercial use. It asks users to complete the download form and cite the relevant publications. However, I could not find an explicit licence granting redistribution rights. So I wouldn't redistribute the file through the PyPSA-Earth code itself, but let users obtain it through the download form and provide it locally to this script.

scripts/non_workflow/build_us_line_types.py
danielelerede-oet: Totally understand it @ekatef , it's a concern to me for generalization, too. Of course, ACTIVSg82k doesn't provide an explicit categorical linetype identifier to be used to determine the most frequently occurring type, but only the electrical parameters of individual branches, so there is no directly defined linetype for which a frequency or mode could be calculated.

I used the median because it is less affected by outliers than the mean. To check whether this choice was mixing clearly different line families, I tried dividing the branches at each voltage level into two groups based on their electrical parameters, and I then measured how clearly those two groups were separated.  I obtained values ranging from 0.26 and 0.46 ( on a scale -1 to 1, with values close to 1 indicating distinct groups and values close to 0 indicating substantial overlaps).

As an additional check, I compared each parameter set produced by the medians with the actual branches in the dataset.  For every voltage level, the resulting parameter set was very close to at least one branch actually present in the source data.

For future similar works, I wouldn't apply the median automatically. If explicit line-type identifiers are available, the representative type can be selected directly, using an appropriate frequency measure such as line count, circuit length or capacity. If no identifiers are available, the parameter distributions should first be checked for distinct families. When distinct families are present and the median is not applicable due to many outliers, selecting a representative observed branch from the dominant family, or retaining multiple types for the same voltage level, would be more appropriate.

scripts/simplify_network.py
danielelerede-oet: This is closely related to the discussion above with Davide (see https://github.com/pypsa-meets-earth/pypsa-earth/pull/1933#issuecomment-5145598890). The main advantage of keeping `base_voltage` configurable is that the highest available line-type voltage is not necessarily the appropriate voltage for the equivalent network. For example, the US definitions include a 765 kV type, but automatically selecting it would map the entire US network to 765 kV. Based on the network representation and the tests described above, 345 kV was selected instead as the more representative simplification voltage (and PyPSA-USA, for instance, is using 230 kV).

I would therefore keep `base_voltage` configurable and use the country-specific linetype mappings to select the closest available type for that configured voltage.

danielelerede-oet:
> > Hi @ekatef Your maps are indeed useful for inspecting the spatial distribution and identifying local outliers. My screening was more limited and intended as a quick global consistency check, especially as I expected substantial changes to happen with respect to the default behaviour, but I can definitely highlight the differences in the US case through graphical tests.
> > In my tests, for each attribute I reduced the network to its aggregate value and calculated:
> > `relative change = 100 \times (sum_after / sum_before − 1)`
> > I applied this independently to `s_nom `, `num_parallel`, `r`, `x`, and `b`. Then I checked that `s_nom` was preserved by the transformation.
> > For this particular change, I would additionally compare the `type` attribute of `n.lines`, since line-type assignment is the behaviour being modified.
> 
> Hello @danielelerede-oet, great, thank you for the detailed explanation! As the aggregate value, you mean the sum across all the network branches, correct?
> 
> For `type` attribute, completely agree. I'd also add `v_nom` to the testing scope for the work which is aiming to enhance representation of the transmission network.
> 
> For maps, I have sketched some options in this notebook https://github.com/open-energy-transition/zambia-dev-notebooks/blob/linetypes_exploration/linetypes_exploration.ipynb
> 
> As a disclaimer: all the scanning discussion relate to the follow-ups of this PR, whenever in PyPSA-ZM, or as a part of GridBuilder project. I'll open a dedicated issue in GridBuilder to continue a discussion there.

Yes, exactly: by aggregate value I mean the sum of the attribute across all network branches before and after the transformation. I also included `v_nom` in the check. In both the clustered network and the solved network, all 10 transmission lines and all 9 buses connected to them remain at the configured 380 kV base voltage. No line has a nominal voltage inconsistent with either of its endpoint buses (this could be a concern in case of a model with two neighboring countries with a different configured `base_voltage`)

Thank you for sharing the notebook, I'll check it and expand it with the checks I introduced in this PR. Also, a dedicated GridBuilder issue sounds like the appropriate place to continue the broader discussion on graphical and quantitative screening, and having the issue addressed for Zambia would also give us additional indications to understand if the approach is solid enough for reproduction.

scripts/base_network.py
```diff
@@ -842,12 +876,39 @@ def base_network(
...
+    dc_types = lines_config["dc_types"]
+
+    use_country_specific_ac_types = use_country_specific_types and all(
+        country in ac_types for country in countries_config
+    )
+
+    use_country_specific_dc_types = use_country_specific_types and all(
+        country in dc_types for country in countries_config
+    )
+
```
davide-f: This needs revision similarly to above: use_country_specific_dc/ac_types are no longer needed as handled in _set_electrical_parameters_ac/dc_lines


```suggestion
```

scripts/base_network.py
```diff
@@ -842,12 +876,39 @@ def base_network(
...
     lines_ac = lines[~lines.dc].copy()
     lines_dc = lines[lines.dc].copy()
-    lines_ac = _set_electrical_parameters_lines(lines_config, voltages_config, lines_ac)
+
+    lines_ac = _set_electrical_parameters_ac_lines(
+        lines_config,
+        buses,
+        lines_ac,
+        ac_types,
+        use_country_specific_ac_types,
```
davide-f:
```suggestion
        use_country_specific_types,
```

scripts/base_network.py
```diff
@@ -842,12 +876,39 @@ def base_network(
...
+    )
 
     lines_dc = _set_electrical_parameters_dc_lines(
-        lines_config, voltages_config, lines_dc
+        lines_config,
+        buses,
+        lines_dc,
+        dc_types,
+        use_country_specific_dc_types,
```
davide-f:
```suggestion
        use_country_specific_types,
```

scripts/augmented_line_connections.py
```diff
@@ -135,14 +147,52 @@ def haversine(p):
...
+
+    new_long_lines["country"] = new_long_lines["bus0"].map(n.buses["country"])
+    new_long_lines["v_nom"] = new_long_lines["bus0"].map(n.buses["v_nom"])
+    new_long_lines["type"] = new_long_lines.apply(
+        lambda line: get_linetype_by_voltage_and_country(
+            line.v_nom,
+            line.country,
+            dc_linetypes,
+            use_country_specific_dc_types,
```
davide-f:
```suggestion
            use_country_specific_types,
```

scripts/augmented_line_connections.py
```diff
@@ -135,14 +147,52 @@ def haversine(p):
...
+    use_country_specific_dc_types = use_country_specific_types
+
+    new_kedge_lines["country"] = new_kedge_lines["bus0"].map(n.buses["country"])
+    new_kedge_lines["v_nom"] = new_kedge_lines["bus0"].map(n.buses["v_nom"])
+    new_kedge_lines["type"] = new_kedge_lines.apply(
+        lambda line: get_linetype_by_voltage_and_country(
+            line.v_nom,
+            line.country,
+            ac_linetypes,
+            use_country_specific_ac_types,
```
davide-f:
```suggestion
            use_country_specific_types,
```

scripts/augmented_line_connections.py
```diff
@@ -135,14 +147,52 @@ def haversine(p):
...
+    dc_linetypes = lines_config["dc_types"]
+
+    use_country_specific_types = lines_config.get(
+        "use_country_specific_types",
+        False,
+    )
+
+    use_country_specific_ac_types = use_country_specific_types
+    use_country_specific_dc_types = use_country_specific_types
+
```
davide-f: No need to define two use_country_specific_ac_types and use_country_specific_dc_types


```suggestion
```

davide-f (commented):
Hello and thanks for the revisions!

Attached few minor comments, I feel a section of code appears twice and needs harmonization.

I have no other comments

danielelerede-oet:
Thanks @davide-f I applied all the suggestions. I removed the redundant AC/DC-specific flags and now pass `use_country_specific_types` directly, since AC/DC selection and the per-country fallback are already handled by `_set_electrical_parameters_ac_dc_lines()`.

davide-f (approved):
Thanks Daniele!
I have no further comments, and PR is ready be merged as CI passes, in my opinion if @ekatef 's comments have been answered.

@ekatef do you have comments or @danielelerede-oet have you aligned with her during a meeting on the comments I see?

danielelerede-oet:
> Thanks Daniele! I have no further comments, and PR is ready be merged as CI passes, in my opinion if @ekatef 's comments have been answered.
> 
> @ekatef do you have comments or @danielelerede-oet have you aligned with her during a meeting on the comments I see?

Thanks @davide-f , with @ekatef we are working on the choice of custom line types for Zambia and the aim is to also develop more accurate selection criteria in case no straightforward choice can be made, which is instead the case for the US given the detailed TAMU dataset. I think this can be addressed in a future, smaller PR  where we can clearly define these criteria.

ekatef:
Hello! As @danielelerede-oet  has mentioned, my comments were aiming to sketch a plan for further enhancement. @davide-f  feel absolutely free to merge when you see fit. An awesome feature and amazing team work!

davide-f:
Thanks for the confirmations! Merging with thanks

PR #1941, opened 2026-07-14
Derive weather-dependent configuration from top-level `weather_year`

danielelerede-oet:

This PR simplifies the configuration of a model run by resolving the weather year, snapshot range, and atlite cutout configuration consistently.

Previously, users had to manually keep several options aligned when changing the weather year, including:

* `load_options.weather_year`;
* `snapshots.start` and `snapshots.end`;
* the default atlite cutout name.

With this PR, `load_options.weather_year` accepts two modes:

* an explicit integer, e.g., `2013` or `2018` (as it is already)
* or the optional value `derive_from_snapshots`

When an explicit year is provided, the workflow automatically derives:

* `snapshots.start`;
* `snapshots.end`;
* the default atlite cutout name;
* the internal atlite cutout configuration.

For example:

```yaml
load_options:
  source: "gegis"
  weather_year: 2018
```

resolves to:

* snapshots from `2018-01-01` to `2019-01-01`;
* the atlite cutout `cutout-2018-era5`;
* renewable profiles configured with `cutout: auto` using the resolved cutout.

Alternatively:

```yaml
snapshots:
  start: "2013-03-01"
  end: "2013-03-07"
  inclusive: "left"

load_options:
  source: "gegis"
  weather_year: derive_from_snapshots
```

preserves the configured snapshot range and infers the demand weather year from `snapshots.start`.

Snapshot-derived weather years are only accepted when the configured range belongs to a single calendar year. 

The PR also simplifies the atlite configuration by replacing:

```yaml
atlite:
  default: ...
  cutouts:
    ...
```

with:

```yaml
atlite:
  cutout:
    module: era5
    dx: 0.3
    dy: 0.3
```

The resolved `atlite.default` and `atlite.cutouts` entries are reconstructed internally by the workflow. Legacy `atlite.cutouts` and `atlite.default` configurations are migrated automatically to `atlite.cutout` with a deprecation warning.

### User impact

Users can choose which configuration should be authoritative:

* use an explicit `load_options.weather_year` to derive an annual snapshot range and matching cutout;
* use `derive_from_snapshots` to preserve a custom snapshot range and infer the matching weather year.

This removes the need to manually synchronize the demand weather year, snapshots, atlite cutout, and renewable profile configuration.

## Checklist


davide-f (commented):
Hello and thanks @danielelerede-oet !
Some merge conflicts have unfortunately arrived.

Regarding the proposal, what about keeping weather_year in load_options with optional value "derive_from_snapshots" to achieve the same goal?

This is a good point; good to keep in mind that weather_year is not too flexible depending on the data source and we can improve the documentation to improve that.
I see value in having the option to choose a different weather year to enable more customization and I see also the benefit in having the option you suggest.

How do you see having an optional non-default value like "weather_year: derive_from_snapshots" to select the weather_year from snapshots?

danielelerede-oet:
Hi @davide-f , thanks for the suggestion. I have resolved the merge conflicts for the time being, but regarding your though I agree that keeping `weather_year` under `load_options` while supporting `derive_from_snapshots` as an optional, non-default value is a good compromise. I'll work on it

danielelerede-oet:
Hi @davide-f the requested changes are ready and I updated the documentation and PR description accordingly

scripts/_helpers.py
```diff
@@ -358,20 +394,169 @@ def _warn(old: str, new: str) -> None:
...
+        raise ValueError(
+            f"`retrieve_cutout` is enabled, but the requested cutout `{cutout_name}` is not available as a pre-built cutout. Available pre-built cutouts: {available_cutouts}. Disable `retrieve_cutout` and enable `build_cutout`."
+        )
+
+    if build_cutout and cutout_name in PREBUILT_CUTOUTS:
+        warnings.warn(
+            f"The requested cutout `{cutout_name}` is available as a pre-built cutout. It will still be built because `build_cutout` is enabled.",
+            UserWarning,
+            stacklevel=2,
+        )
```
ekatef, line 542: The regional-focused cutouts may be more suitable in many cases as compared with the pre-built ones. For example, geographic boundaries may need adjustment to avoid fetching islands, or some features dropped such as soil temperature for power-only model. 

I think it could be a good idea to decrease a level of the message from warning to info. We can also think about providing a Zenodo link to a pre-built cutout but it may be an over-complication for this PR.

scripts/_helpers.py
```diff
@@ -358,20 +394,169 @@ def _warn(old: str, new: str) -> None:
...
+        config["snapshots"]["start"] = f"{weather_year}-01-01"
+        config["snapshots"]["end"] = f"{weather_year + 1}-01-01"
+
+    cutout_config = config["atlite"]["cutout"].copy()
+    module = cutout_config["module"]
+
+    if config.get("tutorial", False):
+        cutout_name = f"cutout-{weather_year}-{module}-tutorial"
+    else:
+        cutout_name = f"cutout-{weather_year}-{module}"
```
ekatef, line 493: I don't think it's really a good idea to hardcode a pattern for the cutout name. Having a generic `cutout-{year}-{module}.nc` name for all the cutouts excludes an opportunity to have different cutouts for different regions which is a crucial feature when you are running analysis for different regions. Also, it doesn't allow for flexibility in providing different versions of weather data (e.g. differentiated by a downscaling approach used). 

That is a very good idea to read a cutout year from the weather data but we absolutely must leave flexibility for cutout definitions.

scripts/_helpers.py
```diff
@@ -80,6 +82,11 @@
 # absolute path to config.default.yaml
 CONFIG_DEFAULT_PATH = os.path.join(BASE_DIR, "config.default.yaml")
 
+# Cutouts currently available for `retrieve_cutout`. Additional entries can be added here as new pre-built cutouts become available.
+PREBUILT_CUTOUTS = {
+    "cutout-2013-era5",
+}
```
ekatef, line 88: I think it would be more reliable to read available pre-built cutouts from `bundle_config` to avoid duplication of inputs. Names of the pre-built cutouts are available under `output` key of `cutouts` category. Additionally, it's possible to use `get_best_bundles` to find out which specific cutout is minimal fully-coverage cutout for countries of interest.

ekatef (changes requested):
Hello @danielelerede-oet, I know the challenge of having plenty of weather years to be modified and totally agree that some automation could help on that. However, I think that we absolutely must keep an option to provide a custom cutout name as it's essential for a number of use cases. 

Basically, that is a similar comment to what Davide has shared for the demand year: automation shouldn't compromise flexibility. Probably, `derive_from_snapshots` could be a good solution also to automate handling the cutouts?

danielelerede-oet:
> Hello @danielelerede-oet, I know the challenge of having plenty of weather years to be modified and totally agree that some automation could help on that. However, I think that we absolutely must keep an option to provide a custom cutout name as it's essential for a number of use cases.
> 
> Basically, that is a similar comment to what Davide has shared for the demand year: automation shouldn't compromise flexibility. Probably, `derive_from_snapshots` could be a good solution also to automate handling the cutouts?

Thanks @ekatef , I agree. I would propose keeping `atlite.default` and `atlite.cutouts`, so users can retain custom cutout names and definitions.

Setting `atlite.default: derive_from_snapshots` would explicitly enable automatic naming based on the resolved snapshot year. An explicit cutout name would preserve the user’s selection and configuration. This would make the automation optional without restricting existing use cases. What do you think?

scripts/_helpers.py
danielelerede-oet: Agreed, building a custom cutout is a valid choice. Changed the message to `logger.info` and updated the docstring accordingly.

scripts/_helpers.py
danielelerede-oet: I agree @ekatef , it's something I hadn't thought about. I removed the hardcoded list, so that now the available pre-built cutout names are read from the output entries of non-tutorial bundles in the cutouts category.

scripts/_helpers.py
danielelerede-oet: I agree @ekatef , it's something I hadn't thought about. I removed the hardcoded list, so that now the available pre-built cutout names are read from the output entries of non-tutorial bundles in the cutouts category.

scripts/_helpers.py
ekatef: Super!

scripts/_helpers.py
ekatef: Thanks for the revision! Looks very neat.

config.default.yaml
```diff
@@ -189,7 +189,7 @@ base_network:
 # ------------------- Demand -------------------
 load_options:
   source: "gegis" # "gegis" (scenario-based: 2011, 2013, 2018 available) or "demcast" (hourly 2000–2024 available): view [issue #1724](https://github.com/pypsa-meets-earth/pypsa-earth/issues/1724) for more details
-  weather_year: 2013 # Weather year of the load profile
+  weather_year: 2013 # Integer or "derive_from_snapshots"
```
ekatef: To my understanding, the original idea has been to remove redundancy of the weather years from the configuration. Given that, I wonder if using `derive_from_snapshots` could be a good idea for the default config. What do you think?

scripts/_helpers.py
ekatef: I guess the idea is to add here also a case when `atlite.default` is present but is not `derive_from_snapshots`? (apologies if the comment is premature!)

ekatef:
> > Hello @danielelerede-oet, I know the challenge of having plenty of weather years to be modified and totally agree that some automation could help on that. However, I think that we absolutely must keep an option to provide a custom cutout name as it's essential for a number of use cases.
> > Basically, that is a similar comment to what Davide has shared for the demand year: automation shouldn't compromise flexibility. Probably, `derive_from_snapshots` could be a good solution also to automate handling the cutouts?
> 
> Thanks @ekatef , I agree. I would propose keeping `atlite.default` and `atlite.cutouts`, so users can retain custom cutout names and definitions.
> 
> Setting `atlite.default: derive_from_snapshots` would explicitly enable automatic naming based on the resolved snapshot year. An explicit cutout name would preserve the user’s selection and configuration. This would make the automation optional without restricting existing use cases. What do you think?

Hi @danielelerede-oet, sounds a very nice automatisation concept! Let me recap to make sure we see the whole picture:
1) the default version is to have `atlite.default: derive_from_snapshots` and `load_options.weather_year: derive_from_snapshots` to remove redundancy for the weather year definition
2) if a custom load year is needed, `load_options.weather_year` can be set as an integer
3) if a custom cutout is used, it can be enabled with `atlite.default: my_fancy_cutout`.

Let me know please if any corrections are needed for this understanding.

Ping me please when the PR will be ready for a final review round 🙂

ekatef:
@yerbol-akhmetov tagging you in the PR on the weather year we have discussed. This PR is resolving one of the most annoying issues we currently have in the model. Given the importance of this work for usability of the model, I think it would be perfect to have your view on that.

PR #1945, opened 2026-07-16
Support country-specific custom powerplant datasets

danielelerede-oet:

This PR extends the handling of custom powerplants datasets by introducing country-specific custom files and configurable merge/replace strategies on a per-country basis (quite needed when modelling more than one country per model run or managing different countries in the same pypsa-earth local build).

Instead of relying on a single `data/custom_powerplants.csv` file with a global merge/replace behavior, the workflow now automatically detects country-specific datasets following the naming convention:

```text
data/custom_powerplants_{country}.csv
```

The configuration has also been updated to support a general behavior (same as current) together with optional country-specific overrides.

Example:

```yaml
electricity:
  custom_powerplants:
    general: false

    overrides:
      JP: replace
      DE: merge
```

Indeed, the previous implementation assumed a single custom powerplants dataset for the entire model domain. While this worked for single-country studies, it became restrictive for multi-country workflows (e.g. PyPSA-ASEAN / PyPSA-NorthAmerica if extending to Canada and Mexico).

In particular, it was not possible to:

- maintain separate custom datasets for different countries;
- replace powerplantmatching data for one country while merging custom data for another;
- organize custom datasets in a scalable way without manually combining multiple files into a single `custom_powerplants.csv`.

## Changes

- Introduce country-specific custom powerplant files using the naming convention:
  - `data/custom_powerplants_{country}.csv`
- Automatically detect custom datasets for the countries included in the current run.
- Introduce a new configuration structure:
  - `general` defines the default behavior (`false`, `merge`, or `replace`);
  - optional `overrides` allow selecting a different behavior for individual countries.
- Apply `replace` or `merge` only to the corresponding country instead of replacing / merging the entire powerplantmatching dataset.

## Checklist


data/custom_powerplants.csv
davide-f, line 1: I'd suggest to keep custom_powerplants.csv without renaming to minimize braking changes

doc/configtables/snippets/electricity.yaml
```diff
@@ -25,7 +25,10 @@ electricity:
     Link: [] # H2 pipeline
 
   powerplants_filter: (DateOut >= 2022 or DateOut != DateOut) and (DateIn <= 2023 or DateIn != DateIn)
-  custom_powerplants: false #  "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
+  custom_powerplants:
+    general: false #  "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
+    # overrides:
+    #   country: merge # allows choosing different strategies for different countries. "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
```
davide-f: What about having the option:
```suggestion
  custom_powerplants:
    method: false #  "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
    filepaths=[data/custom_powerplants.csv]  # list of custom powerplant files to be included
```
this can preserve the previous version and allows multiple files to support, e.g. _US/_BR etc. It will be up to the user on how to customize the config files and use the feature to include selected files only when needed.
the filepaths can be included as input to snakemake and then we can concatenate those files in the script.

How do you see this proposal?

scripts/build_powerplants.py
```diff
@@ -242,37 +243,73 @@ def convert_osm_to_pm(filepath_ppl_osm, filepath_ppl_pm):
...
 ) -> pd.DataFrame:
     """
-    Adds custom powerplants to the powerplants dataframe.
+    Apply country-specific custom powerplant files.
+
+    Files must follow the naming convention:
+
+        custom_powerplants_{country}.csv
+
+    where ``country`` is the ISO2 country code.
```
davide-f: With the proposal below we could support any file name and concatenate them. This could save some here and in the snakefile.
What do you think?

config.default.yaml
```diff
@@ -226,7 +226,10 @@ electricity:
     Link: [] # H2 pipeline
 
   powerplants_filter: (DateOut >= 2022 or DateOut != DateOut) and (DateIn <= 2023 or DateIn != DateIn)
-  custom_powerplants: false #  "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
+  custom_powerplants:
+    general: false #  "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
+    # overrides:
+    #   country: merge # allows choosing different strategies for different countries. "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
```
davide-f: What about having the option:
```suggestion
  custom_powerplants:
    method: false #  "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
    filepaths: ["data/custom_powerplants.csv"]  # list of custom powerplant files to be included
```
this can preserve the previous version and allows multiple files to support, e.g. _US/_BR etc. It will be up to the user on how to customize the config files and use the feature to include selected files only when needed.
the filepaths can be included as input to snakemake and then we can concatenate those files in the script.

How do you see this proposal?

davide-f (changes requested):
Thanks @danielelerede-oet for the proposal and shared a few comments. Thanks!

doc/configtables/snippets/electricity.yaml
danielelerede-oet: @davide-f this proposal removes a crucial functionality of my change: supporting an explicit list of file paths would preserve compatibility with the existing implementation, but  a single global method would not support the main purpose of this change, such as using different strategies for different countries in the same run. For example, we may want to replace the PPM data for the US, merge custom plants for Canada, and keep only powerplantmatching data for Mexico.

We could combine both approaches by keeping the country-specific strategies while allowing the corresponding files to be configured explicitly. Otherwise, the user would need to preprocess the inputs, and replace could again affect countries for which no replacement was intended.

davide-f: > @davide-f this proposal removes a crucial functionality of my change: supporting an explicit list of file paths would preserve compatibility with the existing implementation, but a single global method would not support the main purpose of this change, such as using different strategies for different countries in the same run. For example, we may want to replace the PPM data for the US, merge custom plants for Canada, and keep only powerplantmatching data for Mexico.
> 
> We could combine both approaches by keeping the country-specific strategies while allowing the corresponding files to be configured explicitly. Otherwise, the user would need to preprocess the inputs, and replace could again affect countries for which no replacement was intended.

I see. What about enabling "method:" to be a list one for each element of the filepaths?
To enable back-compatibility, we can assume that if method is not a list, then that approach is used for all files

danielelerede-oet: This was my initial thought as well, but I wanted to avoid hardcoding file paths. With your proposal, I would essentially associate each file path with a method, which is fine for the aim of the PR.

davide-f: The data of powerplants can be filtered using "powerplants_filter" with a "and (country != 'US')".
Do you think we need an additional option for that?

In the case we have the option "replace" by config file, we need to define the rule on how data are replaced. Following your thought, we could drop data for any country in the list, which is equivalent to (country != 'XX').
I'm thinking of possible downsides, e.g. what if someone wants to drop all technologies for one country. In that case "replace" as it is in upstream would work well, while with the revision it would no longer work. It is a kind-of edge case, though

danielelerede-oet: Yes, I agree. Country-specific replacement can already be achieved by excluding the relevant countries through powerplants_filter and then using merge for the custom files. Therefore, I dont't think we need an additional option. I would keep `replace` with its existing upstream behaviour (ppm dataset is discarded), and document `powerplants_filter` + `merge` as the way to perform country-specific replacements.

danielelerede-oet:
@davide-f I have revised the implementation based on the discussion above. `replace` now preserves its original meaning and replaces the complete current dataset, while country-specific replacement can be achieved using `powerplants_filter` together with `merge`. Multiple custom files and per-file methods are also still supported, and methods are applied in filepath order. I also updated the configuration examples and documentation to clarify this behaviour.

Snakefile
```diff
@@ -626,18 +626,49 @@ rule build_renewable_profiles:
...
+    invalid_methods = [
+        method for method in configured_methods if method not in allowed_methods
+    ]
+    if invalid_methods:
+        raise ValueError(
+            "electricity.custom_powerplants.method accepts only false, "
+            f"'merge', or 'replace'; found {invalid_methods}."
+        )
+
+    return filepaths
```
davide-f: Let's shrink this; ideally we could have a 1 or 2 line directly inside the rule. I'm commenting below.

Snakefile
```diff
@@ -626,18 +626,49 @@ rule build_renewable_profiles:
...
         gadm_layer_id=config["build_shape_options"]["gadm_layer_id"],
         alternative_clustering=config["clustering"]["alternative_clustering"],
         powerplants_filter=config["electricity"]["powerplants_filter"],
-        custom_powerplants_option=config["electricity"]["custom_powerplants"],
+        custom_powerplants=config["electricity"]["custom_powerplants"],
     input:
         base_network="networks/" + RDIR + "base.nc",
         pm_config="configs/powerplantmatching_config.yaml",
-        custom_powerplants="data/custom_powerplants.csv",
+        custom_powerplants=get_custom_powerplants_files,
```
davide-f: What do you think of something like this? (needs polishing and alignment with final implementation)
```suggestion
        custom_powerplants=branch(
              config["electricity"]["custom_powerplants"]["method"] != false,
              config["electricity"]["custom_powerplants"]["filepaths"],
              []
        ),
```

config.default.yaml
```diff
@@ -204,7 +204,20 @@ electricity:
 
   # ----------------- Existing powerplants -------------------
   powerplants_filter: (DateOut >= 2022 or DateOut != DateOut) and (DateIn <= 2023 or DateIn != DateIn)
-  custom_powerplants: false #  "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
+  custom_powerplants:
+    filepaths:
+    - data/custom_powerplants.csv
+    method: false # false, merge, or replace; may also be a list matching filepaths
```
davide-f: I'd suggest to drop the list-based approach.
Since we agreed the country-based filter stays in powerplants_filter, the list option can be confusing.
I'd preserve a description of the meaning of false, "merge" and "replace", similarly to in upstream

Given current implementation, using [false, merge, replace], for files [A,B,C] would have the following effect:
- A would be ignored
- B would be merged with base ppl and
- "replace" would disregard the previous base, A and B and use C.
It would be equivalent to use only "replace" with file [C]

davide-f (commented):
Given the exchange, I believe the list-based approach has become redundant; I've provided an example.
Almost ready to go for me

danielelerede-oet:
Hi @davide-f thanks for your comments. 

I have replaced the helper with the inline `branch` expression and removed support for assigning different methods to individual files. A single method is now applied once to the complete set of configured custom powerplant files: `false' ignores them, `merge` appends all of them to the filtered ppm dataset, and `replace` uses all of them instead of the complete ppm dataset. I updated the configuration examples and the `build_powerplants.py` docstring accordingly, including an example showing how country-specific replacement can be achieved by excluding the relevant country through `powerplants_filter` and then using `method: merge`. 

NOTE: Country conditions in `powerplants_filter` must use the full name from the original ppm dataset, such as `Country != 'United States'`, because the filter is evaluated before country names are converted to ISO alpha-2 codes. Custom input files and the generated `powerplants.csv` continue to use codes such as US.

scripts/build_powerplants.py
```diff
@@ -353,21 +437,21 @@ def replace_natural_gas_technology(df: pd.DataFrame):
...
-    ppl = add_custom_powerplants(ppl, snakemake.input, custom_ppl_query).query(
-        ppl_query
-    )  # add carriers from own powerplant files
+    ppl = (
+        pm.powerplants(from_url=False, update=True, config_update=config)
+        .powerplant.fill_missing_decommissioning_years()
+        .query("Country in @countries_names")
+        .powerplant.convert_country_to_alpha2()
+        .pipe(replace_natural_gas_technology)
+    )
```
davide-f: Please, restore the if/else: the if allows to skip the ppl creation when replace option is requested

config.default.yaml
```diff
@@ -204,7 +204,30 @@ electricity:
...
-  custom_powerplants: false #  "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
+  custom_powerplants:
+    filepaths:
+    - data/custom_powerplants.csv
+    # Applied globally to all configured custom powerplant files:
+    # false: ignore custom files and use only PowerplantMatching data
+    # merge: add custom files to the filtered PowerplantMatching dataset
+    # replace: discard the complete PowerplantMatching dataset and use only
+    #          the configured custom files
+    method: false
```
davide-f: Propose to simplify:
```suggestion
    filepaths:  # list of custom powerplant files used in "merge" and "replace" options
    - data/custom_powerplants.csv
    method: false  # "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
```

config.default.yaml
```diff
@@ -204,7 +204,30 @@ electricity:
...
+  # custom_powerplants:
+  #   filepaths:
+  #   - data/custom_powerplants_US.csv
+  #   - data/custom_powerplants_CA.csv
+  #   method: merge
+  #
+  # Result:
+  # - US: custom data only, because US plants are excluded from PPM
+  # - CA: PPM and custom data are merged
+  # - MX: PPM data only, because no custom file is provided
```
davide-f: Propose to shrink the text slightly. Feel free to edit as you see fit.
```suggestion
  # If you wish to replace the data for just one country (e.g. United States), you can use the option powerplants_filter as follows:
  # powerplants_filter: ... and (Country != 'United States')
  # custom_powerplants:
  #   filepaths:
  #   - data/custom_powerplants_US.csv
  #   method: merge
```

doc/configtables/snippets/electricity.yaml
```diff
@@ -15,7 +15,30 @@ electricity:
 
   # ----------------- Existing powerplants -------------------
   powerplants_filter: (DateOut >= 2022 or DateOut != DateOut) and (DateIn <= 2023 or DateIn != DateIn)
-  custom_powerplants: false #  "false" use only powerplantmatching (ppm) data, "merge" combines ppm and custom powerplants, "replace" use only custom powerplants
+  custom_powerplants:
```
davide-f, line 18: This file needs revision in agreement to the revision of config.default.yaml

davide-f (commented):
Minor comments left, many thanks @danielelerede-oet ! :)

davide-f:
> NOTE: Country conditions in `powerplants_filter` must use the full name from the original ppm dataset, such as `Country != 'United States'`

This is a good point and it is the effect of this line:
https://github.com/pypsa-meets-earth/pypsa-earth/blob/cd3d07b5aaa874fa203731e3a66094c3cb0ecc6c/scripts/build_powerplants.py#L362

For harmonization, I'd suggest to move that line, after the custom powerplants are included because it would lead to inconsistencies:
- The "main" powerplantmatching query requires "United States" as the query is applied before the conversion,
- The "custom_powerplants" do not have that conversion after the df is created thus potentially leading to inconsistencies in the filtering

danielelerede-oet:
> > NOTE: Country conditions in `powerplants_filter` must use the full name from the original ppm dataset, such as `Country != 'United States'`
> 
> This is a good point and it is the effect of this line:
> 
> https://github.com/pypsa-meets-earth/pypsa-earth/blob/cd3d07b5aaa874fa203731e3a66094c3cb0ecc6c/scripts/build_powerplants.py#L362
> 
> For harmonization, I'd suggest to move that line, after the custom powerplants are included because it would lead to inconsistencies:
> 
> * The "main" powerplantmatching query requires "United States" as the query is applied before the conversion,
> * The "custom_powerplants" do not have that conversion after the df is created thus potentially leading to inconsistencies in the filtering

Yes, that was also my initial idea, and I wanted to avoid it other changes, but if you agree about that I can also do that

davide-f:
> Yes, that was also my initial idea, and I wanted to avoid it other changes, but if you agree about that I can also do that

Fully agree, many thanks!

danielelerede-oet:
@davide-f I moved `convert_country_to_alpha2` after `add_custom_powerplants`, so country conversion is now applied consistently to the complete resulting dataframe, including the custom powerplants file

scripts/build_powerplants.py
```diff
@@ -353,21 +443,29 @@ def replace_natural_gas_technology(df: pd.DataFrame):
...
-            pm.powerplants(from_url=False, update=True, config_update=config)
+            pm.powerplants(
+                from_url=False,
+                update=True,
+                config_update=config,
+            )
             .powerplant.fill_missing_decommissioning_years()
             .query("Country in @countries_names")
-            .powerplant.convert_country_to_alpha2()
-            .pipe(replace_natural_gas_technology)
```
davide-f, line 363: This line has been dropped; please, restore it

scripts/build_powerplants.py
davide-f: Solved now

davide-f (commented):
Thanks @danielelerede-oet !
A minor comment, please also check some previous comments that are still relevant:
- https://github.com/pypsa-meets-earth/pypsa-earth/pull/1945#discussion_r3656476086
- https://github.com/pypsa-meets-earth/pypsa-earth/pull/1945#discussion_r3656456937
Their core idea is to limit the lines and text in the config file, while keeping the content; exact wording is not necessary
Thank you!

danielelerede-oet:
Thanks @davide-f I addressed both the latest and the older comments

davide-f (approved):
Thanks @danielelerede-oet! , merging

PR #1960, opened 2026-07-30
Add deep-copy for config migration for back-compatibility

### LINKED ISSUE #1958: Increase stability of configuration migration infrastracture

We have infrastructure which allows for backward compatibility when the config parameters are modified in the upstream. There is `migrate_config` function which takes care about parsing old configuration parameters into new ones.

One usability point which could be improved is processing a situation when a custom config file defines only some part of parameters in a config block.

### Example
A local config defines only this line from the whole `cluster_options/clustering` (which is a usual way of defining config-diffs):

```yaml
cluster_options:
  alternative_clustering: false # "False" use Voronoi shapes, "True" use GADM shapes
```
That leads to an error `KeyError in file ~/pypsa-earth/Snakefile, line 883:
'aggregation_strategies'` which is supported by a warning `FutureWarning: 'cluster_options' is deprecated, use 'clustering' instead`.

Fixing the name of the config key in the custom config resolves the issue, and a warning provides a hint on that. Potentially, we could improve user experience further by accounting for such edge cases in `migrate_config`. Another valid approach is providing more specific instructions in the documentation of how to update the config  and adding a link to the warning

yerbol-akhmetov:
# Closes #1958

Good day. This PR aims to add deep-copy of the config during migration to prevent issues when user only has old configuration file with only some keys defined.  

## Checklist


ekatef (commented):
Thanks a lot for the prompt fix @yerbol-akhmetov! I confirm that it resolves #1958 and correctly transfers a value of the overwriting key. The implementation looks quite neat. Could you please also add a release note? Then I think the PR is basically ready to merge once CI passes unless you have any other points to add.

yerbol-akhmetov:
> Thanks a lot for the prompt fix @yerbol-akhmetov! I confirm that it resolves #1958 and correctly transfers a value of the overwriting key. The implementation looks quite neat. Could you please also add a release note? Then I think the PR is basically ready to merge once CI passes unless you have any other points to add.

Thank you for review, @ekatef.  I have added the release notes.

yerbol-akhmetov:
@ekatef, I have updated the PR based on `main`. Do you think we can merge it?

ekatef:
> @ekatef, I have updated the PR based on `main`. Do you think we can merge it?

Absolutely. Thanks a lot for the prompt implementation.

ekatef (approved):

PR #1844, opened 2026-05-27
Add build_natura_raster docstring

Eric-Nitschke:
This is part of #1831. As this is the first time I am writing docstring, I'd appreaciate your feedback.

I've tested this locally with `mkdocs serve` and it looks generally good. Still I noticed a few things that also appear in the provided example (`add_electricity`):
- links are not properly implemented
- `.. code:: yaml` is visible in the "Relevant Settings" section
- function examples look strange. I think none of the other rules provide examples yet, so I am not sure if it's just an issue with my docstring.

## Changes proposed in this Pull Request
- update the docstring for the build_natura_raster rule
- add docstring to all functions in build_natura_raster

## Checklist

- [x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [x] I tested my contribution locally and it seems to work fine.
- [x] Code and workflow changes are sufficiently documented, including updates to docstrings for meaningful functions.
- [x] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`.
- [ ] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [ ] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/user-guide/configuration.md` and `doc/tutorials/electricity-model.md`.
- [ ] If config sections were added, renamed, or removed, update `doc/assets/scripts/extract_config_snippets.py` accordingly.
- [ ] Archives of the uploaded data do not have an enclosing folder and archive names correspond to the conventions of `configs/bundle_config.yaml`.
- [x] A note for the release notes `doc/release-notes.md` is amended in the format of previous release notes, including reference to the requested PR.

PR #1971, opened 2026-08-05
Remove oil reserves

danielelerede-oet:

This PR removes the parameter `oil.reserves: 100` introduced in [PyPSA-Earth-sec #170](https://github.com/pypsa-meets-earth/pypsa-earth-sec/pull/170) and not clearly documented, which is causing mismatches in sector-coupled network result wrt power-only networks as oil comes for free (and can potentially, uncontrollably increase oil power generation).

The current `main` implementation initializes every oil Store with 100 TWh of free oil per bus. Removing these reserves exposed an existing transport-profile scaling error, while making the Store non-cyclic only hid the problem by allowing synthetic oil to accumulate as an implicit carbon sink. The transport correction normalizes demand profiles using the network snapshot weights and explicitly aggregates hourly profiles to the network resolution. 

## Checklist


ekatef (commented):
Thank you @danielelerede-oet! Agree that it's a point which needs some clean-up, and agree that it may be simpler just to remove this part. What I'm not quite clear how to deal with this specific fragment:

https://github.com/pypsa-meets-earth/pypsa-earth/blob/e3848844e790efc0c2f3632c968f105a49e6711e/scripts/prepare_sector_network.py#L301-L312

As we have discussed, `fossil_reserves` is used a switch to present different fuels in a different ways. Do you think it's alright to keep only the one?

yerbol-akhmetov (changes requested):
Hi, @danielelerede-oet. I agree with @ekatef. I think the changes required in `prepare_sector_network` to make Store `extendable` and `cyclic` without setting `e_initial`. Maybe you have implemented it already, but did not push.

danielelerede-oet:
Hi @yerbol-akhmetov @ekatef  I removed the oil reserve entry from the configuration and simplified the shared code with oil and gas by removing `e_initial` and setting `e_cyclic=True` for all fossil fuel stores. Therefore, this changes the behavior only for oil, but for coal and gas it's just code cleanup.

scripts/prepare_sector_network.py
```diff
@@ -298,18 +298,13 @@ def add_carrier_buses(n: pypsa.Network, carrier: str, nodes: list = None) -> Non
     n.madd("Bus", nodes, location=location, carrier=carrier)
 
     # initial fossil reserves
```
yerbol-akhmetov:
```suggestion
```

ekatef:
Hey @danielelerede-oet @yerbol-akhmetov, agree it makes perfect sense to remove a depleting store mode for oil for now. 

A small additional comment: we also have this line in the config migration documentation which I think must be updated:
https://github.com/pypsa-meets-earth/pypsa-earth/blob/5e936fa713fc26058e2fa2479b9969d981030ada/doc/user-guide/configuration.md?plain=1#L47

What is more serious I see CI is complaining when checking the model for infeasibilities `NotImplementedError("Solver model must be a Gurobi Model.")`. Do you have any idea on a possible reason? My first (most obvious) thought is that the load shedding value may be not enough after we have removed "free oil".

danielelerede-oet:
@ekatef thanks for pointing out the CI failure. The full model gets easily solved with the new configuration, but I reproduced the sector test locally with Gurobi so that I could get an IIS, since the current GLPK run only reports `infeasible_or_unbounded` and then fails while trying to run the gurobi infeasibility diagnostics.

The IIS contains 53 constraints, including:

* the nodal balances of all oil buses;
* the oil Store energy-balance constraints;
* the atmospheric CO2 balance;
* the CO2 limit;
* the geological sequestration limit.

With the previous configuration, oil had:

```python
e_initial = 100 #TWh per bus
e_cyclic = False
```

And the PR changes this to:

```python
e_initial = 0
e_cyclic = True
```

In the sector CI test, temporal aggregation leaves a single snapshot with a weighting of 8760 hours. For a cyclic Store, the energy-balance constraint becomes:

```text
-e - 8760 * p + e = 0
```

and therefore enforces `p = 0`. The oil Store cannot have any net charge or discharge over the model horizon.

This is due to Fischer–Tropsch consuming hydrogen and CO2 while producing oil. Under the very tight CO2 constraint, the model appears to require net production and accumulation of synthetic oil. Once the oil Store is made cyclic, that terminal inventory is no longer possible. The IIS then shows a conflict between the oil balances, atmospheric CO2 balance, CO2 cap, and sequestration limit.

Gas Stores also appear in the IIS, but gas was already cyclic before this PR and the base commit passed CI. The behavioural change introduced by this PR is specifically the oil Store becoming cyclic.

I see three possible approaches:

1. Minimal change: remove the free initial oil reserve but retain a non-cyclic, initially empty oil Store.

```python
e_initial = (
    snakemake.params.sector_options.get(carrier, {}).get("reserves", 0)
) * 1e6

n.madd(
    "Store",
    nodes + " Store",
    bus=nodes,
    e_nom_extendable=True,
    e_cyclic=e_initial == 0 and carrier != "oil",
    e_initial=e_initial,
    carrier=carrier,
)
```

This would prevent free fossil-oil supply while allowing oil from FT to remain as terminal inventory. Before accepting this solution, we should inspect the resulting `e_nom_opt` and final oil inventory. If they are very large, the model is effectively using synthetic oil accumulation as an implicit CO2 sequestration option.

2. Keep the oil Store cyclic and revise the test/model assumptions. This could mean relaxing the CO2 limit, increasing the geological sequestration potential, or reconsidering the role of Fischer–Tropsch under this test configuration. This avoids terminal synthetic-oil accumulation but changes the scenario assumptions.

3. Cleaner long-term formulation: separate fossil-oil supply from synthetic-oil storage. Fossil oil can be represented by the existing Generator at its fuel marginal cost, while a distinct, initially empty Store represents synthetic-oil inventory with an explicit capacity cost, terminal condition, or upper limit. This avoids mixing fossil reserves and synthetic-fuel storage in the same component.

Though I like option 3, my proposal is to test option 1 first and inspect the optimized terminal oil stock. If it is reasonable, it is the smallest fix consistent with the purpose of this PR. If it is excessive, the CI failure is exposing a broader modelling issue and option 3 would be safer.

On a side note, the infeasibility reporting should also be guarded by the solver name because `compute_infeasibilities()` is only available for Gurobi; otherwise GLPK masks the actual failure with `NotImplementedError("Solver model must be a Gurobi Model.")`.

ekatef:
Hey @danielelerede-oet, as mentioned that is really an exemplary analysis which we must find a way to implement into the documentation. It is not so rare you encounter Interactions between CO2 management and synthetic fuel production, but it's not so often you get such a clear explanation on the reasons and mechanisms.

On the strategy, completely agree that point 3 sounds a good long-term solution. Would be good however first to refactor and ideally modularise/tabularize the sector-coupled implementation as otherwise we are risking to loose control over complexity. 

Choosing between other solutions,  my personal preferences belongs to the point 2 as that sounds a way to make the model more physically sensible. But I guess it would requite some iterations to adjust parameters which may be waste of resources (unlsee we'll find some way integrate it into some educational activities). So, agree that point 1 sounds a way to go. It's additional advantage is avoiding abrupt changes in the model behaviour which is generally a good thing.

danielelerede-oet:
Hi @ekatef . I implemented option 1 by removing the initial oil reserves and defining the oil store as initially empty and non-cyclic. The CI now solves successfully all the scenarios. However, I inspected the solved networks to understand why feasibility was restored, and the result confirms the concern above:

CI scenario | Terminal oil inventory | Fischer–Tropsch production | Share accumulated
-- | -- | -- | --
2030 | 11,359 TWh | 14,070 TWh | 80.7%
2050 | 13,825 TWh | 17,001 TWh | 81.3%
CCL-2030 | 11,661 TWh | 14,371 TWh | 81.1%

In every case, the optimized oil Store capacity is exactly equal to its terminal inventory. So, option 1 restores feasibility because Fischer–Tropsch can remove CO2 from the system and store most of the resulting synthetic oil indefinitely. 

Since the Store has neither a terminal condition nor an explicit storage cost or capacity limit, it effectively becomes an implicit carbon sink (though these values come from the reduced CI tests, so they should not be interpreted as the actual results).

In any case, the test passes because terminal oil accumulation relaxes the interaction between the CO2 balance, geological sequestration and Fischer–Tropsch production. In this case, I don't think we should proceed with option 1. The results seem to support your preference for option 2, despite the additional work it may require.

Would you agree that we should proceed with option 2 before merging, while keeping option 3 as the longer-term structural solution?

danielelerede-oet:
@ekatef following the issues registered in option 1, I implemented option 2: oil stores are now initialized empty and remain cyclic. This removes the initial fossil-oil reserve while preventing synthetic oil from accumulating at the end of the optimization horizon.

The initial cyclic implementation made the reduced CI scenario infeasible. The issue was not the CO2 sequestration limit, but I have identified two errors in the temporal handling of transport profiles:

* Transport profiles were normalized without snapshot weights. In temporally reduced networks (such as in the sector test), this caused the annual demand to be multiplied again by the representative-snapshot weights.
* Hourly transport profiles were passed to a network with coarser snapshots without explicit aggregation. The model therefore retained matching timestamps instead of averaging the complete intervals.

Together, these errors increased land-transport oil demand from approximately 246 TWh to 2687 TWh in the CI sector scenario.

The correction adopted now normalizes transport profiles using `snapshot_weightings`, aligns profiles with the network resolution explicitly, uses the mean for transport demand and vehicle availability, and the max for DSM minimum-state-of-charge constraints. In addition, it  defines all oil Stores with `e_initial=0.0` and `e_cyclic=True` as from our initial point.

The model now remains feasible even without initial oil reserves and without using synthetic-oil accumulation as an implicit carbon sink.

doc/release-notes.md
```diff
@@ -61,6 +61,9 @@ This part of documentation collects descriptive release notes to capture the mai
 * Reorganize config for ``co2``, ``solar_thermal``, and line length settings. Old config keys will be depreciated in future releases [PR #1863](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1863)
 
 **Minor Changes and bug-fixing**
+* Remove oil reserves [PR #1971](https://github.com/pypsa-meets-earth/pypsa-earth/pull/1971)
+
+* Add docstrings to build_natura_raster script [PR ##1844](https://github.com/pypsa-meets-earth/pypsa-earth/pull/#1844)
```
yerbol-akhmetov: I think it is duplicated occasionally, @danielelerede-oet

scripts/prepare_sector_network.py
```diff
@@ -1725,6 +1720,7 @@ def add_shipping(
             [oil_bus + " Store" for oil_bus in spatial.oil.nodes],
             bus=spatial.oil.nodes,
             e_nom_extendable=True,
+            e_initial=0.0,
```
yerbol-akhmetov: Hi, @danielelerede-oet. If we have resolved the resampling of transport profiles, why do we need to force `e_initial` as 0. I mean here and above. Maybe we do not force it to be 0 at the first snapshot?

yerbol-akhmetov (commented):
Hi, @danielelerede-oet. Great thanks for tackling the issue. I have left one comment.

doc/release-notes.md
danielelerede-oet: @yerbol-akhmetov do you mean the duplicated # in the PR #1844 reference, or that one of the release-note entries appears twice? I have also updated the PR #1971 entry to include the transport-profile weighting and temporal aggregation fixes added after the investigation

scripts/prepare_sector_network.py
danielelerede-oet: @yerbol-akhmetov sure, you're right

doc/release-notes.md
yerbol-akhmetov: Hi, @danielelerede-oet. `Add docstrings to build_natura_raster script [PR ##1844](https://github.com/pypsa-meets-earth/pypsa-earth/pull/#1844` appears twice in the `release_notes.md`.

scripts/prepare_sector_network.py
```diff
@@ -297,19 +297,14 @@ def add_carrier_buses(n: pypsa.Network, carrier: str, nodes: list = None) -> Non
...
-        snakemake.params.sector_options.get(carrier, {}).get("reserves", 0)
-    ) * 1e6
-    # capital cost could be corrected to e.g. 0.2 EUR/kWh * annuity and O&M
     n.madd(
         "Store",
         nodes + " Store",
         bus=nodes,
         e_nom_extendable=True,
-        e_cyclic=True if e_initial == 0 else False,
+        e_initial=0.0,
```
yerbol-akhmetov: Hi, @danielelerede-oet. What about here? I think we can remove `e_initial` here too.

doc/release-notes.md
yerbol-akhmetov: Here, natura raster PR is duplicated.

yerbol-akhmetov:
Hi, @danielelerede-oet. Sorry, I forgot to mention about the changes needed to documentation as we no longer need `fossil_reserves`. We have recently changed keys for fossil reserves and added function for back compatibility. I think we no longer need it.
https://github.com/pypsa-meets-earth/pypsa-earth/blob/0c4e02f4a650fa9c4e1336039f1a077f0c974c39/scripts/_helpers.py#L304-L318
Also, 
https://github.com/pypsa-meets-earth/pypsa-earth/blob/0c4e02f4a650fa9c4e1336039f1a077f0c974c39/scripts/_helpers.py#L361

In addition, we need to modify `configuration.md`
https://github.com/pypsa-meets-earth/pypsa-earth/blob/0c4e02f4a650fa9c4e1336039f1a077f0c974c39/doc/user-guide/configuration.md?plain=1#L47

https://github.com/pypsa-meets-earth/pypsa-earth/blob/0c4e02f4a650fa9c4e1336039f1a077f0c974c39/doc/user-guide/configuration.md?plain=1#L530

and here
https://github.com/pypsa-meets-earth/pypsa-earth/blob/0c4e02f4a650fa9c4e1336039f1a077f0c974c39/doc/configtables/sector_toplevel.csv?plain=1#L42

Please feel free to scan whole repository for leftovers.

danielelerede-oet:
> Hi, @danielelerede-oet. Sorry, I forgot to mention about the changes needed to documentation as we no longer need `fossil_reserves`. We have recently changed keys for fossil reserves and added function for back compatibility. I think we no longer need it.
> 
> https://github.com/pypsa-meets-earth/pypsa-earth/blob/0c4e02f4a650fa9c4e1336039f1a077f0c974c39/scripts/_helpers.py#L304-L318
> 
> 
> Also,
> https://github.com/pypsa-meets-earth/pypsa-earth/blob/0c4e02f4a650fa9c4e1336039f1a077f0c974c39/scripts/_helpers.py#L361
> 
> In addition, we need to modify `configuration.md`
> 
> https://github.com/pypsa-meets-earth/pypsa-earth/blob/0c4e02f4a650fa9c4e1336039f1a077f0c974c39/doc/user-guide/configuration.md?plain=1#L47
> 
> https://github.com/pypsa-meets-earth/pypsa-earth/blob/0c4e02f4a650fa9c4e1336039f1a077f0c974c39/doc/user-guide/configuration.md?plain=1#L530
> 
> and here
> 
> https://github.com/pypsa-meets-earth/pypsa-earth/blob/0c4e02f4a650fa9c4e1336039f1a077f0c974c39/doc/configtables/sector_toplevel.csv?plain=1#L42
> 
> Please feel free to scan whole repository for leftovers.

Thanks @yerbol-akhmetov I have now removed the obsolete fossil-reserve documentation, the renamed-key entry, the sector_toplevel.csv entry, and the _migrate_fossil_reserves compatibility helper and call.

yerbol-akhmetov (approved):
Thank you, @danielelerede-oet, for the revision. I think the PR is ready to be merged now.

PR #1809, opened 2026-05-04
Polishing time-dependent quantities in sector network

davide-f:
# Closes #1808

## Checklist

- [x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [ ] I tested my contribution locally and it seems to work fine.
- [ ] Code and workflow changes are sufficiently documented, including updates to docstrings for meaningful functions.
- [ ] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`.
- [ ] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [ ] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/user-guide/configuration.md` and `doc/tutorials/electricity-model.md`.
- [ ] If config sections were added, renamed, or removed, update `doc/assets/scripts/extract_config_snippets.py` accordingly.
- [ ] Archives of the uploaded data do not have an enclosing folder and archive names correspond to the conventions of `configs/bundle_config.yaml`.
- [x] A note for the release notes `doc/release-notes.md` is amended in the format of previous release notes, including reference to the requested PR.

PR #2027, opened 2026-09-07
Represent fixed sector emissions as time-dependent loads

danielelerede-oet:

This PR represents fixed sectoral CO2 emissions as time-dependent atmospheric loads.

The issue became relevant while validating the sector-specific CO2 constraints introduced in PyPSA-NorthAmerica [PR #60](https://github.com/pypsa-meets-earth/pypsa-northamerica/pull/60).

In `prepare_sector_network.py`, emissions from land transport, residential and services were reduced to a single average value even though the underlying fossil-fuel demand is time-dependent. The services implementation already contained a `TODO` noting that this should be checked with different snapshot settings.

This becomes particularly relevant when snapshot durations are non-uniform, since an unweighted average does not represent the annual emissions correctly.

The change applies to:

- land transport;
- residential oil and gas;
- services oil and gas.

The corresponding atmospheric `Load` components now receive a time-dependent `p_set`, preserving the emission profile of the underlying fuel demand. Snapshot weightings can then be applied when an aggregated annual emission quantity is needed.

## Checklist


davide-f (commented):
Thanks @danielelerede-oet ! This links well with #1809 and the discussion on the time-dependency of co2, we had with @Yerbol.

In #1809 we discussed on the possibility of setting co2 as a time-dependent quantity.
I'm raising this as that feature would solve your requirement and make the code clean as it drops the need for the weighted average and several averaging.
On a computational level, it is hard to say the impact of specifying an average vs a time-dependency value, so that shouldn't be a problem.

What do you think?

Pinging @ekatef and @yerbol-akhmetov as it has implications on our standards

danielelerede-oet:
Thanks Davide for flagging that, and I agree indeed. I went back to #1809 and saw that @yerbol-akhmetov had already suggested representing these CO2 emissions as time-dependent quantities when the underlying demand is itself time-dependent.

I think that is cleaner to represent CO2 emissions as time-dependent quantities when the underlying demand is itself time dependent, rather than introducing a weighted average here. 

As a side note ,this should also fit well with the sector-specific constraints in PyPSA-NorthAmerica (that will be raised upstream soon). There, fixed atmospheric Load emissions are already retrieved as a time series and integrated using the snapshot weights, so time-dependent CO2 loads can be handled directly.

I can update this PR accordingly and then adapt the country-level emission loads in [#60](https://github.com/pypsa-meets-earth/pypsa-northamerica/pull/60) in the same way.

danielelerede-oet:
@davide-f I have updated the PR according to the comments above: land transport, residential and services CO2 emissions are now kept as time-dependent atmospheric Load profiles instead of being reduced to an average.

danielelerede-oet:
@davide-f I also noticed that introducing time-dependent atmospheric CO2 loads made the model infeasible when load shedding was enabled. This was caused by the load-shedding capacity calculation assigning a negative capacity to the `co2 atmosphere` bus.

As a cross-check, disabling load shedding made the same model solve to optimality, confirming that the issue was unrelated to the time-dependent CO2 accounting itself.

I updated `get_load_shedding_capacity()` to use the dense load representation, so static and time-dependent `p_set` values are handled consistently, aggregate them by bus, and clip the resulting peak load to non-negative values. As a result, the `co2 atmosphere` bus receives zero load-shedding capacity instead of a negative one. Please let me know if you think that's a feasible solution

ekatef:
Hello! Completely support treatment of CO2 as a time-series. I wonder if it could make sense to generalise this representation also for power-only model? Out of scope of this PR obviously but can be added as an issue in case you don't see any major drawbacks.

danielelerede-oet:
Good point @ekatef in the power-only model, CO2 emissions are already linked to dispatch through the carrier `co2_emissions` attribute, so the situation is slightly different from the sectoral case fixed here.
Still, it could make sense to unify the representation. I can look into this separately and, if it remains clean, propose it in a follow-up PR. @davide-f if you think the approach for the sector model is ok I would keep it as a standalone PR and merge it :)

davide-f (approved):
Good to go for me with thanks! Waiting CI to pass

PR #2028, opened 2026-09-08
Modularize Snakefile

yerbol-akhmetov:
Good day. This PR aims to make Snakefile mode modular by grouping the rules into individual `.smk` files.

## Checklist


yerbol-akhmetov:
@davide-f, @GbotemiB. I would be happy, if you can review the PR and suggest any improvements if possible.

scripts/_helpers.py
```diff
@@ -2475,3 +2475,24 @@ def get_linetype_by_voltage_and_country(
...
+        The root directory of the project (where the script directory is located).
+
+    Returns
+    -------
+    Callable[[str], Path]
+        A function that takes a script name as input and returns the full path to the script.
+    """
+
+    def _get_script_path(script: str) -> Path:
+        return Path("file://") / project_dir / "scripts" / script
```
GbotemiB: Do we really need this `file://`

GbotemiB (approved):
Looks good to me. I am approving with an added minor comment.

scripts/_helpers.py
yerbol-akhmetov: Thank you for the comment, @GbotemiB. It is not needed because we provide input as absolute path ( `Path(BASE_DIR)`). I will remove it.

ekatef (changes requested):
Hello @yerbol-akhmetov, totally support the idea to make Snakemake more structured. I think it is perfectly alright to move into smk rules functionality which is relatively independent of other rules, such as `retrieve.smk` or `postprocess.smk`.

However, it doesn't look a good design for me if we do bulk externalising of all the rules as it damages integrity. If you have code scattered through separate smk rules, it became quite a challenge to spot overall logic of the workflow. While it may be an essential part of debug since we have quite a complex workflow and using Snakemake DAG is not really handy.

davide-f (commented):
Hello!

Thanks for the PR and great comments!
I agree on the need to find a compromise to reduce the number of files while enhancing understanding on the rules.

I propose to merge all solve_*.smk and run_scenarios into a solve.smk, this saves three files.

The monte_carlo file could be disaggregated into build_electricity and solve.smk too to preserve the logic, otherwise it contains parts of both. This could save an additional file.

collect.smk is quite short, though inherently coherent. I'm fine with keeping it and if you have better ideas feel free to share

yerbol-akhmetov:
> Hello!
> 
> Thanks for the PR and great comments! I agree on the need to find a compromise to reduce the number of files while enhancing understanding on the rules.
> 
> I propose to merge all solve_*.smk and run_scenarios into a solve.smk, this saves three files.
> 
> The monte_carlo file could be disaggregated into build_electricity and solve.smk too to preserve the logic, otherwise it contains parts of both. This could save an additional file.
> 
> collect.smk is quite short, though inherently coherent. I'm fine with keeping it and if you have better ideas feel free to share

Thank you, @davide-f. I have applied your suggestions.

yerbol-akhmetov:
Thank you. I have applied the comments, @davide-f. Could you please take a look? What do you think regarding the updates? @ekatef

davide-f (approved):
Thanks Yerbol! Neat implementation. I have no further question

PR #2022, opened 2026-09-04
Remove support to pydoe2 in monte_carlo

davide-f:
# Closes #2021 

## Changes proposed in this Pull Request
Remove pydoe2 dependency that is outdated, is not a default and has alternative options (chaospy and scipy)

## Checklist

- [x] I consent to the release of this PR's code under the AGPLv3 license and non-code contributions under CC0-1.0 and CC-BY-4.0.
- [ ] I tested my contribution locally and it seems to work fine.
- [ ] Code and workflow changes are sufficiently documented, including updates to docstrings for meaningful functions.
- [ ] Newly introduced dependencies are added to `envs/environment.yaml` and `doc/requirements.txt`.
- [ ] Changes in configuration options are added in all of `config.default.yaml` and `config.tutorial.yaml`.
- [ ] Add a test config or line additions to `test/` (note tests are changing the config.tutorial.yaml)
- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/user-guide/configuration.md` and `doc/tutorials/electricity-model.md`.
- [ ] If config sections were added, renamed, or removed, update `doc/assets/scripts/extract_config_snippets.py` accordingly.
- [ ] Archives of the uploaded data do not have an enclosing folder and archive names correspond to the conventions of `configs/bundle_config.yaml`.
- [x] A note for the release notes `doc/release-notes.md` is amended in the format of previous release notes, including reference to the requested PR.

PR #2032, opened 2026-09-13
Unlock snakemake 9

davide-f:
# Closes # (if applicable)

This PR builds on top of #1676 and https://github.com/pypsa-meets-earth/pypsa-earth/pull/1312 to enable snakemake 9. It also includes https://github.com/pypsa-meets-earth/pypsa-earth/pull/2023 and https://github.com/pypsa-meets-earth/pypsa-earth/pull/2022
I'm keeping it in draft mode as it requires the others to be merged first

## Checklist


ekatef:
@davide-f would you mind to resolve the conflicts and update the PR branch here and in #1676? At the moment, I'm a bit struggling with checking the diff.

davide-f:
For convenience, you can see the differences between this PR and #1676 here: https://github.com/davide-f/pypsa-earth/pull/84/changes#top

Snakefile
```diff
@@ -11,11 +11,9 @@ sys.path.append("./scripts")
 from pathlib import Path
 from shutil import copyfile, move, unpack_archive
 
-from snakemake.remote.HTTP import RemoteProvider as HTTPRemoteProvider
-
-from _helpers import branch  # Remove if Snakemake >= 8.3.0
 from _helpers import (
     BASE_DIR,
+    branch,
```
ekatef: I thought Snakemake v9+ has in-built `branch` method? We have currently a comment on the need to replace with it the self-implemented `branch` function:
 https://github.com/pypsa-meets-earth/pypsa-earth/blob/37fafda6858e02a0986508d07c0b028cd6745c2b/Snakefile#L16
Do you see any reasons to avoid it?

Snakefile
```diff
@@ -149,25 +151,46 @@ rule plot_all_summaries:
...
-            "benchmarks/" + RDIR + "retrieve_databundle_light"
-        script:
-            "scripts/retrieve_databundle_light.py"
+    bundle_dict = {
+        category: get_best_bundles_in_snakemake(config, include_categories=[category])
+        for category in databundle_categories
+    }
+
+    # List selected bundles for each category
+    print("\n====================================================")
```
ekatef: Please, remember to replace prints with loggers

scripts/_helpers.py
```diff
@@ -928,73 +931,107 @@ def mock_snakemake(
...
 
     import snakemake as sm
-
-    try:
-        from pypsa.descriptors import Dict
-    except:
-        from pypsa.definitions.structures import Dict  # from pypsa version v0.31
+    from packaging import version
+    from pypsa.definitions.structures import Dict
+    from snakemake import __version__ as sm_version
```
ekatef: I guess `sm_version` is something string-like, not boolean? Not sure it is well consistent with using [PYPSA_V1](https://github.com/davide-f/pypsa-earth/blob/8beefba289b3b78e7d98c4314a75344554b7ecff/scripts/_helpers.py#L60) variable

Snakefile
```diff
@@ -149,25 +151,46 @@ rule plot_all_summaries:
 
 if config["enable"].get("retrieve_databundle", True):
 
-    bundles_to_download = get_best_bundles_in_snakemake(
-        config, exclude_categories=["cutouts"]
+    databundle_categories = get_databundle_categories(
+        config["databundles"], exclude_categories=["cutouts"]
```
ekatef: From the user experience I have on hand, retrieve_databundle infrastructure is very powerful but still has a level of abstraction which is higher than comfortable one even for experience developers. Creating a dedicated function to get a list of categories adds even more complexity to that. I agree that it looks elegant but fear it can be counterproductive.

My feeling is that a hard-coded list would work in a better way, at least to a moment when the whole retrieve_databundle part will be well documented, and it's major stability issues will be fixed.

scripts/retrieve_databundle_light.py
```diff
@@ -1086,9 +1120,12 @@ def retrieve_databundle(
...
     missing_bundles = set(bundles_to_download) - set(downloaded_bundles)
     if missing_bundles:
-        logger.warning(
-            "The following bundles could not be downloaded:\n\t"
+        raise RuntimeError(
+            "Databundle retrieval was incomplete. The following bundles could not be downloaded:\n\t"
             + "\n\t".join(list(missing_bundles))
+            + "\n"
+            "Retry or run `python scripts/non_workflow/databundle_cli.py "
+            "--diagnostic logs/databundle_cli.yaml` for inspection."
```
ekatef: To your best knowledge, has been usage of the `databundle_cli` tested for this specific command. As far as I'm aware we still have some pieces to polish in the integration of `databundle_cli` into the workflow. It would be a bit awkward  to recommend as a retrieval fix a solution which leads to issues itself.

ekatef (commented):
Thank you @davide-f, a really great work! Have added a few comments which are though rather minor. Looking forward to see it completed.

Snakefile
davide-f: The custom function handled better the types: if the input is a dictionary, the else case returns an empty list. I can explore removing it

Snakefile
davide-f: I can move this section of code in a helper function in the retrieve_databundle script and import it here.
Currently, as it is in the snakefile I cannot add it, but if the code is in the script I can.
Does it work for you?

scripts/_helpers.py
davide-f: Hard to understand: I propose str_snakemake_version or str_sm_version if we need a shorter version

scripts/retrieve_databundle_light.py
davide-f: Revised

Snakefile
davide-f: This implies that when a user needs to add a category in the bundle, it needs to add the name of the category into the script as well, otherwise this functionality doesn't work. I see it quite error-prone.
The current list of categories is about 10, it will be in the same order of length as the function. Do you see it easier to maintain? May a different naming or some comments help the readability?
Example: "Collect all databundle categories, except the cutouts.

Optionally, the function could collect all databundles and drop cutouts afterwards e.g. `set(get_databundle_categories(config["databundles"])) - set(["cutouts"])`

Snakefile
davide-f: Confirmed: when the entry is a dictionary, branch(..) returns an empty list anyway. To fix it, we need to explicitly add the default (..., {}) always. I'll explore it

davide-f: see https://github.com/pypsa-meets-earth/pypsa-earth/actions/runs/36063842780/job/107849200571?pr=2032

PR #2037, opened 2026-09-16
Fix country lookup for country-specific line type mappings

danielelerede-oet:

This PR fixes country-specific transmission line-type selection introduced in #1933.

`base_network()` currently resets the bus index immediately after loading the OSM buses:

```python
buses = _load_buses_from_osm(inputs.osm_buses).reset_index(drop=True)
```

Country-specific line types are then selected by mapping each line's `bus0` identifier to the corresponding bus country:

```python
lines["country"] = lines["bus0"].map(buses["country"])
```

Since `bus0` and `bus1` retain the original OSM bus identifiers, resetting the bus index breaks this lookup. The line country becomes `NaN`, causing `get_linetype_by_voltage_and_country()` to fall back to the default line-type mapping instead of using the country-specific mapping.

The issue was identified while integrating country-specific line types into PyPSA-Zambia. Before the fix, all 205 transmission lines in the Zambia base network had `country = NaN` and therefore used the default mappings. After preserving the original bus index, all 205 lines were assigned to `ZM` and the expected Zambia-specific line types were selected, while topology and `num_parallel` remained unchanged.

## Checklist


scripts/base_network.py
```diff
@@ -871,7 +871,7 @@ def base_network(
         Fully assembled base network with buses, lines, links, transformers,
         converters, country assignments, and underwater fractions.
     """
-    buses = _load_buses_from_osm(inputs.osm_buses).reset_index(drop=True)
+    buses = _load_buses_from_osm(inputs.osm_buses)
     lines = _load_lines_from_osm(inputs.osm_lines).reset_index(drop=True)
```
davide-f: On the same principle, I'd remove it also here
```suggestion
    lines = _load_lines_from_osm(inputs.osm_lines)
```

davide-f (commented):
Nice catch! This fix is indeed needed.
It would be useful to introduce a reset of the index in build_osm_network to improve the index.
I think before set_line_ids may be a good location e.g. https://github.com/pypsa-meets-earth/pypsa-earth/blob/b88e4127397c92ec760fdfd4dae3bdf015221c67/scripts/build_osm_network.py#L523

If you don't have time to investigate, we can add a new issue to track this point

danielelerede-oet:
Thanks @davide-f I think it makes sense to apply your suggestion.

Regarding the index reset in `build_osm_network`, I agree that handling it would be cleaner, but I’d prefer to keep it as a separate follow-up since it affects the OSM network-building step. I'm opening another PR for that.

davide-f (approved):
Ready to merge as CI passes with thanks!

